PR #4, opened 2024-12-06
Add clang format

zhjwpku:
Copied from apache arrow

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+---
+BasedOnStyle: Google
+ColumnLimit: 90
+DerivePointerAlignment: false
+IncludeBlocks: Preserve
+IndentPPDirectives: AfterHash
--- a/.clang-format
+++ b/.clang-format
@@ -19,5 +19,14 @@ Language: Cpp
 BasedOnStyle: Google
 ColumnLimit: 90
 DerivePointerAlignment: false
-IncludeBlocks: Preserve
+IncludeBlocks: Regroup
+IncludeCategories:
+- Regex: '^<.*/' # Like <boost/lexical_cast.hpp>
+  Priority: 3
+- Regex: '^<.*\.' # Like <fcntl.h>
+  Priority: 1
+- Regex: '^<' # Like <vector>
+  Priority: 2
+- Regex: '^"' # Like "util/auth-util.h"
+  Priority: 4
 IndentPPDirectives: AfterHash
--- a/.clang-format
+++ b/.clang-format
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 ---
+Language: Cpp
 BasedOnStyle: Google
 ColumnLimit: 90
 DerivePointerAlignment: false


.clang-format
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+---
+BasedOnStyle: Google
+ColumnLimit: 90
```
gaborkaszab, line 20: It's 100 for the Java project. We also have 90 as a limit for Impala but I find it annoyingly short tbh :)

.clang-format
```diff
@@ -0,0 +1,22 @@
...
+DerivePointerAlignment: false
+IncludeBlocks: Preserve
```
gaborkaszab: I personally find the output of 'Regroup' more readable

.clang-format
```diff
@@ -0,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
gaborkaszab, line 1: Could you add some instructions into the README file about how to execute clang checks/rewrites?

.clang-format
pitrou: The problem with very long lines is that it degraded the PR review experience IMHO.

.clang-format
pitrou: How about adding a pre-commit file?

.clang-format
wgtmac: I have experienced 80, 90, 100, 120 from different projects. Personally I prefer 90 but no strong opinion.

.clang-format
wgtmac: Can we integrate a Github action like https://github.com/cpp-linter/cpp-linter-action? FYI I have filed an INFRA JIRA to whitelist it but it seems not resolved yet: https://issues.apache.org/jira/browse/INFRA-26318

I have also added a C++ format check via Github action `jidicula/clang-format-action`: https://github.com/apache/orc/pull/1308/files#diff-48c0ee97c53013d18d6bbae44648f7fab9af2e0bf5b0dc1ca761e18ec5c478f2 But I'm in favor of `cpp-linter/cpp-linter-action` above.

cc @raulcd

.clang-format
zhjwpku: I also prefer 90 since it's good for side by side PR review. I will mark this as resovled if you don't mind.

.clang-format
zhjwpku: I agree we should have a pre-commit file and Github action. I use .clang-format for auto format when source file edited in vscode. TBH I don't know how to set a pre-commit and cpp-linter-action, so do you mind we did that in a seperate PR?

wgtmac: I don't think they are blockers at this moment. Eventually we should have them.

raulcd: I am not familiar with the `cpp-linter-action` or the `clang-format-action` but they don't seem complex to integrate. At Arrow we use pre-commit as suggested by @pitrou .

https://github.com/apache/arrow/blob/main/.pre-commit-config.yaml#L61-L101

https://github.com/apache/arrow/blob/main/.github/workflows/dev.yml#L43-L67

I am happy to integrate any, should be pretty straight forward once we agree on a configuration.

.clang-format
gaborkaszab: I don't think we reached consensus on this front, so would prefer not to resolve this ATM. I just want to learn what difficulties people find with 100 long lines when it comes to PR reviewing before moving on.
I left the same comment on on you other review too. I'm sure that Iceberg Java has a limit of 100. I haven't found any difficulties so far reviewing code on that project. Hence the curiosity from my side.

pitrou: > I just want to learn what difficulties people find with 100 long lines when it comes to PR reviewing before moving on.

The main problem is that it can imply horizontal scrolling which is usually tedious and destroys readability. That depends on font size, window width, etc. Also, as someone else mentioned, if you like side by side reviews then 90 is already quite long.

gaborkaszab: Thanks for sharing your view! I think I can live with 90 then.

.clang-format
zhjwpku: pre-commit or cpp-linter-action will be integrated in later PR, so I will mark this as resolved, thanks.

.clang-format
zhjwpku: Hi, @wgtmac @pitrou @raulcd, what's your opinions on this one. I copied this from arrow project, and I see ORC also has the same setting, so I myself prefer this as is.

.clang-format
Fokko: I'll leave this here: https://github.com/apache/iceberg-python/blob/1c73b7c97d8db5e76dd34e57debee3753439f0eb/pyproject.toml#L915

However, it is language-specific. I don't find any issues with reviewing PRs with longer-length lines, but I think it is similar to what you're used to.

zhjwpku: > I'll leave this here: https://github.com/apache/iceberg-python/blob/1c73b7c97d8db5e76dd34e57debee3753439f0eb/pyproject.toml#L915
> 
> However, it is language-specific. I don't find any issues with reviewing PRs with longer-length lines, but I think it is similar to what you're used to.

Agree, I just pushed a commit that make clang-format language-specific, now the rules only effect C++ source code.
Interestingly the ORC project has `ColumnLimit: 100`, it's not only language-specific but also project-specific, so I suggest we use `ColumnLimit: 90` for now and do some change when we find it's not proper later.

.clang-format
gaborkaszab: Other than another project uses this, is there any reason you prefer this setting? With 'Regroup' the includes are more structured and organised. What are the pros for 'Preserve'?

.clang-format
gaborkaszab: "90 for now and do some change when we find it's not proper later."

Changing this later seems painful for me. In case we, for any reason decide to increase this to 100 later, we gonna have to do alignments in the code for this. That would kind of mess up the usage of "git blame" because many lines will show the refactor and not the actual change on that line.
My point here is that we have to be reasonable from the beginning and not just rush forward by copy-pasting stuff from other projects. As Fokko mentioned, the Java library uses 100 and the python uses even more, and he didn't find any issues with that. Im pretty sure it would work for the C++ project too.

With the current world of decent sized screens I find it strange that 10 extra chars causes issues for PR reviews.
Would you mind re-opening the conversation here. Let's see if anyone else wants to chime in on this matter.

pitrou: > In case we, for any reason decide to increase this to 100 later, we gonna have to do alignments in the code for this.

We don't. It's in the case where we _decrease_ the value that adjustments are needed.

> As Fokko mentioned, the Java library uses 100 and the python uses even more, and he didn't find any issues with that. 

Different people have different preferences (and screen/window/font size setups too!), but the longer the line length, the more people will encounter difficulties reading/reviewing code.

.clang-format
zhjwpku: > Other than another project uses this, is there any reason you prefer this setting? With 'Regroup' the includes are more structured and organised. What are the pros for 'Preserve'?

We use Preserve in our own projects, developers who add the header files should give them a reasonable position.

.clang-format
zhjwpku: > "90 for now and do some change when we find it's not proper later."
> 
> Changing this later seems painful for me. In case we, for any reason decide to increase this to 100 later, we gonna have to do alignments in the code for this. That would kind of mess up the usage of "git blame" because many lines will show the refactor and not the actual change on that line. My point here is that we have to be reasonable from the beginning and not just rush forward by copy-pasting stuff from other projects. As Fokko mentioned, the Java library uses 100 and the python uses even more, and he didn't find any issues with that. Im pretty sure it would work for the C++ project too.
> 
> With the current world of decent sized screens I find it strange that 10 extra chars causes issues for PR reviews. Would you mind re-opening the conversation here. Let's see if anyone else wants to chime in on this matter.

Sure, I can reopen this conversation. I once suggested starting a vote process on slack, but you seems not agree. I'm wondering if someone does not agree with others, what's the right way to reach consensus?
I just checked impala's .clang-format, https://github.com/apache/impala/blob/master/.clang-format, it has ColumnLimit: 90 for both Cpp and Java.

.clang-format
pitrou: Using Preserve allows us to have separate include blocks for 1) stdlib includes 2) third-party includes 3) own project includes.

However, it seems the same effect could be achieved with Regroup and a carefully-defined IncludeCategories? (not sure how to detect stdlib includes that way, though)

raulcd: even though I don't have much experience with it, `Regroup` seems like a good approach to me

zhjwpku: The doc says `IncludeCategories` can be used with any `IncludeBlocks` strategy, there is a additional field `SortPriority` which can be used with `Regroup`[0].  Here is the setting from Impala, do you think this works for us?

```
IncludeCategories:
- Priority: 3
  Regex: '^<.*/' # Like <boost/lexical_cast.hpp>
- Priority: 1
  Regex: '^<.*\.' # Like <fcntl.h>
- Priority: 2
  Regex: '^<' # Like <vector>
- Priority: 4
  Regex: '^"' # Like "util/auth-util.h"
```

[0] https://rocm.docs.amd.com/projects/llvm-project/en/latest/LLVM/clang/html/ClangFormatStyleOptions.html#includecategories

pitrou: As long as it allows dissociating stdlib, third-party, and own project imports, then I'm fine with it.

.clang-format
gaborkaszab: > I just checked impala's .clang-format, it has ColumnLimit: 90 for both Cpp and Java.
Yes, and I'm involved with the Impala project and can tell you that whoever I talked to on the community hates the 90 char limit and would prefer to raise it to 100.

> We don't. It's in the case where we decrease the value that adjustments are needed.
I don't think this statement is correct. If you increase the line length then there are lines what are broken into 2 but with the new limit will fit into one. In case you have auto formatting those will be rewritten, loosing the git blame info.

I don't want to block this PR, don't get me wrong. I just want to be more careful than simply rushing commits like this as these will have an effect on the project forever.

> someone does not agree with others
This gives me the impression that you say it's me against everyone else :) Fokko also politely mentioned the char limits for other projects too.

I see that there is support for the 90 char limit by people coming from the Arrow community, so let's just simply give other set of people a bit more time to chime in and then we can resolve this comment if nothing changes.

pitrou: > If you increase the line length then there are lines what are broken into 2 but with the new limit will fit into one. In case you have auto formatting those will be rewritten, loosing the git blame info.

Hmm, perhaps one can prevent clang-format from removing existing line breaks? The maximum line length is just a maximum after all; the goal is not to make lines as long as possible...

Fokko: In the Iceberg community, we try to be consensus-driven. Here it is important to have a conversation _about why_ anyone thinks it should be 90, 100, or something else, so we understand each others' reasoning and potential concerns. Neither shutting down a discussion by closing a comment, or jumping to a vote is not helping to reach that, and is hopefully not needed because it adds procedure. I think, just pointing to other projects is not a strong argument, because it might be like that for historical reasons. When starting a new project, it is a good moment to reconsider some of the decisions made in the past. We try to reach a state where everyone can live with the decision being made, give their approval, and move forward.

raulcd: From my side, both 90 and 100 sound reasonable. I don't have a strong preference. With current screen widths and resolutions, I don't see 100 being a problem.

pitrou: I still think we should be mindful of diverse situations:
* people with a smaller screen (such as a laptop)
* side-by-side diffs
* various windows arrangements
* people who need a larger font because of eyesight issues

There is also the issue that the longer the line, the farther the eye has to scan horizontally. It's not always very comfortable to do so. This is why newspapers typically use several narrow columns side-by-side.

In the end, of course, the difference between 90 and 100 will not be fundamental.

zhjwpku: I come from postgres, postgres has a line limit of 80, I found it's really easy to read(especially the code comments), so personally I prefer 90.

Fokko (approved):

.clang-format
zhjwpku: I plan to make the following changes after we reached a consensus.

```
-IncludeBlocks: Preserve
+IncludeBlocks: Regroup
+IncludeCategories:
+- Regex: '^<.*/' # Like <boost/lexical_cast.hpp>
+  Priority: 3
+- Regex: '^<.*\.' # Like <fcntl.h>
+  Priority: 1
+- Regex: '^<' # Like <vector>
+  Priority: 2
+- Regex: '^"' # Like "util/auth-util.h"
+  Priority: 4
```

Before I push a new commit, I'd like to explain why I think this makes sense.

I did a experiment, `Regroup` seems to be the default setting if we don't specify a `IncludeBlocks` setting, I put it here explicitly because I think it's more clear for the readers. With the `IncludeCategories` settings, the following:

```
#include <iostream>
#include <string>

#include <boost/lexical_cast.hpp>

#include "util/auth-util.h"

#include <vector>

#include <fcntl.h>

#include "demo_table.h"
```

goes to:

```
#include "demo_table.h"

#include <fcntl.h>

#include <iostream>
#include <string>
#include <vector>

#include <boost/lexical_cast.hpp>

#include "util/auth-util.h"
```

It's slightly different from Impala's setting because I think putting the `Regex` as the first line of each rule is more readable, "If match this regex, then goes priority xxx".

What do you think about this proposal? @gaborkaszab @pitrou @Fokko @wgtmac @raulcd

gaborkaszab: Seems reasonable for me. Thanks!

zhjwpku: Pushed the change, thanks for your reaction.

gaborkaszab (approved):
Since I'm the only one against 90 long rows, let's go with that! Thanks for the PR @zhjwpku

Fokko:
Let's move this forward, thanks everyone for chiming in here!

PR #5, opened 2024-12-06
Add CMake format

zhjwpku:
use the following command to apply the rules to all existing CMakeLists.txt

find . -name CMakeLists.txt | xargs cmake-format -c cmake-format.py -i

### DIFF
--- a/cmake-format.py
+++ b/cmake-format.py
@@ -1,0 +1,74 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# cmake-format configuration file
+
+# -----------------------------
+# Options affecting formatting.
+# -----------------------------
+with section("format"):
+    # How wide to allow formatted cmake files
+    line_width = 90
+
+    # How many spaces to tab for indent
+    tab_size = 2
+
+    # If a positional argument group contains more than this many arguments,
+    # then force it to a vertical layout.
+    max_pargs_hwrap = 4
+
+    # If the statement spelling length (including space and parenthesis) is
+    # smaller than this amount, then force reject nested layouts.
+    # This value only comes into play when considering whether or not to nest
+    # arguments below their parent. If the number of characters in the parent
+    # is less than this value, we will not nest.
+    min_prefix_chars = 32
+
+    # If true, separate flow control names from their parentheses with a space
+    separate_ctrl_name_with_space = False
+
+    # If true, separate function names from parentheses with a space
+    separate_fn_name_with_space = False
+
+    # If a statement is wrapped to more than one line, than dangle the closing
+    # parenthesis on it's own line
+    dangle_parens = False
+
+    # What style line endings to use in the output.
+    line_ending = 'unix'
+
+    # Format command names consistently as 'lower' or 'upper' case
+    command_case = 'lower'
+
+    # Format keywords consistently as 'lower' or 'upper' case
+    keyword_case = 'unchanged'
+
+# ------------------------------------------------
+# Options affecting comment reflow and formatting.
+# ------------------------------------------------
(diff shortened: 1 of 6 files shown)


cmake-format.py
```diff
@@ -0,0 +1,74 @@
...
+# under the License.
+
+# cmake-format configuration file
+
+# -----------------------------
+# Options affecting formatting.
+# -----------------------------
+with section("format"):
+    # How wide to allow formatted cmake files
+    line_width = 90
```
gaborkaszab, line 25: I'd prefer 100 if we see any point of limiting the line width for CMake files at all.

src/puffin/CMakeLists.txt
```diff
@@ -18,11 +18,10 @@
...
 
-ADD_ICEBERG_LIB(
-  iceberg_puffin
-  CMAKE_PACKAGE_NAME
-  puffin
-  SOURCES
-  ${ICEBERG_PUFFIN_SOURCES}
-  PRIVATE_INCLUDES
-  ${ICEBERG_PUFFIN_INCLUDES})
+add_iceberg_lib(iceberg_puffin
```
gaborkaszab, line 21: Might be personal preference but I find the original formatting more readable. I don't like this much indentation. Do you think there is a way to configure that?

wgtmac: I simply used the `cmake-format` plugin in vscode with default setting which formatted these CMake files.

zhjwpku: > I simply used the `cmake-format` plugin in vscode with default setting which formatted these CMake files.

Yeah, but we need to check the format when PRs are created, I did not include the CI part because I not sure what is the best way to do it. We might consider that together with clang-format.

cmake-format.py
zhjwpku: No strong opinion, but as @pitrou mentioned on #4, 90 is better for PR review, so I guess we should keep it as 90?

gaborkaszab: Java and iceberg-python also seem to have 100 for limit. I personally find that with 90 chars you have to break lines too often hurting code readability. Not sure what advantages it have for PR reviewing.

Fokko: `iceberg-python` is at 130: https://github.com/apache/iceberg-python/blob/ede363bd63640f3db6c3b3d6904a5765dafd2600/pyproject.toml#L916

pitrou: Yes, and you get this kind of code that's very annoying to read:
pyiceberg1

gaborkaszab: Obviously it depends on what device you use for reading the code, but for me, not even on a big screen but on a regular sized laptop this whole thing fits into the screen including the navigator panel on the left.

Screenshot 2024-12-11 at 10 33 26

pitrou: Sure. The point is that, the longer the code lines, the more people get annoyed.

zhjwpku: I'm not sure I get you right, this cmake-format.py is for formatting CMakeLists.txt files, not python files. 

I checked some large projects' CMakeLists.txt(impala, arrow, ceph), it seems there can be some long `message` settings in the file, other than that most of the settings seems fairly short. 

I can stand both 90 and 100 but 90 is my first choice though.

pitrou: @zhjwpku You're right, but people's expectations wrt line length is usually the same from one language to another, and this was a good comparison IMHO :)

gaborkaszab: I think on the other PR we concluded (even me :) ) to use the 90 limit. This one should have the same value.

Fokko (approved):
Before adding source files, I think it would be good to conclude this PR. Any pending concerns? If not, please approve the PR 👍

zhjwpku:
> One general question: I see the command in the description that leverages this change to format CMake files. It's up to the contributor to run this manually, and I'm wondering if it makes sense to introduce some build script that runs this, the clang format, and also compiles the project. We'd have more guarantee then that these format requirements are met

Make sense, I will try to find some gh actions that are fit for these tasks, maybe after #7 merged.

gaborkaszab (approved):
As discussed, some follow-up PRs would be great for running the formatting mechanisms.

wgtmac (approved):
+1

nit: should we put cmake-format.py into a sub-folder like `scripts` or something?

pitrou:
> nit: should we put cmake-format.py into a sub-folder like `scripts` or something?

It's not a script, it's a config file for the `cmake-format` utility (yes, it's confusing and awkward).

pitrou:
> > Make sense, I will try to find some gh actions that are fit for these tasks, maybe after #7 merged.
> 
> Sounds great, @zhjwpku ! Also some build script that people can run locally would be a nice help for contributors.

We should just use [pre-commit](https://pre-commit.com/), which is able to do both.

Fokko:
Thanks for raising this PR @zhjwpku and thanks for the review @gaborkaszab @pitrou and @wgtmac 

> We should just use [pre-commit](https://pre-commit.com/), which is able to do both.

I love pre-commit 👍

PR #7, opened 2024-12-09
Add basic CI to build iceberg and example

raulcd:
This PR adds a minimal build for Ubuntu 24.04, macOS 14 and Windows 2019 for both Iceberg CPP and the Example to validate follow up PRs build correctly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,8 +38,8 @@ env:
 
 jobs:
   ubuntu:
-    name: AMD64 Ubuntu latest
-    runs-on: ubuntu-latest
+    name: AMD64 Ubuntu 24.04
+    runs-on: ubuntu-24.04
     timeout-minutes: 30
     strategy:
       fail-fast: false
@@ -55,8 +55,8 @@ jobs:
         shell: bash
         run: ci/scripts/build_example.sh $(pwd)/example
   macos:
-    name: AArch64 macOS latest
-    runs-on: macos-latest
+    name: AArch64 macOS 14
+    runs-on: macos-14
     timeout-minutes: 30
     strategy:
       fail-fast: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,7 +57,7 @@ jobs:
         run: ci/scripts/build_example.sh $(pwd)/example
   macos:
     name: AArch64 macOS 14
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     if: ${{ !contains(github.event.pull_request.title, 'WIP') }}
     timeout-minutes: 30
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,75 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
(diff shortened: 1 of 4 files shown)


raulcd:
I've validated actions run successfully on my fork: https://github.com/raulcd/iceberg-cpp/actions/runs/12233865254

.github/workflows/test.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+      - name: Build Iceberg
+        shell: bash
+        run: ci/scripts/build_iceberg.sh $(pwd)
+      - name: Build Example
+        shell: bash
+        run: ci/scripts/build_example.sh $(pwd)/example
+  macos:
```
wgtmac, line 57: Do we need to add a Windows build as well?

raulcd: I've added Windows. I've had to add a minor fix to the examples CMake for it to work.

raulcd: See the run on my fork here: https://github.com/raulcd/iceberg-cpp/actions/runs/12254974905/job/34187043292

.github/workflows/test.yml
```diff
@@ -0,0 +1,95 @@
...
+permissions:
+  contents: read
+
+env:
+  ICEBERG_HOME: /tmp/iceberg
+
+jobs:
+  ubuntu:
+    name: AMD64 Ubuntu latest
+    runs-on: ubuntu-latest
```
Fokko: Do we want to pin this to a specific version to avoid breaking the build when `latest` gets updated?

.github/workflows/test.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Build Iceberg
+        shell: bash
+        run: ci/scripts/build_iceberg.sh $(pwd)
+      - name: Build Example
+        shell: bash
+        run: ci/scripts/build_example.sh $(pwd)/example
+  macos:
+    name: AArch64 macOS latest
+    runs-on: macos-latest
+    if: ${{ !contains(github.event.pull_request.title, 'WIP') }}
```
Fokko: I would leave this out, even for a WIP PR is might be nice to see if the CI passes

Fokko (approved):
Some small nits, but in general this looks good to me. Thanks for working on this @raulcd

.github/workflows/test.yml
raulcd: I am in two minds with this one but I think I prefer to have `latest` so if a new OS (GitHub runner) is released we find early that there are issues with it.
The other approach might leave us with silent failures on newer systems if we don't update them periodically but both points are valid, I am ok with both approaches.

gaborkaszab: With pinning to a specific version we could document that as the supported version and moving on to a later version could be performed in a more controlled fashion.
With using latest we might have surprise breaks even without changing code in this project. This wouldn't be that easy to investigate either. Also merging code would be blocked as long as we don't fix the incompatibility with a new Ubuntu version even though we could still deliver code using the release before.

I'm +1 to pin this to a specific version.

raulcd: Do we want to pin macOS too?

pitrou: It also depends how often these images change versions. If it's once every 2 years, it's less contentious to use "latest" than if GH bumps their versions every month.

raulcd: `ubuntu-latest` is about to be migrated to `24.04` and it was migrated to `22.04` from `20.04` around 2 years ago, see: https://github.com/actions/runner-images/commit/9c544f47011f3fb7f13ab122927f15939d950b49
`macos-latest` has been having a higher rotation, around once a year (it went from macos-11, to macos-12 and then to macos-14)

raulcd: I've pinned them to ubuntu-24.04 and macos-14

raulcd:
Run is successful as seen here:
https://github.com/raulcd/iceberg-cpp/actions/runs/12314556083
@gaborkaszab @wgtmac if you are :+1:  with this it probably can be merged

gaborkaszab (approved):
Thanks for this works, @raulcd !

Fokko:
Thanks @raulcd for the PR and thanks @wgtmac @gaborkaszab @pitrou for reviewing. It seems like most of us are in agreement, and we can always fine-tune later. I've merged this so this won't block any next steps 👍

PR #8, opened 2024-12-13
add Status data structure

zhjwpku:
This is adapted from Apache Arrow, which a pre-step before integrating the Result data structure.

### DIFF
--- a/src/iceberg/status.h
+++ b/src/iceberg/status.h
@@ -1,0 +1,278 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <cstring>
+#include <format>
+#include <iosfwd>
+#include <memory>
+#include <string>
+#include <utility>
+
+#include "util/compare.h"
+#include "util/macros.h"
+#include "util/to_string_ostreamable.h"
+
+namespace iceberg {
+
+enum class StatusCode : char {
+  OK = 0,
+  IOError = 1,
+  NotImplemented = 2,
+  UnknownError = 127
+};
+
+/// \brief An opaque class that allows subsystems to retain
+/// additional information inside the Status.
+class StatusDetail {
+ public:
+  virtual ~StatusDetail() = default;
+  /// \brief Return a unique id for the type of the StatusDetail
+  /// (effectively a poor man's substitute for RTTI).
+  virtual const char* type_id() const = 0;
+  /// \brief Produce a human-readable description of this status.
+  virtual std::string ToString() const = 0;
+
+  bool operator==(const StatusDetail& other) const noexcept {
+    return std::string(type_id()) == other.type_id() && ToString() == other.ToString();
+  }
+};
+
+/// \brief Status outcome object (success or error)
+///
+/// The Status object is an object holding the outcome of an operation.
+/// The outcome is represented as a StatusCode, either success
+/// (StatusCode::OK) or an error (any other of the StatusCode enumeration values).
+///
+/// Additionally, if an error occurred, a specific error message is generally
+/// attached.
+class [[nodiscard]] Status : public util::EqualityComparable<Status>,
+                             public util::ToStringOstreamable<Status> {
+ public:
+  // Create a success status.
+  constexpr Status() noexcept : state_(nullptr) {}
+  ~Status() noexcept {
+    if (ICEBERG_PREDICT_FALSE(state_ != nullptr)) {
+      DeleteState();
+    }
+  }
+
+  Status(StatusCode code, const std::string& msg);
+  /// \brief Pluggable constructor for use by sub-systems. detail cannot be null.
+  Status(StatusCode code, std::string msg, std::shared_ptr<StatusDetail> detail);
+
+  // Copy the specified status.
+  inline Status(const Status& s);
+  inline Status& operator=(const Status& s);
+
+  // Move the specified status.
+  inline Status(Status&& s) noexcept;
+  inline Status& operator=(Status&& s) noexcept;
+
+  inline bool Equals(const Status& other) const;
+
+  // AND the statuses.
+  inline Status operator&(const Status& s) const noexcept;
+  inline Status operator&(Status&& s) const noexcept;
+  inline Status& operator&=(const Status& s) noexcept;
+  inline Status& operator&=(Status&& s) noexcept;
+
+  /// Return a success status.
+  static Status OK() { return Status(); }
+
+  template <typename... Args>
+  static Status FromArgs(StatusCode code, std::string_view fmt, Args&&... args) {
+    return Status(code, std::vformat(fmt, std::make_format_args(args...)));
+  }
+
+  template <typename... Args>
+  static Status FromDetailAndArgs(StatusCode code, std::shared_ptr<StatusDetail> detail,
+                                  std::string_view fmt, Args&&... args) {
+    return Status(code, std::vformat(fmt, std::make_format_args(args...)),
+                  std::move(detail));
+  }
+
+  /// Return an error status when some IO-related operation failed
+  template <typename... Args>
+  static Status IOError(Args&&... args) {
+    return Status::FromArgs(StatusCode::IOError, std::forward<Args>(args)...);
+  }
+
+  /// Return an error status when an operation or a combination of operation and
+  /// data types is unimplemented
+  template <typename... Args>
+  static Status NotImplemented(Args&&... args) {
+    return Status::FromArgs(StatusCode::NotImplemented, std::forward<Args>(args)...);
+  }
+
+  /// Return an error status for unknown errors
+  template <typename... Args>
+  static Status UnknownError(Args&&... args) {
+    return Status::FromArgs(StatusCode::UnknownError, std::forward<Args>(args)...);
+  }
+
+  /// Return true iff the status indicates success.
+  constexpr bool ok() const { return (state_ == nullptr); }
+
+  /// Return true iff the status indicates an IO-related failure.
+  constexpr bool IsIOError() const { return code() == StatusCode::IOError; }
+  /// Return true iff the status indicates an unimplemented operation.
+  constexpr bool IsNotImplemented() const { return code() == StatusCode::NotImplemented; }
+  /// Return true iff the status indicates an unknown error.
+  constexpr bool IsUnknownError() const { return code() == StatusCode::UnknownError; }
+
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -56,6 +64,11 @@ add_subdirectory(api)
 add_subdirectory(src)
 
 if(ICEBERG_BUILD_TESTS)
+  fetchcontent_declare(googletest
```
pitrou: This adds a GoogleTest dependency. Was there a decision to use GT? A popular alternative is https://github.com/catchorg/Catch2/, I'm not sure which one would be better.

api/iceberg/util/string_builder.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include "visibility.h"
+
+namespace iceberg {
+
+namespace util {
+
+namespace detail {
+
+class ICEBERG_EXPORT StringStreamWrapper {
```
pitrou: Do we really to keep this from Arrow (which was C++11 compatible back then), or do we want to adopt a modern formatting library such as https://en.cppreference.com/w/cpp/utility/format or a backport thereof?

pitrou:
Also @zhjwpku can you make sure you proof-read your PR description? There are spell checkers that can help.

CMakeLists.txt
zhjwpku: Yeah, I see that Catch2 has TDD and BDD, which is better, I will do some work and find out which one fits us better.

api/iceberg/util/string_builder.h
wgtmac: I'm not sure whether `std::format` has been fully implemented by popular compilers. At least we can choose `fmt::format`: https://fmt.dev/

api/iceberg/status.h
```diff
@@ -0,0 +1,435 @@
...
+//
+// A Status encapsulates the result of an operation.  It may indicate success,
+// or it may indicate an error with an associated error message.
+//
+// Multiple threads can invoke const methods on a Status without
+// external synchronization, but if any of the threads may call a
+// non-const method, all threads accessing the same Status must use
+// external synchronization.
+
+// Adapted from Apache Kudu, TensorFlow, Apache Arrow
```
wgtmac: IIRC, status is the recommended convention from Google coding style. We haven't discussed whether to use status or exception (or something else) across the repo. Personally I prefer using exception which provides better debug-ability and requires less lines to deal with error handling.

src/common/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set(ICEBERG_COMMON_SOURCES util/string_builder.cc status.cc)
+set(ICEBERG_COMMON_INCLUDES "${ICEBERG_API_DIR}")
+
+ADD_ICEBERG_LIB(
+  iceberg_common
```
wgtmac: Why do we need a `iceberg_common` library? Isn't `iceberg-core` the common stuff?

CMakeLists.txt
pitrou: Perhaps open a GH issue to discuss it?

api/iceberg/util/string_builder.h
pitrou: Either is fine to me.

api/iceberg/status.h
pitrou: Would exceptions allow a rich exception hierarchy together with something similar to `StatusDetail`?

src/common/CMakeLists.txt
zhjwpku: > Why do we need a `iceberg_common` library? Isn't `iceberg-core` the common stuff?

I was thinking that maybe puffin also need the abilities that this module provide? I'm fine if you think this should be moved to iceberg-core.

api/iceberg/status.h
zhjwpku: > IIRC, status is the recommended convention from Google coding style. We haven't discussed whether to use status or exception (or something else) across the repo. Personally I prefer using exception which provides better debug-ability and requires less lines to deal with error handling.

From users' perspective, I think Status is better, developers tend to handle the Status/Result more likely than the exception. The iceberg-rust repo also adopt this kind of design IIUC.

pitrou: > The iceberg-rust repo also adopt this kind of design IIUC.

Yes, but that's the standard idiom in Rust, and there are syntactical facilities around it that don't exist in C++.
https://doc.rust-lang.org/std/result/

gaborkaszab:
I think this PR makes too much assumptions that we will go with everything that the Arrow community chose. For instance this seems to decide the "exceptions vs status codes" question. Also decides which test framework to use, and additionally pulls in a number of macros too.
I think at this point we should start conversations on these topics, maybe on the Slack channel or as github issues.

I myself planned to come up with the headers for basic classes like Catalog Table and such, but I'd personally prefer exceptions over status codes for many reasons. Method chaining is something I find really powerful in the Java implementation but with status codes you wouldn't be able to do that if you have to check the status for every function call. Not to mention that constantly checking for statuses would make the code less readable. Also Statuses would make the parameter list longer, not to mention that lines would be longer (which somewhat matters now that we went for 90 chars).

Long story short I'm -1 for this PR.

zhjwpku:
> I think this PR makes too much assumptions that we will go with everything that the Arrow community chose. For instance this seems to decide the "exceptions vs status codes" question. Also decides which test framework to use, and additionally pulls in a number of macros too. I think at this point we should start conversations on these topics, maybe on the Slack channel or as github issues.
> 
> I myself planned to come up with the headers for basic classes like Catalog Table and such, but I'd personally prefer exceptions over status codes for many reasons. Method chaining is something I find really powerful in the Java implementation but with status codes you wouldn't be able to do that if you have to check the status for every function call. Not to mention that constantly checking for statuses would make the code less readable. Also Statuses would make the parameter list longer, not to mention that lines would be longer (which somewhat matters now that we went for 90 chars).
> 
> Long story short I'm -1 for this PR.

This PR is not trying to decide which test framework to use, I have started a issue #12 where we can discuss the test framework.

zhjwpku:
> I think this PR makes too much assumptions that we will go with everything that the Arrow community chose. For instance this seems to decide the "exceptions vs status codes" question. Also decides which test framework to use, and additionally pulls in a number of macros too. I think at this point we should start conversations on these topics, maybe on the Slack channel or as github issues.
> 
> I myself planned to come up with the headers for basic classes like Catalog Table and such, but I'd personally prefer exceptions over status codes for many reasons. Method chaining is something I find really powerful in the Java implementation but with status codes you wouldn't be able to do that if you have to check the status for every function call. Not to mention that constantly checking for statuses would make the code less readable. Also Statuses would make the parameter list longer, not to mention that lines would be longer (which somewhat matters now that we went for 90 chars).
> 
> Long story short I'm -1 for this PR.

I have seen some use cases of Method chaining in C++ world, mostly people use it for constructing object I guess. Can you elaborate how we can use it with exception to make the apis easy to use?

I don't like method chaining, it makes debug more difficult, when you set a breakpoint on the method chain and exception happens, how do you know which call give out the exception?

I think it's generally good practice to always write very short and concise lines. Every line should just make one method call.

There is an old stackoverflow question which I think we can take a look, [Method chaining (fluent interfacing) - why is it a good practice, or not?](https://stackoverflow.com/questions/1103985/method-chaining-fluent-interfacing-why-is-it-a-good-practice-or-not)

gaborkaszab:
> There is an old stackoverflow question which I think we can take a look
There is a stackoverflow for everything and for the opposite too. I just googled for "Status codes vs exceptions" and it gives me the first 5 results supporting exceptions over status codes, I guess because the algorithm is biased by my personal preference it learned from me.

> when you set a breakpoint on the method chain and exception happens, how do you know which call give out the exception?
This works well in Java, I don't think it would be any different in C++ world. If you get an exception I guess you can check the trace where it was thrown and then set your debugger accordingly.
Also method chaining doesn't mean that you write everything in a single line, you could break as you want.

I did a quick run through on this PR and my general opinion is this: This PR assumes that we simply want to pull in auxiliary stuff from the Arrow project and without questioning whether we'd need them here or not we just pull entire files into this project. This is pretty difficult to review TBH, and if someone is not from the Arrow project it's even more difficult because they have to figure out the initial motivation of each functionality of another project.

I think this should be the other way around. Let's identify some functionality that we'd like to develop in this project, like interface for catalogs let's say, and if we find something in other projects that could help then pull in only the required part. Or let's have a very simple implementation ourselves, we can still evolve it later on.

wgtmac:
I agree with @gaborkaszab that it would be better to discuss a concrete API design (e.g. Table, FileIO, etc.) before introducing a full-functional status implementation. If we decide to go with status over exception, we can add a basic status class at first and make it powerful later on.

zhjwpku:
> I agree with @gaborkaszab that it would be better to discuss a concrete API design (e.g. Table, FileIO, etc.) before introducing a full-functional status implementation. If we decide to go with status over exception, we can add a basic status class at first and make it powerful later on.

Agreed, maybe we should schedule a sync meeting to discuss some of the issues.

lidavidm:
I'm a bit late here but even if we want to use Status/Result, it may be better to go with a backport of std::expected to at least try to align with the STL.

wgtmac:
@lidavidm I was about to reply to this thread to revive the discussion :)

zhjwpku:
> I'm a bit late here but even if we want to use Status/Result, it may be better to go with a backport of std::expected to at least try to align with the STL.
> 
> Ah sorry, I see discussion moved to #14.

No worry, I totally agree your suggestion, I will work on this way in the following days.

lidavidm:
I have used this before: https://github.com/TartanLlama/expected

I think my main complaint (this applies to Arrow too) is that the structure is a bit opaque in a debugger. Arrow ended up writing a GDB pretty printer script.

api/iceberg/util/string_builder.h
zhjwpku: I have changed this to builtin format, let's see if the CI complaints.

api/iceberg/status.h
zhjwpku: We had some discussion that std::expected might be a better solution[1], we haven't decide whether we should adopt C++23 or write a backport version of std::expected. But no matter which way, we need this Status data structure.

[1] https://github.com/apache/iceberg-cpp/issues/14#issuecomment-2581572706

src/common/CMakeLists.txt
zhjwpku: not a problem any more.

src/iceberg/status.h
```diff
@@ -0,0 +1,278 @@
...
+
+  /// Return true iff the status indicates success.
+  constexpr bool ok() const { return (state_ == nullptr); }
+
+  /// Return true iff the status indicates an IO-related failure.
+  constexpr bool IsIOError() const { return code() == StatusCode::IOError; }
+  /// Return true iff the status indicates an unimplemented operation.
+  constexpr bool IsNotImplemented() const { return code() == StatusCode::NotImplemented; }
+  /// Return true iff the status indicates an unknown error.
+  constexpr bool IsUnknownError() const { return code() == StatusCode::UnknownError; }
```
raulcd:
```suggestion
  /// Return true if the status indicates success.
  constexpr bool ok() const { return (state_ == nullptr); }

  /// Return true if the status indicates an IO-related failure.
  constexpr bool IsIOError() const { return code() == StatusCode::IOError; }
  /// Return true if the status indicates an unimplemented operation.
  constexpr bool IsNotImplemented() const { return code() == StatusCode::NotImplemented; }
  /// Return true if the status indicates an unknown error.
  constexpr bool IsUnknownError() const { return code() == StatusCode::UnknownError; }
```

zhjwpku:
This is obsoleted by #40, though we haven't decided to adopt expected or not, this PR can be closed. Thanks for all the inputs.

PR #11, opened 2024-12-13
Add Dependabot for GitHub Actions

Fokko:
I love dependabot <3

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,3 @@
-#
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,4 +22,5 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    open-pull-requests-limit: 10
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,26 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
+#
```
zhjwpku: It seems we don't have this leading `#` and the trailing `#` in other license blocks that use `#` as comment mark, should we keep it consistent?

Other than that, LGTM.

raulcd (approved):
LGTM, on some other projects we override (and increase) the PR limit to be 10 instead of the default 5:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit
Do we want to do so?

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
wgtmac, line 21: For now it keeps an eye on the Github actions for update?

Fokko: Exactly 👍

.github/dependabot.yml
Fokko: Good one, let's remove it 👍 
```suggestion
```

.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+    directory: "/"
+    schedule:
+      interval: "daily"
+
```
Fokko, line 26:
```suggestion
    open-pull-requests-limit: 10

```

zhjwpku (approved):

raulcd (approved):
Thanks @Fokko

nastra (approved):

Issue #12, opened 2024-12-14
[Discussion] googletest(gtest) or Catch2


zhjwpku:
Both can be easily integrated into our CMake build. As for the popularity, googletest rank 1 (33%) according to the 2023 JetBrains C++ ecosystem survey [1] while Catch2 rank 2 (12%).

Catch2 also provides basic micro-benchmarking features, and simple BDD macros, but I'm not sure do we need these features in our project.

I have no previous experience using Catch2, so I want to hear from the community, please comment which UT framework do you think we should use, and the reason why you prefer it.

[1] https://www.jetbrains.com/lp/devecosystem-2023/cpp/#cpp_unittesting_two_years

pitrou:
As I wrote on your PR, my main concern with GTest is compilation times, but I don't know if Catch2 would improve things on this axis.

One question I have about Catch2 is whether it's possible to have rich error reports, for example I would like `REQUIRE(lhs == rhs)` to display the values of `lhs` and `rhs` if the equality fails.

lidavidm:
Catch2 does have rich errors, but I'm rather rattled by how it implements it:

![image](https://github.com/user-attachments/assets/f4090d7c-28bf-4286-a9ae-feebdee28a1d)

> The individual test assertions are written using the `REQUIRE` macro. It accepts a boolean expression, and uses expression templates to internally decompose it, so that it can be individually stringified on test failure.

gaborkaszab:
No strong opinions from me. I myself gravitate towards gtest just because I have experience on that, but could live with any of them.

zhjwpku:
> No strong opinions from me. I myself gravitate towards gtest just because I have experience on that, but could live with any of them.

Yeah, both test frameworks seems capable of meeting our project's needs. However, it would be most beneficial for us to choose the one that the majority of our development team is already familiar with.

I have no experience on Catch2, so I give my vote to googletest.

wgtmac:
I don't have experience with Catch2 either. But I don't object to adopt it if it brings good stuff.

wgtmac:
Should we move forward? I don't see any objection to googletest and https://github.com/apache/iceberg-cpp/pull/13 looks good now.

cc @Fokko @Xuanwo

zhjwpku:
[...]

PR #13, opened 2024-12-15
Integrate Googletest as Test Framework

zhjwpku:
In the discussion #12 regarding the choice between Googletest and Catch2 for our testing framework, most of the developers have experience with Googletest, and while they are open to considering Catch2, they see no immediate advantage, as no one on the team has experience with it. Given this context, the decision to adopt Googletest was made based on the familiarity and comfort level of the majority of the team.

We can switch to Catch2 with one shot if it proves to be a better fit for our needs in the future.


pitrou:
Thanks @zhjwpku !

What I would be interested to know is the compile and link times, respectively, for both GTest and Catch2.
My experience in Arrow is that GTest's main header `gtest.h` imposes a large compilation overhead when you have many independent test modules, and I'm curious if Catch2 improves the situation.

zhjwpku:
> Thanks @zhjwpku !
> 
> What I would be interested to know is the compile and link times, respectively, for both GTest and Catch2. My experience in Arrow is that GTest's main header `gtest.h` imposes a large compilation overhead when you have many independent test modules, and I'm curious if Catch2 improves the situation.

I did 3 tests:

1. using -E to see the preprocessed file size
   - 4301941   -- gtest
   - 4086315  -- catch2
2. using `time` to see the compile time
   - 0.40s user 0.06s system 78% cpu 0.587 total   -- gtest
   - 0.32s user 0.06s system 76% cpu 0.494 total   -- catch2
 3. using `time` to see the link time
   - 0.04s user 0.02s system 79% cpu 0.080 total   -- gtest
   - 0.09s user 0.03s system 98% cpu 0.120 total   -- catch2

As the results shows, catch2 is better at preprocessed file size and compile time, but gtest is better when it comes to link time. I don't think the above tests is adequate to prove which is better than the other, since the UT in this PR is fairly simple.

I think we should focus closely on the specific benefits that the test framework is expected to deliver for this project.

pitrou:
Thanks for the number @zhjwpku .

> As the results shows, catch2 is better at preprocessed file size and compile time, but gtest is better when it comes to link time. I don't think the above tests is adequate to prove which is better than the other, since the UT in this PR is fairly simple.

Well, it does tell us that they are close in terms of inclusion overhead, which was my concern.

> I think we should focus closely on the specific benefits that the test framework is expected to deliver for this project.

Agreed. Is there a detailed (hopefully unbiased) comparison somewhere?

pitrou:
> I found this link: https://yurigeronimus.medium.com/guide-for-choosing-a-test-framework-for-your-c-project-2a7741b53317

Seems like a lot of fluff with no substance, unfortunately.

cmake_modules/FindGTestAlt.cmake
```diff
@@ -0,0 +1,28 @@
...
+find_package(GTest)
+
+if(GTest_FOUND)
+  return()
+endif()
+
+fetchcontent_declare(googletest
+                     GIT_REPOSITORY https://github.com/google/googletest.git
+                     GIT_TAG b514bdc898e2951020cbdca1304b75f5950d1f59 # release-1.15.2
+)
```
wgtmac:
```suggestion
fetchcontent_declare(googletest
                     GIT_REPOSITORY https://github.com/google/googletest.git
                     GIT_TAG b514bdc898e2951020cbdca1304b75f5950d1f59 # release-1.15.2
                     FIND_PACKAGE_ARGS GTest
)
```
Could you please try this?

zhjwpku: committed with changing 'FIND_PACKAGE_ARGS GTest' to 'FIND_PACKAGE_ARGS NAMES GTest', there is a demo in cmake's document exactly our case [0]

[1] https://cmake.org/cmake/help/latest/module/FetchContent.html#integrating-with-find-package

README.md
```diff
@@ -28,13 +28,14 @@ C++ implementation of [Apache Iceberg™](https://iceberg.apache.org/).
...
 
-### Build and Install Core Libraries
+### Build, Run Test and Install Core Libraries
 
 ```bash
 cd iceberg-cpp
 mkdir build && cd build
 cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/iceberg -DICEBERG_BUILD_STATIC=ON -DICEBERG_BUILD_SHARED=ON
 cmake --build .
+ctest --build .
```
wgtmac: Is better to use `ctest --output-on-failure`?

test/CMakeLists.txt
```diff
@@ -14,3 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+include(FindGTestAlt)
```
wgtmac: Do we actually need a separate FindGTestAlt file?

zhjwpku: I have no strong opinion here, I just thought gathering all the thirdparties' cmake file in cmake_modules might be easy to manage. I admit this is a simple one which could be insert directly here, so will change to that.

wgtmac (approved):
LGTM. Thanks!

We can always improve it when we have better ideas.

zhjwpku:
@raulcd @Fokko can you please review this PR?

wgtmac:
BTW, should we run ctest in build_iceberg.sh (or a separate script) from pre-commit check in the CI?

zhjwpku:
> BTW, should we run ctest in build_iceberg.sh (or a separate script) from pre-commit check in the CI?

I've added the ctest command in build_iceberg.sh, this will run test on all platforms we are targeting.

ci/scripts/build_iceberg.sh
```diff
@@ -31,6 +31,7 @@ cmake \
     -DICEBERG_BUILD_SHARED=ON \
     ${source_dir}
 cmake --build . --target install
+ctest --output-on-failure -C Debug
```
raulcd, line 34: Maybe we want to split testing into a different step on CI but this can be done afterwards

raulcd (approved):
Thanks for the work @zhjwpku ! :+1:

wgtmac:
@zhjwpku It is better to change the PR title to reflect that we have chosen googletest?

zhjwpku:
> @zhjwpku It is better to change the PR title to reflect that we have chosen googletest?

Updated the PR title as well as the PR's first message, thanks for the suggestion ;)

Xuanwo (approved):
Thank you @zhjwpku for working on this, and thank you all for the review. Let's move.

PR #16, opened 2024-12-17
Add pre-commit config

zhjwpku:


Fokko (approved):
Love it!

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,31 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [main]
```
raulcd: I would probably allow this to run on any branch (except dependabot) not only on main. For example we might have branches to track minor / patch releases where we still want to run those jobs in the future.
```suggestion
    branches:
      - '**'
      - '!dependabot/**'
```

raulcd (approved):
just a minor nit but LGTM. Thanks for the work @zhjwpku

zhjwpku:
@Fokko @raulcd there is startup error, do you have any experience on this?
https://github.com/apache/iceberg-cpp/actions/runs/12374293318

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,33 @@
...
+      - '**'
+      - '!dependabot/**'
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+    - uses: pre-commit/action@v3.0.1
```
zhjwpku, line 33: @Fokko @raulcd I'm considering change to the following code snip, but I want to know if you have any better idea.

```suggestion
    - name: Install pre-commit
      run: |
        python -m pip install pre-commit
    - name: Run pre-commit
      run: |
        pre-commit run --all-files --color=always --show-diff-on-failure
```

raulcd:
> @Fokko @raulcd there is startup error, do you have any experience on this? https://github.com/apache/iceberg-cpp/actions/runs/12374293318

Oh, it seems pre-commit is not an allowed GH action to use, yes the snippet you shared looks good to me. That's more or less how we do it in Arrow: https://github.com/apache/arrow/blob/main/.github/workflows/dev.yml#L56-L67

zhjwpku:
> > @Fokko @raulcd there is startup error, do you have any experience on this? https://github.com/apache/iceberg-cpp/actions/runs/12374293318
> 
> Oh, it seems pre-commit is not an allowed GH action to use, yes the snippet you shared looks good to me. That's more or less how we do it in Arrow: https://github.com/apache/arrow/blob/main/.github/workflows/dev.yml#L56-L67

Yeah, I just copied that from Arrow with some reduction.

> _edit_: On another note we could ask infra to whitelist the action but that might take some time.

Do you mean apache infra or github infra? I think we can go the Arrow way first and change to pre-commit action after infra whitelist it.

raulcd:
>  you mean apache infra or github infra?

Apache Infra

Fokko:
Cool, I've raised an issue: https://issues.apache.org/jira/browse/INFRA-26378

zhjwpku:
> > Cool, I've raised an issue: https://issues.apache.org/jira/browse/INFRA-26378
> 
> It got approved :)

Still got a startup error: https://github.com/apache/iceberg-cpp/actions/runs/12384969557, maybe we should just wait for some time ;(

zhjwpku:
> Should we also add some instructions on how to use `pre-commit`? Similar to https://py.iceberg.apache.org/contributing/#linting

I've added some linting instructions to Contribute section of README.md, please check if that is adequate.

wgtmac (approved):

zhjwpku:
> > Cool, I've raised an issue: https://issues.apache.org/jira/browse/INFRA-26378
> 
> It got approved :)

Hi @Fokko, I see the comment that the pre-commit/action@3.0.1 has been whitelisted, do i need to take some action so that this PR can be merged into main?

Fokko:
@zhjwpku no action needed, let me get this in

wgtmac:
> pre-commit/action@v3.0.1 is not allowed to be used in apache/iceberg-cpp.

It seems that pre-commit check never run successfully: https://github.com/apache/iceberg-cpp/actions/workflows/pre-commit.yml. I saw the same thing with cpp-linter-action: https://issues.apache.org/jira/browse/INFRA-26318

zhjwpku:
> > pre-commit/action@v3.0.1 is not allowed to be used in apache/iceberg-cpp.
> 
> It seems that pre-commit check never run successfully: https://github.com/apache/iceberg-cpp/actions/workflows/pre-commit.yml. I saw the same thing with cpp-linter-action: https://issues.apache.org/jira/browse/INFRA-26318

Yeah, I see that too, not sure how the infra works, any chance @Fokko take a look at this?

lidavidm:
Apache allow-lists GitHub actions to use. I think for arrow-adbc I've just manually installed and run pre-commit instead of using the dedicated action.

zhjwpku:
> Apache allow-lists GitHub actions to use. I think for arrow-adbc I've just manually installed and run pre-commit instead of using the dedicated action.

But https://issues.apache.org/jira/browse/INFRA-26378 seems whitelists the pre-commit action. Anyway, if we find that doesn't work in the end, I will file a PR changing to the arrow way.

wgtmac:
I have replied to the JIRA ticket. I guess it might have some misconfigurations.

Fokko:
Looks like an issue on the INFRA end? For PyIceberg we also [just install pre-commit](https://github.com/apache/iceberg-python/blob/main/.github/workflows/python-ci.yml) and this works just as well

wgtmac:
Yes, they have confirmed that there was a typo. Now it works: https://github.com/apache/iceberg-cpp/actions/runs/12512760344

PR #23, opened 2025-01-10
Remove .DS_Store and add it into gitignore

ConeyLiu:
A small update to remove the .DS_Store from repo.


ConeyLiu:
cc @wgtmac @Xuanwo

.gitignore
```diff
@@ -16,3 +16,9 @@
 # under the License.
 
 build/
+cmake-build-debug/
```
wgtmac, line 19: Should we also add `cmake-build-release` as well?

ConeyLiu: added

Xuanwo (approved):
Thanks!

PR #24, opened 2025-01-10
Update cmake instructions in README

zuyu:
Build `Core Libraries` w/o `Arrow`, as there is another section about building `Iceberg Arrow Library` explicitly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 # under the License.
 
 build/
+install/
 cmake-build-debug/
 cmake-build-release/
 .DS_Store
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ C++ implementation of [Apache Iceberg™](https://iceberg.apache.org/).
 
 ```bash
 cd iceberg-cpp
-cmake -S . -B build -DCMAKE_INSTALL_PREFIX=/path/to/install -DICEBERG_BUILD_STATIC=ON -DICEBERG_BUILD_SHARED=ON
+cmake -S . -B build -DCMAKE_INSTALL_PREFIX=/path/to/install -DICEBERG_BUILD_STATIC=ON -DICEBERG_BUILD_SHARED=ON -DICEBERG_BUILD_TESTS=OFF -DICEBERG_ARROW=OFF
 cmake --build build
 cmake --install build
 ```


.gitignore
```diff
@@ -16,6 +16,7 @@
 # under the License.
 
 build/
+install/
```
zhjwpku, line 19: Why this? Is this generated by some IDE?

README.md
```diff
@@ -32,7 +32,7 @@ C++ implementation of [Apache Iceberg™](https://iceberg.apache.org/).
 
 ```bash
 cd iceberg-cpp
-cmake -S . -B build -DCMAKE_INSTALL_PREFIX=/path/to/install -DICEBERG_BUILD_STATIC=ON -DICEBERG_BUILD_SHARED=ON
+cmake -S . -B build -DCMAKE_INSTALL_PREFIX=/path/to/install -DICEBERG_BUILD_STATIC=ON -DICEBERG_BUILD_SHARED=ON -DICEBERG_BUILD_TESTS=OFF -DICEBERG_ARROW=OFF
```
zhjwpku, line 35: Maybe these options should be documented somewhere but not specified in the  same example command?

wgtmac: Agreed. An official site for iceberg-cpp is more suitable for these documentations.

lidavidm: Should we open an issue to choose a documentation/site generator?

wgtmac: It would be good if someone with expertise could lead this :)

lidavidm: https://github.com/apache/iceberg-cpp/issues/26

.gitignore
zuyu: No, since we specify `CMAKE_INSTALL_PREFIX` in cmake, I think someone might specify `CMAKE_INSTALL_PREFIX=install` in the repo directory as well.

README.md
zuyu: This PR is to update cmake to build Core Libraries, and I think we don't need to build `Arrow` for the core libraries, right?

.gitignore
zhjwpku: > No, since we specify `CMAKE_INSTALL_PREFIX` in cmake, I think someone might specify `CMAKE_INSTALL_PREFIX=install` in the repo directory as well.

I doubt anyone would do this, if you really need to install in src tree, you can specify `CMAKE_INSTALL_PREFIX=/build/install`.

PR #31, opened 2025-01-16
Add data type/schema field/schema

lidavidm:
Implement the API for data types and schemas.

### DIFF
--- a/src/iceberg/schema.cc
+++ b/src/iceberg/schema.cc
@@ -22,7 +22,7 @@
 #include <format>
 
 #include "iceberg/type.h"
-#include "iceberg/util/formatter.h"
+#include "iceberg/util/formatter.h"  // IWYU pragma: keep
 
 namespace iceberg {
 
--- a/src/iceberg/schema.cc
+++ b/src/iceberg/schema.cc
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/schema.h"
+
+#include <format>
+
+#include "iceberg/type.h"
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+Schema::Schema(int32_t schema_id, std::vector<SchemaField> fields)
+    : StructType(std::move(fields)), schema_id_(schema_id) {}
+
+int32_t Schema::schema_id() const { return schema_id_; }
+
+std::string Schema::ToString() const {
+  std::string repr = "schema<";
+  for (const auto& field : fields_) {
+    std::format_to(std::back_inserter(repr), "  {}\n", field);
+  }
+  repr += ">";
+  return repr;
+}
+
+bool Schema::Equals(const Schema& other) const {
+  return schema_id_ == other.schema_id_ && fields_ == other.fields_;
+}
+
+}  // namespace iceberg
--- a/src/iceberg/schema.h
+++ b/src/iceberg/schema.h
@@ -44,7 +44,7 @@ class ICEBERG_EXPORT Schema : public StructType {
 
   /// \brief Get the schema ID.
   ///
-  /// Schemas are identified by a unique ID for the purposes of schema
+  /// A schema is identified by a unique ID for the purposes of schema
   /// evolution.
   [[nodiscard]] int32_t schema_id() const;
 
--- a/src/iceberg/schema.h
+++ b/src/iceberg/schema.h
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/schema.h
+/// Schemas for Iceberg tables.  This header contains the definition of Schema
+/// and any utility functions.  See iceberg/type.h and iceberg/field.h as well.
+
+#include <cstdint>
+#include <string>
+#include <vector>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/schema_field.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+/// \brief A schema for a Table.
+///
+/// A schema is a list of typed columns, along with a unique integer ID.  A
+/// Table may have different schemas over its lifetime due to schema
+/// evolution.
+class ICEBERG_EXPORT Schema : public StructType {
+ public:
+  Schema(int32_t schema_id, std::vector<SchemaField> fields);
+
+  /// \brief Get the schema ID.
+  ///
+  /// Schemas are identified by a unique ID for the purposes of schema
+  /// evolution.
+  [[nodiscard]] int32_t schema_id() const;
+
+  [[nodiscard]] std::string ToString() const;
+
+  friend bool operator==(const Schema& lhs, const Schema& rhs) { return lhs.Equals(rhs); }
+
+  friend bool operator!=(const Schema& lhs, const Schema& rhs) { return !(lhs == rhs); }
+
+ private:
+  /// \brief Compare two schemas for equality.
+  [[nodiscard]] bool Equals(const Schema& other) const;
+
+  const int32_t schema_id_;
+};
+
+}  // namespace iceberg
--- a/src/iceberg/schema_field.cc
+++ b/src/iceberg/schema_field.cc
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/schema_field.h"
+
+#include <format>
+
+#include "iceberg/type.h"
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+SchemaField::SchemaField(int32_t field_id, std::string name, std::shared_ptr<Type> type,
+                         bool optional)
+    : field_id_(field_id),
+      name_(std::move(name)),
+      type_(std::move(type)),
+      optional_(optional) {}
+
+SchemaField SchemaField::MakeOptional(int32_t field_id, std::string name,
+                                      std::shared_ptr<Type> type) {
+  return SchemaField(field_id, std::move(name), std::move(type), true);
+}
+
+SchemaField SchemaField::MakeRequired(int32_t field_id, std::string name,
+                                      std::shared_ptr<Type> type) {
+  return SchemaField(field_id, std::move(name), std::move(type), false);
+}
+
+int32_t SchemaField::field_id() const { return field_id_; }
+
+std::string_view SchemaField::name() const { return name_; }
+
+const std::shared_ptr<Type>& SchemaField::type() const { return type_; }
+
+bool SchemaField::optional() const { return optional_; }
+
+std::string SchemaField::ToString() const {
+  return std::format("{} ({}): {}{}", name_, field_id_, *type_,
+                     optional_ ? "" : " (required)");
+}
+
+bool SchemaField::Equals(const SchemaField& other) const {
+  return field_id_ == other.field_id_ && name_ == other.name_ && *type_ == *other.type_ &&
+         optional_ == other.optional_;
+}
+
+}  // namespace iceberg
--- a/src/iceberg/schema_field.h
+++ b/src/iceberg/schema_field.h
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/schema_field.h
+/// A (schema) field is a name and a type and is part of a schema or nested
+/// type (e.g. a struct).
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <string_view>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/type_fwd.h"
+#include "iceberg/util/formattable.h"
+
+namespace iceberg {
+
+/// \brief A type combined with a name.
+class ICEBERG_EXPORT SchemaField : public iceberg::util::Formattable {
+ public:
+  /// \brief Construct a field.
+  /// \param[in] field_id The field ID.
+  /// \param[in] name The field name.
+  /// \param[in] type The field type.
+  /// \param[in] optional Whether values of this field are required or nullable.
+  SchemaField(int32_t field_id, std::string name, std::shared_ptr<Type> type,
+              bool optional);
+
+  /// \brief Construct an optional (nullable) field.
+  static SchemaField MakeOptional(int32_t field_id, std::string name,
+                                  std::shared_ptr<Type> type);
+  /// \brief Construct a required (non-null) field.
+  static SchemaField MakeRequired(int32_t field_id, std::string name,
+                                  std::shared_ptr<Type> type);
+
+  /// \brief Get the field ID.
+  [[nodiscard]] int32_t field_id() const;
+
+  /// \brief Get the field name.
+  [[nodiscard]] std::string_view name() const;
+
+  /// \brief Get the field type.
+  [[nodiscard]] const std::shared_ptr<Type>& type() const;
+
+  /// \brief Get whether the field is optional.
+  [[nodiscard]] bool optional() const;
+
+  [[nodiscard]] std::string ToString() const;
+
+  friend bool operator==(const SchemaField& lhs, const SchemaField& rhs) {
+    return lhs.Equals(rhs);
+  }
+
+  friend bool operator!=(const SchemaField& lhs, const SchemaField& rhs) {
+    return !(lhs == rhs);
+  }
+
+ private:
+  /// \brief Compare two fields for equality.
+  [[nodiscard]] bool Equals(const SchemaField& other) const;
+
+  int32_t field_id_;
+  std::string name_;
+  std::shared_ptr<Type> type_;
+  bool optional_;
+};
+
+}  // namespace iceberg
--- a/src/iceberg/type.cc
+++ b/src/iceberg/type.cc
@@ -216,7 +216,7 @@ std::string MapType::ToString() const {
   std::string repr = "map<";
 
   std::format_to(std::back_inserter(repr), "{}: {}", key(), value());
-  repr += "}";
+  repr += ">";
   return repr;
 }
 std::span<const SchemaField> MapType::fields() const { return fields_; }
--- a/src/iceberg/type.cc
+++ b/src/iceberg/type.cc
@@ -1,0 +1,297 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/type.h"
+
+#include <format>
+#include <iterator>
+#include <stdexcept>
+
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+TypeId BooleanType::type_id() const { return TypeId::kBoolean; }
+std::string BooleanType::ToString() const { return "boolean"; }
+bool BooleanType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kBoolean;
+}
+
+TypeId Int32Type::type_id() const { return TypeId::kInt32; }
+std::string Int32Type::ToString() const { return "int32"; }
+bool Int32Type::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kInt32;
+}
+
+TypeId Int64Type::type_id() const { return TypeId::kInt64; }
+std::string Int64Type::ToString() const { return "int64"; }
+bool Int64Type::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kInt64;
+}
+
+TypeId Float32Type::type_id() const { return TypeId::kFloat32; }
+std::string Float32Type::ToString() const { return "float32"; }
+bool Float32Type::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kFloat32;
+}
+
+TypeId Float64Type::type_id() const { return TypeId::kFloat64; }
+std::string Float64Type::ToString() const { return "float64"; }
+bool Float64Type::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kFloat64;
+}
+
+DecimalType::DecimalType(int32_t precision, int32_t scale)
+    : precision_(precision), scale_(scale) {
+  if (precision < 0 || precision > kMaxPrecision) {
+    throw std::runtime_error(
+        std::format("DecimalType: precision must be in [0, 38], was {}", precision));
+  }
+}
+
+int32_t DecimalType::precision() const { return precision_; }
+int32_t DecimalType::scale() const { return scale_; }
+TypeId DecimalType::type_id() const { return TypeId::kDecimal; }
+std::string DecimalType::ToString() const {
+  return std::format("decimal({}, {})", precision_, scale_);
+}
+bool DecimalType::Equals(const Type& other) const {
+  if (other.type_id() != TypeId::kDecimal) {
+    return false;
+  }
+  const auto& decimal = static_cast<const DecimalType&>(other);
+  return precision_ == decimal.precision_ && scale_ == decimal.scale_;
+}
+
+TypeId TimeType::type_id() const { return TypeId::kTime; }
+std::string TimeType::ToString() const { return "time"; }
+bool TimeType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kTime;
+}
+
+TypeId DateType::type_id() const { return TypeId::kDate; }
+std::string DateType::ToString() const { return "date"; }
+bool DateType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kDate;
+}
+
+TypeId TimestampType::type_id() const { return TypeId::kTimestamp; }
+std::string TimestampType::ToString() const { return "timestamp"; }
+bool TimestampType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kTimestamp;
+}
+
+TypeId TimestampTzType::type_id() const { return TypeId::kTimestampTz; }
+std::string TimestampTzType::ToString() const { return "timestamptz"; }
+bool TimestampTzType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kTimestampTz;
+}
+
+TypeId BinaryType::type_id() const { return TypeId::kBinary; }
+std::string BinaryType::ToString() const { return "binary"; }
+bool BinaryType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kBinary;
+}
+
+TypeId StringType::type_id() const { return TypeId::kString; }
+std::string StringType::ToString() const { return "string"; }
+bool StringType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kString;
+}
+
+FixedType::FixedType(int32_t length) : length_(length) {
+  if (length < 0) {
+    throw std::runtime_error(
+        std::format("FixedType: length must be >= 0, was {}", length));
+  }
+}
+
+int32_t FixedType::length() const { return length_; }
+TypeId FixedType::type_id() const { return TypeId::kFixed; }
+std::string FixedType::ToString() const { return std::format("fixed({})", length_); }
+bool FixedType::Equals(const Type& other) const {
+  if (other.type_id() != TypeId::kFixed) {
+    return false;
+  }
+  const auto& fixed = static_cast<const FixedType&>(other);
+  return length_ == fixed.length_;
+}
+
+TypeId UuidType::type_id() const { return TypeId::kUuid; }
+std::string UuidType::ToString() const { return "uuid"; }
+bool UuidType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kUuid;
+}
+
+ListType::ListType(SchemaField element) : element_(std::move(element)) {
+  if (element_.name() != kElementName) {
+    throw std::runtime_error(
+        std::format("ListType: child field name should be '{}', was '{}'", kElementName,
+                    element_.name()));
+  }
+}
+
+ListType::ListType(int32_t field_id, std::shared_ptr<Type> type, bool optional)
+    : element_(field_id, std::string(kElementName), std::move(type), optional) {}
+
+TypeId ListType::type_id() const { return TypeId::kList; }
+std::string ListType::ToString() const { return std::format("list<{}>", element_); }
+std::span<const SchemaField> ListType::fields() const { return {&element_, 1}; }
+std::optional<std::reference_wrapper<const SchemaField>> ListType::GetFieldById(
+    int32_t field_id) const {
+  if (field_id == element_.field_id()) {
+    return std::cref(element_);
+  }
+  return std::nullopt;
+}
+std::optional<std::reference_wrapper<const SchemaField>> ListType::GetFieldByIndex(
+    int index) const {
+  if (index == 0) {
+    return std::cref(element_);
+  }
+  return std::nullopt;
+}
+std::optional<std::reference_wrapper<const SchemaField>> ListType::GetFieldByName(
+    std::string_view name) const {
+  if (name == element_.name()) {
+    return std::cref(element_);
+  }
+  return std::nullopt;
+}
+bool ListType::Equals(const Type& other) const {
+  if (other.type_id() != TypeId::kList) {
+    return false;
+  }
+  const auto& list = static_cast<const ListType&>(other);
+  return element_ == list.element_;
+}
+
+MapType::MapType(SchemaField key, SchemaField value)
+    : fields_{std::move(key), std::move(value)} {
+  if (this->key().name() != kKeyName) {
+    throw std::runtime_error(
+        std::format("MapType: key field name should be '{}', was '{}'", kKeyName,
+                    this->key().name()));
+  }
+  if (this->value().name() != kValueName) {
+    throw std::runtime_error(
+        std::format("MapType: value field name should be '{}', was '{}'", kValueName,
+                    this->value().name()));
+  }
+}
+
+const SchemaField& MapType::key() const { return fields_[0]; }
+const SchemaField& MapType::value() const { return fields_[1]; }
+TypeId MapType::type_id() const { return TypeId::kMap; }
+std::string MapType::ToString() const {
+  return std::format("map<{}: {}>", key(), value());
+}
+std::span<const SchemaField> MapType::fields() const { return fields_; }
+std::optional<std::reference_wrapper<const SchemaField>> MapType::GetFieldById(
+    int32_t field_id) const {
+  if (field_id == key().field_id()) {
+    return key();
+  } else if (field_id == value().field_id()) {
+    return value();
+  }
+  return std::nullopt;
+}
+std::optional<std::reference_wrapper<const SchemaField>> MapType::GetFieldByIndex(
+    int index) const {
+  if (index == 0) {
+    return key();
+  } else if (index == 0) {
+    return value();
+  }
+  return std::nullopt;
+}
+std::optional<std::reference_wrapper<const SchemaField>> MapType::GetFieldByName(
+    std::string_view name) const {
+  if (name == kKeyName) {
+    return key();
+  } else if (name == kValueName) {
+    return value();
+  }
+  return std::nullopt;
+}
+bool MapType::Equals(const Type& other) const {
+  if (other.type_id() != TypeId::kMap) {
+    return false;
+  }
+  const auto& map = static_cast<const MapType&>(other);
+  return fields_ == map.fields_;
+}
+
+StructType::StructType(std::vector<SchemaField> fields) : fields_(std::move(fields)) {
+  size_t index = 0;
+  for (const auto& field : fields_) {
+    auto [it, inserted] = field_id_to_index_.try_emplace(field.field_id(), index);
+    if (!inserted) {
+      throw std::runtime_error(
+          std::format("StructType: duplicate field ID {} (field indices {} and {})",
+                      field.field_id(), it->second, index));
+    }
+
+    index++;
+  }
+}
+
+TypeId StructType::type_id() const { return TypeId::kStruct; }
+std::string StructType::ToString() const {
+  std::string repr = "struct<\n";
+  for (const auto& field : fields_) {
+    std::format_to(std::back_inserter(repr), "  {}\n", field);
+  }
+  repr += ">";
+  return repr;
+}
+std::span<const SchemaField> StructType::fields() const { return fields_; }
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldById(
+    int32_t field_id) const {
+  auto it = field_id_to_index_.find(field_id);
+  if (it == field_id_to_index_.end()) return std::nullopt;
+  return fields_[it->second];
+}
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByIndex(
+    int index) const {
+  if (index < 0 || index >= static_cast<int>(fields_.size())) {
+    return std::nullopt;
+  }
+  return fields_[index];
+}
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByName(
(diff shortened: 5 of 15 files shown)


lidavidm:
Note: we could simplify the data type definition greatly if we adopt something like Arrow Java/[cuDF](https://docs.rapids.ai/api/libcudf/stable/classcudf_1_1data__type). Type would be a POD containing type ID and fields for all parameterized types (decimal scale, fixed length, etc.). Field would contain a Type and a (possibly empty/invalid) list of nested fields (to handle things like Struct). This loses type safety but removes a layer of indirection/allocations.

lidavidm:
For here I've chosen something somewhat resembling Arrow C++. However I think Schema and Field (unlike Arrow) should not always be wrapped in smart pointers. Type is still in a smart pointer for type erasure.

src/iceberg/type_fwd.h
```diff
@@ -0,0 +1,55 @@
...
+/// \file iceberg/type_fwd.h
+/// Forward declarations and enum definitions.
+
+namespace iceberg {
+
+/// \brief A data type.
+///
+/// This is not a complete data type by itself because some types are nested
+/// and/or parameterized.
+enum class TypeId {
```
lidavidm, line 35: This needs the Iceberg V3 types added as well.

src/iceberg/schema.h
```diff
@@ -0,0 +1,60 @@
...
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+/// \brief A schema for a Table.
+///
+/// A schema is a list of typed columns, along with a unique integer ID.  A
+/// Table may have different schemas over its lifetime due to schema
+/// evolution.
+class ICEBERG_EXPORT Schema : public StructType {
```
lidavidm, line 41: In iceberg-arrow we can add functions from iceberg::Schema to arrow::Schema.

Fokko: image

You can take inspiration from PyIceberg:

- [`schema_to_pyarrow`](https://github.com/apache/iceberg-python/blob/f4caa3ac927c626eeba5d0408f80ddd9b95214e0/pyiceberg/io/pyarrow.py#L557)
- [`pyarrow_to_schema`](https://github.com/apache/iceberg-python/blob/f4caa3ac927c626eeba5d0408f80ddd9b95214e0/pyiceberg/io/pyarrow.py#L905C5-L905C22)

src/iceberg/type.h
```diff
@@ -0,0 +1,120 @@
...
+ public:
+  explicit StructType(std::vector<Field> fields);
+
+  StructType() = default;
+  ~StructType() = default;
+
+  [[nodiscard]] std::span<Field> fields() const;
+
+  const Field& field(int i) const;
+  const Field& field(std::string_view name) const;
```
Fokko: I would add case-sensitive here as well

src/iceberg/type_fwd.h
Fokko: I think we can go two ways;

- Leave as is, and first go with everything in V2 since V3 is still not finalized (eg Variant support is still pending in Java, Geometry/Geography is still under discussion).
- Add everything, and fail when a user tries to read a table with an unsupported type.

I think I'll prefer the first one to make sure that we don't expand too much on the surface area.

src/iceberg/type.h
```diff
@@ -0,0 +1,120 @@
...
+  /// \brief Compare two fields for equality.
+  [[nodiscard]] bool Equals(const Field& other) const;
+
+  friend bool operator==(const Field& lhs, const Field& rhs) { return lhs.Equals(rhs); }
+
+  friend bool operator!=(const Field& lhs, const Field& rhs) { return !(lhs == rhs); }
+};
+
+/// \brief A data type representing a boolean.
+class ICEBERG_EXPORT BooleanType : public Type {
```
Fokko: In [PyIceberg](https://github.com/apache/iceberg-python/blob/f4caa3ac927c626eeba5d0408f80ddd9b95214e0/pyiceberg/types.py#L179)/[Java](https://github.com/apache/iceberg/blob/d96901b843395fe669f6bd4f618f8e5e46c0eed4/api/src/main/java/org/apache/iceberg/types/Type.java#L104) we add a `PrimitiveType` to the OOP hierarchy. This makes it a bit easier down the road, maybe also something to consider here.

Fokko (commented):
Looking good, left a few comments

src/iceberg/type_fwd.h
wgtmac: +1 for the 1st approach. Supporting v2 is already a big milestone.

src/iceberg/type.h
wgtmac: I believe this follows the Arrow style where trivial accessor functions are lower-cased.

src/iceberg/type.h
```diff
@@ -0,0 +1,120 @@
...
+#include <string>
+#include <vector>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Interface for a data type for a field.
+class ICEBERG_EXPORT Type {
```
wgtmac: Should we add default value or later?

src/iceberg/schema.h
wgtmac: Do we need to add `KeyValueMetadata` to Schema and Field, though the Iceberg spec does not have it yet.

src/iceberg/type.h
lidavidm: I could change this to be FieldByIndex/FieldByName perhaps

src/iceberg/type.h
lidavidm: I don't see that in the Java API. We could add it as it becomes useful?

src/iceberg/schema.h
lidavidm: I'd rather not add it if it's not in the spec.

lidavidm: (Also if Iceberg is adding it, consider the issues that Arrow has: Arrow currently doesn't allow binary metadata values due to an oversight.)

src/iceberg/type.h
lidavidm: Changed to GetFieldBy{Index,Name}

lidavidm:
I added Primitive/NestedType; I'll start filling out definitions and the rest of the types next, along with adding the visitors to iceberg-arrow.

src/iceberg/schema.h
```diff
@@ -0,0 +1,60 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/schema.h
```
gaborkaszab, line 22: I'm not familiar with this style of commenting. Could you please explain me why including the file name into a comment needed here?

src/iceberg/schema.h
```diff
@@ -0,0 +1,60 @@
...
+/// evolution.
+class ICEBERG_EXPORT Schema : public StructType {
+ public:
+  Schema(int32_t schema_id, std::vector<Field> fields);
+
+  /// \brief Get the schema ID.
+  ///
+  /// Schemas are identified by a unique ID for the purposes of schema
+  /// evolution.
+  [[nodiscard]] int32_t schema_id() const;
```
gaborkaszab, line 49: The functions within this class may not follow the same naming convention: See schema_id vs ToString, Equals
I found the same mismatch below too, so I'm wondering if it's me missing something here.

Update: I've just checked the style guide, and I guess this is fine: "accessors and mutators may be named like variables"
https://google.github.io/styleguide/cppguide.html#Function_Names

src/iceberg/schema.h
```diff
@@ -0,0 +1,60 @@
...
+
+  /// \brief Get a user-readable string representation of the schema.
+  [[nodiscard]] std::string ToString() const;
+
+  /// \brief Compare two schemas for equality.
+  [[nodiscard]] bool Equals(const Schema& other) const;
```
gaborkaszab: If we have operators for equality, can't we make the Equals() function private? I don't see the reason to expose both.

src/iceberg/type.h
```diff
@@ -0,0 +1,149 @@
...
+
+/// \brief A type combined with a name.
+class ICEBERG_EXPORT Field {
+ public:
+  /// \brief Construct a field.
+  /// \param[in] field_id The field ID.
+  /// \param[in] name The field name.
+  /// \param[in] type The field type.
+  /// \param[in] optional Whether values of this field are required or nullable.
+  Field(int32_t field_id, std::string name, std::shared_ptr<Type> type, bool optional);
```
gaborkaszab: > std::shared_ptr<Type> type
Does this assume that for each type there is going to be an instance and the Fields will share these instances? Would it be too much overhead to keep a copy of a Type for each Field?

src/iceberg/type.h
```diff
@@ -0,0 +1,149 @@
...
+#include <string>
+#include <vector>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief A type combined with a name.
+class ICEBERG_EXPORT Field {
```
gaborkaszab: I think similarly to Schema, I'd put Field into a separate file. Now this is in type.h but Field isn't really a type.

gaborkaszab (commented):
Great to see progress on this area, @lidavidm ! I went through for my own education and left some comments or questions. Let me know what you think!

src/iceberg/schema.h
lidavidm: I'm not done with it yet, this is for Doxygen which will generate a page to document the header.

src/iceberg/schema.h
lidavidm: Google style has trivial accessors using snake_case.

src/iceberg/schema.h
lidavidm: It may be useful to add overloads later (e.g. if we want to ignore metadata or IDs)

src/iceberg/type.h
lidavidm: See the very first two comments here - there is an alternate design.

We need some sort of pointer with the current design to achieve type erasure. Unless you want me to do something with std::variant (I don't mind exploring that actually!)

src/iceberg/schema.h
gaborkaszab: Then we can extend this later. I believe in adding what is required ATM, and if there is a need later, we can still extend.
Also, we are defining a public interface here, should keep it as simple as possible. Exposing the Equals method seems a redundancy for me ATM.

src/iceberg/type.h
lidavidm: There's a bit of a circular dependency between Field (which wants Type) and StructType (which wants Field). Unless (again) you see the alternate design in the first two comments of this PR which puts child Fields always on the Field (this means that Field has a potentially redundant...field, but avoids nesting of types - I personally slightly prefer this but it's conceptually dependent on whether you model the type itself as nested or the field as the nested object)

src/iceberg/schema.h
```diff
@@ -0,0 +1,60 @@
...
+namespace iceberg {
+
+/// \brief A schema for a Table.
+///
+/// A schema is a list of typed columns, along with a unique integer ID.  A
+/// Table may have different schemas over its lifetime due to schema
+/// evolution.
+class ICEBERG_EXPORT Schema : public StructType {
+ public:
+  Schema(int32_t schema_id, std::vector<Field> fields);
```
gaborkaszab: One thing I saw in the Java implementation, is that there is a name to field ID mapping there, and there are accessors getting Field(s) by ID or by name. I think that would be useful here too.

src/iceberg/schema.h
gaborkaszab, line 41: > : public StructType
I don't think that Schema should derive from StructType. I get that it will hold a list of Fields such as StructType, but still, Schema in general is not a Type, would be pretty unnatural to derive from it.

src/iceberg/type.h
gaborkaszab: Sure, will take a look a bit later.
Generally asking, can't we solve the circular dependency with forward declarations and using pointers/references?

lidavidm: We can if we continue to type-erase Type, but not if we want it to not be behind a smart pointer.

src/iceberg/schema.h
lidavidm: (Note that various type systems do actually just represent a schema as a struct type, including nanoarrow/Arrow C Data Interface. At the very least they are isomorphic and you can see methods that convert between them often. That said I don't feel strongly about this.)

wgtmac: I'm also inclined to inherit Schema from StructType. In practice this inheritance avoid an unnecessary conversion from Schema to StructType to implement type visitors.

lidavidm:
Just to make sure @gaborkaszab @wgtmac: are we ok with the Arrow-style type representation here? (Types are represented by a class hierarchy, erased behind smart pointers; nested types store the fields in the type object, not the field object)

The alternatives are:
- cuDF style: type objects are minimal, just a type ID, and there is no hierarchy, just the base `DataType` class which is not type-erased and therefore there is no need for a smart pointer. As a trade-off, nested fields have to be extracted from the field and the base DataType has to have fields for all possible parameterized types
- arrow-java style: there is still a hierarchy of type objects, but child fields are stored on the Field, not the DataType. Avoids a conceptual dependency cycle between DataType and Field
- variant style: like arrow-java, but instead of type erasure and a type hierarchy (or conversely, like cuDF but avoids redundant fields), we have a `std::variant` with all possibilities. Avoids boxing but the object is larger (in e.g. a vector)

lidavidm:
Arrow-style does let you do a bunch of compile-time metaprogramming (e.g. see arrow::TypeTraits)

src/iceberg/type.h
```diff
@@ -0,0 +1,149 @@
...
+  /// \brief Get a user-readable string representation of the type.
+  [[nodiscard]] virtual std::string ToString() const = 0;
+
+  /// \brief Compare two types for equality.
+  [[nodiscard]] virtual bool Equals(const Type& other) const = 0;
+
+  /// \brief Is this a primitive type (may not have child fields)?
+  [[nodiscard]] virtual bool is_primitive() const = 0;
+
+  /// \brief Is this a nested type (may have child fields)?
```
mapleFU, line 52: What raise my interests is that:
1. Does iceberg has a "void" ( or null ) bottom type?
2. Does iceberg allowed to define an empty structure?

lidavidm: There is `unknown`. 

I don't think the spec says anything either way about an empty struct. I didn't check any implementations. If allowed, you could think of it as a `void` type.

src/iceberg/schema.h
gaborkaszab: Just curious: Could visitors work with Schema if we had a Schema::asStructType or such? Doesn't seem a big deal to do that conversion for me and then the class hierarchy would be more intuitive in my opinion. Or I miss something :)

src/iceberg/type.h
gaborkaszab: Well, I think it's been too long for me since I found myself writing Java code and not C++ on a daily basis :) After taking another look, sharep_ptr<Type> seems fine here. Sorry :)

src/iceberg/type.h
gaborkaszab: Putting them behind smart pointers makes sense. See my other comment. And the this could go into a separate file.
Meanwhile I'm sketching some code for the PartitionSpec and PartitionField and I'm wondering if this one here is intuitive to be called Field, or could be better called SchemaField (considering there is going to be a PartitionField too).

src/iceberg/type.h
gaborkaszab: I quickly checked the Java code, and apparently there is nothing preventing us from creating a StructType with zero fields.

gaborkaszab:
> Just to make sure @gaborkaszab @wgtmac: are we ok with the Arrow-style type representation here? (Types are represented by a class hierarchy, erased behind smart pointers; nested types store the fields in the type object, not the field object)

If I get it right, Arrow-style mes sense for me too.

src/iceberg/schema.h
lidavidm: That would also work, it would just add the conversion, and then duplicating all the methods. No big deal either way to me at least.

(I suppose the OOP approach is to add a StructLike interface that both Struct and Schema implement.)

src/iceberg/type.h
lidavidm: I'll rename

lidavidm:
I've implemented the Type subclasses. I've also added a trait for ToString/std::format formatting (might as well use the new C++ features if we're requiring C++20). Although it appears clang is not happy about what I did.

I've elected to make getters fallible (e.g. GetFieldById) via `optional<reference_wrapper<T>>` which looks a bit messy but should be safer than returning `T const&` and throwing, or returning `T const*` and admitting `nullptr`.

lidavidm:
Ok, hopefully that fixes macOS. It appears C++20 support on the version of clang shipped by macOS is still flaky:

https://godbolt.org/z/W9rTexc13

Maybe we want to avoid C++20 (sadly) or maybe we need to encourage people to get a newer clang on macOS. (It appears to be clang 15, whereas the latest is clang 19.) Or maybe it's just an issue with the version the CI environment has.

lidavidm:
I'll add some unit tests as long as the API looks good to everyone.

src/iceberg/schema_field.h
```diff
@@ -0,0 +1,87 @@
...
+
+  /// \brief Construct an optional (nullable) field.
+  static SchemaField MakeOptional(int32_t field_id, std::string name,
+                                  std::shared_ptr<Type> type);
+  /// \brief Construct a required (non-null) field.
+  static SchemaField MakeRequired(int32_t field_id, std::string name,
+                                  std::shared_ptr<Type> type);
+
+  /// \brief Get the field ID.
+  [[nodiscard]] int32_t field_id() const;
```
wgtmac, line 56: Is it an overkill to add `[[nodiscard]]` to these accessors?

src/iceberg/type.h
```diff
@@ -0,0 +1,397 @@
...
+  bool is_nested() const override { return true; }
+
+  /// \brief Get a view of the child fields.
+  [[nodiscard]] virtual std::span<const SchemaField> fields() const = 0;
+  /// \brief Get a field by field ID.
+  [[nodiscard]] virtual std::optional<std::reference_wrapper<const SchemaField>>
+  GetFieldById(int32_t field_id) const = 0;
+  /// \brief Get a field by index.
+  [[nodiscard]] virtual std::optional<std::reference_wrapper<const SchemaField>>
+  GetFieldByIndex(int i) const = 0;
```
wgtmac:
```suggestion
  GetFieldByIndex(int32_t i) const = 0;
```

I'd prefer explicit width

src/iceberg/type.h
```diff
@@ -0,0 +1,397 @@
...
+  std::string ToString() const override;
+
+ protected:
+  bool Equals(const Type& other) const override;
+};
+
+/// \brief A data type representing a fixed-precision decimal.
+class ICEBERG_EXPORT DecimalType : public PrimitiveType {
+ public:
+  constexpr static const int32_t kMaxPrecision = 38;
```
wgtmac, line 258: Is it worth adding enforcement of constant style to `.clang-tidy`? Now it seems that we prefer `kMaxPrecision` over `MAX_PRECISION`

src/iceberg/type.h
```diff
@@ -0,0 +1,397 @@
...
+  TypeId type_id() const override;
+  std::string ToString() const override;
+
+ protected:
+  bool Equals(const Type& other) const override;
+};
+
+/// \brief A data type representing a timestamp in microseconds in a
+///   particular timezone.
+class ICEBERG_EXPORT TimestampTzType : public PrimitiveType {
```
wgtmac: Should we add an abstract base class for all timestamp types? We will need `timestamp_ns` and `timestamptz_ns` for v3. The only differences are `timezone` and `precision`.

wgtmac (commented):
The interface looks good to me. Thanks!

src/iceberg/schema_field.h
lidavidm: I don't think it hurts

src/iceberg/type.h
lidavidm: Do we want signed or unsigned?

src/iceberg/type.h
wgtmac: Do you suppose `PartitionField` to inherit `SchemaField` or it is totally a separate thing? @gaborkaszab

src/iceberg/type.h
lidavidm: We should add it. Maybe this isn't in line with the style guide

src/iceberg/type.h
lidavidm: Are timezone types parameterized? The Java implementation is weird and only has `adjustToUTC` (so I guess there are only naive and UTC timestamps in Iceberg?): https://github.com/apache/iceberg/blob/ab9e05e7369c55a95632b856729a05ea876cf788/api/src/main/java/org/apache/iceberg/types/Types.java#L211

wgtmac: Right, Parquet and Iceberg only care about whether this is normalized to UTC.

src/iceberg/type.h
wgtmac: Perhaps int32_t as my suggested change.

src/iceberg/type.h
lidavidm: Hmm, I can add a base class then.

src/iceberg/type.h
lidavidm: This is correct: https://google.github.io/styleguide/cppguide.html#Constant_Names

lidavidm:
I added a bunch of unit tests + fixed some copy-paste errors I discovered in the process.

src/iceberg/schema_field.h
wgtmac: Right, it is just personal flavor. I have commented at https://github.com/apache/iceberg-cpp/pull/32#issuecomment-2611657586

src/iceberg/schema_field.cc
```diff
@@ -0,0 +1,64 @@
...
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+SchemaField::SchemaField(int32_t field_id, std::string name, std::shared_ptr<Type> type,
+                         bool optional)
+    : field_id_(field_id),
+      name_(std::move(name)),
+      type_(std::move(type)),
+      optional_(optional) {}
```
raulcd, line 34: I am wondering why python moved to `required` instead of `optional` here:
https://github.com/apache/iceberg-python/blob/main/pyiceberg/types.py#L308
The change from `optional` to `required` seemed intentional [here](https://github.com/apache/iceberg-python/commit/9c8e387dea1c324dfa47acc504d7307c3f9acfd5), @Fokko any thoughts?
BTW it seems Java went for `optional` too.

src/iceberg/type.cc
```diff
@@ -0,0 +1,314 @@
...
+bool Float64Type::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kFloat64;
+}
+
+DecimalType::DecimalType(int32_t precision, int32_t scale)
+    : precision_(precision), scale_(scale) {
+  if (precision < 0 || precision > kMaxPrecision) {
+    throw std::runtime_error(
+        std::format("DecimalType: precision must be in [0, 38], was {}", precision));
+  }
```
mapleFU, line 41: Now that we've checked the precision, should we also check the scale here?

src/iceberg/type.cc
```diff
@@ -0,0 +1,314 @@
...
+}
+
+FixedType::FixedType(int32_t length) : length_(length) {
+  if (length < 0) {
+    throw std::runtime_error(
+        std::format("FixedType: length must be >= 0, was {}", length));
+  }
+}
+
+int32_t FixedType::length() const { return length_; }
```
mapleFU: an unimportant issue: should we just use uint32_t?

src/iceberg/type.cc
```diff
@@ -0,0 +1,314 @@
...
+  size_t index = 0;
+  for (const auto& field : fields_) {
+    auto [it, inserted] = field_id_to_index_.try_emplace(field.field_id(), index);
+    if (!inserted) {
+      throw std::runtime_error(
+          std::format("StructType: duplicate field ID {} (field indices {} and {})",
+                      field.field_id(), it->second, index));
+    }
+
+    index++;
```
mapleFU: nit: ++index;

src/iceberg/type.cc
```diff
@@ -0,0 +1,314 @@
...
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByIndex(
+    int32_t index) const {
+  if (index < 0 || index >= static_cast<int>(fields_.size())) {
+    return std::nullopt;
+  }
+  return fields_[index];
+}
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByName(
+    std::string_view name) const {
+  // TODO: what is the right behavior if there are duplicate names? (Are
```
mapleFU: I found that the java and the rust implement has a lazy initialization and locked hash map in get field by name, but I don't understand the reason...

mapleFU (commented):
Great work!

src/iceberg/type.cc
lidavidm: Java uses int, so I'd rather be consistent

src/iceberg/type.cc
lidavidm: Java actually doesn't validate the scale, and the spec is silent on this part. (Also, Java actually allows negative precision, though I don't think that makes any sense?)

src/iceberg/type.cc
lidavidm: The reason for the laziness, or the reason for the map? (I could change this to be lazy but tbh that introduces quite a bit of complication...)

src/iceberg/type.cc
zhjwpku: `scale` should be greater or equal than 0 and less or equal than `precision`, we can be more strict if we do the right thing. Apache arrow-rs has this this restriction[0], but I don't see this in Apache arrow cpp.

[0] https://github.com/apache/arrow-rs/blob/main/parquet/src/schema/types.rs#L505-L538

src/iceberg/type_fwd.h
```diff
@@ -0,0 +1,85 @@
...
+class Int32Type;
+class Int64Type;
+class ListType;
+class MapType;
+class NestedType;
+class PrimitiveType;
+class Schema;
+class SchemaField;
+class StringType;
+class StructType;
```
zhjwpku, line 76: typo: redundant

src/iceberg/schema.h
```diff
@@ -0,0 +1,64 @@
...
+/// A schema is a list of typed columns, along with a unique integer ID.  A
+/// Table may have different schemas over its lifetime due to schema
+/// evolution.
+class ICEBERG_EXPORT Schema : public StructType {
+ public:
+  Schema(int32_t schema_id, std::vector<SchemaField> fields);
+
+  /// \brief Get the schema ID.
+  ///
+  /// Schemas are identified by a unique ID for the purposes of schema
```
zhjwpku: nit: should this be `A Schema is identified by a unique ID` or `Schemas are identified by unique IDs`?

src/iceberg/type.cc
```diff
@@ -0,0 +1,314 @@
...
+
+namespace iceberg {
+
+TypeId BooleanType::type_id() const { return TypeId::kBoolean; }
+std::string BooleanType::ToString() const { return "boolean"; }
+bool BooleanType::Equals(const Type& other) const {
+  return other.type_id() == TypeId::kBoolean;
+}
+
+TypeId Int32Type::type_id() const { return TypeId::kInt32; }
```
zhjwpku: Should we use Int/Long/Float/Double instead? I guess this arrow style, but I think we should be consistent with spec? I checked iceberg and pyiceberg, they do the spec way.

src/iceberg/type.cc
```diff
@@ -0,0 +1,314 @@
...
+}
+bool DecimalType::Equals(const Type& other) const {
+  if (other.type_id() != TypeId::kDecimal) {
+    return false;
+  }
+  const auto& decimal = static_cast<const DecimalType&>(other);
+  return precision_ == decimal.precision_ && scale_ == decimal.scale_;
+}
+
+TypeId TimeType::type_id() const { return TypeId::kTime; }
```
zhjwpku: nit: place TimeType after DateType

src/iceberg/type.h
```diff
@@ -0,0 +1,411 @@
...
+
+  /// \brief Compare two types for inequality.
+  friend bool operator!=(const Type& lhs, const Type& rhs) { return !(lhs == rhs); }
+
+ protected:
+  /// \brief Compare two types for equality.
+  [[nodiscard]] virtual bool Equals(const Type& other) const = 0;
+};
+
+/// \brief A data type that may not have child fields.
```
zhjwpku: how about: A data type that does not have child fields.

I think the word `may` is not precise, the same applied to some other places.

src/iceberg/type.h
```diff
@@ -0,0 +1,411 @@
...
+  virtual ~Type() = default;
+
+  /// \brief Get the type ID.
+  [[nodiscard]] virtual TypeId type_id() const = 0;
+
+  /// \brief Is this a primitive type (may not have child fields)?
+  [[nodiscard]] virtual bool is_primitive() const = 0;
+
+  /// \brief Is this a nested type (may have child fields)?
+  [[nodiscard]] virtual bool is_nested() const = 0;
```
zhjwpku, line 53: Isn't this redundant with is_primitive()?

src/iceberg/type.h
```diff
@@ -0,0 +1,411 @@
...
+  ~Int64Type() = default;
+
+  TypeId type_id() const override;
+  std::string ToString() const override;
+
+ protected:
+  bool Equals(const Type& other) const override;
+};
+
+/// \brief A data type representing a 32-bit (single precision) float.
```
zhjwpku: 32-bit IEEE 754 floating point

src/iceberg/type.h
```diff
@@ -0,0 +1,411 @@
...
+  ~Float32Type() = default;
+
+  TypeId type_id() const override;
+  std::string ToString() const override;
+
+ protected:
+  bool Equals(const Type& other) const override;
+};
+
+/// \brief A data type representing a 64-bit (double precision) float.
```
zhjwpku: 64-bit IEEE 754 floating point
It will great we just copy past the type description from the spec. The same applied to other types.

src/iceberg/type.h
```diff
@@ -0,0 +1,411 @@
...
+  [[nodiscard]] int32_t length() const;
+
+  TypeId type_id() const override;
+  std::string ToString() const override;
+
+ protected:
+  bool Equals(const Type& other) const override;
+
+ private:
+  int32_t length_;
```
zhjwpku, line 397: iceberg-rust has the length of `u64`[1], `int32_t` can represent about 2 gigabytes, maybe we should change to `int64_t` to be compatible with other language implementation?

[1] https://github.com/apache/iceberg-rust/blob/main/crates/iceberg/src/spec/datatypes.rs#L241

src/iceberg/type_fwd.h
```diff
@@ -0,0 +1,85 @@
...
+  kInt32,
+  kInt64,
+  kFloat32,
+  kFloat64,
+  kDecimal,
+  kDate,
+  kTime,
+  kTimestamp,
+  kTimestampTz,
+  kBinary,
```
zhjwpku: nit: list these the same order as spec.

src/iceberg/type.cc
mapleFU: I prefer just do like this until someone need optimize this!

src/iceberg/type.cc
mapleFU: From https://github.com/apache/iceberg/issues/7420 I think setting it or not is all possible, it's ok to keep this, since underlying other storage type may throw exception

src/iceberg/type.cc
mapleFU: Seems that Java not has unsigned int? But both part lgtm

src/iceberg/type.cc
zhjwpku: Blue give some examples which scale can be greater than precision and smaller than 0[1], though I don't think it makes sense, I agree we should keep this as is.

[1] https://github.com/apache/iceberg/issues/7420#issuecomment-1526522715

src/iceberg/type.h
lidavidm: Java uses a 32 bit int. https://github.com/apache/iceberg/blob/d693f83b6ad6e306550d117ff0bf0014356500b2/api/src/main/java/org/apache/iceberg/types/Types.java#L360

Are the Iceberg maintainers interested in making the spec actually lay out these details? I've mostly copied Java since the spec itself is extremely vague.

src/iceberg/type_fwd.h
lidavidm: Is the spec actually canonical? It seems the implementations all disagree...

src/iceberg/type.h
lidavidm: To me it's clearer to be able to say `is_nested` or `is_primitive` and they can just be implemented in terms of each other.

src/iceberg/type.cc
lidavidm: Can we fix this in the spec? The spec itself says nothing about these types which is a little frustrating since I feel like the Java implementation is the actual de-facto spec

lidavidm: Oh wow, the issue asking for spec clarification was closed. Alright then, I will leave this as-is

src/iceberg/type.cc
lidavidm: Java does not have unsigned types but you can manipulate signed types as if they were unsigned when necessary. I'll close this since there's another thread about this now...

src/iceberg/type_fwd.h
lidavidm: And the spec doesn't actually have any single listing of all the types (since nested types are split off), so I am electing to put the nested types first to be consistent with the spec (in as far as it defines any order)

src/iceberg/type.cc
lidavidm: Fixed

src/iceberg/type_fwd.h
lidavidm: Fixed

src/iceberg/schema.h
lidavidm: Changed to "A schema..."

src/iceberg/type.cc
lidavidm: Updated

src/iceberg/type.cc
lidavidm: Reordered headers/impls

src/iceberg/type.h
lidavidm: Updated

src/iceberg/type.h
lidavidm: Updated

src/iceberg/type_fwd.h
lidavidm: Updated the order

src/iceberg/type.h
zhjwpku: Ok, let's keep both.

zhjwpku (approved):

src/iceberg/schema.cc
```diff
@@ -0,0 +1,47 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/schema.h"
+
+#include <format>
```
wgtmac, line 22: This is redundant.

src/iceberg/util/formatter.h
```diff
@@ -0,0 +1,43 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/util/formatter.h
```
wgtmac, line 22: Should this file be renamed to `formatter_internal.h`? We don't have to export and install this file.

FYI: I have followed the same pattern from Apache Arrow: https://github.com/apache/iceberg-cpp/blob/main/cmake_modules/IcebergBuildUtils.cmake#L251

src/iceberg/type.h
```diff
@@ -0,0 +1,414 @@
...
+
+/// \brief A data type that does not have child fields.
+class ICEBERG_EXPORT PrimitiveType : public Type {
+ public:
+  bool is_primitive() const override { return true; }
+  bool is_nested() const override { return false; }
+};
+
+/// \brief A data type that has child fields.
+class ICEBERG_EXPORT NestedType : public Type {
```
wgtmac, line 74: Do we need a function for getting number of fields?

src/iceberg/type.h
```diff
@@ -0,0 +1,414 @@
...
+
+  /// \brief Get a view of the child fields.
+  [[nodiscard]] virtual std::span<const SchemaField> fields() const = 0;
+  /// \brief Get a field by field ID.
+  [[nodiscard]] virtual std::optional<std::reference_wrapper<const SchemaField>>
+  GetFieldById(int32_t field_id) const = 0;
+  /// \brief Get a field by index.
+  [[nodiscard]] virtual std::optional<std::reference_wrapper<const SchemaField>>
+  GetFieldByIndex(int32_t index) const = 0;
+  /// \brief Get a field by name.
```
wgtmac: Is this case sensitive?

src/iceberg/type.cc
```diff
@@ -0,0 +1,312 @@
...
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+StructType::StructType(std::vector<SchemaField> fields) : fields_(std::move(fields)) {
+  size_t index = 0;
+  for (const auto& field : fields_) {
+    auto [it, inserted] = field_id_to_index_.try_emplace(field.field_id(), index);
+    if (!inserted) {
+      throw std::runtime_error(
```
wgtmac, line 35: Should we define `IcebergException` or even a more specific exception to replace it?

src/iceberg/type.cc
```diff
@@ -0,0 +1,312 @@
...
+                      field.field_id(), it->second, index));
+    }
+
+    ++index;
+  }
+}
+
+TypeId StructType::type_id() const { return TypeId::kStruct; }
+std::string StructType::ToString() const {
+  std::string repr = "struct<\n";
```
wgtmac, line 46: Is there a specific reason to use a separate line for each sub-field? An alternative is to use the Hive-style where all fields are concatenated in a single line.

src/iceberg/type.cc
```diff
@@ -0,0 +1,312 @@
...
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByIndex(
+    int32_t index) const {
+  if (index < 0 || index >= static_cast<int>(fields_.size())) {
+    return std::nullopt;
+  }
+  return fields_[index];
+}
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByName(
+    std::string_view name) const {
+  // TODO: what is the right behavior if there are duplicate names? (Are
```
wgtmac, line 69: It seems that the spec is unclear about this so the behavior is undefined and implementation-specific.

src/iceberg/util/formatter.h
lidavidm: No, you do need this file externally if you yourself wish to use std::format.

src/iceberg/type.cc
lidavidm: Are we using exceptions in general? I am only using an exception here because this is a constructor. We could also hide the constructor, or change to assertions.

src/iceberg/type.cc
lidavidm: That gets messy IMO once there are more than a couple fields

src/iceberg/type.cc
lidavidm: There seem to be a lot of spec issues...

src/iceberg/type.h
lidavidm: `fields().size()`, or do you think a separate accessor is useful?

src/iceberg/type.h
lidavidm: Updated

src/iceberg/type.cc
wgtmac: Do we want to support restoring a Schema object by deserializing the formatted string?

src/iceberg/schema.cc
lidavidm: Isn't it good practice to IWYU?

src/iceberg/type.h
wgtmac: That's enough I think.

wgtmac (approved):
+1

src/iceberg/schema.cc
wgtmac: I'm not that familiar with IWYU yet...

src/iceberg/type.cc
wgtmac: No, my point is that a more specific exception is better than the `std::runtime_error`. 

BTW, should we build `field_id_to_index_` lazily when the particular getter is called for the 1st time?

src/iceberg/type.cc
lidavidm: I'd rather add JSON serializers than try to parse the repr

src/iceberg/type.cc
lidavidm: That's discussed above, but it complicates the implementation a lot

src/iceberg/schema.cc
lidavidm: We use std::format in the .cc file so we should include format, even if other headers happen to include it already

lidavidm:
Is anything else required here?

mapleFU (approved):

zhjwpku:
> Is anything else required here?

+1 to merge this PR, thanks for working on this.

wgtmac:
@Fokko @Xuanwo I think this is ready to merge. cc @gaborkaszab in case there is any comment.

gaborkaszab:
> @Fokko @Xuanwo I think this is ready to merge. cc @gaborkaszab in case there is any comment.

I got a bit busy with other stuff, so wouldn't block this PR for me to find time here. Thanks for these changes!

src/iceberg/schema_field.cc
Fokko: It is because the spec serializes it as `required`: https://iceberg.apache.org/spec/#schemas This way it is more straightforward to serialize it using the library that we use (Pydantic). See the original issue: https://github.com/apache/iceberg/issues/4985. In PyIceberg we have a property called `optional` as well, so you can choose.

src/iceberg/schema_field.cc
```diff
@@ -0,0 +1,64 @@
...
+
+std::string_view SchemaField::name() const { return name_; }
+
+const std::shared_ptr<Type>& SchemaField::type() const { return type_; }
+
+bool SchemaField::optional() const { return optional_; }
+
+std::string SchemaField::ToString() const {
+  return std::format("{} ({}): {}{}", name_, field_id_, *type_,
+                     optional_ ? "" : " (required)");
```
Fokko, line 56: Maybe it is just me, but I'm not a big fan of the implied `(optional)`

src/iceberg/type.cc
```diff
@@ -0,0 +1,312 @@
...
+  return fields_[it->second];
+}
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByIndex(
+    int32_t index) const {
+  if (index < 0 || index >= static_cast<int>(fields_.size())) {
+    return std::nullopt;
+  }
+  return fields_[index];
+}
+std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByName(
```
Fokko, line 67: We also want to distinguish upper- and lower-case, or start with case-sensitive here.

src/iceberg/type.cc
Fokko: Please consider the Java implementation as the reference one. There we raise a `ValidationException` when we see a duplicate name:

https://github.com/apache/iceberg/blob/507e2a9e39f93356bdc0dae1521bbdc46629681c/api/src/main/java/org/apache/iceberg/types/IndexByName.java#L199-L205

> There seem to be a lot of spec issues...

@lidavidm Please elaborate

src/iceberg/type.h
Fokko: The `UnknownType` will be part of the V3 specification.

An empty struct is pretty common (for example, an unpartitioned table has an empty struct as a partition).

Fokko (approved):
Looks like there is consensus, let's move this forward 🚀

src/iceberg/type.cc
lidavidm: Is this in the written spec? Or is the canonical-ness of the Java implementation documented somewhere? Just so I know what to reference

lidavidm: Also see above e.g. https://github.com/apache/iceberg-cpp/pull/31#discussion_r1929559679
https://github.com/apache/iceberg-cpp/pull/31#discussion_r1928875579

possibly I overlooked something

lidavidm:
Thanks @Fokko I'll follow up to correct those things

src/iceberg/type.cc
lidavidm: The docstring states this is case-sensitive already. Should we add something like Java's `StructType#caseInsensitiveField`?

src/iceberg/type.cc
Fokko: The spec describes the table format itself, for the behavioral things, we tend to check the Java implementation (considered the reference implementation). If we would need to formalize all this as well, then the spec would become unreadable.

lidavidm: IMO, at least something like "duplicate field names are not allowed" should be part of the spec, no? I wouldn't really qualify that as "behavioral" any more than "decimal precision is limited to 38" is behavioral

src/iceberg/type.h
Fokko: The int32 is fine. No one should go beyond 2GB values for a single value.

src/iceberg/type.cc
Fokko: Since Iceberg comes from the database world, not having duplicate names is more or less implied to me, but feel free to raise a PR to add this to the spec itself.

I was surprised to see that this is allowed:

```python
>>> import pyarrow as pa
>>> pa.table([
...     pa.array(["foo", "bar"]),
...     pa.array(["Parrot", "Dog"])
... ], names=['animals', 'animals'])
pyarrow.Table
animals: string
animals: string
----
animals: [["foo","bar"]]
animals: [["Parrot","Dog"]]
```

wgtmac: > I was surprised to see that this is allowed:

I think Arrow is different. It is totally valid to project the same column more than once in an Arrow Table or RecordBatch. For example, returning query result via Arrow for `select animals, animals from foo.bar`;

PR #45, opened 2025-03-18
Build/link to Rust iceberg-c library

lidavidm:
Requires https://github.com/apache/iceberg-rust/pull/966 with changes from https://github.com/lidavidm/iceberg-rust/tree/add-c-binding.

### DIFF
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -21,7 +21,11 @@ add_iceberg_lib(iceberg
                 SOURCES
                 ${ICEBERG_SOURCES}
                 PRIVATE_INCLUDES
-                ${ICEBERG_INCLUDES})
+                ${ICEBERG_INCLUDES}
+                SHARED_LINK_LIBS
+                IcebergC::iceberg_c_shared
+                STATIC_LINK_LIBS
+                IcebergC::iceberg_c_static)
 
 iceberg_install_all_headers(iceberg)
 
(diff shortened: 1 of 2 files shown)


Xuanwo:
Hi, @lidavidm, maybe we should link against the iceberg-rust cpp bindings?  I'll prepare another PR to demonstrate how CXX works.

src/iceberg/CMakeLists.txt
```diff
@@ -21,7 +21,11 @@ add_iceberg_lib(iceberg
                 SOURCES
                 ${ICEBERG_SOURCES}
                 PRIVATE_INCLUDES
-                ${ICEBERG_INCLUDES})
+                ${ICEBERG_INCLUDES}
+                SHARED_LINK_LIBS
+                IcebergC::iceberg_c_shared
```
wgtmac, line 26: Do we need to make `iceberg` library mainly interface-oriented and add a new `iceberg-rust-cxx-binding` library to plug in the implementation? It is still unclear to me the extent of plug-ability in this approach.

lidavidm:
For sure, I just wanted to prove that the CMake work I did on the other side works properly.

src/iceberg/CMakeLists.txt
lidavidm: It depends on if we want to use Rust definitions for things like data types too, or if we just want to use it to implement talking to catalogs etc.

wgtmac: I think that leveraging rust library is a good approach to move fast. However, there are still users that don't have a rust toolchain, or do not want to add extra dependencies introduced by the rust library.

PR #47, opened 2025-03-21
feat: add pure virtual classes for Catalog, Table, etc.

wgtmac:
# What are done in this PR
- Added pure virtual classes of `Catalog`, `Table`, `TableOperations`, `Transaction` and `LocationProvider` with limited features like `AppendFiles` and `TableScan`.
- Added basic definitions of `Namespace` and `TableIdentifier`.
- Defined initial `Error` enum.

# What are undecided
- How to define an IO-less `FileIO`.
- How to define `StructLike`.


src/iceberg/table.h
```diff
@@ -19,17 +19,91 @@
...
-  virtual std::string print() const = 0;
+
+  /// \brief Return the full name for this table
+  virtual std::string name() const = 0;
+
+  /// \brief Returns the UUID of the table
+  virtual std::string uuid() const = 0;
+
+  /// \brief Refresh the current table metadata
+  virtual void Refresh() = 0;
```
lidavidm: Hmm, do we want a table to be able to do I/O, or should it effectively be a POD (albeit not actually a POD)?

src/iceberg/table.h
```diff
@@ -19,17 +19,91 @@
...
+  /// \return a vector of history entries
+  virtual std::vector<std::shared_ptr<HistoryEntry>> history() const = 0;
+
+  /// \brief Create a new table scan for this table
+  ///
+  /// Once a table scan is created, it can be refined to project columns and filter data.
+  virtual std::shared_ptr<TableScan> NewScan() const = 0;
+
+  /// \brief Create a new append API to add files to this table and commit
+  virtual std::shared_ptr<AppendFiles> NewAppend() = 0;
```
lidavidm, line 97: Why are some of these shared_ptr and some of these unique_ptr? It feels like AppendFiles and Transaction are morally the same kind of operation?

src/iceberg/table_identifier.h
```diff
@@ -0,0 +1,43 @@
...
+namespace iceberg {
+
+/// \brief A namespace in a catalog.
+struct ICEBERG_EXPORT Namespace {
+  std::vector<std::string> levels;
+};
+
+/// \brief Identifies a table in iceberg catalog.
+struct ICEBERG_EXPORT TableIdentifier {
+  Namespace name_space;
```
lidavidm: `name_space` -> `namespace`?

src/iceberg/type_fwd.h
```diff
@@ -81,4 +81,36 @@ class TimestampTzType;
 class Type;
 class UuidType;
 
+/// \brief Error types for iceberg.
+/// TODO: add more and sort them based on some rules.
+enum class Error {
```
lidavidm: nit: maybe ErrorKind, ErrorType?

src/iceberg/table.h
```diff
@@ -19,17 +19,91 @@
...
+
+  /// \brief Get the snapshot history of this table
+  ///
+  /// \return a vector of history entries
+  virtual std::vector<std::shared_ptr<HistoryEntry>> history() const = 0;
+
+  /// \brief Create a new table scan for this table
+  ///
+  /// Once a table scan is created, it can be refined to project columns and filter data.
+  virtual std::shared_ptr<TableScan> NewScan() const = 0;
```
lidavidm: I feel like this can be unique_ptr?

src/iceberg/table.h
```diff
@@ -19,17 +19,91 @@
...
+  virtual std::map<int32_t, std::shared_ptr<PartitionSpec>> specs() const = 0;
+
+  /// \brief Return the sort order for this table
+  virtual const std::shared_ptr<SortOrder>& sort_order() const = 0;
+
+  /// \brief Return a map of sort order IDs to sort orders for this table
+  virtual std::map<int32_t, std::shared_ptr<SortOrder>> sort_orders() const = 0;
+
+  /// \brief Return a map of string properties for this table
+  virtual std::map<std::string, std::string> properties() const = 0;
```
lidavidm: maybe `map const&`?

src/iceberg/table.h
```diff
@@ -19,17 +19,91 @@
...
+
+  /// \brief Return the table's base location
+  virtual std::string location() const = 0;
```
lidavidm: maybe `string const&`?

lidavidm: (ditto for various others)

lidavidm (commented):
Seems reasonable overall

src/iceberg/table_identifier.h
wgtmac: `namespace` is a C++ keyword...

lidavidm: Ah, doh. That's kind of annoying here.

src/iceberg/table.h
wgtmac: `Transaction` might have more than one operations and `AppendFiles` is a part of it. I see that the Transaction object in the Java implementation holds all associated operations including AppendFiles so that's why I make it a shared_ptr.

wgtmac: I think I wrote TableScan to be in a std::unique_ptr (it might be changed by Cursor...)

src/iceberg/table.h
wgtmac: In my mind, `Table` only triggers the (JSON and/or Avro) readers to perform I/O and it is the implementation's responsibility to do that. TBH, I haven't thought about it in detail yet.

lidavidm: I was thinking Table would be a wrapper around already-parsed data, and so to refresh the table, you'd just load the table again (or is there a reason why it has to be the same object?)

lidavidm: (In that model, there would be no need for Table to be abstract, except for perhaps the parts relating to scans/transactions)

src/iceberg/table_identifier.h
wgtmac: Perhaps `ns` looks better than `name_space`

lidavidm: Or `table_namespace` and `table_name` (perhaps kind of redundant)

src/iceberg/catalog.h
```diff
@@ -0,0 +1,178 @@
...
+
+namespace iceberg {
+
+/// \brief A Catalog API for table create, drop, and load operations.
+class ICEBERG_EXPORT Catalog {
+ public:
+  virtual ~Catalog() = default;
+
+  /// \brief Return the name for this catalog
+  virtual std::string name() const = 0;
```
lidavidm: const& here too? (Or string_view?)

src/iceberg/table.h
wgtmac: > I was thinking Table would be a wrapper around already-parsed data, and so to refresh the table

I think you are referring to `SerializableTable` in the java impl. I think a Table abstraction is still needed because we want to differentiate a real table versus metadata table. Table internally holds a `TableOperations` object which offloads all I/O operations to the implementation.

src/iceberg/table_operations.h
```diff
@@ -0,0 +1,94 @@
...
+
+namespace iceberg {
+
+/// \brief Interface for table operations including metadata refresh and commit
+class ICEBERG_EXPORT TableOperations {
+ public:
+  virtual ~TableOperations() = default;
+
+  /// \brief Return the currently loaded table metadata, without checking for updates.
+  virtual std::shared_ptr<TableMetadata> Current() = 0;
```
lidavidm: const&?

src/iceberg/catalog.h
```diff
@@ -0,0 +1,178 @@
...
+
+    /// \brief Creates the table
+    ///
+    /// \return the created table
+    virtual std::unique_ptr<Table> Create() = 0;
+
+    /// \brief Starts a transaction to create the table
+    ///
+    /// \return the Transaction to create the table
+    virtual std::unique_ptr<Transaction> CreateTransaction() = 0;
```
Fokko: Just a thought. To make this more Iceberg REST catalog agnostic, we could also call this `Stage`:

https://github.com/apache/iceberg/blob/a4816c1c99063770473920cd6d62f88f90a292dc/open-api/rest-catalog-open-api.yaml#L556-L561

src/iceberg/table_operations.h
```diff
@@ -0,0 +1,94 @@
...
+  /// This is called by transactions when uncommitted table metadata should be used; for
+  /// example, to create a metadata file location based on metadata in the transaction
+  /// that has not been committed.
+  ///
+  /// Transactions will not call `Refresh()` or `Commit()`.
+  ///
+  /// \param uncommittedMetadata uncommitted table metadata
+  /// \return a temporary table operations that behaves like the uncommitted metadata is
+  /// current
+  virtual std::unique_ptr<TableOperations> Temp(
```
wgtmac: I thought this one is more appropriate to be renamed as `Stage`... @Fokko

src/iceberg/table_identifier.h
wgtmac: I like shorter names...

src/iceberg/table.h
Fokko: Maybe good to also look at PyIceberg, where we left out the `TableOperations`. There is more abstraction in Java because supports [File System tables](https://iceberg.apache.org/spec/#file-system-tables). In PyIceberg we've designed everything around the catalog. The `refresh` operation will refresh the `TableMetadata` that (de)serializes from/into the JSON metadata.

src/iceberg/table_operations.h
Fokko: The staged table creation is where you first call the catalog to create a table. The table won't be visible yet but will be reserved for the client that called it. Next, you can write all your data and once you commit the snapshot, it will be visible in the catalog. This way:

- If you do a `CREATE TABLE AS SELECT * ...` which might take some time, you know that you're not going to get into conflicts.
- The catalog might provide S3 credentials the client needs to write the data to S3.

src/iceberg/table.h
Fokko: In PyIceberg, the Table also has a reference back to the catalog, which makes it easier to refresh the underlying TableMetadata.

src/iceberg/type_fwd.h
```diff
@@ -81,4 +81,36 @@ class TimestampTzType;
 class Type;
 class UuidType;
 
+/// \brief Error types for iceberg.
+/// TODO: add more and sort them based on some rules.
+enum class ErrorKind {
```
zhjwpku: Should we create a new file for this ErrorKind?

src/iceberg/type_fwd.h
```diff
@@ -81,4 +81,36 @@ class TimestampTzType;
...
+  kNoSuchNamespace,
+  kAlreadyExists,
+  kNoSuchTable,
+  kCommitStateUnknown,
+};
+
+struct Namespace;
```
zhjwpku, line 84: Since we are putting these forward declarations here, should we change the name type_fwd.h to something like spec_fwd.h?

.github/workflows/cpp-linter.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           thread-comments: true
           ignore: 'build|cmake_modules|ci'
           database: build
+          extra-args: '-I$PWD/src -I$PWD/build/src'
```
zhjwpku: Need rebase.

src/iceberg/catalog.h
wgtmac: @Fokko Quick question: `Transaction` has a comment `A transaction for performing multiple updates to a table` but the `BaseTransaction` has a check `checkLastOperationCommitted` which throws when previous update is uncommitted. Will multiple updates be supported in the future to remove this inconsistency?

src/iceberg/table.h
wgtmac: To @lidavidm:

> Hmm, do we want a table to be able to do I/O, or should it effectively be a POD (albeit not actually a POD)?

I believe we have to support I/O at least to parse metadata file. See my comment at https://github.com/apache/iceberg-cpp/pull/30#pullrequestreview-2710525598.

> I was thinking Table would be a wrapper around already-parsed data, and so to refresh the table, you'd just load the table again

I think we can still define an abstract `Table` class. A subclass `StaticTable` accepts a deserialized `TableMetadata` and `StaticTable::Refresh()` is a no-op.

We still need the Table abstraction for catalog-backed tables and metadata tables.

To @Fokko:

> In PyIceberg, the Table also has a reference back to the catalog, which makes it easier to refresh the underlying TableMetadata.

That makes sense! We can add a `CatalogTable` to implement a catalog-backed table.

src/iceberg/type_fwd.h
wgtmac: `type_fwd.h` is a convention used by arrow-cpp to hold forward declarations and enums in a single header file. It might be a coincident that it starts with the data types...

src/iceberg/type_fwd.h
wgtmac: Yes, I was also thinking of adding a detail error message like:

```
struct Error {
  ErrorKind kind;
  std::string message;
};
```

mapleFU: This seems same as `Status`?

src/iceberg/type_fwd.h
lidavidm: Yeah, it's not "Iceberg data type forward declarations", it's "C++ data type forward declarations". Open to renaming it (though STL at least uses the `fwd` convention too with `iosfwd`)

src/iceberg/type_fwd.h
wgtmac: Status can be OK but here we don't want to include that.

wgtmac:
I have removed `TableOperations` as suggested. Now the `Catalog` API reflects the `operationId` from the Rest Catalog API to be consistent. Please note that these APIs are still subject to change when we find any issue. With these minimal set of APIs, developers can work on different parts without blocking each other.

Let me know what you think. @lidavidm @zhjwpku @Fokko @Xuanwo

src/iceberg/table_operations.h
wgtmac: You convinced me. I have renamed them.

src/iceberg/catalog.h
Fokko: > Will multiple updates be supported in the future to remove this inconsistency?

Yes, there is no reason not to support this.

Fokko (approved):

src/iceberg/catalog.h
```diff
@@ -0,0 +1,194 @@
...
+/// \brief A Catalog API for table create, drop, and load operations.
+///
+/// Note that these functions are named after the corresponding operationId
+/// specified by the Iceberg Rest Catalog API.
+class ICEBERG_EXPORT Catalog {
+ public:
+  virtual ~Catalog() = default;
+
+  /// \brief Return the name for this catalog
+  virtual std::string_view name() const = 0;
```
gty404, line 44: maybe Name() ?

src/iceberg/catalog.h
```diff
@@ -0,0 +1,194 @@
...
+
+  /// \brief Return all the identifiers under this namespace
+  ///
+  /// \param ns a namespace
+  /// \return a list of identifiers for tables or ErrorKind::kNoSuchNamespace
+  /// if the namespace does not exist
+  virtual expected<std::vector<TableIdentifier>, Error> ListTables(
```
gty404, line 51: should std::unique_ptr<std::vector<TableIdentifier>> be used 
 instead of std::vector<TableIdentifier>  to avoid copying ?

gty404: We can forward declare TableIdentifierList / TableIdentifierListPtr.

src/iceberg/catalog.h
wgtmac: This is the coding style convention from arrow-cpp. For non-trivial functions, we use  camel case with capitalized initial. For trivial functions (e.g. getters), we simply use lowercased snake case.

src/iceberg/table.h
```diff
@@ -19,17 +19,93 @@
...
+  /// \brief Get the snapshot of this table with the given id, or null if there is no
+  /// matching snapshot
+  ///
+  /// \param snapshot_id the ID of the snapshot to get
+  /// \return the Snapshot with the given id
+  virtual expected<std::shared_ptr<Snapshot>, Error> snapshot(
+      int64_t snapshot_id) const = 0;
+
+  /// \brief Get the snapshots of this table
+  virtual const std::vector<std::shared_ptr<Snapshot>>& snapshots() const = 0;
```
gty404, line 84: Should we return an Enumerator or Iterator type if there are too many snapshots?

src/iceberg/transaction.h
```diff
@@ -0,0 +1,53 @@
...
+  ///
+  /// \return a new AppendFiles
+  virtual std::shared_ptr<AppendFiles> NewAppend() = 0;
+
+  /// \brief Apply the pending changes from all actions and commit
+  ///
+  /// May throw ValidationException if any update cannot be applied to the current table
+  /// metadata. May throw CommitFailedException if the updates cannot be committed due to
+  /// conflicts.
+  virtual void CommitTransaction() = 0;
```
gty404, line 50: Abort or rollback is automatically called when a commit fails ?

src/iceberg/table.h
```diff
@@ -19,17 +19,93 @@
...
+
+  /// \brief Get the snapshot history of this table
+  ///
+  /// \return a vector of history entries
+  virtual const std::vector<std::shared_ptr<HistoryEntry>>& history() const = 0;
```
gty404, line 89: Enumerator or Iterator  here too ?

src/iceberg/catalog.h
wgtmac: I think the compiler is smart enough to do RVO. BTW, I still prefer `std::vector<TableIdentifier>` to `TableIdentifierList` because it is short enough and more readable (do not need an extra jump to see its full definition)

src/iceberg/table.h
wgtmac: That's a good question. Maybe you can add it later? I think the current one makes sense because the `Table` object has a full state of `TableMetadata` so it anyway caches all `Snapshot`s in it.

src/iceberg/transaction.h
wgtmac: I'm not 100% sure about this. But checking the java impl it at least does some cleanup work: https://github.com/apache/iceberg/blob/e1e0a7404740b2bf9e6638afb5f0ff19f2536713/core/src/main/java/org/apache/iceberg/BaseTransaction.java#L324-L349

src/iceberg/table.h
wgtmac: Same reason as above.

src/iceberg/transaction.h
gty404: If that’s the case, then the abort/rollback methods don’t need to exist on the transaction interface.

gt-yu (approved):

PR #53, opened 2025-03-28
feat: convert iceberg schema to arrow schema

wgtmac:


wgtmac:
@lidavidm @zhjwpku PTAL :)

src/iceberg/arrow_c_data.h
```diff
@@ -66,6 +71,47 @@ struct ArrowArray {
...
 };
 
-}  // extern "C"
-
 #endif  // ARROW_C_DATA_INTERFACE
+
+#ifndef ARROW_C_STREAM_INTERFACE
+#  define ARROW_C_STREAM_INTERFACE
+
+struct ArrowArrayStream {
```
zhjwpku: Should we mention [0] in the \file description?

I have a feeling that this is not related to PR $titile, so maybe another PR for this?

[0] https://arrow.apache.org/docs/format/CStreamInterface.html

test/arrow_test.cc
```diff
@@ -59,4 +67,192 @@ TEST(ArrowCDataTest, CheckArrowSchemaAndArrayByNanoarrow) {
...
+                           kStructFieldId, std::string(kStructFieldName), struct_type)});
+
+  ArrowSchema arrow_schema;
+  ASSERT_OK(ToArrowSchema(schema, &arrow_schema));
+
+  auto imported_schema = ::arrow::ImportSchema(&arrow_schema).ValueOrDie();
+  ASSERT_EQ(imported_schema->num_fields(), 1);
+
+  auto field = imported_schema->field(0);
+  CheckArrowField(*field, ::arrow::Type::STRUCT, kStructFieldName, /*nullable=*/false,
```
lidavidm: technically you should `ASSERT_NO_FATAL_FAILURE(CheckArrowField(...))`

test/test_util.h
```diff
@@ -0,0 +1,26 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#define EXPECT_OK(value) \
```
lidavidm: Note that you can write a GMock matcher using one of the macros GMock provides, then you can do something like `ASSERT_THAT(Foo(), IsOk())`. A macro might be simpler, though.

src/iceberg/schema_internal.cc
```diff
@@ -0,0 +1,193 @@
...
+
+constexpr const char* kArrowExtensionName = "ARROW:extension:name";
+constexpr const char* kArrowExtensionMetadata = "ARROW:extension:metadata";
+
+// Convert an Iceberg type to Arrow schema. Return value is Nanoarrow error code.
+ArrowErrorCode ConvertToArrowSchema(const Type& type, ArrowSchema* schema, bool optional,
+                                    std::string_view name = "", int32_t field_id = -1) {
+  ArrowBuffer metadata_buffer;
+  NANOARROW_RETURN_NOT_OK(ArrowMetadataBuilderInit(&metadata_buffer, nullptr));
+  if (field_id > 0) {
```
lidavidm: BTW, where in the Iceberg spec does it say that field IDs must be nonzero? (I don't even see a bit width defined for field ID in the spec...)

`optional` might be a more idiomatic way to write the type but this works (if we can find a reference for the field ID type)

lidavidm (approved):

src/iceberg/schema_internal.cc
wgtmac: Good question! In practice, parquet-cpp rejects negative `field_id`: https://github.com/apache/arrow/blob/618ef501a21375abfaeee19e393eb64dee83ef0d/cpp/src/parquet/arrow/schema.cc#L248-L278

@Fokko @rdblue Has this been discussed before? Should we restrict `field_id` to be non-negative from the spec?

src/iceberg/arrow_c_data.h
wgtmac: I have to add this, otherwise compiling files with headers from `nanoarrow` will complain missing definition of `ArrowArrayStream`.

zhjwpku: OK, then the patch LGTM.

zhjwpku (approved):
LGTM

src/iceberg/schema_internal.cc
lidavidm: Hmm, should field ID 0 be valid then though?

test/arrow_test.cc
wgtmac: You're right. I forgot to add them.

src/iceberg/arrow_c_data.h
wgtmac: I have removed ArrowArrayStream. I think internally we can just include `nanoarrow.h` so we don't have to add ArrowArrayStream and confuse the downstream.

test/test_util.h
wgtmac: Good suggestion! Added `matchers.h` for this.

src/iceberg/schema_internal.cc
wgtmac: I've switched to use std::optional<int32_t> for now.

wgtmac:
Thanks @lidavidm @zhjwpku! I believe I have addressed all the comments now.

lidavidm (approved):

wgtmac:
@Fokko @Xuanwo Could you help review and merge it? Thanks!

src/iceberg/schema_internal.cc
```diff
@@ -0,0 +1,193 @@
...
+
+namespace iceberg {
+
+namespace {
+
+constexpr const char* kArrowExtensionName = "ARROW:extension:name";
+constexpr const char* kArrowExtensionMetadata = "ARROW:extension:metadata";
+
+// Convert an Iceberg type to Arrow schema. Return value is Nanoarrow error code.
+ArrowErrorCode ConvertToArrowSchema(const Type& type, ArrowSchema* schema, bool optional,
```
Xuanwo: `ConvertToArrowSchema` and `ToArrowSchema` are a bit confusing to me.  Do we need to make the API naming more clear or split them in different namespaces?

---

By the way, do we have a policy for our API args' order?

`input_field_a, out, input_field_b, input_field_c` seems not good to me and hard to follow.

Is it widely adopted in cpp community?

Xuanwo (commented):
Hi, thank @wgtmac for working on this. I have some dumb questions over this PR.

src/iceberg/schema_internal.cc
wgtmac: That makes sense. Actually they are in different namespaces (the former one is in the anonymous namespace). I have renamed them to `ToArrowSchema` to be consistent.

W.r.t. the order of arguments, those with default values must be placed at the end. Since they are all internal functions, I have just removed any default argument and put the out param to the end.

PR #54, opened 2025-03-30
feat: add partition field/partition spec

gty404:
- Implement the API for partition spec/field.
- add pure virtual classes for transform

### DIFF
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -23,6 +23,9 @@ set(ICEBERG_SOURCES
     schema.cc
     schema_field.cc
     schema_internal.cc
+    partition_field.cc
+    partition_spec.cc
+    transform.cc
     type.cc)
 
 set(ICEBERG_STATIC_BUILD_INTERFACE_LIBS)
--- a/src/iceberg/exception.h
+++ b/src/iceberg/exception.h
@@ -38,4 +38,11 @@ class ICEBERG_EXPORT IcebergError : public std::runtime_error {
   explicit IcebergError(const std::string& what) : std::runtime_error(what) {}
 };
 
+#define ICEBERG_CHECK(condition, ...)                        \
+  do {                                                       \
+    if (!(condition)) [[unlikely]] {                         \
+      throw iceberg::IcebergError(std::format(__VA_ARGS__)); \
+    }                                                        \
+  } while (0)
+
 }  // namespace iceberg
--- a/src/iceberg/partition_field.h
+++ b/src/iceberg/partition_field.h
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/partition_field.h
+/// A partition field in a partition spec
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/type_fwd.h"
+#include "iceberg/util/formattable.h"
+
+namespace iceberg {
+
+/// \brief a field with its transform.
+class ICEBERG_EXPORT PartitionField : public util::Formattable {
+ public:
+  /// \brief Construct a field.
+  /// \param[in] source_id The source field ID.
+  /// \param[in] field_id The partition field ID.
+  /// \param[in] name The partition field name.
+  /// \param[in] transform The transform function.
+  PartitionField(int32_t source_id, int32_t field_id, std::string name,
+                 std::shared_ptr<TransformFunction> transform);
+
+  /// \brief Get the source field ID.
+  int32_t source_id() const;
+
+  /// \brief Get the partition field ID.
+  int32_t field_id() const;
+
+  /// \brief Get the partition field name.
+  std::string_view name() const;
+
+  /// \brief Get the transform type.
+  std::shared_ptr<TransformFunction> const& transform() const;
+
+  std::string ToString() const override;
+
+  friend bool operator==(const PartitionField& lhs, const PartitionField& rhs) {
+    return lhs.Equals(rhs);
+  }
+
+  friend bool operator!=(const PartitionField& lhs, const PartitionField& rhs) {
+    return !(lhs == rhs);
+  }
+
+ private:
+  /// \brief Compare two fields for equality.
+  [[nodiscard]] bool Equals(const PartitionField& other) const;
+
+  int32_t source_id_;
+  int32_t field_id_;
+  std::string name_;
+  std::shared_ptr<TransformFunction> transform_;
+};
+
+}  // namespace iceberg
--- a/src/iceberg/partition_spec.cc
+++ b/src/iceberg/partition_spec.cc
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/partition_spec.h"
+
+#include <format>
+
+#include "iceberg/schema.h"
+#include "iceberg/type.h"
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+PartitionSpec::PartitionSpec(std::shared_ptr<Schema> schema, int32_t spec_id,
+                             std::vector<PartitionField> fields)
+    : schema_(std::move(schema)), spec_id_(spec_id), fields_(std::move(fields)) {}
+
+const std::shared_ptr<Schema>& PartitionSpec::schema() const { return schema_; }
+
+int32_t PartitionSpec::spec_id() const { return spec_id_; }
+
+std::span<const PartitionField> PartitionSpec::fields() const { return fields_; }
+
+std::string PartitionSpec::ToString() const {
+  std::string repr = std::format("partition_spec[spec_id<{}>,\n", spec_id_);
+  for (const auto& field : fields_) {
+    std::format_to(std::back_inserter(repr), "  {}\n", field);
+  }
+  repr += "]";
+  return repr;
+}
+
+bool PartitionSpec::Equals(const PartitionSpec& other) const {
+  return *schema_ == *other.schema_ && spec_id_ == other.spec_id_ &&
+         fields_ == other.fields_;
+}
+
+}  // namespace iceberg
--- a/src/iceberg/partition_spec.h
+++ b/src/iceberg/partition_spec.h
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/partition_spec.h
+/// Partition specs for Iceberg tables.
+
+#include <cstdint>
+#include <span>
+#include <string>
+#include <vector>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/partition_field.h"
+#include "iceberg/util/formattable.h"
+
+namespace iceberg {
+
+/// \brief A partition spec for a Table.
+///
+/// A partition spec is a list of partition fields, along with a unique integer ID.  A
+/// Table may have different partition specs over its lifetime due to partition spec
+/// evolution.
+class ICEBERG_EXPORT PartitionSpec : public util::Formattable {
+ public:
+  PartitionSpec(std::shared_ptr<Schema> schema, int32_t spec_id,
+                std::vector<PartitionField> fields);
+  /// \brief Get the table schema
+  const std::shared_ptr<Schema>& schema() const;
+  /// \brief Get the spec ID.
+  int32_t spec_id() const;
+  /// \brief Get a view of the partition fields.
+  std::span<const PartitionField> fields() const;
+
+  std::string ToString() const override;
+
+  friend bool operator==(const PartitionSpec& lhs, const PartitionSpec& rhs) {
+    return lhs.Equals(rhs);
+  }
+
+  friend bool operator!=(const PartitionSpec& lhs, const PartitionSpec& rhs) {
+    return !(lhs == rhs);
+  }
+
+ private:
+  /// \brief Compare two partition specs for equality.
+  [[nodiscard]] bool Equals(const PartitionSpec& other) const;
+
+  std::shared_ptr<Schema> schema_;
+  const int32_t spec_id_;
+  std::vector<PartitionField> fields_;
+};
+
+}  // namespace iceberg
--- a/src/iceberg/transform.h
+++ b/src/iceberg/transform.h
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/transform.h
+
+#include <cstdint>
+#include <memory>
+
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/error.h"
+#include "iceberg/expected.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/type_fwd.h"
+#include "iceberg/util/formattable.h"
+
+namespace iceberg {
+
+/// \brief Transform types used for partitioning
+enum class TransformType {
+  /// Used to represent some customized transform that can't be recognized or supported
+  /// now.
+  kUnknown,
+  /// Equal to source value, unmodified
+  kIdentity,
+  /// Hash of value, mod `N`
+  kBucket,
+  /// Value truncated to width `W`
+  kTruncate,
+  /// Extract a date or timestamp year, as years from 1970
+  kYear,
+  /// Extract a date or timestamp month, as months from 1970-01
+  kMonth,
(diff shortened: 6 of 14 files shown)


src/iceberg/partition_spec.h
```diff
@@ -0,0 +1,73 @@
...
+/// evolution.
+class ICEBERG_EXPORT PartitionSpec {
+ public:
+  virtual ~PartitionSpec() = default;
+  PartitionSpec(int32_t spec_id, std::vector<PartitionField> fields);
+  /// \brief Get the spec ID.
+  [[nodiscard]] int32_t spec_id() const;
+  /// \brief Get a view of the partition fields.
+  [[nodiscard]] virtual std::span<const PartitionField> fields() const;
+  /// \brief Get a field by field ID.
```
lidavidm: Should we clarify whether these are source/transform field IDs?

src/iceberg/partition_field.h
```diff
@@ -0,0 +1,78 @@
...
+  [[nodiscard]] int32_t source_field_id() const;
+
+  /// \brief Get the partition field ID.
+  [[nodiscard]] int32_t field_id() const;
+
+  /// \brief Get the partition field name.
+  [[nodiscard]] std::string_view name() const;
+
+  /// \brief Get the transform type.
+  [[nodiscard]] TransformType transform_type() const;
```
lidavidm: Java stores the full transform, not the type; do we need to do the same?

src/iceberg/transform.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#pragma once
+
+/// \file iceberg/transform.h
+
+#include <cstdint>
+#include <memory>
+
+#include "iceberg/arrow_c_data_internal.h"
```
wgtmac: Do not include any internal header in the public header file. To use Arrow C data, we can include `arrow_c_data.h`.

src/iceberg/transform.h
```diff
@@ -0,0 +1,81 @@
...
+  /// Extract a date or timestamp month, as months from 1970-01-01
+  kMonth,
+  /// Extract a date or timestamp day, as days from 1970-01-01
+  kDay,
+  /// Extract a timestamp hour, as hours from 1970-01-01 00:00:00
+  kHour,
+  /// Always produces `null`
+  kVoid,
+  /// Used to represent some customized transform that can't be recognized or supported now.
+  kUnknown,
```
wgtmac: What about putting `kUnknown` to the beginning?

src/iceberg/transform.h
```diff
@@ -0,0 +1,81 @@
...
+};
+
+/// \brief Get the relative transform name
+constexpr std::string_view ToString(TransformType type) {
+  /// Transform names used for formatting
+  constexpr std::string_view TransformTypeNames[] = {
+    "Identity", "Bucket", "Truncate", "Year", "Month", "Day", "Hour", "Void", "Unknown"
+  };
+  return TransformTypeNames[static_cast<size_t>(type)];
```
wgtmac: Is it undefined behavior to do this?

src/iceberg/transform.h
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/// \brief A transform function used for partitioning.
+class ICEBERG_EXPORT TransformFunction : public util::Formattable {
+ public:
+  explicit TransformFunction(TransformType type);
+  /// \brief Transform an input array to a new array
+  [[nodiscard]] virtual expected<std::unique_ptr<ArrowArray>, Error> Transform(const ArrowArray& inputArrowArray) = 0;
```
```suggestion
  virtual expected<ArrowArray, Error> Transform(const ArrowArray& data) = 0;
```

`expected` already has `nodiscard` attribute in its definition, we don't need it here.

For `ArrowArray`, usually we don't use `std::unique_ptr` since we have to explicitly call its release function.

src/iceberg/transform.h
```diff
@@ -0,0 +1,81 @@
...
+  /// \brief Get the transform type
+  [[nodiscard]] virtual TransformType transform_type() const;
+  [[nodiscard]] std::string ToString() const override;
```
```suggestion
  virtual TransformType transform_type() const;
  std::string ToString() const override;
```

I don't think we need to add nodiscard to these functions.

src/iceberg/partition_field.h
wgtmac: I think so. Some transform types required additional parameters like bucket number or truncate length.

src/iceberg/partition_field.h
```diff
@@ -0,0 +1,78 @@
...
+
+  friend bool operator!=(const PartitionField& lhs, const PartitionField& rhs) {
+    return !(lhs == rhs);
+  }
+
+ private:
+  /// \brief Compare two fields for equality.
+  [[nodiscard]] bool Equals(const PartitionField& other) const;
+
+  int32_t source_field_id_;
```
wgtmac: Should we consider multi-argument transform here? In that case, a partition field might have more than one source field ids.

See https://docs.google.com/document/d/1aDoZqRgvDOOUVAGhvKZbp5vFstjsAMY4EFCyjlxpaaw

src/iceberg/partition_spec.h
wgtmac: Perhaps we don't need these `GetFieldByXXX` at this moment?

src/iceberg/partition_spec.h
```diff
@@ -0,0 +1,73 @@
...
+
+/// \brief A partition spec for a Table.
+///
+/// A partition spec is a list of partition fields, along with a unique integer ID.  A
+/// Table may have different partition specs over its lifetime due to partition spec
+/// evolution.
+class ICEBERG_EXPORT PartitionSpec {
+ public:
+  virtual ~PartitionSpec() = default;
+  PartitionSpec(int32_t spec_id, std::vector<PartitionField> fields);
```
wgtmac: IMO, we need an `Schema` in the constructor to denote to the partition schema.

test/transform_test.cc
```diff
@@ -0,0 +1,66 @@
...
+    EXPECT_EQ("Identity", iceberg::ToString(iceberg::TransformType::kIdentity));
+    EXPECT_EQ("Bucket", iceberg::ToString(iceberg::TransformType::kBucket));
+    EXPECT_EQ("Truncate", iceberg::ToString(iceberg::TransformType::kTruncate));
+    EXPECT_EQ("Year", iceberg::ToString(iceberg::TransformType::kYear));
+    EXPECT_EQ("Month", iceberg::ToString(iceberg::TransformType::kMonth));
+    EXPECT_EQ("Day", iceberg::ToString(iceberg::TransformType::kDay));
+    EXPECT_EQ("Hour", iceberg::ToString(iceberg::TransformType::kHour));
+    EXPECT_EQ("Void", iceberg::ToString(iceberg::TransformType::kVoid));
+    EXPECT_EQ("Unknown", iceberg::ToString(iceberg::TransformType::kUnknown));
+  }
```
wgtmac:
```suggestion
    EXPECT_EQ("Identity", iceberg::ToString(iceberg::TransformType::kIdentity));
    EXPECT_EQ("Bucket", iceberg::ToString(iceberg::TransformType::kBucket));
    EXPECT_EQ("Truncate", iceberg::ToString(iceberg::TransformType::kTruncate));
    EXPECT_EQ("Year", iceberg::ToString(iceberg::TransformType::kYear));
    EXPECT_EQ("Month", iceberg::ToString(iceberg::TransformType::kMonth));
    EXPECT_EQ("Day", iceberg::ToString(iceberg::TransformType::kDay));
    EXPECT_EQ("Hour", iceberg::ToString(iceberg::TransformType::kHour));
    EXPECT_EQ("Void", iceberg::ToString(iceberg::TransformType::kVoid));
    EXPECT_EQ("Unknown", iceberg::ToString(iceberg::TransformType::kUnknown));
```

wgtmac: Add the `iceberg` namespace so we don't have to add `iceberg::` everywhere.

test/transform_test.cc
```diff
@@ -0,0 +1,66 @@
...
+    }
+  };
+
+  TestTransformFunction transform;
+  EXPECT_EQ(iceberg::TransformType::kUnknown, transform.transform_type());
+  EXPECT_EQ("Unknown", transform.ToString());
+  EXPECT_EQ("Unknown", std::format("{}", transform));
+
+  auto [_, arrow_array] =
+      iceberg::internal::CreateExampleArrowSchemaAndArrayByNanoarrow();
```
wgtmac: I will remove CreateExampleArrowSchemaAndArrayByNanoarrow some day. Perhaps add a TODO here?

src/iceberg/exception.h
```diff
@@ -38,4 +38,11 @@ class ICEBERG_EXPORT IcebergError : public std::runtime_error {
   explicit IcebergError(const std::string& what) : std::runtime_error(what) {}
 };
 
+#define ICEBERG_CHECK_FMT(condition, ...)                    \
```
wgtmac:
```suggestion
#define ICEBERG_CHECK(condition, ...)                    \
```

src/iceberg/partition_field.h
```diff
@@ -0,0 +1,78 @@
...
+
+/// \brief a field with its transform.
+class ICEBERG_EXPORT PartitionField : public util::Formattable {
+ public:
+  /// \brief Construct a field.
+  /// \param[in] source_id The source field ID.
+  /// \param[in] field_id The partition field ID.
+  /// \param[in] name The partition field name.
+  /// \param[in] transformType The transform type.
+  PartitionField(int32_t source_id, int32_t field_id, std::string name, TransformType transformType);
```
yingcai-cy: source_id->source_field_id? keep consistent with impl

wgtmac: I think it is ok to be consistent with the spec: https://iceberg.apache.org/spec/#partition-specs

```

| V1       | V2       | V3       | Field            | JSON representation | Example      |
|----------|----------|----------|------------------|---------------------|--------------|
| required | required | omitted  | **`source-id`**  | `JSON int`          | 1            |
|          |          | required | **`source-ids`** | `JSON list of ints` | `[1,2]`      |
|          | required | required | **`field-id`**   | `JSON int`          | 1000         |
| required | required | required | **`name`**       | `JSON string`       | `id_bucket`  |
| required | required | required | **`transform`**  | `JSON string`       | `bucket[16]` |
```

src/iceberg/partition_spec.h
yingcai-cy: > Perhaps we don't need these `GetFieldByXXX` at this moment?

+1

gty404: > Should we clarify whether these are source/transform field IDs?

I will change it to transform_fields, and the PartitionField class will define source_field_id/transform_field_id

gty404: > Perhaps we don't need these `GetFieldByXXX` at this moment?

Okay, I'll delete this part of the interface first

src/iceberg/partition_field.h
gty404: I will change it to TransformFunction

src/iceberg/transform.h
gty404: As long as the order of TransformTypes is not modified, there is no undefined behavior

test/transform_test.cc
gty404: Okay, I'll build a simple arrow array to test

src/iceberg/partition_field.h
gty404: iceberg-java used Transform, iceberg-rust used Transform, which is an enum type, and TransformFunction serves as the implementation interface for Transform. Which one does iceberg-cpp tend to use?  @wgtmac @yingcai-cy

src/iceberg/partition_spec.h
gty404: From the spec, it seems we should keep the fields.
image

wgtmac: Yes, fields are required

src/iceberg/CMakeLists.txt
```diff
@@ -23,7 +23,11 @@ set(ICEBERG_SOURCES
     schema.cc
     schema_field.cc
     schema_internal.cc
-    type.cc)
+    type.cc
```
wgtmac:
```suggestion
```

src/iceberg/CMakeLists.txt
```diff
@@ -23,7 +23,11 @@ set(ICEBERG_SOURCES
...
+    type.cc
+    transform.cc
+    partition_field.cc
```
wgtmac, line 26: Sort them alphabetically

src/iceberg/transform.h
```diff
@@ -0,0 +1,94 @@
...
+  kVoid,
+};
+
+/// \brief Get the relative transform name
+constexpr std::string_view ToString(TransformType type) {
+  /// Transform names used for formatting
+  constexpr std::string_view TransformTypeNames[] = {"Unknown",  "Identity", "Bucket",
+                                                     "Truncate", "Year",     "Month",
+                                                     "Day",      "Hour",     "Void"};
+  return TransformTypeNames[static_cast<size_t>(type)];
```
```suggestion
  constexpr std::string_view kTransformTypeNames[] = {"Unknown",  "Identity", "Bucket",
                                                      "Truncate", "Year",     "Month",
                                                      "Day",      "Hour",     "Void"};
  return kTransformTypeNames[static_cast<size_t>(type)];
```

src/iceberg/transform.h
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/// \brief A transform function used for partitioning.
+class ICEBERG_EXPORT TransformFunction : public util::Formattable {
+ public:
+  explicit TransformFunction(TransformType type);
+  /// \brief Transform an input array to a new array
+  virtual expected<ArrowArray, Error> Transform(const ArrowArray& inputArrowArray) = 0;
```
wgtmac: We use snake case for parameter name. So either `input_arrow_array` or simply `data`?

src/iceberg/partition_spec.h
```diff
@@ -0,0 +1,71 @@
...
+/// A partition spec is a list of partition fields, along with a unique integer ID.  A
+/// Table may have different partition specs over its lifetime due to partition spec
+/// evolution.
+class ICEBERG_EXPORT PartitionSpec {
+ public:
+  virtual ~PartitionSpec() = default;
+  PartitionSpec(std::shared_ptr<Schema> schema, int32_t spec_id,
+                std::vector<PartitionField> fields);
+  /// \brief Get the partition schema
+  [[nodiscard]] const std::shared_ptr<Schema>& schema() const;
```
```suggestion
  const std::shared_ptr<Schema>& schema() const;
```

src/iceberg/partition_spec.h
```diff
@@ -0,0 +1,71 @@
...
+  /// \brief Get the spec ID.
+  [[nodiscard]] int32_t spec_id() const;
```
```suggestion
  int32_t spec_id() const;
```

src/iceberg/partition_spec.h
```diff
@@ -0,0 +1,71 @@
...
+  /// \brief Get a view of the partition fields.
+  [[nodiscard]] virtual std::span<const PartitionField> fields() const;
```
```suggestion
  virtual std::span<const PartitionField> fields() const;
```

src/iceberg/partition_spec.h
```diff
@@ -0,0 +1,71 @@
...
+
+  [[nodiscard]] std::string ToString() const;
```
```suggestion
  std::string ToString() const;
```

src/iceberg/partition_spec.h
wgtmac: I think these trivial getters do not need to add `nodiscard`. Callers should know what they are doing.

src/iceberg/partition_spec.cc
```diff
@@ -0,0 +1,55 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/partition_spec.h"
+
+#include <format>
+
+#include "iceberg/exception.h"
```
```suggestion
```

src/iceberg/partition_spec.cc
```diff
@@ -0,0 +1,55 @@
...
+const std::shared_ptr<Schema>& PartitionSpec::schema() const { return schema_; }
+
+int32_t PartitionSpec::spec_id() const { return spec_id_; }
+
+std::span<const PartitionField> PartitionSpec::fields() const { return fields_; }
+
+std::string PartitionSpec::ToString() const {
+  std::string repr = "partition_spec(<";
+  for (const auto& field : fields_) {
+    std::format_to(std::back_inserter(repr), "  {}\n", field);
```
wgtmac, line 43: We'd better eliminate the whitespace before the 1st partition field.

BTW, should we include spec_id in it?

src/iceberg/partition_field.h
```diff
@@ -0,0 +1,79 @@
...
+  /// \brief Get the partition field ID.
+  [[nodiscard]] int32_t field_id() const;
+
+  /// \brief Get the partition field name.
+  [[nodiscard]] std::string_view name() const;
+
+  /// \brief Get the transform type.
+  [[nodiscard]] std::shared_ptr<TransformFunction> const& transform() const;
+
+  [[nodiscard]] std::string ToString() const override;
```
wgtmac:
```suggestion
  int32_t source_field_id() const;

  /// \brief Get the partition field ID.
  int32_t field_id() const;

  /// \brief Get the partition field name.
  std::string_view name() const;

  /// \brief Get the transform type.
  std::shared_ptr<TransformFunction> const& transform() const;

  std::string ToString() const override;
```

src/iceberg/partition_field.h
```diff
@@ -0,0 +1,79 @@
...
+  /// \brief Construct a field.
+  /// \param[in] source_id The source field ID.
+  /// \param[in] field_id The partition field ID.
+  /// \param[in] name The partition field name.
+  /// \param[in] transform The transform function.
+  PartitionField(int32_t source_id, int32_t field_id, std::string name,
+                 std::shared_ptr<TransformFunction> transform);
+
+  /// \brief Get the source field ID.
+  [[nodiscard]] int32_t source_field_id() const;
```
wgtmac: I think `source_id` is better and it aligns with the naming in the spec.

src/iceberg/transform.h
```diff
@@ -0,0 +1,94 @@
...
+constexpr std::string_view ToString(TransformType type) {
+  /// Transform names used for formatting
+  constexpr std::string_view TransformTypeNames[] = {"Unknown",  "Identity", "Bucket",
+                                                     "Truncate", "Year",     "Month",
+                                                     "Day",      "Hour",     "Void"};
+  return TransformTypeNames[static_cast<size_t>(type)];
+}
+
+/// \brief A transform function used for partitioning.
+class ICEBERG_EXPORT TransformFunction : public util::Formattable {
```
wgtmac, line 60: Why `TransformFunction` but not `Transform`?

test/partition_field_test.cc
```diff
@@ -0,0 +1,84 @@
...
+class TestTransformFunction : public TransformFunction {
+ public:
+  TestTransformFunction() : TransformFunction(TransformType::kUnknown) {}
+  expected<ArrowArray, Error> Transform(const ArrowArray& input) override {
+    return unexpected(
+        Error{.kind = ErrorKind::kNotSupported, .message = "test transform function"});
+  }
+};
+
+class IdentityTransformFunction : public TransformFunction {
```
wgtmac: Why not implementing this in the source code?

wgtmac (approved):
+1

wgtmac:
@lidavidm @zhjwpku @yingcai-cy Do you want to take a look?

src/iceberg/transform.h
```diff
@@ -0,0 +1,119 @@
...
+  [[nodiscard]] virtual bool Equals(const TransformFunction& other) const;
+
+  TransformType transform_type_;
+};
+
+class IdentityTransformFunction : public TransformFunction {
+ public:
+  IdentityTransformFunction() : TransformFunction(TransformType::kIdentity) {}
+  expected<ArrowArray, Error> Transform(const ArrowArray& input) override {
+    return input;
```
lidavidm: Is this actually correct? The output will share a lifetime with the input, which may be unexpected.

gty404: The implementation of identify transform is currently only used in test, I will implement these Transforms in the next pr.

zhjwpku (approved):
Looking good, thanks.

wgtmac:
@Fokko @Xuanwo Could you help review and merge it?

src/iceberg/transform.h
```diff
@@ -0,0 +1,118 @@
...
+  kUnknown,
+  /// Equal to source value, unmodified
+  kIdentity,
+  /// Hash of value, mod `N`
+  kBucket,
+  /// Value truncated to width `W`
+  kTruncate,
+  /// Extract a date or timestamp year, as years from 1970
+  kYear,
+  /// Extract a date or timestamp month, as months from 1970-01-01
```
Fokko: Nit, if we want to be consistent:
```suggestion
  /// Extract a date or timestamp month, as months from 1970-01
```

src/iceberg/transform.h
```diff
@@ -0,0 +1,118 @@
...
+
+/// \brief Get the relative transform name
+constexpr std::string_view ToString(TransformType type) {
+  switch (type) {
+    case TransformType::kUnknown:
+      return "Unknown";
+    case TransformType::kIdentity:
+      return "Identity";
+    case TransformType::kBucket:
+      return "Bucket";
```
Fokko: Should we align this with the [names in the spec](https://iceberg.apache.org/spec/#partition-transforms). Having `bucket[16]` instead of `Bucket`?

gty404: I understand that BucketTransformFunction::ToString returns bucket[16] because the number of buckets is specified when the partition table is created.

Fokko: That's correct. The same applies for `truncate[16]`, which will truncate until the first 16 characters (in the case of a String). Also, in the spec, everything is lowercase.

gty404: > That's correct. The same applies for `truncate[16]`, which will truncate until the first 16 characters (in the case of a String). Also, in the spec, everything is lowercase.

Sure, I have changed it to lowercase. Can you help me review it again?

Fokko: Would it be possible also to add the number of buckets for the bucketing transform, and the width for the truncate? Or is that complicated?

gty404: > Would it be possible also to add the number of buckets for the bucketing transform, and the width for the truncate? Or is that complicated?

As stated above, these values can only exist in the subclasses of TransformFunction. Here, we are just enumerating the string representations of the values. In another PR, I will implement BucketTransform and TruncateTransform and implement the ToString method to return the corresponding bucket[16] or truncate[16]. This is how I understand and plan it.

wgtmac: @Fokko This function is to print the transform type only. `bucket[N]` and `truncate[N]` will be printed via the `ToString` function of each concrete transform function implementation which is not yet implemented.

Fokko: Ah, I see, thanks! 👍

Fokko (approved):

PR #57, opened 2025-04-01
feat: convert arrow schema to iceberg schema

wgtmac:


wgtmac:
@lidavidm @zhjwpku Please take a look when you have time. Thanks!

src/iceberg/schema_internal.cc
```diff
@@ -183,11 +186,171 @@ expected<void, Error> ToArrowSchema(const Schema& schema, ArrowSchema* out) {
...
+      for (int i = 0; i < schema.n_children; i++) {
+        auto field_result = to_schema_field(*schema.children[i]);
+        if (!field_result) {
+          return unexpected<Error>(field_result.error());
+        }
+        fields.emplace_back(std::move(*field_result.value()));
+      }
+
+      return std::make_shared<StructType>(std::move(fields));
+    } break;
```
zhjwpku: Do we need this `break` here?

wgtmac: Good catch! I forgot to delete them...

lidavidm (approved):

zhjwpku (approved):
LGTM, thanks

wgtmac:
Thanks @lidavidm @zhjwpku!

@Fokko @Xuanwo Could you help review and merge this?

src/iceberg/schema_internal.cc
```diff
@@ -183,11 +186,171 @@ expected<void, Error> ToArrowSchema(const Schema& schema, ArrowSchema* out) {
   return {};
 }
 
+namespace {
+
+int32_t GetFieldId(const ArrowSchema& schema) {
+  if (schema.metadata == nullptr) {
+    return kUnknownFieldId;
```
Fokko, line 193: I think this is fine for now. Later on, we can fetch the field IDs from name mapping if they are not in the metadata. Or we can assign fresh IDs in case we want to create a new table from an Arrow schema 👍

Fokko:
Thanks @wgtmac for working on this, and thanks @lidavidm and @zhjwpku for the review 🙌

PR #58, opened 2025-04-02
feat:add init expression interface.

yingcai-cy:


src/iceberg/expression.h
```diff
@@ -0,0 +1,101 @@
...
+    kCount,
+    kCountStar,
+    kMax,
+    kMin
+  };
+
+  virtual ~Expression() = default;
+
+  /// \brief Returns the operation for an expression node.
+  [[nodiscard]] virtual Operation Op() const = 0;
```
lidavidm: What's the intent of having so many virtual methods? I would assume the expression tree itself (maybe not evaluation) should be concrete?

yingcai-cy: It is the interface of other specific expressions. Borrowed interface abstraction from Java api

lidavidm: Java seems to do that because each operation is its own subclass. That seems expensive and unnecessary for C++?

yingcai-cy: Use such inheritance abstraction makes the code clearer, rather than putting all the code into a single large class. Additionally, the use of virtual functions here will not cause significant performance issues because the expression evaluation is not intended for the engine—the engine can implement its own version may be codegen or something.

src/iceberg/error.h
```diff
@@ -38,6 +38,8 @@ enum class ErrorKind {
   kNotImplemented,
   kUnknownError,
   kNotSupported,
+  kInvalidExpression,
+  kInvalidOperatorType,
```
gty404: Sort them alphabetically ?

src/iceberg/expression.cc
```diff
@@ -0,0 +1,129 @@
...
+      return Operation::kAnd;
+    case Operation::kOr:
+      return Operation::kOr;
+    default:
+      return unexpected<Error>(
+          {ErrorKind::kInvalidOperatorType, "No left-right flip for operation"});
+  }
+}
+
+}  // namespace iceberg
```
zhjwpku: nit: keep this empty line at the EOF.
same for expression_test.cc

zhjwpku (commented):
I'm wondering should we add some subdirs below iceberg, this feature seems not be in the spec?

src/iceberg/expression.h
lidavidm: I'm not sure a 1:1 port of Java is necessarily more readable. Given most code should be trivial, one "large class" would be easier to manage than 20 copy-pasted ones

src/iceberg/expression.cc
lidavidm: We could set up pre-commit or something to enforce this automatically

src/iceberg/error.h
wgtmac:
```suggestion
  kInvalidOperation,
```

test/CMakeLists.txt
```diff
@@ -39,6 +39,11 @@ target_sources(expected_test PRIVATE expected_test.cc)
 target_link_libraries(expected_test PRIVATE iceberg_static GTest::gtest_main GTest::gmock)
 add_test(NAME expected_test COMMAND expected_test)
 
+add_executable(expression_test)
```
wgtmac, line 47: Should we reuse some test executables? I'm afraid that we are creating too many.

src/iceberg/expression.h
```diff
@@ -0,0 +1,101 @@
...
+#include "iceberg/expected.h"
+#include "iceberg/iceberg_export.h"
+
+namespace iceberg {
+
+/// \brief Represents a boolean expression tree.
+class ICEBERG_EXPORT Expression {
+ public:
+  /// Operation types for expressions
+  enum class Operation {
```
wgtmac, line 38:
```suggestion
  enum class Operation : uint8_t {
```

src/iceberg/expression.h
wgtmac:
```suggestion
  virtual Operation Op() const = 0;
```

src/iceberg/expression.h
```diff
@@ -0,0 +1,101 @@
...
+    kMin
+  };
+
+  virtual ~Expression() = default;
+
+  /// \brief Returns the operation for an expression node.
+  [[nodiscard]] virtual Operation Op() const = 0;
+
+  /// \brief Returns the negation of this expression, equivalent to not(this).
+  [[nodiscard]] virtual expected<std::shared_ptr<Expression>, Error> Negate() const {
```
wgtmac:
```suggestion
  virtual expected<std::shared_ptr<Expression>, Error> Negate() const {
```

wgtmac: Do we want to create a new one or change itself in place?

src/iceberg/expression.h
wgtmac: Should we move the enum definition to `type_fwd.h`?

src/iceberg/expression.h
```diff
@@ -0,0 +1,101 @@
...
+  ///
+  /// \param other another expression
+  /// \return true if the expressions are equivalent
+  [[nodiscard]] virtual bool IsEquivalentTo(const Expression& other) const {
+    // only bound predicates can be equivalent
+    return false;
+  }
+
+  /// \brief Convert operation string to enum
+  static expected<Operation, Error> FromString(const std::string& operation_type);
```
wgtmac: What about creating a separate `struct OperationUtil` or something for functions here and below? Or we can just use functions without any class.

wgtmac:
```suggestion
  static expected<Operation, Error> FromString(std::string_view op);
```

Use string_view when possible.

src/iceberg/expression.h
wgtmac: If we really want to use inheritance, what about CRTP? Something like:

```cpp
template <typename Derived>
class Expression {
 public:
  Operation op() const { return static_cast<const Derived*>(this)->op_impl(); }

  expected<Derived, Error> Negate() const {
    if constexpr (requires { static_cast<const Derived*>(this)->NegateImpl(); }) {
      return static_cast<const Derived*>(this)->NegateImpl();
    } else {
      return unexpected<Error>(
          {.kind = ErrorKind::kInvalidExpression, .message = "Cannot negate expression ..."});
    }
  }

  template <typename OtherDerived>
  bool IsEquivalentTo(const Expression<OtherDerived>& other) const {
    if constexpr (requires {
                    static_cast<const Derived*>(this)->IsEquivalentToImpl(other);
                  }) {
      return static_cast<const Derived*>(this)->IsEquivalentToImpl(other);
    } else {
      return false;
    }
  }

 protected:
  Expression() = default;
};
```

yingcai-cy: > I'm not sure a 1:1 port of Java is necessarily more readable. Given most code should be trivial, one "large class" would be easier to manage than 20 copy-pasted ones

Yes, that is a good point. I think Java already has clean simple API,  C++ can also have various small classes and interfaces, which is friendly to compilation speed and easy to read.

yingcai-cy: > If we really want to use inheritance, what about CRTP? Something like:
> 
> ```c++
> template <typename Derived>
> class Expression {
>  public:
>   Operation op() const { return static_cast<const Derived*>(this)->op_impl(); }
> 
>   expected<Derived, Error> Negate() const {
>     if constexpr (requires { static_cast<const Derived*>(this)->NegateImpl(); }) {
>       return static_cast<const Derived*>(this)->NegateImpl();
>     } else {
>       return unexpected<Error>(
>           {.kind = ErrorKind::kInvalidExpression, .message = "Cannot negate expression ..."});
>     }
>   }
> 
>   template <typename OtherDerived>
>   bool IsEquivalentTo(const Expression<OtherDerived>& other) const {
>     if constexpr (requires {
>                     static_cast<const Derived*>(this)->IsEquivalentToImpl(other);
>                   }) {
>       return static_cast<const Derived*>(this)->IsEquivalentToImpl(other);
>     } else {
>       return false;
>     }
>   }
> 
>  protected:
>   Expression() = default;
> };
> ```

IMO, we should keep the interface small clean to avoid any weird template compilation issues.

src/iceberg/error.h
yingcai-cy: Got it

src/iceberg/expression.h
lidavidm: Small classes aren't necessarily good for compilation speed (you pay the overhead of reading and parsing the headers over and over, which is why CMake etc. offer unity builds to _speed up_ compilation)

src/iceberg/expression.h
lidavidm: IMO we should avoid mutability

src/iceberg/expression.cc
yingcai-cy: Got it

src/iceberg/expression.h
alonesniper: Remove those functions for now. Add them back later when really needed.

test/CMakeLists.txt
yingcai-cy: OK, expresion could be a huge test because all kinds of operations. Will fix this later.

src/iceberg/expression.h
yingcai-cy: Will move to type_fwd.h later if necessary

src/iceberg/expression.h
```diff
@@ -0,0 +1,83 @@
...
+    kCount,
+    kCountStar,
+    kMax,
+    kMin
+  };
+
+  virtual ~Expression() = default;
+
+  /// \brief Returns the operation for an expression node.
+  virtual Operation Op() const = 0;
```
wgtmac:
```suggestion
  virtual Operation op() const = 0;
```

This is a trivial function so we can use lowercase name.

src/iceberg/expression.h
```diff
@@ -0,0 +1,83 @@
...
+
+  /// \brief Returns the negation of this expression, equivalent to not(this).
+  virtual Result<std::shared_ptr<Expression>> Negate() const {
```
wgtmac, line 70:
```suggestion
  virtual Result<std::unique_ptr<Expression>> Negate() const {
```

I'm inclined to use `unique_ptr` because it can be accepted as a `shared_ptr` but not vice versa.

src/iceberg/result.h
```diff
@@ -39,6 +39,8 @@ enum class ErrorKind {
   kNotImplemented,
   kUnknownError,
   kNotSupported,
+  kInvalidExpression,
+  kInvalidOperatorType,
```
wgtmac: kInvalidOperationType ? The enum is `Operation`.

src/iceberg/expressions/and.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include "iceberg/expression.h"
+
+namespace iceberg {
+
+/// \brief An Expression that represents a logical AND operation between two expressions.
+///
+/// This expression evaluates to true if and only if both of its child expressions
+/// evaluate to true.
+class ICEBERG_EXPORT And : public Expression {
```
wgtmac: Can we put all these expr impls into a single file? I'm afraid that we'll end up with too many files merely for exprs.

src/iceberg/expressions/false.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "iceberg/expression.h"
+
+namespace iceberg {
+
+/// \brief An expression that is always false.
+class ICEBERG_EXPORT False : public Expression {
+ public:
+  static const False& instance() {
+    static False instance;
```
wgtmac: We have to define static variables in the .cc file. Otherwise we'll have many of them in different compilation units and crash the program.

src/iceberg/expressions/false.h
```diff
@@ -0,0 +1,55 @@
...
+#include <string>
+
+#include "iceberg/expression.h"
+
+namespace iceberg {
+
+/// \brief An expression that is always false.
+class ICEBERG_EXPORT False : public Expression {
+ public:
+  static const False& instance() {
```
wgtmac:
```suggestion
  static const False& Instance() {
```

src/iceberg/expressions/false.h
```diff
@@ -0,0 +1,55 @@
...
+    static False instance;
+    return instance;
+  }
+
+  static const std::shared_ptr<Expression>& shared_instance() {
```
wgtmac: Is it enough to delete the above one and keep this one?

wgtmac: We can also extend std::enabled_from_this to use the same object.

src/iceberg/expression.h
```diff
@@ -0,0 +1,83 @@
...
+/// Expression interface for Iceberg table operations.
+
+#include "iceberg/expected.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+/// \brief Represents a boolean expression tree.
+class ICEBERG_EXPORT Expression {
```
wgtmac, line 35: Is it better to define all the subclass interfaces in this PR (without implementing them) to see the big picture of the class hierarchy and interface design?

src/iceberg/expressions/false.h
yingcai-cy: Actually it is safe after c++ 11, inline function and local static variable only initiallize once; I'll move implmentation to cc.

src/iceberg/expressions/false.h
yingcai-cy: Got it

src/iceberg/expression.h
yingcai-cy: If True/False objects are singletons, then they cannot return unique_ptr to change ownership. Maybe we should use shared_ptr for all expression methods, because sub expressions can be shared in many cases.

src/iceberg/result.h
yingcai-cy: Will remove it later.

src/iceberg/expression.h
```diff
@@ -0,0 +1,167 @@
...
+
+#include "iceberg/expected.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+/// \brief Type alias for shared pointer to Expression
+class Expression;
+using ExpressionPtr = std::shared_ptr<Expression>;
```
wgtmac: I'd suggest not to use this type alias because users are not sure whether it is unique_ptr or shared_ptr without another jump.

wgtmac (approved):

test/expression_test.cc
```diff
@@ -0,0 +1,67 @@
...
+
+#include "iceberg/expression.h"
+
+#include <memory>
+
+#include <gtest/gtest.h>
+
+namespace iceberg {
+
+TEST(TrueFalseTest, Basi) {
```
zhjwpku: %s/Basi/Basic/

test/expression_test.cc
```diff
@@ -0,0 +1,67 @@
...
+  // Test negation of False returns True
+  const std::shared_ptr<False> false_instance = False::Instance();
```
zhjwpku: maybe use auto to deduce? I see `False::Instance()` returns a ref of shared_ptr.

src/iceberg/expression.h
```diff
@@ -0,0 +1,167 @@
...
+
+/// \brief An Expression that represents a logical AND operation between two expressions.
+///
+/// This expression evaluates to true if and only if both of its child expressions
+/// evaluate to true.
+class ICEBERG_EXPORT And : public Expression {
+ public:
+  /// \brief Constructs an And expression from two sub-expressions.
+  ///
+  /// @param left The left operand of the AND expression
```
zhjwpku: We use backslash other than @ for the comments, maybe stick to the same rule.
I see you are using backslash in `Expression`, I guess it's an oversight here ;)

src/iceberg/expression.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include "expression.h"
+
+#include <format>
+
+namespace iceberg {
+
+// True implementation
+const std::shared_ptr<True>& True::Instance() {
+  static const std::shared_ptr<True> instance = std::shared_ptr<True>(new True());
```
lidavidm: why not make_shared?

test/expression_test.cc
lidavidm, line 1: This could benefit from a parameterized test

src/iceberg/expression.cc
```diff
@@ -0,0 +1,65 @@
...
+    : left_(std::move(left)), right_(std::move(right)) {}
+
+std::string And::ToString() const {
+  return std::format("({} and {})", left_->ToString(), right_->ToString());
+}
+
+Result<ExpressionPtr> And::Negate() const {
+  // TODO(yingcai-cy): Implement Or expression
+  return unexpected(
+      Error(ErrorKind::kInvalidExpression, "And negation not yet implemented"));
```
Fokko, line 53: This is pretty straightforward following Morgan's law: `not (A and B) = (not A) or (not B)`

Why not also add `Or` in this PR?

src/iceberg/expression.cc
```diff
@@ -0,0 +1,65 @@
...
+}
+
+bool And::Equals(const Expression& expr) const {
+  if (expr.op() == Operation::kAnd) {
+    const auto& other = static_cast<const And&>(expr);
+    return (left_->Equals(*other.left()) && right_->Equals(*other.right())) ||
+           (left_->Equals(*other.right()) && right_->Equals(*other.left()));
```
Fokko, line 60: In Java we call this `isEquivalentTo`: https://github.com/apache/iceberg/blob/97d1107d60b42d29306c663f23f869fc1d439e6b/api/src/main/java/org/apache/iceberg/expressions/Or.java#L43-L52, which is different from being equal. Should we follow the same naming for clarity?

src/iceberg/expression.h
```diff
@@ -0,0 +1,167 @@
...
+    kNotIn,
+    kNot,
+    kAnd,
+    kOr,
+    kStartsWith,
+    kNotStartsWith,
+    kCount,
+    kCountStar,
+    kMax,
+    kMin
```
Fokko, line 61: Do we want to add these right away? I would suggest adding these in a later PR, because:

- `Min`/`Max` can not always be obtained from the metadata, because the metadata is missing, or the value might be truncated (Iceberg applies `truncate[16]` by default on min/max statistics.
- Counts can be problematic to do purely from the metadata when you have merge on read deletes, or it cannot be determined from the metadata that the whole file needs to be included in the scan.

Therefore, I think it is more up to the query engine to implement this kind of operations.

Fokko: Ah, ignore this comment. I see that this has been added to Java as well 👍

src/iceberg/expression.cc
yingcai-cy: because True's constructor is private, can not call it from make_shared.

src/iceberg/expression.cc
yingcai-cy: I wanted to keep this PR simple by just providing the main interfaces and framework for everyone to review. Will add Or and other operators in follow-up PRs.

src/iceberg/expression.h
yingcai-cy: Thanks for pointing this out. Fixed.

test/expression_test.cc
yingcai-cy: Got it.

src/iceberg/expression.cc
yingcai-cy: Good suggestion! In the first version, it was isEquivalentTo. I followed the style of other classes in the codebase and switched to Equals for consistency.

lidavidm (approved):

zhjwpku (approved):
LGTM, thanks

Fokko (approved):
@yingcai-cy Can you rebase against main to resolve the conflict? Thanks!

Xuanwo (approved):
Thank you for this great work!

Fokko:
Thanks @yingcai-cy for working on this, and thanks @alonesniper, @gty404, @lidavidm, @wgtmac, @Xuanwo and @zhjwpku for the review 🙌 Great to have this in!

PR #59, opened 2025-04-02
feat: sort order

zhjwpku:


src/iceberg/sort_field.cc
```diff
@@ -0,0 +1,85 @@
...
+  return transform_;
+}
+
+SortDirection SortField::sort_direction() const { return sort_direction_; }
+
+NullOrder SortField::null_order() const { return null_order_; }
+
+std::string SortField::ToString() const {
+  return std::format(
+      "SortField(source_id={}, transform={}, sort_direction={}, null_order={})",
```
gty404: https://iceberg.apache.org/spec/#sort-orders

Should sort_direction be changed to direction according to the spec? Including member variable names.

zhjwpku: Reasonable, changed. Thanks.

test/sort_field_test.cc
```diff
@@ -0,0 +1,82 @@
...
+        "null_order=nulls-first)",
+        std::format("{}", field));
+  }
+}
+
+TEST(SortFieldTest, Equality) {
+  auto test_transform = std::make_shared<TestTransformFunction>();
+  auto identity_transform = std::make_shared<IdentityTransformFunction>();
+
+  SortField field1(1, test_transform, SortDirection::kAscending, NullOrder::kFirst);
```
wgtmac, line 66: nit: using number as variable name suffix really degrades the readability...

wgtmac (approved):
LGTM

src/iceberg/sort_field.cc
```diff
@@ -0,0 +1,84 @@
...
+  return transform_;
+}
+
+SortDirection SortField::direction() const { return direction_; }
+
+NullOrder SortField::null_order() const { return null_order_; }
+
+std::string SortField::ToString() const {
+  return std::format(
+      "sort_field(source_id={}, transform={}, direction={}, null_order={})", source_id_,
```
wgtmac, line 75: nit: it seems that the spec prints `transform` before `source_id`

zhjwpku: There a slight inconsistency in the spec, in `#sorting` section `source id` appears before `transform`, while in appendix `transform` before `source id`. iceberg-rust put `source_id` before `transform` in its display fmt[1], so I prefer to keep it as is ;)

[1] https://github.com/apache/iceberg-rust/blob/main/crates/iceberg/src/spec/sort.rs#L89-L97

test/sort_field_test.cc
zhjwpku: I copied this from partition_filed_test.c, we seem to have adopted this style in quite a few places, no idea how to make it better though I do agree this is bad for readability.

zhjwpku:
@Fokko @Xuanwo Please take a look at this PR when you are available, thanks ;)

wgtmac:
Do you have any comment? @lidavidm

lidavidm (approved):

Fokko (approved):

Fokko:
Thanks @zhjwpku for working on this, and thanks @wgtmac, @lidavidm and @gty404 for the reviews 🙌

PR #60, opened 2025-04-07
feat: snapshot

zhjwpku:


src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+#include <optional>
+#include <string>
+#include <unordered_map>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/util/formattable.h"
+
+namespace iceberg {
+
+/// Optional Snapshot Summary Fields
```
lidavidm: These will all go in the toplevel namespace. Should we have a subnamespace just for the constants?

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+/// \brief The original id of a cherry-picked snapshot
+constexpr std::string_view kSourceSnapshotIDKey = "source-snapshot-id";
+/// \brief Name of the engine that created the snapshot
+constexpr std::string_view kEngineNameKey = "engine-name";
+/// \brief Version of the engine that created the snapshot
+constexpr std::string_view kEngineVersionKey = "engine-version";
+
+/// \brief The operation field is used by some operations, like snapshot expiration, to
+/// skip processing certain snapshots.
+enum class Operation {
```
lidavidm: Same here - is there ever a chance we will have another "Operation"-like enum?

src/iceberg/snapshot.cc
```diff
@@ -0,0 +1,123 @@
...
+const std::string& Snapshot::manifest_list() const { return manifest_list_; }
+
+const std::shared_ptr<Summary>& Snapshot::summary() const { return summary_; }
+
+std::optional<int32_t> Snapshot::schema_id() const { return schema_id_; }
+
+std::string Snapshot::ToString() const {
+  std::string repr = "snapshot: { id: " + std::to_string(snapshot_id_);
+  if (parent_snapshot_id_.has_value()) {
+    repr += ", parent_id: " + std::to_string(parent_snapshot_id_.value());
```
lidavidm: You can use `std::format_to(std::back_inserter(repr), ...)`

src/iceberg/snapshot.cc
```diff
@@ -0,0 +1,123 @@
...
+
+const std::unordered_map<std::string, std::string>& Summary::properties() const {
+  return additional_properties_;
+}
+
+std::string Summary::ToString() const {
+  std::string repr =
+      "summary: { operation: " + std::string(iceberg::ToString(operation_));
+  for (const auto& [key, value] : additional_properties_) {
+    repr += ", " + key + ": " + value;
```
lidavidm: Same here, use format_to

src/iceberg/snapshot.cc
```diff
@@ -0,0 +1,123 @@
...
+
+#include "iceberg/snapshot.h"
+
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+namespace {
+/// \brief Get the relative Operation name
+constexpr std::string_view ToString(Operation operation) {
```
lidavidm: Is this useful to expose at the top level? Maybe we can call it OperationToString? (Or just let overloading do its thing...)

We could also write a std::format specialization so that you can just pass Operation directly to `format`

src/iceberg/snapshot_ref.h
```diff
@@ -0,0 +1,57 @@
...
+enum class SnapshotRefType {
+  /// Branches are mutable named references that can be updated by committing a new
+  /// snapshot as the branch’s referenced snapshot using the Commit Conflict Resolution
+  /// and Retry procedures.
+  BRANCH,
+  /// Tags are labels for individual snapshots
+  TAG,
+};
+
+struct ICEBERG_EXPORT SnapshotRef {
```
lidavidm: nit: docstring?

src/iceberg/snapshot_ref.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "iceberg/iceberg_export.h"
+
+namespace iceberg {
+
+enum class SnapshotRefType {
+  /// Branches are mutable named references that can be updated by committing a new
+  /// snapshot as the branch’s referenced snapshot using the Commit Conflict Resolution
+  /// and Retry procedures.
+  BRANCH,
```
lidavidm: Shouldn't it be kBranch for consistency?

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+///
+/// A snapshot consist of one or more file manifests, and the complete table contents is
+/// the union of all the data files in those manifests.
+///
+/// Snapshots are created by table operations.
+class ICEBERG_EXPORT Snapshot : public iceberg::util::Formattable {
+ public:
+  Snapshot(int64_t snapshot_id, std::optional<int64_t> parent_snapshot_id,
+           int64_t sequence_number, int64_t timestamp_ms, std::string manifest_list,
+           std::shared_ptr<Summary> summary, std::optional<int64_t> schema_id);
```
lidavidm: Will a Summary actually get shared across multiple snapshots? Shouldn't it be unique_ptr or even better just use Summary directly?

src/iceberg/snapshot.h
wgtmac: There is an `Operation` type in the `Expression`.

wgtmac:
```suggestion
enum class DataOperation {
```

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+  /// A timestamp when the snapshot was created, used for garbage collection and table
+  /// inspection.
+  int64_t timestamp_ms_;
+  /// The location of a manifest list for this snapshot that tracks manifest files with
+  /// additional metadata.
+  std::string manifest_list_;
+  /// A string map that summaries the snapshot changes, including operation.
+  std::shared_ptr<Summary> summary_;
+  /// ID of the table's current schema when the snapshot was created.
+  std::optional<int32_t> schema_id_;
```
wgtmac: Should we reserve `first_row_id_` and `added_rows_` for v3 here?

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+  /// A monotonically increasing long that tracks the order of changes to a table.
+  int64_t sequence_number_;
+  /// A timestamp when the snapshot was created, used for garbage collection and table
+  /// inspection.
+  int64_t timestamp_ms_;
+  /// The location of a manifest list for this snapshot that tracks manifest files with
+  /// additional metadata.
+  std::string manifest_list_;
+  /// A string map that summaries the snapshot changes, including operation.
+  std::shared_ptr<Summary> summary_;
```
wgtmac: Why not directly using `std::map`?

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+  /// The snapshot ID of the snapshot's parent. Omitted for any snapshot with no parent.
+  std::optional<int64_t> parent_snapshot_id_;
+  /// A monotonically increasing long that tracks the order of changes to a table.
+  int64_t sequence_number_;
+  /// A timestamp when the snapshot was created, used for garbage collection and table
+  /// inspection.
+  int64_t timestamp_ms_;
+  /// The location of a manifest list for this snapshot that tracks manifest files with
+  /// additional metadata.
+  std::string manifest_list_;
```
wgtmac: How should we deal with a list of manifest files for v1 (where it does not have a manifest list file)?

src/iceberg/snapshot.h
wgtmac: Agreed. It might be better to use a sub-namespace or group them as static variables in classes like:
```cpp
struct SnapshotSummaryFields {
  constexpr static std::string_view kOperation = "operation";
  constexpr static std::string_view kAddedDataFiles = "added-data-files";
  ...
};
```

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+///
+/// Snapshots are created by table operations.
+class ICEBERG_EXPORT Snapshot : public iceberg::util::Formattable {
+ public:
+  Snapshot(int64_t snapshot_id, std::optional<int64_t> parent_snapshot_id,
+           int64_t sequence_number, int64_t timestamp_ms, std::string manifest_list,
+           std::shared_ptr<Summary> summary, std::optional<int64_t> schema_id);
+
+  /// \brief Get the id of the snapshot.
+  [[nodiscard]] int64_t snapshot_id() const;
```
wgtmac:
```suggestion
  int64_t snapshot_id() const;
```

I would suggest removing all these `nodiscard` attributes here

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,217 @@
...
+  [[nodiscard]] std::optional<int64_t> parent_snapshot_id() const;
+
+  /// \brief Get the sequence number of the snapshot.
+  [[nodiscard]] int64_t sequence_number() const;
+
+  /// \brief Get the timestamp of the snapshot.
+  [[nodiscard]] int64_t timestamp_ms() const;
+
+  /// \brief Get the manifest list of the snapshot.
+  [[nodiscard]] const std::string& manifest_list() const;
```
wgtmac: This might be optional in v1

src/iceberg/snapshot.cc
```diff
@@ -0,0 +1,123 @@
...
+
+  return repr;
+}
+
+bool Snapshot::Equals(const Snapshot& other) const {
+  return snapshot_id_ == other.snapshot_id_ &&
+         parent_snapshot_id_ == other.parent_snapshot_id_ &&
+         sequence_number_ == other.sequence_number_ &&
+         timestamp_ms_ == other.timestamp_ms_ && manifest_list_ == other.manifest_list_ &&
+         *summary_ == *other.summary_ && schema_id_ == other.schema_id_;
```
wgtmac: Do we really want to compare summary here?

FYI the Java impl is 
```java
  public boolean equals(Object o) {
    if (this == o) {
      return true;
    }

    if (o instanceof BaseSnapshot) {
      BaseSnapshot other = (BaseSnapshot) o;
      return this.snapshotId == other.snapshotId()
          && Objects.equal(this.parentId, other.parentId())
          && this.sequenceNumber == other.sequenceNumber()
          && this.timestampMillis == other.timestampMillis()
          && Objects.equal(this.schemaId, other.schemaId());
    }

    return false;
  }
```

wgtmac (changes requested):

src/iceberg/snapshot.h
lidavidm: In that case we need a separate namespace

src/iceberg/snapshot_ref.h
```diff
@@ -0,0 +1,57 @@
...
+  /// Tags are labels for individual snapshots
+  TAG,
+};
+
+struct ICEBERG_EXPORT SnapshotRef {
+  /// A reference's snapshot ID. The tagged snapshot or latest snapshot of a branch.
+  int64_t snapshot_id_;
```
gty404: In the struct, the variable name should be snapshot_id? The following are also

src/iceberg/snapshot.h
zhjwpku: Done this way.

src/iceberg/snapshot.h
zhjwpku: Is it acceptable by embedding the Operation in Summary class?

src/iceberg/snapshot.cc
zhjwpku: done.

src/iceberg/snapshot.cc
zhjwpku: done

src/iceberg/snapshot.cc
zhjwpku: It's in an anonymous namespace, so I guess it would not be exposed outside this cc file?

src/iceberg/snapshot_ref.h
zhjwpku: done.

src/iceberg/snapshot.h
zhjwpku: I ended with just use Summary directly.

src/iceberg/snapshot.h
zhjwpku: I embed this Operation in Summary class. I can change to your suggestion if my way is not acceptable.

src/iceberg/snapshot.h
zhjwpku: I'm not sure about this, maybe leave this for future work? I remembered we agreed only touch v1 and v2 for the MVP, so I didn't add these two fields.

src/iceberg/snapshot.h
zhjwpku: I changed to just use Summary here, do you think it's ok?

src/iceberg/snapshot.h
zhjwpku: removed.

src/iceberg/snapshot.cc
zhjwpku: changed to same as java impl.

src/iceberg/snapshot_ref.h
zhjwpku: agree, thanks.

src/iceberg/snapshot.h
zhjwpku: Maybe we can define two more structures SnapshotV1 and SnapshotV2, and do some transformation during json serialization? iceberg-rust implement this way, @Xuanwo do you have any suggestions here?

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,213 @@
...
+  std::unordered_map<std::string, std::string> additional_properties_;
+};
+
+/// \brief A snapshot of the data in a table at a point in time.
+///
+/// A snapshot consist of one or more file manifests, and the complete table contents is
+/// the union of all the data files in those manifests.
+///
+/// Snapshots are created by table operations.
+class ICEBERG_EXPORT Snapshot : public iceberg::util::Formattable {
```
wgtmac: Is it much simpler to use a struct without those trivial getters and constructor for `Snapshot`? I did this to `TableMetadata` because it is a metadata-carrying object without any logic. We can also use [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization) to deal with optional fields in a more flexible way. I think `Snapshot` and even `SortOrder` share the same attribute.

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,213 @@
...
+    kDelete,
+  };
+  Summary() = default;
+  /// \brief Construct a summary with the given operation and properties.
+  Summary(Operation op, std::unordered_map<std::string, std::string> props);
+
+  /// \brief Get the operation type of the snapshot.
+  Operation operation() const;
+
+  /// \brief Get the additional properties of the snapshot.
```
wgtmac: For `additional`, we assume that `operation` is not included in the map? If that's true, we need extra handling to deal with its JSON serialization.

src/iceberg/snapshot_ref.h
```diff
@@ -0,0 +1,59 @@
...
+  /// Branches are mutable named references that can be updated by committing a new
+  /// snapshot as the branch’s referenced snapshot using the Commit Conflict Resolution
+  /// and Retry procedures.
+  kBranch,
+  /// Tags are labels for individual snapshots
+  kTag,
+};
+
+/// \brief A reference to a snapshot, either a branch or a tag.
+struct ICEBERG_EXPORT SnapshotRef {
```
wgtmac: What about moving this to `snapshot.h`?

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,213 @@
...
+  const std::unordered_map<std::string, std::string>& properties() const;
+
+  std::string ToString() const override;
+
+ private:
+  /// The type of operation in the snapshot
+  Operation operation_{Operation::kAppend};
```
wgtmac: I know that iceberg-rust has defined `Summary` and `Operation`. What should we do if "operation" key is not set in the map, or if it is an unknown operation key? The Java implementation has the following signature:

```
  /**
   * Return the name of the {@link DataOperations data operation} that produced this snapshot.
   *
   * @return the operation that produced this snapshot, or null if the operation is unknown
   * @see DataOperations
   */
  String operation();
```

wgtmac: Is it simpler to remove the `Summary` class and define `std::unordered_map<std::string, std::string> summary` to the `Snapshot` class by adding a function to get operation from the summary map?

src/iceberg/snapshot.h
wgtmac: What about this:

```cpp
using manifest_list_t = std::string;
using manifests_t = std::vector<std::string>;
std::variant<manifest_list_t, manifests_t> manifest_list;

std::optional<std::reference_wrapper<manifest_list_t>> ManifestList() const;
std::optional<std::reference_wrapper<manifests_t>> Manifests() const;
```

zhjwpku: done this way.

src/iceberg/snapshot.h
zhjwpku: changed.

src/iceberg/snapshot_ref.h
zhjwpku: done

src/iceberg/snapshot.h
zhjwpku: changed to this.

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,224 @@
...
+  std::optional<int32_t> min_snapshots_to_keep;
+  /// For branch type only, a positive number for the max age of snapshots to keep when
+  /// expiring, including the latest snapshot. Defaults to table property
+  /// history.expire.max-snapshot-age-ms.
+  std::optional<int64_t> max_snapshot_age_ms;
+  /// For snapshot references except the main branch, a positive number for the max age of
+  /// the snapshot reference to keep while expiring snapshots. Defaults to table property
+  /// history.expire.max-ref-age-ms. The main branch never expires.
+  std::optional<int64_t> max_ref_age_ms;
+};
```
lidavidm: Instead of optional, would this make more sense? Then we can avoid having 3 `optional` fields that must stay in sync and be individually checked

```cpp
struct SnapshotRef {
  int64_t snapshot_id;

  std::variant<Branch, Tag> snapshot_type;

  struct Branch {
    int32_t min_snapshots_to_keep;
    // ...
  };
  struct Tag {};

  SnapshotRefType type() const noexcept {
    return std::visit(...);
  }
};
```

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,224 @@
...
+
+/// \brief A snapshot of the data in a table at a point in time.
+///
+/// A snapshot consist of one or more file manifests, and the complete table contents is
+/// the union of all the data files in those manifests.
+///
+/// Snapshots are created by table operations.
+struct ICEBERG_EXPORT Snapshot {
+  using manifest_list_t = std::string;
+  using manifests_t = std::vector<std::string>;
```
lidavidm: If the idea is to use these in a variant, maybe a newtype would be better? 

```cpp
struct ManifestList {
  std::string manifest_list_path;
};
struct Manifests {
  std::vector<std::string> manifest_paths;
};
std::variant<ManifestList, Manifests> manifest_list;
```

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,224 @@
...
+  std::unordered_map<std::string, std::string> summary;
+  /// ID of the table's current schema when the snapshot was created.
+  std::optional<int32_t> schema_id;
+
+  /// \brief Return the name of the DataOperations data operation that produced this
+  /// snapshot.
+  ///
+  /// \return the operation that produced this snapshot, or nullopt if the operation is
+  /// unknown.
+  std::optional<std::string> operation() const;
```
lidavidm: nit: maybe string_view? Assuming this is a reference into `summary`

src/iceberg/snapshot.cc
```diff
@@ -0,0 +1,62 @@
...
+#include "iceberg/snapshot.h"
+
+#include <format>
+
+#include "iceberg/util/formatter.h"
+
+namespace iceberg {
+
+std::optional<std::string> Snapshot::operation() const {
+  auto it = summary.find(std::string(SnapshotSummaryFields::kOperation));
```
lidavidm: (I thought modern C++ allowed you to index with string_view directly?)

If we're going to be wrapping with `std::string` all the time maybe let's just make them actual `static std::string`?

src/iceberg/snapshot.h
zhjwpku: Just checked the java and rust implementation, even after splitting Branch and Tag, we still need the field as optional? 

```
public class SnapshotRef implements Serializable {

  public static final String MAIN_BRANCH = "main";

  private final long snapshotId;
  private final SnapshotRefType type;
  private final Integer minSnapshotsToKeep;
  private final Long maxSnapshotAgeMs;
  private final Long maxRefAgeMs;
```

src/iceberg/snapshot.cc
zhjwpku: I tried to change to `static constexpr std::string`, but for string longer than 24 bytes(including the null-terminator), the compiler complains `Constexpr variable 'xxxx' must be initialized by a constant`, so for `SnapshotSummaryFields`, I end up with `static const std::string`.

src/iceberg/snapshot.h
zhjwpku: changed to this.

src/iceberg/snapshot.h
zhjwpku: done.

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,229 @@
...
+  /// \brief Number of equality delete files added in the snapshot
+  static const std::string kAddedEqDeleteFiles;
+  /// \brief Number of equality delete files removed in the snapshot
+  static const std::string kRemovedEqDeleteFiles;
+  /// \brief Number of position delete files added in the snapshot
+  static const std::string kAddedPosDeleteFiles;
+  /// \brief Number of position delete files removed in the snapshot
+  static const std::string kRemovedPosDeleteFiles;
+  /// \brief Number of deletion vectors added in the snapshot
+  static const std::string kAddedDVS;
```
Fokko: nit: Shouldn't `s` be lowercase?
```suggestion
  static const std::string kAddedDVs;
```

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,229 @@
...
+  int64_t snapshot_id;
+  /// The snapshot ID of the snapshot's parent. Omitted for any snapshot with no parent.
+  std::optional<int64_t> parent_snapshot_id;
+  /// A monotonically increasing long that tracks the order of changes to a table.
+  int64_t sequence_number;
+  /// A timestamp when the snapshot was created, used for garbage collection and table
+  /// inspection.
+  int64_t timestamp_ms;
+  /// The location of a manifest list for this snapshot that tracks manifest files with
+  /// additional metadata(v2) or a list of manifest file locations(v1).
```
Fokko: I think what this is referring to is the `manifests field`:

image

I think we should leave this out entirely, like we did with PyIceberg.

```suggestion
  /// The location of a manifest list for this snapshot
```

The `manifests` is used only used before it was open source, but it quickly turned out that using the manifest-list is the better option to keep the `manifest.json` within reasonable size.

src/iceberg/snapshot.h
lidavidm: I don't think Rust does that? Rust does what I suggested

https://github.com/apache/iceberg-rust/blob/a229d89db8ae8f443a5475a060a7016f9584230b/crates/iceberg/src/spec/snapshot.rs#L359-L388

Java can do what I suggested but would require a relatively recent Java (via sealed interfaces)

lidavidm: Oh I see. Hmm, I guess we do have to keep optional, annoyingly. But at least we can delineate which fields are valid or not.

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,206 @@
...
+  /// For snapshot references except the main branch, a positive number for the max age of
+  /// the snapshot reference to keep while expiring snapshots. Defaults to table property
+  /// history.expire.max-ref-age-ms. The main branch never expires.
+  std::optional<int64_t> max_ref_age_ms;
+};
+
+/// \brief Optional Snapshot Summary Fields
+struct SnapshotSummaryFields {
+  /// \brief The operation field key
+  static const std::string kOperation;
```
wgtmac: Why not using `constexpr` for these fields like you did in `DataOperation`?

src/iceberg/snapshot.h
wgtmac: Thanks @Fokko! This context is very helpful!

src/iceberg/snapshot.cc
wgtmac: What about `inline static const std::string kFoo = "bar";` to avoid defining these in the .cc file?

zhjwpku: good point, will change.

src/iceberg/snapshot.h
zhjwpku: reasonable, will try, thanks

src/iceberg/snapshot.h
lidavidm: You can't (really) have a constexpr string

zhjwpku: So for consistency, I will change the `static constexpr string` to `inline static const string`, WDYT?

lidavidm: SGTM

wgtmac: We can do this for strings shorter than 24 bytes because of SSO? That behavior may depend on the compilers.

lidavidm: Ah...Yeah, I'd rather not depend on compiler/libc++ behavior if possible

src/iceberg/snapshot.h
zhjwpku: @lidavidm It would be great if you could help check whether I did it right.

src/iceberg/snapshot.h
```diff
@@ -0,0 +1,218 @@
...
+    /// For snapshot references except the main branch, a positive number for the max age
+    /// of the snapshot reference to keep while expiring snapshots. Defaults to table
+    /// property history.expire.max-ref-age-ms. The main branch never expires.
+    std::optional<int64_t> max_ref_age_ms;
+  };
+
+  /// A reference's snapshot ID. The tagged snapshot or latest snapshot of a branch.
+  int64_t snapshot_id;
+  /// Snapshot retention policy
+  std::variant<Branch, Tag> retention;
```
wgtmac, line 69: It is a little bit weird to name it `retention` though now it is all about retention. Another weird thing is that both `snapshot_id` and `max_ref_age_ms` are common fields to branch/tag but they appear in the different places.

wgtmac:
```
struct Branch {
  int64_t snapshot_id;
  std::optional<int32_t> min_snapshots_to_keep;
  std::optional<int64_t> max_snapshot_age_ms;
  std::optional<int64_t> max_ref_age_ms;
};

struct Tag {
  int64_t snapshot_id;
  std::optional<int64_t> max_ref_age_ms;
};

using SnapshotRef = std::variant<Branch, Tag>;

or 

struct SnapshotRef {
  std::variant<Branch, Tag> ref;
  SnapshotRefType ref_type() const {
    if (std::holds_alternative<Branch>(ref)) {
      return SnapshotRefType::kBranch;
    } else {
      return SnapshotRefType::kTag;
    }
  }
};
```

I'm not sure if this has been discussed.

zhjwpku: Yeah, I just copied the rust impl. I don’t have a strong opinion on this, let's wait for @lidavidm's comments.

lidavidm: I would prefer not having `using` for this. I think if `retention` weird we can call it `snapshot_type` or something?

src/iceberg/snapshot.cc
```diff
@@ -0,0 +1,55 @@
...
+    return it->second;
+  }
+  return std::nullopt;
+}
+
+bool Snapshot::Equals(const Snapshot& other) const {
+  if (this == &other) {
+    return true;
+  }
+  return snapshot_id == other.snapshot_id &&
```
lidavidm, line 49: Compare `retention`?

lidavidm (approved):

src/iceberg/snapshot.h
wgtmac: I'm fine with `retention` for now. I don't think we need to waste too much time on this.

wgtmac (approved):

src/iceberg/snapshot.cc
zhjwpku: Here I just follow the java impl, suggested by @wgtmac. We can add the retention if needed in the future.

```
  public boolean equals(Object o) {
    if (this == o) {
      return true;
    }

    if (o instanceof BaseSnapshot) {
      BaseSnapshot other = (BaseSnapshot) o;
      return this.snapshotId == other.snapshotId()
          && Objects.equal(this.parentId, other.parentId())
          && this.sequenceNumber == other.sequenceNumber()
          && this.timestampMillis == other.timestampMillis()
          && Objects.equal(this.schemaId, other.schemaId());
    }

    return false;
  }
```

zhjwpku:
@Fokko @Xuanwo I think the community have reached the consensus, and I have resolved the merge conflict, it would be great if you can help merge this, thanks.

Xuanwo (approved):
The community has reached a consensus—let's move forward!

PR #61, opened 2025-04-08
feat: transform function

gty404:


src/iceberg/transform/transform_factory.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include <memory>
+
+#include "iceberg/error.h"
+#include "iceberg/expected.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Factory class for creating TransformFunction instances from TransformSpec.
```
lidavidm: Can we just have `TransformFunction::Make`? Especially because `create` is static anyways so there is no value to having a factory (+factory is very much a Java-ism that doesn't offer anything here)

src/iceberg/transform.h
```diff
@@ -56,14 +56,45 @@ enum class TransformType {
...
-  explicit TransformFunction(TransformType type);
+  explicit TransformFunction(TransformType transform_type,
+                             std::shared_ptr<Type> source_type);
   /// \brief Transform an input array to a new array
   virtual expected<ArrowArray, Error> Transform(const ArrowArray& data) = 0;
   /// \brief Get the transform type
-  virtual TransformType transform_type() const;
+  TransformType transform_type() const;
+  /// \brief Get the source type of transform function
+  std::shared_ptr<Type> source_type() const;
```
lidavidm: `const&` to avoid forcing a copy

src/iceberg/transform.h
```diff
@@ -56,14 +56,45 @@ enum class TransformType {
...
+  /// \brief Get the result type of transform function
+  virtual expected<std::shared_ptr<Type>, Error> result_type() const = 0;
```
lidavidm: `snake_case` is only for trivial getters. This should either be `shared_ptr<Type> const& result_type() const` or `expected<...> ResultType() const`.

src/iceberg/transform/transform_spec.h
```diff
@@ -0,0 +1,117 @@
...
+  /// \brief The type of the transform (e.g., kBucket, kTruncate, kIdentity, etc.)
+  TransformType transform_type;
+
+  /// \brief Optional parameter values passed to the transform, represented as an
+  /// ArrowArray.
+  ///
+  /// For transforms that require parameters (e.g., Bucket(N)), this holds the arguments
+  /// as a primitive ArrowArray (e.g., INT32 for num_buckets or width).
+  /// If the transform does not require parameters, this will be std::nullopt.
+  std::optional<ArrowArray> params_opt;
```
lidavidm: Hmm, do the parameters need to be Arrow data? I suppose what we need is a standard for scalars or row-wise data...

lidavidm: Do we actually need TransformSpec though?

src/iceberg/transform.h
```diff
@@ -56,14 +56,45 @@ enum class TransformType {
...
+      return "year";
+    case TransformType::kMonth:
+      return "month";
+    case TransformType::kDay:
+      return "day";
+    case TransformType::kHour:
+      return "hour";
+    case TransformType::kVoid:
+      return "void";
+    default:
```
wgtmac: It seems that we can remove the `default` branch

src/iceberg/arrow_c_data.h
```diff
@@ -73,4 +73,44 @@ struct ArrowArray {
 
 #endif  // ARROW_C_DATA_INTERFACE
 
+#ifndef ARROW_C_STREAM_INTERFACE
+#  define ARROW_C_STREAM_INTERFACE
+
+struct ArrowArrayStream {
```
wgtmac: Why do we need this?

src/iceberg/transform.h
```diff
@@ -56,14 +56,45 @@ enum class TransformType {
...
+      return "invalid";
+  }
+}
+
 /// \brief A transform function used for partitioning.
 class ICEBERG_EXPORT TransformFunction : public util::Formattable {
  public:
-  explicit TransformFunction(TransformType type);
+  explicit TransformFunction(TransformType transform_type,
```
wgtmac: We can remove `explicit` now

src/iceberg/transform/transform_spec.h
wgtmac: I was thinking about the following approach:

```cpp
struct TransformUtil {
  // TODO: source and result types may be required here
  static expected<ArrowArray, Error> Copy(const ArrowArray& input);
  static expected<ArrowArray, Error> Bucket(const ArrowArray& input, int32_t num_buckets);
  static expected<ArrowArray, Error> Truncate(const ArrowArray& input, int32_t width);
};

class Transform {
 public:
  virtual ~Transform() = default;

  expected<ArrowArray, Error> Apply(const ArrowArray& input) const {
    return transform_function_(input);
  }

 protected:
  Transform(
      TransformType transform_type, std::shared_ptr<Type> source_type,
      std::shared_ptr<Type> result_type,
      std::function<expected<ArrowArray, Error>(const ArrowArray&)> transform_function)
      : Transform(transform_type,
                  std::vector<std::shared_ptr<Type>>{std::move(source_type)},
                  std::move(result_type), std::move(transform_function)) {}

  Transform(
      TransformType transform_type, std::vector<std::shared_ptr<Type>> source_types,
      std::shared_ptr<Type> result_type,
      std::function<expected<ArrowArray, Error>(const ArrowArray&)> transform_function)
      : transform_type_(transform_type),
        source_types_(std::move(source_types)),
        result_type_(std::move(result_type)),
        transform_function_(std::move(transform_function)) {}

 private:
  TransformType transform_type_;
  std::vector<std::shared_ptr<Type>> source_types_;
  std::shared_ptr<Type> result_type_;
  std::function<expected<ArrowArray, Error>(const ArrowArray&)> transform_function_;
};

class IdentityTransform : public Transform {
 public:
  explicit IdentityTransform(std::shared_ptr<Type> source_type)
      : Transform(TransformType::kIdentity, source_type, source_type,
                  [](const ArrowArray& input) -> expected<ArrowArray, Error> {
                    return TransformUtil::Copy(input);
                  }) {}
};

class BucketTransform : public Transform {
 public:
  BucketTransform(std::shared_ptr<Type> source_type, int32_t num_buckets)
      : BucketTransform(std::vector<std::shared_ptr<Type>>{std::move(source_type)},
                        num_buckets) {}

  BucketTransform(std::vector<std::shared_ptr<Type>> source_types, int32_t num_buckets)
      : Transform(TransformType::kBucket, std::move(source_types),
                  std::make_shared<IntType>(),
                  [num_buckets](const ArrowArray& input) -> expected<ArrowArray, Error> {
                    return TransformUtil::Bucket(input, num_buckets);
                  }),
        num_buckets_{num_buckets} {}

  int32_t num_buckets() const { return num_buckets_; }

 private:
  int32_t num_buckets_;
};

```

We can even overload `expected<ArrowArray, Error> operator()(const ArrowArray& input) const` if you don't like the name of `Apply`.

WDYT? @gty404 @lidavidm @yingcai-cy

src/iceberg/transform/transform_factory.h
gty404: Reasonable, I will modify it. Thank you.

src/iceberg/arrow_c_data.h
gty404: I have the impression that I relied on it when testing and copying ArrowArray, and I would delete it

src/iceberg/transform.h
gty404: Reasonable, I will modify it. Thank you.

src/iceberg/transform.h
gty404: Reasonable, I will modify it. Thank you.

src/iceberg/transform.h
gty404: Reasonable, I will modify it. Thank you.

src/iceberg/transform/transform_spec.h
gty404: 1. Using ArrowArray is necessary, as different Transforms may use different types and numbers of parameters, which should be needed within the Transform or when creating an instance of the Transform.

2. TransformSpec is used to describe a transform, which can be derived from JSON and is also convenient for future extensions. 

@lidavidm

gty404: The main difference lies in the fact that your approach sub-classes uniformly provide a lambda function. I am concerned that the function might become too complex, as the same Transform type may need to implement corresponding logic based on different source types. However, iceberg-java has a similar approach; it generates specific function objects by binding the source type.  @wgtmac

lidavidm: > 1. Using ArrowArray is necessary, as different Transforms may use different types and numbers of parameters, which should be needed within the Transform or when creating an instance of the Transform.

Yes, but why not (say) `vector<variant<...>>`?

gty404: > > 1. Using ArrowArray is necessary, as different Transforms may use different types and numbers of parameters, which should be needed within the Transform or when creating an instance of the Transform.
> 
> Yes, but why not (say) `vector<variant<...>>`?

The parameter list for this variant might be quite long. Additionally, I understand that there is also logic to convert from JSON to ArrowArray in other places, and using ArrowArray would be more natural. However, this is just my personal opinion.

lidavidm: It somewhat feels like overengineering to me; there are only two transforms listed with parameters, and in both it's a single integer. Allocating an entire Arrow table to represent this seems like extreme overkill and will make the code much more complicated for that use case, and as noted here it causes lifetime issues too.

gty404: > It somewhat feels like overengineering to me; there are only two transforms listed with parameters, and in both it's a single integer. Allocating an entire Arrow table to represent this seems like extreme overkill and will make the code much more complicated for that use case, and as noted here it causes lifetime issues too.

From the perspective of complexity and meeting current needs, it is a bit over-designed. No problem, I will revise this part again.

gty404: I have changed the type to vector<variant<int32_t>>, please review it in your free time.   @lidavidm @wgtmac

wgtmac: > I suppose what we need is a standard for scalars or row-wise data...

I feel concerned of the complexity to introduce another row-wise data. Can't we reuse `ArrowArray` and pretend it as a scalar when its size is 1? @lidavidm

lidavidm: That's why I was questioning if we need this class in the first place vs. directly constructing the transforms which would sidestep this complexity

You _could_ use ArrowArray as a scalar but it is an extremely expensive scalar all things considered. At the very least you should pull in or write a RAII helper to manage its lifetime IMO if you are going down that path.

lidavidm: If the argument is that you can't instantiate the full transform during parsing, you could split up the transform's type (aka the operation + parameters) which you should be able to construct immediately from the fully instantiated transform

gty404: > If the argument is that you can't instantiate the full transform during parsing, you could split up the transform's type (aka the operation + parameters) which you should be able to construct immediately from the fully instantiated transform

When creating a TransformFunction, you need to specify the transform type, optional parameters, and the types of one or more source columns. These types need to be retrievable from the context to get the table's schema. My intention was to use TransformSpec as a parsed data block that can be easily passed around and delay the creation of the TransformFunction until the context meets the requirements. This might just be a hypothetical scenario, and currently, there might not be a need to provide an entry to create all TransformFunctions?

lidavidm: Right, which is why I said 

> you could split up the transform's type (aka the operation + parameters)[...]from the fully instantiated transform

You could even still have something like TransformSpec, I'm just arguing that it might make more sense to directly instantiate a transform-like object (with parameters, without types), just like how a Field separates the (parameterized) Type out.

gty404: > Right, which is why I said
> 
> > you could split up the transform's type (aka the operation + parameters)[...]from the fully instantiated transform
> 
> You could even still have something like TransformSpec, I'm just arguing that it might make more sense to directly instantiate a transform-like object (with parameters, without types), just like how a Field separates the (parameterized) Type out.

I see. I will provide UnboundTransform, which includes the transform type and parameters. At the same time, TransformFunction will provide a Bind interface to receive UnboundTransform and source type, generating a specific TransformFunction. What do you think?

lidavidm: That sounds good to me. @wgtmac?

src/iceberg/transform/transform_spec.h
```diff
@@ -0,0 +1,100 @@
...
+struct TransformSpec {
+  /// \brief The type of the transform (e.g., kBucket, kTruncate, kIdentity, etc.)
+  TransformType transform_type;
+
+  /// \brief Optional parameter values passed to the transform.
+  ///
+  /// For transforms that require parameters (e.g., Bucket(N)), this holds the arguments
+  /// as a int32 value (e.g., INT32 for num_buckets or width).
+  /// If the transform does not require parameters, this will be empty.
+  std::vector<std::variant<int32_t>> params;
```
wgtmac: Why not simply `std::variant<std::monostate, int32_t> param`?

wgtmac: The only difference between `TransformFunction` and `TransformSpec` is this addition param. Should we delete `TransformSpec` by adding `param` to `TransformFunction`?

src/iceberg/transform/transform_spec.h
wgtmac: SGTM!

src/iceberg/transform/transform_spec.h
gty404: 1. Could there be multiple param?
2. TransformSpec is just the result after json parsing, and TransformFunction is the specific implementation for the transformation.

wgtmac: For each transform function, there is unlikely to have more than one param in the near future. Am I correct, @Fokko?

wgtmac: I don't like the name of `TransformSpec` because it does not have a spec_id or something. What about renaming `TransformSpec` to `Transform`? For `TransformFunction`, we can use `Apply` or even `operator()` to resolve the name conflict.

gty404: > I don't like the name of `TransformSpec` because it does not have a spec_id or something. What about renaming `TransformSpec` to `Transform`? For `TransformFunction`, we can use `Apply` or even `operator()` to resolve the name conflict.

How about UnboundTransform?

wgtmac: I suppose `UnboundTransform` will be used in the expression

src/iceberg/transform.cc
```diff
@@ -21,55 +21,26 @@
...
-TransformFunction::TransformFunction(TransformType type) : transform_type_(type) {}
+TransformFunction::TransformFunction(TransformType transform_type,
+                                     std::shared_ptr<Type> source_type)
+    : transform_type_(transform_type), source_type_(std::move(source_type)) {}
 
 TransformType TransformFunction::transform_type() const { return transform_type_; }
 
-std::string TransformFunction::ToString() const {
-  return std::format("{}", iceberg::ToString(transform_type_));
+std::shared_ptr<Type> const& TransformFunction::source_type() const {
```
zhjwpku, line 186: It seems we have two different styles for const placement: `const std::shared_ptr<Type>&` vs `std::shared_ptr<Type> const&`. Do we want to stick to one style? 

You don’t need to make changes in this PR; I’m just curious if this is something we should pay attention to. @wgtmac @lidavidm WDYT

src/iceberg/type_fwd.h
```diff
@@ -101,6 +101,7 @@ class StructLike;
 class TableMetadata;
 enum class TransformType;
 class TransformFunction;
+struct TransformSpec;
```
zhjwpku: I don't see this struct elsewhere, do we need this?

gty404: It's not needed anymore, I will delete it.

src/iceberg/transform.cc
lidavidm: Ideally the formatter would just pick one style and stick to it...I think `T const&` is cleaner but `const T&` is far more common

lidavidm: I suppose there's https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifieralignment but with some caveats

test/transform_test.cc
```diff
@@ -25,30 +25,48 @@
...
-      return unexpected(
-          Error{.kind = ErrorKind::kNotSupported, .message = "test transform function"});
-    }
-  };
-
-  TestTransformFunction transform;
-  EXPECT_EQ(TransformType::kUnknown, transform.transform_type());
-  EXPECT_EQ("unknown", transform.ToString());
-  EXPECT_EQ("unknown", std::format("{}", transform));
+  IdentityTransform transform{std::make_shared<StringType>()};
```
lidavidm: not for this PR but we should consider singleton type instances for common types (though maybe sharing a single refcount is actually a net negative)

gty404:
I made the following modifications based on the previous comment:

1. The Transform object is used for JSON serialization/deserialization.

2. Transform::Bind can get the TransformFunction, which is used for data transformation.

Please review in your spare time. Thanks  @lidavidm @wgtmac @zhjwpku

src/iceberg/transform.h
```diff
@@ -56,16 +57,98 @@ enum class TransformType {
...
+  /// \brief Checks equality with another Transform instance.
+  [[nodiscard]] virtual bool Equals(const Transform& other) const;
+
+  TransformType transform_type_;
+  ///< Optional parameter (e.g., num_buckets, width)
+  std::variant<std::monostate, int32_t> param_;
+};
+
 /// \brief A transform function used for partitioning.
-class ICEBERG_EXPORT TransformFunction : public util::Formattable {
```
lidavidm, line 60: nit: why not formattable anymore?

lidavidm (approved):

src/iceberg/transform.h
gty404: Transform is already formattable, I haven't thought of the need for TransformFunction to be printable yet, possibly outputting the source type/result type is a requirement. In the next PR, I will support this interface.

lidavidm: Sounds good to me.

src/iceberg/transform.h
```diff
@@ -56,16 +57,98 @@ enum class TransformType {
...
+  /// \brief Returns the transform type.
+  TransformType transform_type() const;
+
+  /// \brief Binds this transform to a source type, returning a typed TransformFunction.
+  ///
+  /// This creates a concrete transform implementation based on the transform type and
+  /// parameter.
+  /// \param source_type The source column type to bind to.
+  /// \return A TransformFunction instance wrapped in `expected`, or an error on failure.
+  expected<std::unique_ptr<TransformFunction>, Error> Bind(
```
wgtmac: Can we use forward declaration here and define TransformFunction in the `transform_function.h`?

src/iceberg/transform.cc
```diff
@@ -21,65 +21,113 @@
...
-constexpr std::string_view ToString(TransformType type) {
-  switch (type) {
-    case TransformType::kUnknown:
-      return "unknown";
+std::shared_ptr<Transform> Transform::Identity() {
+  static auto instance = std::make_shared<Transform>(TransformType::kIdentity);
+  return instance;
+}
+
+Transform::Transform(TransformType transform_type) : transform_type_(transform_type) {}
```
wgtmac, line 103: Throw for parameterized transform_type?

src/iceberg/transform.cc
```diff
@@ -21,65 +21,113 @@
...
+
+Transform::Transform(TransformType transform_type, int32_t param)
```
wgtmac, line 105: Throw for non-parameterized transform_type?

wgtmac: If we want to avoid invalid inputs. Perhaps we should define a separate static functions to create each transform type?

src/iceberg/transform.cc
```diff
@@ -21,65 +21,113 @@
...
+    : transform_type_(transform_type), param_(param) {}
+
+TransformType Transform::transform_type() const { return transform_type_; }
+
+expected<std::unique_ptr<TransformFunction>, Error> Transform::Bind(
```
wgtmac: It seems that we should merge `transform_function.h/cc` to `transform.h/cc`. They have strong connections.

src/iceberg/transform.h
```diff
@@ -56,16 +57,98 @@ enum class TransformType {
...
   virtual expected<ArrowArray, Error> Transform(const ArrowArray& data) = 0;
   /// \brief Get the transform type
-  virtual TransformType transform_type() const;
-
-  std::string ToString() const override;
+  TransformType transform_type() const;
+  /// \brief Get the source type of transform function
+  const std::shared_ptr<Type>& source_type() const;
+  /// \brief Get the result type of transform function
+  virtual expected<std::shared_ptr<Type>, Error> ResultType() const = 0;
```
wgtmac: Use `Result`?

src/iceberg/transform.cc
gty404: Yes, currently I only added Transform::Identity, I will add the others as well.

src/iceberg/transform.cc
gty404: I suggest it is separate, with users of Transform only needing to depend on transform.h and not needing to concern themselves with the implementation details of the transform function.

src/iceberg/transform.h
gty404: My idea is that transform.h contains all the interfaces that transform users depend on, without exposing the implementation details of the transform functions.

src/iceberg/transform.h
gty404: Okay, I will merge the latest code.

src/iceberg/transform.h
```diff
@@ -56,16 +57,133 @@ enum class TransformType {
...
+  /// \brief Returns the transform type.
+  TransformType transform_type() const;
+
+  /// \brief Binds this transform to a source type, returning a typed TransformFunction.
+  ///
+  /// This creates a concrete transform implementation based on the transform type and
+  /// parameter.
+  /// \param source_type The source column type to bind to.
+  /// \return A TransformFunction instance wrapped in `expected`, or an error on failure.
+  expected<std::unique_ptr<TransformFunction>, Error> Bind(
```
```suggestion
  Result<std::unique_ptr<TransformFunction>> Bind(
```

nit: we can leave these as-is and fix all together in a separate PR.

Xuanwo (approved):
The community has reached a consensus. Let's go!

PR #62, opened 2025-04-08
feat: add table metadata definition

wgtmac:


src/iceberg/table_metadata.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/type_fwd.h"
+#include "iceberg/util/formattable.h"
+
+namespace iceberg {
+
+/// \brief Represents a snapshot log entry
+struct ICEBERG_EXPORT SnapshotLogEntry : public util::Formattable {
+  /// The timestamp in milliseconds of the change
```
lidavidm, line 42: Would it make more sense to store as a `std::chrono` type?

src/iceberg/statistics_file.h
```diff
@@ -0,0 +1,104 @@
...
+  /// Type of the blob
+  std::string type;
+  /// ID of the Iceberg table's snapshot the blob was computed from
+  int64_t source_snapshot_id;
+  /// Sequence number of the Iceberg table's snapshot the blob was computed from
+  int64_t source_snapshot_sequence_number;
+  /// Ordered list of fields the blob was calculated from
+  std::vector<int32_t> fields;
+  /// Additional properties of the blob, specific to the blob type
+  std::map<std::string, std::string> properties;
```
lidavidm: I think some PRs use unordered_map and some use map - should we be consistent?

lidavidm (approved):

src/iceberg/statistics_file.h
wgtmac: I have switched to `unordered_map` for now.

src/iceberg/table_metadata.h
wgtmac: Sounds good. Switched to use time_point

lidavidm (approved):

src/iceberg/table_metadata.h
```diff
@@ -0,0 +1,120 @@
...
+/// \brief Represents the metadata for an Iceberg table
+///
+/// Note that it only contains table metadata from the spec.  Compared to the Java
+/// implementation, missing pieces including: 1) Map<Integer,
+/// Schema|PartitionSpec|SortOrder> 2) List<MetadataUpdate> 3) Map<Long, Snapshot> 4)
+/// Map<String, SnapshotRef>
+///
+/// TODO(wgtmac): Implement Equals and ToString once SortOrder and Snapshot are
+/// implemented.
+struct ICEBERG_EXPORT TableMetadata {
```
gty404, line 69: Do you need to mark some fields as optional?

wgtmac: Yes, I think we need but I'm still unclear which one should change. I will postpone this decision until we implement the json serialization from/to different table format versions.

src/iceberg/table_metadata.h
```diff
@@ -0,0 +1,120 @@
...
+  /// Default sort order id of the table
+  int32_t default_sort_order_id;
+  /// A map of snapshot references
+  std::unordered_map<std::string, std::string> refs;
+  /// A list of table statistics
+  std::vector<std::shared_ptr<struct StatisticsFile>> statistics;
+  /// A list of partition statistics
+  std::vector<std::shared_ptr<struct PartitionStatisticsFile>> partition_statistics;
+  /// whether or not to track the creation and updates to rows in the table
+  bool row_lineage_enabled = false;
```
Fokko: Row lineage will always be enabled from V3 onwards: https://github.com/apache/iceberg/pull/12593 So I think we can remove this 👍

Fokko (approved):

src/iceberg/table_metadata.h
wgtmac: Thanks for the heads-up! I found that the Java impl still has this field which looks weird to me.

```suggestion
```

zhjwpku (approved):
LGTM

src/iceberg/table_metadata.cc
```diff
@@ -0,0 +1,38 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/table_metadata.h"
+
+#include <format>
+#include <string>
+
+#include "iceberg/statistics_file.h"
```
raulcd, line 25: is there a reason why this include is here and not in `table_metadata.h` where we seem to require the following?
```
  std::vector<std::shared_ptr<struct StatisticsFile>> statistics;
  /// A list of partition statistics
  std::vector<std::shared_ptr<struct PartitionStatisticsFile>> partition_statistics;
```

wgtmac: I just want to use forward declaration as much as possible. The implementation is not yet complete due to missing the concrete implementation of `Snapshot` and other classes so it looks weird that `iceberg/statistics_file.h` is included but not used at the moment.

raulcd: but as those would still be part of the table metadata API wouldn't it be better to be part of the header file? I am learning C++, I just thought it would still be better there even when we add the implementation part but I might be wrong :)

raulcd (approved):

src/iceberg/table_metadata.cc
wgtmac: It is a common practice to use forward declaration to speed up compilation if the implementation detail is not required in the header file, though I suspect that modern compilers are smart enough to optimize this.

Fokko:
Thanks @wgtmac for working on this, and thanks @zhjwpku for the review 🙌

PR #64, opened 2025-04-09
feat: sort field/order json serialize/deserialization

gty404:


src/iceberg/type_fwd.h
```diff
@@ -97,6 +97,7 @@ class HistoryEntry;
 class PartitionSpec;
 class Snapshot;
 class SortOrder;
+class SortField;
```
wgtmac, line 99: Sort alphabetically.

src/iceberg/json_internal.h
```diff
@@ -0,0 +1,73 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <memory>
+
+#include <nlohmann/json.hpp>
```
wgtmac:
```suggestion
#include <nlohmann/json_fwd.hpp>
```

src/iceberg/transform.cc
```diff
@@ -72,4 +72,13 @@ expected<ArrowArray, Error> IdentityTransformFunction::Transform(
                             .message = "IdentityTransformFunction::Transform"});
 }
 
+expected<std::unique_ptr<TransformFunction>, Error> TransformFunctionFromString(
```
lidavidm, line 75: Not related, but should we do something like define `iceberg::Result<T> = expected<T, iceberg::Error>`?

src/iceberg/transform.h
```diff
@@ -82,6 +82,9 @@ class ICEBERG_EXPORT TransformFunction : public util::Formattable {
   TransformType transform_type_;
 };
 
+expected<std::unique_ptr<TransformFunction>, Error> TransformFunctionFromString(
```
```suggestion
ICEBERG_EXPORT expected<std::unique_ptr<TransformFunction>, Error> TransformFunctionFromString(
```

src/iceberg/transform.h
```diff
@@ -82,6 +82,9 @@ class ICEBERG_EXPORT TransformFunction : public util::Formattable {
...
+    const std::string& str);
+
 class IdentityTransformFunction : public TransformFunction {
```
```suggestion
class ICEBERG_EXPORT IdentityTransformFunction : public TransformFunction {
```

src/iceberg/transform.h
wgtmac: Can we use `std::string_view`?

src/iceberg/sort_field.h
```diff
@@ -41,13 +43,49 @@ enum class SortDirection {
...
+}
 
 enum class NullOrder {
   /// Nulls are sorted first
   kFirst,
   /// Nulls are sorted last
   kLast,
 };
+/// \brief Get the relative null order name
+constexpr std::string_view NullOrderToString(NullOrder null_order) {
```
wgtmac:
```suggestion
ICEBERG_EXPORT constexpr std::string_view NullOrderToString(NullOrder null_order) {
```

src/iceberg/sort_field.h
```diff
@@ -41,13 +43,49 @@ enum class SortDirection {
...
+    case NullOrder::kFirst:
+      return "nulls-first";
+    case NullOrder::kLast:
+      return "nulls-last";
+    default:
+      return "invalid";
+  }
+}
+/// \brief Get the relative null order from name
+constexpr expected<NullOrder, Error> NullOrderFromString(const std::string& str) {
```
wgtmac:
```suggestion
ICEBERG_EXPORT constexpr expected<NullOrder, Error> NullOrderFromString(const std::string& str) {
```

wgtmac (commented):
Generally LGTM.

@lidavidm @zhjwpku @yingcai-cy WDYT?

src/iceberg/sort_field.h
```diff
@@ -41,13 +43,52 @@ enum class SortDirection {
   /// Descending
   kDescending,
 };
+/// \brief Get the relative sort direction name
+ICEBERG_EXPORT constexpr std::string_view SortDirectionToString(SortDirection direction) {
```
wgtmac, line 47: nit: we should probably only declare these functions here and put all implementation in the .cc file.

wgtmac (approved):

src/iceberg/json_internal.cc
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+#define TRY_ASSIGN(json_value, expr)                                    \
+  auto _tmp_##json_value = (expr);                                      \
+  if (!_tmp_##json_value) return unexpected(_tmp_##json_value.error()); \
+  auto json_value = std::move(_tmp_##json_value.value());
+}  // namespace
+
+nlohmann::json ToJson(const SortField& sort_field) {
```
zhjwpku, line 64: I was thinking about adding ToJson and FromJson member functions to each Spec classes, but your way might be better, users don't need to touch the json ser/der functions, right?

zhjwpku (approved):
LGTM

src/iceberg/json_internal.cc
gty404: Yes, there is no need to introduce json either.

wgtmac: Yes, in this way we don't need to expose our internal json dependency.

src/iceberg/json_internal.cc
```diff
@@ -0,0 +1,112 @@
...
+  try {
+    return json.at(key).get<T>();
+  } catch (const std::exception& ex) {
+    return unexpected<Error>({.kind = ErrorKind::kInvalidArgument,
+                              .message = std::string("Failed to parse key: ") +
+                                         key.data() + ", " + ex.what()});
+  }
+}
+
+#define TRY_ASSIGN(json_value, expr)                                    \
```
lidavidm, line 58: We'll probably want to make a broader version of this macro. (Not necessarily in this PR)

FWIW, you could steal the Arrow version, which uses a counter to autogenerate the name, so it can be used like `ARROW_ASSIGN_OR_RAISE(auto value, ...)` or `ARROW_ASSIGN_OR_RAISE(foo.bar, ...)` too

lidavidm (approved):

src/iceberg/json_internal.cc
wgtmac: I did this exactly: https://github.com/apache/iceberg-cpp/pull/65/files#diff-d8a08fcbce5230d86b8b253508bd5d8dc6216a2d0f5049b58b6e7792f2243040

Xuanwo (approved):
Thank you for working on this! There is no other concern from the community, I'm going to merge this one now.

wgtmac:
Thank you @Xuanwo!

PR #67, opened 2025-04-10
refactor: consolidate result type

gty404:


lidavidm (approved):
Thanks!

src/iceberg/result.h
```diff
@@ -47,4 +48,14 @@ struct ICEBERG_EXPORT [[nodiscard]] Error {
...
 
+/// /brief Default error trait
+template <typename T>
+struct DefaultError {
+  using type = Error;
+};
+
+/// \brief Result alias
+template <typename T, typename E = typename DefaultError<T>::type>
+using Result = expected<T, E>;
```
wgtmac, line 59: Should we default `T` to `void` so we can simply use `Result` as the return type if nothing to return?

wgtmac (approved):

src/iceberg/result.h
lidavidm: We could also add `using Status = expected<void, Error>`?

gty404: Good idea, I will uniformly modify the places where expected<void, Error> is used

zhjwpku (approved):
Finally, it's Result and Status, thanks ;)

lidavidm (approved):

wgtmac:
@Xuanwo @Fokko Could you help merge this? Thanks!

Xuanwo (approved):
The community has reached a consensus. Let's move forward!

PR #69, opened 2025-04-14
chore: don't include expected.h but result.h

zhjwpku:
It is suggested that we include result.h instead of expected.h so that in the future we can simply replace this ported iceberg::expected with std::expected.

Besides that, I fix some IWYU warning in this PR.


lidavidm (approved):

wgtmac (approved):

zhjwpku:
@Fokko @Xuanwo Please take a look at this PR when you are available, thanks.

Xuanwo (approved):
Simple changes, let's move.

PR #74, opened 2025-04-14
feat: snapshot serde

zhjwpku:


src/iceberg/snapshot.h
```diff
@@ -39,6 +40,28 @@ enum class SnapshotRefType {
...
+/// \brief Get the relative snapshot reference type name
+ICEBERG_EXPORT constexpr std::string_view SnapshotRefTypeToString(
+    SnapshotRefType type) noexcept {
+  switch (type) {
+    case SnapshotRefType::kBranch:
+      return "branch";
+    case SnapshotRefType::kTag:
+      return "tag";
+    default:
+      return "invalid";
```
wgtmac:
```suggestion
```

src/iceberg/snapshot.h
```diff
@@ -54,13 +77,35 @@ struct ICEBERG_EXPORT SnapshotRef {
     /// of the snapshot reference to keep while expiring snapshots. Defaults to table
     /// property history.expire.max-ref-age-ms. The main branch never expires.
     std::optional<int64_t> max_ref_age_ms;
+
+    /// \brief Compare two branches for equality.
+    friend bool operator==(const Branch& lhs, const Branch& rhs) {
```
wgtmac, line 80: Aren't these trivial `operator==` and `operator!=` generated by compilers?

src/iceberg/json_internal.cc
```diff
@@ -231,6 +296,53 @@ nlohmann::json SchemaToJson(const Schema& schema) {
...
 
+nlohmann::json SnapshotRefToJson(const SnapshotRef& ref) {
+  nlohmann::json json;
+  json[kSnapshotId] = ref.snapshot_id;
+  json[kType] = SnapshotRefTypeToString(ref.type());
+  if (ref.type() == SnapshotRefType::kBranch) {
+    const auto& branch = std::get<SnapshotRef::Branch>(ref.retention);
+    if (branch.min_snapshots_to_keep.has_value()) {
+      json[kMinSnapshotsToKeep] = *branch.min_snapshots_to_keep;
+    }
```
lidavidm: Worth writing a helper for this?

test/json_internal_test.cc
```diff
@@ -148,4 +148,96 @@ TEST(JsonPartitionTest, PartitionSpec) {
...
+            "invalid-field":"value"
+          },
+          "schema-id":42})"_json;
+
+  auto result = SnapshotFromJson(invalid_json_snapshot);
+  ASSERT_FALSE(result.has_value());
+
+  EXPECT_EQ(result.error().kind, ErrorKind::kJsonParseError);
+  EXPECT_TRUE(result.error().message.find("Invalid snapshot summary field") !=
+              std::string::npos);
```
lidavidm: Note that for this, you can do `EXPECT_THAT(result.error().message, ::testing::HasSubstr("Invalid snapshot summary field"));`

lidavidm (approved):

src/iceberg/json_internal.cc
```diff
@@ -70,6 +72,53 @@ constexpr std::string_view kValueRequired = "value-required";
...
+    SnapshotSummaryFields::kRemovedPosDeletes,
+    SnapshotSummaryFields::kTotalPosDeletes,
+    SnapshotSummaryFields::kAddedEqDeletes,
+    SnapshotSummaryFields::kRemovedEqDeletes,
+    SnapshotSummaryFields::kTotalEqDeletes,
+    SnapshotSummaryFields::kDeletedDuplicatedFiles,
+    SnapshotSummaryFields::kChangedPartitionCountProp,
+    SnapshotSummaryFields::kWAPID,
+    SnapshotSummaryFields::kPublishedWAPID,
+    SnapshotSummaryFields::kSourceSnapshotID,
```
wgtmac:
```suggestion
    SnapshotSummaryFields::kWAPId,
    SnapshotSummaryFields::kPublishedWAPId,
    SnapshotSummaryFields::kSourceSnapshotId,
```

wgtmac (approved):

src/iceberg/json_internal.cc
```diff
@@ -231,6 +296,53 @@ nlohmann::json SchemaToJson(const Schema& schema) {
...
+}
+
+nlohmann::json SnapshotToJson(const Snapshot& snapshot) {
+  nlohmann::json json;
+  json[kSnapshotId] = snapshot.snapshot_id;
+  if (snapshot.parent_snapshot_id.has_value()) {
+    json[kParentSnapshotId] = *snapshot.parent_snapshot_id;
+  }
+  json[kSequenceNumber] = snapshot.sequence_number;
+  json[kTimestampMs] = snapshot.timestamp_ms;
```
wgtmac, line 333: Should we use something like `std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds>`?

src/iceberg/json_internal.cc
```diff
@@ -231,6 +296,53 @@ nlohmann::json SchemaToJson(const Schema& schema) {
...
+  json[kManifestList] = snapshot.manifest_list;
+
+  nlohmann::json summary_json;
+  for (const auto& [key, value] : snapshot.summary) {
+    summary_json[key] = value;
+  }
+  json[kSummary] = summary_json;
```
wgtmac:
```suggestion
  json[kSummary] = snapshot.summary;
```

src/iceberg/json_internal.cc
zhjwpku: I noticed there is a `using TimePointMs` in table_metadata.h, do you mind I do this in a separate PR, I think we should put this in a common header file, but I haven't think too hard about the header name.

src/iceberg/snapshot.h
zhjwpku: Tried, but failed with:

```
error: invalid operands to binary expression ('const iceberg::SnapshotRef::Tag' and 'const iceberg::SnapshotRef::Tag')
   61 |            std::get<Tag>(retention) == std::get<Tag>(other.retention);
```

src/iceberg/json_internal.cc
zhjwpku: brilliant, done that way.

src/iceberg/json_internal.cc
zhjwpku: Add a helper named `SetOptionalField`, not sure if the name is good enough ;)

test/json_internal_test.cc
zhjwpku: Good catch, done the matchers way, thanks.

wgtmac (approved):
LGTM, Thanks!

src/iceberg/snapshot.h
```diff
@@ -139,11 +196,11 @@ struct SnapshotSummaryFields {
   /// Other Fields, see https://iceberg.apache.org/spec/#other-fields
 
   /// \brief The Write-Audit-Publish id of a staged snapshot
-  inline static const std::string kWAPID = "wap.id";
+  inline static const std::string kWAPId = "wap.id";
```
yingcai-cy, line 199: Just curious，is there a reason for using a different style of const string here? why not use “inline constexpr std::string_view
”

zhjwpku: Hah, we discussed about this in #60, because of SSO(Small String Optimization), `inline constexpr std::string_view` can only handle small string(typically up to 15-23 characters, depending on the implementation), thus changed to `inline static const std::string`.

src/iceberg/json_internal.cc
```diff
@@ -419,4 +520,75 @@ Result<std::unique_ptr<PartitionSpec>> PartitionSpecFromJson(
...
+  } else {
+    ICEBERG_ASSIGN_OR_RAISE(auto max_ref_age_ms,
+                            GetJsonValueOptional<int64_t>(json, kMaxRefAgeMs));
+
+    return std::make_unique<SnapshotRef>(
+        snapshot_id, SnapshotRef::Tag{.max_ref_age_ms = max_ref_age_ms});
+  }
+}
+
+Result<std::unique_ptr<Snapshot>> SnapshotFromJson(const nlohmann::json& json) {
```
wgtmac, line 557: I think we need to be consistent with the Java impl: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SnapshotParser.java. Specifically, we need to deal with cases where sequence number or summary is missing.

@Fokko Will it actually happen that a snapshot does not have `summary` (and thus `operation` is also missing)?

wgtmac: From the spec, `summary` is required for v2 and v3 but optional for v1. So I believe the spec answers my question. We have to handle this.

zhjwpku:
I've updated the PR to use `ToJson` as the serialization function name, also only write the summary map when there is an operation, this is the same behavior from java impl[1]

[1] https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SnapshotParser.java#L69-L82

src/iceberg/json_internal.cc
```diff
@@ -419,4 +523,77 @@ Result<std::unique_ptr<PartitionSpec>> PartitionSpecFromJson(
...
+  ICEBERG_ASSIGN_OR_RAISE(auto snapshot_id, GetJsonValue<int64_t>(json, kSnapshotId));
+  ICEBERG_ASSIGN_OR_RAISE(auto sequence_number,
```
wgtmac, line 559: `sequence_number` is required by v2 and v3 but is nonexistent in v1.

zhjwpku: Fixed, thanks for the reminder.

wgtmac (approved):

src/iceberg/json_internal.cc
Fokko: You're correct!

- For V1 the summary is optional.
- For V2/V3 the summary is required, and also the operation. Some writers produced some malformed metadata in the past. Instead of throwing an exception, we would it is an overwrite operation, since that's the most generic one.

wgtmac: It sounds like we need to set `operation` to `overwrite` when `summary` is available but `operation` is missing. @zhjwpku

zhjwpku: Fixed, please take a look.

Fokko (approved):

Fokko:
Thanks for working on this @zhjwpku, and thanks for the review @lidavidm, @yingcai-cy and @wgtmac 🙌

PR #75, opened 2025-04-14
feat: add json serde for table metadata

wgtmac:


wgtmac:
@lidavidm @zhjwpku @gty404 Could you help review this?

I will add test cases once https://github.com/apache/iceberg-cpp/pull/74 has been merged.

Note that I have removed `Formatable` inheritance to facilitate aggregate initialization of simple structs and add `ToString` functions instead.

src/iceberg/statistics_file.h
```diff
@@ -99,6 +93,29 @@ struct ICEBERG_EXPORT PartitionStatisticsFile {
...
+  }
+
+  /// \brief Compare two PartitionStatisticsFiles for inequality.
+  friend bool operator!=(const PartitionStatisticsFile& lhs,
+                         const PartitionStatisticsFile& rhs) {
+    return !(lhs == rhs);
+  }
 };
 
+/// \brief Returns a string representation of a BlobMetadata
```
gty404, line 111: The method name may need to use BlobMetadataToString to avoid conflicts with the ToString method in member methods ？

gty404 (approved):

src/iceberg/statistics_file.h
wgtmac: I have removed `ToString` member function because inheritance prohibits aggregate initialization.

lidavidm: SGTM. We could extend the std::format specialization to look for an overload of iceberg::ToString or something (or try argument-dependent lookup? :thinking:)

lidavidm (approved):

src/iceberg/statistics_file.h
wgtmac: Yes, I think we can specialize std::formatter in the `iceberg/util/formatter.h` for classes that have `iceberg::ToString(const T&)` at hands.

zhjwpku (approved):
LGTM

src/iceberg/statistics_file.cc
```diff
@@ -23,29 +23,26 @@
...
   std::format_to(std::back_inserter(repr),
-                 "type='{}',sourceSnapshotId={},sourceSnapshotSequenceNumber={},", type,
-                 source_snapshot_id, source_snapshot_sequence_number);
+                 "type='{}',sourceSnapshotId={},sourceSnapshotSequenceNumber={},",
+                 blob_metadata.type, blob_metadata.source_snapshot_id,
+                 blob_metadata.source_snapshot_sequence_number);
   std::format_to(std::back_inserter(repr), "fields=[");
-  for (auto iter = fields.cbegin(); iter != fields.cend(); ++iter) {
-    if (iter != fields.cbegin()) {
+  for (auto iter = blob_metadata.fields.cbegin(); iter != blob_metadata.fields.cend();
```
yingcai-cy, line 33: how about:
```suggestion
  if (!blob_metadata.fields.empty()) {
    std::format_to(std::back_inserter(repr), "{}", blob_metadata.fields.front());
  for (auto iter = std::next(blob_metadata.fields.cbegin()); iter != blob_metadata.fields.cend(); ++iter) {
    std::format_to(std::back_inserter(repr), ",{}", *iter);
  }
  }
```
Remove the if/else in the for loop or we should wrap a helper function.

src/iceberg/json_internal.h
```diff
@@ -74,44 +76,32 @@ Result<std::unique_ptr<SortOrder>> SortOrderFromJson(const nlohmann::json& json)
...
-nlohmann::json ToJson(const Snapshot& snapshot);
+nlohmann::json ToJson(const Schema& schema);
 
 /// \brief Convert JSON to an Iceberg Schema.
 ///
 /// \param[in] json The JSON representation of the schema.
 /// \return The Iceberg schema or an error if the conversion fails.
 Result<std::unique_ptr<Schema>> SchemaFromJson(const nlohmann::json& json);
 
+/// \brief Convert an Iceberg Type to JSON.
```
zhjwpku, line 87: nit: should we stick to the same doc pattern, `Serializes a 'xxx' object to JSON`.

zhjwpku (approved):

gty404 (approved):
LGTM

src/iceberg/json_internal.h
wgtmac: Sure, let me do this in a follow-up PR. This one gets large enough.

src/iceberg/json_internal.cc
```diff
@@ -602,8 +757,431 @@ Result<std::unique_ptr<Snapshot>> SnapshotFromJson(const nlohmann::json& json) {
...
+  json[kProperties] = table_metadata.properties;
+
+  if (std::ranges::find_if(table_metadata.snapshots, [&](const auto& snapshot) {
+        return snapshot->snapshot_id == table_metadata.current_snapshot_id;
+      }) != table_metadata.snapshots.cend()) {
+    json[kCurrentSnapshotId] = table_metadata.current_snapshot_id;
+  } else if (table_metadata.format_version >= kMinNullCurrentSnapshotVersion) {
+    json[kCurrentSnapshotId] = nullptr;
+  } else {
+    json[kCurrentSnapshotId] = TableMetadata::kInvalidSnapshotId;
```
Fokko: Let's avoid writing `-1`, and write `null` instead: https://iceberg.apache.org/spec/#assignment-of-snapshot-ids-and-current-snapshot-id

src/iceberg/statistics_file.cc
Fokko: @wgtmac Thoughts?

Fokko (commented):
@wgtmac Thanks for raising this PR, this looks great already! How do you feel about adding some tests? 

We have some example JSONs here: https://github.com/apache/iceberg/tree/main/core/src/test/resources

src/iceberg/statistics_file.cc
wgtmac: I'll refactor the formatter code later so let's keep it simple here.

src/iceberg/json_internal.cc
wgtmac: I did this in line 939 above. It can be verified by the test below:

```
TEST(JsonInternalTest, JsonNullTest) {
  nlohmann::json json;
  json["json-null"] = nlohmann::json::value_t::null;
  json["null-ptr"] = nullptr;
  std::cout << json.dump() << std::endl;
}

[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from JsonInternalTest
[ RUN      ] JsonInternalTest.JsonNullTest
{"json-null":null,"null-ptr":null}
[       OK ] JsonInternalTest.JsonNullTest (0 ms)
[----------] 1 test from JsonInternalTest (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.

```

wgtmac:
Thanks @Fokko for the view! I will add test cases in a separate PR. Otherwise @zhjwpku might have a headache to rewrite a lot of test cases when working on this issue: https://github.com/apache/iceberg-cpp/issues/79

src/iceberg/json_internal.cc
wgtmac: Now I have removed the check by always writing null for invalid snapshot ids.

Fokko: Thanks!

Fokko (approved):
@wgtmac Sounds good to me 👍 Thanks for working on this, this looks great!

PR #77, opened 2025-04-16
feat: introduce Scalar and StructLike interfaces for Transform evaluation

gty404:
Problems to Solve
1. nanoarrow does not provide a copy interface for ArrowArray, which is needed when implementing IdentityTransform.
2. Both Transform and Expression would benefit from a lightweight, uniform data structure for input values — ideally something like a simplified StructLike.

Proposed Solution
1. Introduce an abstract Scalar interface to represent values (both primitive and complex).
2. Define a StructLike interface to represent structured values with field-based access.
3. Allow classes such as ManifestEntity, DataFile, etc., to implement StructLike and serve as evaluation inputs for Transform and Expression.

Benefits
1. Decouples Transform/Expression from ArrowArray, improving flexibility.
2. Enables easier testing, mocking, and transformation logic using in-memory representations.
3. Lays the foundation for future enhancements like expression optimization or code generation.

### DIFF
--- a/src/iceberg/struct_like.h
+++ b/src/iceberg/struct_like.h
@@ -1,0 +1,278 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+class Scalar;
+class StructScalar;
+class ArrayScalar;
+class ScalarVisitor;
+class StructLike;
+class StructLikeVisitor;
+/**
+ * @brief Base class representing a generic scalar value.
+ */
+class ICEBERG_EXPORT Scalar {
+ public:
+  virtual ~Scalar() = default;
+
+  /**
+   * @brief Checks if the scalar represents a null value.
+   * @return true if the value is null, false otherwise.
+   */
+  virtual bool IsNull() const = 0;
+
+  /**
+   * @brief Casts the scalar to a specific derived type.
+   * @tparam T The target derived Scalar type.
+   * @return Reference to the casted value.
+   * @throws std::bad_cast if the cast fails.
+   */
+  template <typename T>
+  const T& as() const {
+    auto ptr = dynamic_cast<const T*>(this);
+    if (!ptr) throw std::bad_cast();
+    return *ptr;
+  }
+
+  /**
+   * @brief Creates a deep copy of the scalar.
+   * @return A shared pointer to the cloned scalar.
+   */
+  virtual std::shared_ptr<Scalar> clone() const = 0;
+
+  /**
+   * @brief Accepts a visitor for dispatching based on scalar type.
+   * @param visitor The visitor to accept.
+   */
+  virtual void accept(ScalarVisitor& visitor) const = 0;
+};
+
+/**
+ * @brief Visitor interface for Scalar, supporting type-dispatched operations.
+ */
+class ICEBERG_EXPORT ScalarVisitor {
+ public:
+  virtual ~ScalarVisitor() = default;
+
+  /// Visit methods for primitive types
+  virtual void visit(int32_t value) = 0;
+  virtual void visit(int64_t value) = 0;
+  virtual void visit(float value) = 0;
+  virtual void visit(double value) = 0;
+  virtual void visit(bool value) = 0;
+  virtual void visit(const std::string& value) = 0;
+
+  /**
+   * @brief Called when visiting a null scalar.
+   */
+  virtual void visitNull() = 0;
+
+  /// Visit methods for complex types
+  virtual void visitStruct(const StructScalar& value) = 0;
+  virtual void visitArray(const ArrayScalar& value) = 0;
+};
+
+/**
+ * @brief Represents a scalar that holds a primitive value.
+ * @tparam T The type of the primitive value.
+ */
+template <typename T>
+class ICEBERG_EXPORT PrimitiveScalar : public Scalar {
+ public:
+  /**
+   * @brief Constructs a PrimitiveScalar from an optional value.
+   * @param value The optional primitive value.
+   */
+  explicit PrimitiveScalar(std::optional<T> value) : value_(std::move(value)) {}
+
+  bool IsNull() const override { return !value_.has_value(); }
+
+  /**
+   * @brief Gets the actual value held by the scalar.
+   * @return Reference to the value.
+   * @note Behavior is undefined if the scalar is null.
+   */
+  const T& value() const { return *value_; }
+
+  /**
+   * @brief (Deprecated) Type-specific accessor for raw value.
+   * @tparam U Must match T.
+   * @return Reference to the value if types match.
+   * @throws std::bad_cast if U does not match T.
+   */
+  template <typename U>
+  const U& as() const {
+    if constexpr (std::is_same_v<T, U>) return *value_;
+    throw std::bad_cast();
+  }
+
+  std::shared_ptr<Scalar> clone() const override {
+    return std::make_shared<PrimitiveScalar<T>>(value_);
+  }
+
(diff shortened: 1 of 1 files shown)


src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * @brief Visitor interface for Scalar, supporting type-dispatched operations.
+ */
+class ICEBERG_EXPORT ScalarVisitor {
+ public:
+  virtual ~ScalarVisitor() = default;
+
+  /// Visit methods for primitive types
+  virtual void visit(int32_t value) = 0;
```
wgtmac, line 84: Shouldn't it accept `Int32Scalar`?

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+    } else {
+      visitor.visit(*value_);
+    }
+  }
+
+ private:
+  std::optional<T> value_;
+};
+
+// Aliases for common primitive scalar types
```
wgtmac, line 151: We are supposed to specialize each type like `DecimalType`, `BinaryType`, etc.?

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+class Scalar;
+class StructScalar;
+class ArrayScalar;
+class ScalarVisitor;
+class StructLike;
+class StructLikeVisitor;
+/**
+ * @brief Base class representing a generic scalar value.
+ */
+class ICEBERG_EXPORT Scalar {
```
wgtmac, line 40: Should we make `std::shared_ptr<Type>` as a member variable to `Scalar`?

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+using DoubleScalar = PrimitiveScalar<double>;
+using StringScalar = PrimitiveScalar<std::string>;
+using BoolScalar = PrimitiveScalar<bool>;
+
+/**
+ * @brief Base class for scalars that contain a vector of child Scalars.
+ * @tparam Derived The concrete subclass type (CRTP).
+ */
+template <typename Derived>
+class VectorScalarBase : public Scalar {
```
wgtmac, line 163: It is a little bit weird that values in it have different meanings in the `StructScalar` (horizontal) and `ArrayScalar` (vertical).

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+    } else {
+      visitor.visitStruct(*this);
+    }
+  }
+};
+
+/**
+ * @brief Represents an array-like scalar (list of values).
+ */
+class ArrayScalar : public VectorScalarBase<ArrayScalar> {
```
wgtmac, line 214: How do we define `MapScalar`? Should it be `array<struct<key,value>>`?

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+    } else {
+      visitor.visitArray(*this);
+    }
+  }
+};
+
+/**
+ * @brief Abstract interface for structured field access.
+ */
+class ICEBERG_EXPORT StructLike {
```
wgtmac, line 230: Can't we simply reuse `StructScalar`?

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+   * @param value The scalar to assign.
+   * @return Status indicating success or failure.
+   */
+  virtual Status set(int32_t pos, std::shared_ptr<Scalar> value) = 0;
+};
+
+/**
+ * @brief Visitor interface for traversing StructLike fields.
+ */
+class ICEBERG_EXPORT StructLikeVisitor {
```
wgtmac, line 265: ditto, we can remove this as well.

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+ */
+class ICEBERG_EXPORT Scalar {
+ public:
+  virtual ~Scalar() = default;
+
+  /**
+   * @brief Checks if the scalar represents a null value.
+   * @return true if the value is null, false otherwise.
+   */
+  virtual bool IsNull() const = 0;
```
lidavidm, line 48: Is the implication that this is not a non-trivial operation? (Else it might be `is_null` to imply it's a simple getter?)

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+    auto ptr = dynamic_cast<const T*>(this);
+    if (!ptr) throw std::bad_cast();
+    return *ptr;
+  }
+
+  /**
+   * @brief Creates a deep copy of the scalar.
+   * @return A shared pointer to the cloned scalar.
+   */
+  virtual std::shared_ptr<Scalar> clone() const = 0;
```
lidavidm, line 67: Shouldn't this be `Clone`? (Ditto in general for the PR)

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+  virtual void visit(int64_t value) = 0;
+  virtual void visit(float value) = 0;
+  virtual void visit(double value) = 0;
+  virtual void visit(bool value) = 0;
+  virtual void visit(const std::string& value) = 0;
+
+  /**
+   * @brief Called when visiting a null scalar.
+   */
+  virtual void visitNull() = 0;
```
lidavidm, line 94: Is this a specific NullScalar type, or any scalar with a null value?

src/iceberg/struct_like.h
```diff
@@ -0,0 +1,278 @@
...
+  explicit PrimitiveScalar(std::optional<T> value) : value_(std::move(value)) {}
+
+  bool IsNull() const override { return !value_.has_value(); }
+
+  /**
+   * @brief Gets the actual value held by the scalar.
+   * @return Reference to the value.
+   * @note Behavior is undefined if the scalar is null.
+   */
+  const T& value() const { return *value_; }
```
lidavidm, line 121: nit but perhaps in DEBUG it should use `optional::value()` to force an exception and in NDEBUG it can use `operator*` to bypass that

src/iceberg/struct_like.h
lidavidm: Ok I see the behavior below. IMO I'd rather we pass the scalars to the visitor so that the users can tell between different types of nulls

src/iceberg/struct_like.h
gty404: There are three possible approaches to organizing type information:
1. Scalar holds built-in types independently, and StructLike does not rely on StructType.
2. Scalar does not contain any type information, and its type is inferred from the associated StructType within StructLike.
3. Both Scalar and StructType carry type information, resulting in some redundancy.
I have chosen the first approach. Which one would you prefer?

src/iceberg/struct_like.h
gty404: Thanks, I will modify it.

src/iceberg/struct_like.h
gty404: In the Scalar class, `is_null` can be a member variable, or the `IsNull` interface can be implemented by subclasses, where subclasses might use `optional` or `pointer` to determine if it is null. Here, the second option was chosen; I actually don't have a strong preference. `StructScalar` also uses the `is_null` field, so it might be more appropriate to directly make `is_null` a member variable of Scalar. I'll make a modification.

src/iceberg/struct_like.h
gty404: image
The PrimitiveScalar::Accept method directly uses value, and using PrimitiveScalar is also acceptable. I will make some modifications.

src/iceberg/struct_like.h
gty404: Good suggestion, visit I will modify it to accept Scalar types as input

src/iceberg/struct_like.h
gty404: Good suggestion, I will modify here

src/iceberg/struct_like.h
gty404: I think it should be specialized.

src/iceberg/struct_like.h
gty404: It is a bit strange because the type information is all in StructLike, with StructScalar/ArrayScalar both being a group of Scalar members and there is not much difference.

src/iceberg/struct_like.h
gty404: Currently, I haven't thought of a good design, perhaps map can still be used, and Scalar needs to add interfaces for hash and equality judgment?

src/iceberg/struct_like.h
gty404: StructLike has a StructType, while StructScalar does not have one, I think this should depend on the discussion result regarding whether a Scalar should have its own type information.

src/iceberg/struct_like.h
wgtmac: If it accepts physical values instead of typed values, how can it differentiate between the actual types? For example, long and timestamp both use `int64_t`.

src/iceberg/struct_like.h
wgtmac: If Scalar does not carry the type info, we have to pass Schema wherever Scalar is used. In the case of schema evolution, then we need both Scalar schema and expected (table) schema.

src/iceberg/struct_like.h
wgtmac: But this sounds like an anti-pattern in the OOP. You cannot use the base `VectorScalarBase` class without knowing its subclass implementation.

PR #78, opened 2025-04-16
feat: adopt MurmurHash3

zhjwpku:
Which will be used by BucketTransform.


Fokko (approved):

Fokko:
Thanks @zhjwpku for adding this 👍

PR #80, opened 2025-04-17
feat: implement initial MemoryCatalog functionality with namespace and table support

gty404:
- Added NamespaceContainer class to manage namespace hierarchy and table metadata
- Implemented MemoryCatalog methods: ListTables, TableExists, DropTable, RegisterTable


src/iceberg/catalog/memory_catalog.cc
```diff
@@ -0,0 +1,248 @@
...
+      table_names, std::back_inserter(table_idents),
+      [&ns](auto const& table_name) { return TableIdentifier(ns, table_name); });
+  return table_idents;
+}
+
+Result<std::unique_ptr<Table>> MemoryCatalog::CreateTable(
+    const TableIdentifier& identifier, const Schema& schema, const PartitionSpec& spec,
+    const std::string& location,
+    const std::unordered_map<std::string, std::string>& properties) {
+  throw IcebergError("not implemented");
```
zhjwpku: maybe return NotImplementedError after #75 merged.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,228 @@
...
+  std::string catalog_name_;
+  std::unordered_map<std::string, std::string> properties_;
+  std::shared_ptr<FileIO> file_io_;
+  std::optional<std::string> warehouse_location_;
+  std::unique_ptr<NamespaceContainer> root_container_;
+  mutable std::recursive_mutex mutex_;
+};
+
+/**
+ * @brief A hierarchical container that manages namespaces and table metadata in-memory.
```
zhjwpku: We seem to use `\` other than `@`.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,228 @@
...
+      const NamespaceContainer* root, const Namespace& namespace_ident);
+
+  /**
+   * @brief Templated implementation for retrieving a NamespaceContainer node.
+   * @tparam ContainerPtr Pointer type to NamespaceContainer (const or non-const).
+   * @param root The root node.
+   * @param namespace_ident The namespace path.
+   * @return A pointer to the container node if found; nullptr otherwise.
+   */
+  template <typename ContainerPtr>
```
zhjwpku: Why not just NamespaceContainer here?

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,228 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+#include <mutex>
+#include <optional>
+#include <unordered_map>
+
+#include "iceberg/catalog.h"
```
wgtmac: If we add a new subdirectory, we need to add a CMakeLists.txt in it to make sure new headers are installed.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,228 @@
...
+
+namespace iceberg {
+
+class NamespaceContainer;
+
+class ICEBERG_EXPORT MemoryCatalog : public Catalog {
+ public:
+  MemoryCatalog(std::shared_ptr<FileIO> file_io,
+                std::optional<std::string> warehouse_location);
```
wgtmac: Why this is optional?

src/iceberg/catalog/memory_catalog.h
wgtmac: +1

We should use doxygen format: https://www.doxygen.nl/manual/docblocks.html

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,228 @@
...
+   */
+  bool NamespaceExists(const Namespace& namespace_ident) const;
+
+  /**
+   * @brief Lists immediate child namespaces under the given parent namespace.
+   * @param parent_namespace_ident The optional parent namespace. If not provided,
+   *                                the children of the root are returned.
+   * @return A vector of child namespace names.
+   */
+  std::vector<std::string> ListChildrenNamespaces(
```
wgtmac: Why not return `std::vector<Namespace>`? Does it return next level or all levels?

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,228 @@
...
+
+ private:
+  /**
+   * @brief Helper function to retrieve the container node for a given namespace.
+   * @param root The root of the tree to search.
+   * @param namespace_ident The namespace path to traverse.
+   * @return A pointer to the corresponding NamespaceContainer if it exists; nullptr
+   * otherwise.
+   */
+  static NamespaceContainer* GetNamespaceContainer(NamespaceContainer* root,
```
wgtmac: Functions like these can be moved to memory_catalog.cc in an anonymous namespace.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,228 @@
...
+                                                   const Namespace& namespace_ident);
+
+  /**
+   * @brief Const version of GetNamespaceContainer.
+   */
+  static const NamespaceContainer* GetNamespaceContainer(
```
wgtmac: ditto

src/iceberg/catalog/memory_catalog.cc
```diff
@@ -0,0 +1,248 @@
...
+    if (auto it = container->children_.find(part_level);
+        it == container->children_.end()) {
+      container = &container->children_[part_level];
+      newly_created = true;
+    } else {
+      container = &it->second;
+    }
+  }
+
+  if (!newly_created) return false;
```
wgtmac:
```suggestion
  if (!newly_created) {
    return false;
  }
```

I'd suggest to add curly braces for better readability. Same for below.

src/iceberg/catalog/memory_catalog.h
gty404: It is returning the namespace of the next level, returning std::string instead of Namespace, I think this makes it easier to get the name.

src/iceberg/catalog/memory_catalog.h
gty404: ContainerPtr can be either NamespaceContainer* or const NamespaceContainer*.

zhjwpku: Good to know, thanks

Fokko:
I know that you already took the time and effort to write this, but I did want to put this out here. In PyIceberg we've removed the MemoryCatalog from the codebase, mostly because the SQLCatalog (JDBCCatalog in Java) was proven to be much better. The SQLCatalog also allows running in-memory using a SQLite in-memory instance, but also provides the option to easily keep the state after the process terminates. Next to that, it also provides locking out of the box.

wgtmac:
@Fokko AFAIK, the C++ ecosystem does not yet have a good ORM framework so it might be challenging to implement a `SqlCatalog` at the moment. @gty404 is an expert on this and he may have insight in it.

gty404:
@Fokko When implementing JDBCCatalog, ORM and SQLite need to be introduced, as @wgtmac  also mentioned. It's difficult to find a suitable ORM framework in the C++ system, which might be due to the license or ease of use.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
...
+      const TableIdentifier& identifier,
+      const std::vector<std::unique_ptr<UpdateRequirement>>& requirements,
+      const std::vector<std::unique_ptr<MetadataUpdate>>& updates) override;
+
+  Result<std::shared_ptr<Transaction>> StageCreateTable(
+      const TableIdentifier& identifier, const Schema& schema, const PartitionSpec& spec,
+      const std::string& location,
+      const std::unordered_map<std::string, std::string>& properties) override;
+
+  bool TableExists(const TableIdentifier& identifier) const override;
```
yingcai-cy: Can these two interfaces also return a result like above, so that we can return clear errors like backend database errors.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
...
+};
+
+/**
+ * \brief A hierarchical container that manages namespaces and table metadata in-memory.
+ *
+ * Each NamespaceContainer represents a namespace level and can contain properties,
+ * tables, and child namespaces. This structure enables a tree-like representation
+ * of nested namespaces.
+ */
+class ICEBERG_EXPORT NamespaceContainer {
```
yingcai-cy: InMemoryNamespace?

Fokko:
@gty404 @wgtmac I see, got it. Thanks for considering it, then the InMemoryCatalog is the next best option 👍

src/iceberg/catalog/memory_catalog.h
gty404: I think it's okay, but the Catalog interface is defined like this, what does @wgtmac  think?

wgtmac: Feel free to change this. This interface was added before we have the `Result` type.

gty404:
@yingcai-cy @wgtmac @lidavidm @zhjwpku Could you help me review it again?

src/iceberg/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set(ICEBERG_INCLUDES "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/src>"
                      "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>")
 set(ICEBERG_SOURCES
     arrow_c_data_internal.cc
+    catalog/memory_catalog.cc
```
wgtmac: Rename it to `in_memory_catalog.cc`

src/iceberg/catalog.h
```diff
@@ -90,8 +90,11 @@ class ICEBERG_EXPORT Catalog {
   /// \brief Check whether table exists
   ///
   /// \param identifier a table identifier
-  /// \return true if the table exists, false otherwise
-  virtual bool TableExists(const TableIdentifier& identifier) const = 0;
+  /// \return Status indicating success or failure.
+  ///         - On success, the table existence was successfully checked (actual existence
+  ///         may be inferred elsewhere).
+  ///         - On failure, contains error information.
+  virtual Status TableExists(const TableIdentifier& identifier) const = 0;
```
wgtmac: What about `Result<bool>`? I don't think it is an error when table does not exist. Same for below.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
...
+      const TableIdentifier& identifier) const override;
+
+  Result<std::shared_ptr<Table>> RegisterTable(
+      const TableIdentifier& identifier,
+      const std::string& metadata_file_location) override;
+
+  std::unique_ptr<iceberg::TableBuilder> BuildTable(const TableIdentifier& identifier,
+                                                    const Schema& schema) const override;
+
+ private:
```
wgtmac: nit: what about using `pimpl` idiom to hide the implementation detail like `InMemoryNamespace` and reduce compile time?

```
class ICEBERG_EXPORT InMemoryCatalog : public Catalog {
public:
  ...

private:
  std::unique_ptr<class InMemoryCatalogImpl> impl_;
};
```

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
...
+
+namespace iceberg {
+
+class InMemoryNamespace;
+
+class ICEBERG_EXPORT InMemoryCatalog : public Catalog {
+ public:
+  InMemoryCatalog(std::shared_ptr<FileIO> file_io, std::string warehouse_location);
+
+  void Initialize(
```
wgtmac: The Java impl provides a default constructor to facilitate creating instance via reflection. That's why it has a `initialize` method. Perhaps we can remove it and move these parameters to the constructor?

src/iceberg/catalog/memory_catalog.cc
```diff
@@ -0,0 +1,275 @@
...
+      table_names.value(), std::back_inserter(table_idents),
+      [&ns](auto const& table_name) { return TableIdentifier(ns, table_name); });
+  return table_idents;
+}
+
+Result<std::unique_ptr<Table>> InMemoryCatalog::CreateTable(
+    const TableIdentifier& identifier, const Schema& schema, const PartitionSpec& spec,
+    const std::string& location,
+    const std::unordered_map<std::string, std::string>& properties) {
+  return unexpected<Error>(
```
zhjwpku: return NotImplemented would be better. The same with other places that using unexpected.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
+/*
```
zhjwpku: rename the filename to in_memory_catalog.h/cc?

zhjwpku (approved):
I have two minor review comments, other than that, LGTM.

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
...
+
+#include <mutex>
+#include <optional>
+#include <unordered_map>
+
+#include "iceberg/catalog.h"
+
+namespace iceberg {
+
+class InMemoryNamespace;
```
wgtmac: nit: you can forward declare it when it appears for the 1st time to use less lines :)

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
...
+   */
+  Status NamespaceExists(const Namespace& namespace_ident) const;
+
+  /**
+   * \brief Lists immediate child namespaces under the given parent namespace.
+   * \param[in] parent_namespace_ident The optional parent namespace. If not provided,
+   *                                the children of the root are returned.
+   * \return A vector of child namespace names.
+   */
+  Result<std::vector<std::string>> ListChildrenNamespaces(
```
wgtmac: nit: use std::span

src/iceberg/catalog/memory_catalog.h
```diff
@@ -0,0 +1,204 @@
...
+};
+
+/**
+ * \brief A hierarchical namespace that manages namespaces and table metadata in-memory.
+ *
+ * Each InMemoryNamespace represents a namespace level and can contain properties,
+ * tables, and child namespaces. This structure enables a tree-like representation
+ * of nested namespaces.
+ */
+class ICEBERG_EXPORT InMemoryNamespace {
```
wgtmac: I was simply mirroring the Java impl when I added the `Catalog` without notice of `SupportsNamespace`: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/catalog/SupportsNamespaces.java. Does it make sense to add all interfaces from `SupportsNamespace` to our `Catalog` API? `iceberg-python` did the same thing: https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/__init__.py#L588 This aligns with the rest catalog api.

@Fokko Do you have an opinion?

Fokko: Hey @wgtmac I think it makes sense to add all of them to the `Catalog` API. Keep in mind that some implementations (Rest, SQL) support nested namespaces, and some don't (Hive, Glue etc)

src/iceberg/catalog/memory_catalog.h
gty404: How about placing InMemoryCatalogImpl and InMemoryNamespace in another header file? Their unit tests are also easy to add.

src/iceberg/catalog/memory_catalog.h
gty404: The returned Namespace list is a temporary variable.

src/iceberg/catalog/memory_catalog.h
wgtmac: Why unit tests are required to have access to InMemoryCatalogImpl? Isn't the `Catalog` interface enough to test?

gty404: Don't InMemoryCatalogImpl and InMemoryNamespace require corresponding unit tests? The interfaces provided by these two may not be completely covered by the InMemoryCatalog test

wgtmac: Yes, that's reasonable.

src/iceberg/catalog/in_memory_catalog.h
```diff
@@ -0,0 +1,70 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+#include "iceberg/catalog.h"
+
+namespace iceberg {
+
+class ICEBERG_EXPORT InMemoryCatalog : public Catalog {
```
wgtmac, line 35: Add some comments to describe its purpose. Especially to clarify that it is not supposed for production use.

src/iceberg/catalog/in_memory_catalog.cc
```diff
@@ -0,0 +1,496 @@
...
+namespace {
+
+/**
+ * \brief A hierarchical namespace that manages namespaces and table metadata in-memory.
+ *
+ * Each InMemoryNamespace represents a namespace level and can contain properties,
+ * tables, and child namespaces. This structure enables a tree-like representation
+ * of nested namespaces.
+ */
+class ICEBERG_EXPORT InMemoryNamespace {
```
wgtmac, line 41: As discussed in https://github.com/apache/iceberg-cpp/pull/80#discussion_r2059685178, we need to move some of these functions (which are included in the rest catalog spec) as a part of the public Catalog API.

src/iceberg/catalog/in_memory_catalog.cc
```diff
@@ -0,0 +1,496 @@
...
+ public:
+  /**
+   * \brief Checks whether the given namespace exists.
+   * \param[in] namespace_ident The namespace to check.
+   * \return Status indicating success or failure.
+   */
+  Status NamespaceExists(const Namespace& namespace_ident) const;
```
wgtmac: Use `Result<bool>`

src/iceberg/catalog/in_memory_catalog.cc
```diff
@@ -0,0 +1,496 @@
...
+
+  /**
+   * \brief Lists immediate child namespaces under the given parent namespace.
+   * \param[in] parent_namespace_ident The optional parent namespace. If not provided,
+   *                                the children of the root are returned.
+   * \return A vector of child namespace names.
+   */
+  Result<std::vector<std::string>> ListChildrenNamespaces(
```
wgtmac: If parent can be optional, is it better to rename it to `ListNamespaces`?

src/iceberg/catalog.h
```diff
@@ -42,6 +43,67 @@ class ICEBERG_EXPORT Catalog {
...
+  virtual Result<bool> NamespaceExists(const Namespace& ns) const = 0;
+
+  /// \brief Set metadata properties on a namespace.
+  ///
+  /// \param ns the namespace to modify
+  /// \param properties the properties to set or update
+  /// \return Status::OK if updated successfully;
+  ///         ErrorKind::kNoSuchNamespace if the namespace does not exist;
+  ///         ErrorKind::kNotSupported if the operation is not supported
+  virtual Status SetNamespaceProperties(
```
wgtmac: What about merging `SetNamespaceProperties` and `RemoveNamespaceProperties` into a single `UpdateNamespaceProperties(updates, removals)`? This is a single call in the rest catalog spec.

src/iceberg/catalog/in_memory_catalog.h
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * @brief An in-memory implementation of the Iceberg Catalog interface.
+ *
+ * This catalog stores all metadata purely in memory, with no persistence to disk
+ * or external systems. It is primarily intended for unit tests, prototyping, or
+ * demonstration purposes.
+ *
+ * @note This class is **not** suitable for production use.
+ *       All data will be lost when the process exits.
+ */
```
wgtmac, line 34:
```suggestion
 /// \brief An in-memory implementation of the Iceberg Catalog interface.
 ///
 /// This catalog stores all metadata purely in memory, with no persistence to disk
 /// or external systems. It is primarily intended for unit tests, prototyping, or
 /// demonstration purposes.
 ///
 /// \note This class is **not** suitable for production use.
 ///       All data will be lost when the process exits.
```

wgtmac (approved):
Thanks for updating it! I've left some minor comments.

src/iceberg/catalog.h
gty404: I am referring to the implementation in java implementations, which has two interfaces. Should we follow the REST catalog spec?

wgtmac: @Fokko WDYT?

wgtmac (approved):

Fokko:
Yes, looks good @gty404 thanks!

PR #81, opened 2025-04-18
refactor: add expression subdirectory

wgtmac:


src/iceberg/result.h
```diff
@@ -78,4 +78,12 @@ auto JsonParseError(const std::format_string<Args...> fmt, Args&&... args)
...
 }
 
+/// \brief Create an unexpected error with kInvalidExpression
+template <typename... Args>
+auto InvalidExpressionError(const std::format_string<Args...> fmt, Args&&... args)
+    -> unexpected<Error> {
+  return unexpected<Error>({.kind = ErrorKind::kInvalidExpression,
+                            .message = std::format(fmt, std::forward<Args>(args)...)});
+}
+
```
zhjwpku, line 88: Should we add an Error wrapper for each ErrorKind? I'd suggest including all those in this PR if needed. #80 might need NoSuchNamespaceError.

zhjwpku (approved):

src/iceberg/result.h
wgtmac: That's a good idea! We can use a separate PR for this.

yingcai-cy (approved):
Thanks for doing this.

Xuanwo (approved):
The community has reached a consensus. Let’s move forward!

wgtmac:
Thank you @Xuanwo!

PR #82, opened 2025-04-18
test: add table metadata deserialization test

wgtmac:
- Added a set of test files from https://github.com/apache/iceberg/tree/main/core/src/test/resources
- Fixed table metadata deserialization
- Added singleton for unsorted sort order and unpartitioned spec
- Added test case for deserializing valid v1 and v2 metadata files


src/iceberg/sort_order.cc
```diff
@@ -21,14 +21,21 @@
...
 namespace iceberg {
 
-SortOrder::SortOrder(int64_t order_id, std::vector<SortField> fields)
+SortOrder::SortOrder(int32_t order_id, std::vector<SortField> fields)
     : order_id_(order_id), fields_(std::move(fields)) {}
 
-int64_t SortOrder::order_id() const { return order_id_; }
+const std::shared_ptr<SortOrder>& SortOrder::Unsorted() {
+  static const std::shared_ptr<SortOrder> unsorted =
+      std::make_shared<SortOrder>(kInitialSortOrderId, std::vector<SortField>{});
```
zhjwpku:
```suggestion
      std::make_shared<SortOrder>(kInitialSortOrderId - 1, std::vector<SortField>{});
```

Java impl use 0 as UNSORTED_ORDER order_id, so here should minus 1.

zhjwpku (changes requested):

src/iceberg/sort_order.cc
wgtmac: Good catch!

zhjwpku (approved):
LGTM

src/iceberg/table_metadata.cc
```diff
@@ -34,6 +38,33 @@ std::string ToString(const MetadataLogEntry& entry) {
                      entry.metadata_file);
 }
 
+const std::shared_ptr<Schema>& TableMetadata::Schema() const {
+  static const std::shared_ptr<::iceberg::Schema> empty_schema = nullptr;
```
gty404: Why not return a truly empty schema object, but null?

zhjwpku: I once had the same concern, but this method returns a reference, I guess that's the reason?

wgtmac: I switched to wrap the return in a Result so we can get an explicit `NotFound` error.

.pre-commit-config.yaml
```diff
@@ -33,6 +33,7 @@ repos:
     rev: v19.1.5
     hooks:
       - id: clang-format
+        exclude: ^test/resources/.*\.json$
```
Fokko, line 36: Does clang want to format json? 🤔

Fokko (approved):
This looks great @wgtmac, thanks for adding these test-cases 👍

.pre-commit-config.yaml
wgtmac: Yes, without this change the pre-commit fails and complains that `clang-format` does not support JSON file.

Xuanwo (approved):
The community has reached a consensus. Let's move forward!

PR #83, opened 2025-04-18
chore: use chrono::milliseconds in snapshot and consolidate error usage

zhjwpku:
reduce direct unexpected usage by using more Error wrappers defined in result.h

### DIFF
--- a/src/iceberg/result.h
+++ b/src/iceberg/result.h
@@ -28,20 +28,19 @@
 namespace iceberg {
 
 /// \brief Error types for iceberg.
-/// TODO: add more and sort them based on some rules.
 enum class ErrorKind {
-  kNoSuchNamespace,
   kAlreadyExists,
-  kNoSuchTable,
   kCommitStateUnknown,
-  kInvalidSchema,
   kInvalidArgument,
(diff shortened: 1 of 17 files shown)


src/iceberg/result.h
```diff
@@ -86,4 +93,67 @@ auto InvalidExpressionError(const std::format_string<Args...> fmt, Args&&... arg
                             .message = std::format(fmt, std::forward<Args>(args)...)});
 }
 
+/// \brief Create an unexpected error with kInvalidSchema
```
wgtmac: Perhaps we can use a macro to define these to avoid duplicates?

zhjwpku: done, thanks

lidavidm (approved):

wgtmac (approved):

zhjwpku:
@Fokko @Xuanwo This PR should benefit future dev with better error handling, I'd appreciate if you can take a look, thanks.

Fokko:
Looks great @zhjwpku Thanks for working on this, and thanks @lidavidm and @wgtmac for the review 🙌

PR #85, opened 2025-04-23
feat: add table metadata reader and writer

wgtmac:


src/iceberg/table_metadata.h
```diff
@@ -140,4 +141,29 @@ ICEBERG_EXPORT std::string ToString(const SnapshotLogEntry& entry);
...
+  kGzip,
+};
+
+/// \brief Utility class for table metadata
+struct ICEBERG_EXPORT TableMetadataUtil {
+  /// \brief Get the codec type from the table metadata file name.
+  ///
+  /// \param file_name The name of the table metadata file.
+  /// \return The codec type of the table metadata file.
+  static Result<MetadataFileCodecType> FromFileName(std::string_view file_name);
```
lidavidm: nit: maybe CodecFromFileName?

lidavidm (approved):

src/iceberg/table_metadata.h
wgtmac: Make sense. Changed.

lidavidm (approved):

src/iceberg/json_internal.cc
```diff
@@ -1209,4 +1209,20 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataFromJson(const nlohmann::jso
   return table_metadata;
 }
 
+Result<nlohmann::json> FromJsonString(const std::string& json_string) {
+  try {
+    return nlohmann::json::parse(json_string);
+  } catch (const std::exception& e) {
```
yingcai-cy, line 1215: nit:should we only catch json::parse_error exception?

wgtmac: I agree that the official documentation says it only throws `parse_error`. I just want to be safe enough to catch all just in case. Or we can add a special branch for `parse_error` for better error message.

src/iceberg/table_metadata.cc
```diff
@@ -69,4 +76,100 @@ Result<std::shared_ptr<SortOrder>> TableMetadata::SortOrder() const {
...
+namespace {
+
+template <typename T>
+bool SharedPtrVectorEquals(const std::vector<std::shared_ptr<T>>& lhs,
+                           const std::vector<std::shared_ptr<T>>& rhs) {
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+  for (size_t i = 0; i < lhs.size(); ++i) {
+    if (*lhs[i] != *rhs[i]) {
```
yingcai-cy, line 88: Could be null pointer?

wgtmac: It could be null if there is something wrong. Since `operator==` is only used in the test case, it should be fine. I can add a `NullSafeEquals` later if this is required.

yingcai-cy (approved):

zhjwpku (approved):
LGTM, thanks

src/iceberg/json_internal.cc
lidavidm: You can also call an overload that doesn't throw an exception, and explicitly check whether the resulting JSON object is `discarded()` (IIRC)

wgtmac: I've created https://github.com/apache/iceberg-cpp/issues/87 to track it.

Fokko (approved):

Fokko:
Thanks @wgtmac for working on this, and thanks @lidavidm, @zhjwpku and @yingcai-cy for the review 🙌

PR #86, opened 2025-04-23
feat: add formatter specialization

wgtmac:
- Added std::formatter specialization for std::map, std::unordered_map and std::vector when their elements are also formattable.
- Added std::formatter specialization for classes that have a ToString at hands.

### DIFF
--- a/src/iceberg/util/formatter.h
+++ b/src/iceberg/util/formatter.h
@@ -27,10 +27,7 @@
 
 #include <concepts>
 #include <format>
-#include <map>
 #include <string_view>
-#include <unordered_map>
-#include <vector>
 
 #include "iceberg/util/formattable.h"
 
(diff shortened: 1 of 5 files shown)


src/iceberg/util/formatter.h
```diff
@@ -40,3 +43,102 @@ struct std::formatter<Derived> : std::formatter<std::string_view> {
     return std::formatter<string_view>::format(obj.ToString(), ctx);
   }
 };
+
+/// \brief std::formatter specialization for std::vector
+template <typename T>
+struct std::formatter<std::vector<T>> : std::formatter<std::string_view> {
```
lidavidm: FWIW, I _think_ the "idiomatic" way is supposed to be `std::format::join` and ranges. I wonder if specializing on std types may come back to bite us later.

Example: https://stackoverflow.com/a/77993463

lidavidm: Or another example, albeit C++23: https://stackoverflow.com/a/71196142

test/formatter_test.cc
```diff
@@ -0,0 +1,163 @@
...
+TEST(FormatterTest, NestedContainersFormat) {
+  std::vector<std::map<std::string, int>> nested = {{{"a", 1}, {"b", 2}},
+                                                    {{"c", 3}, {"d", 4}}};
+
+  EXPECT_EQ("[{a: 1, b: 2}, {c: 3, d: 4}]", std::format("{}", nested));
+
+  std::map<std::string, std::vector<int>> nested_map = {
+      {"primes", {2, 3, 5, 7, 11}}, {"fibonacci", {1, 1, 2, 3, 5, 8, 13}}};
+  std::string result = std::format("{}", nested_map);
+  EXPECT_TRUE(result.find("primes") != std::string::npos);
```
lidavidm: Note that you can do `EXPECT_THAT(result, testing::HasSubstr("primes"));` which will also give a much better error on failure

src/iceberg/util/formatter.h
```diff
@@ -40,3 +43,102 @@ struct std::formatter<Derived> : std::formatter<std::string_view> {
...
+  std::string result = "{";
+
+  bool first = true;
+  for (const auto& [key, value] : map) {
+    if (!first) {
+      std::format_to(std::back_inserter(result), ", ");
+    }
+
+    // Format key (handle if it's a smart pointer)
+    if constexpr (requires { *key; }) {
```
lidavidm: Won't this also fire on something like `std::optional` that might be undesirable?

src/iceberg/util/formatter.h
wgtmac: I think we need to rename this file to `formatter_internal.h` so we don't expose it to the downstream. In this approach, we take full control of it.

src/iceberg/util/formatter.h
wgtmac: I've modified the concept to strictly check smart pointer type.

src/iceberg/util/formatter.h
lidavidm: SGTM.

It might be good to expose formatter specializations on our own types, just so long as they're separate from the specializations for std types

lidavidm: Ah ok I see that's what you've done.

lidavidm (approved):

wgtmac:
I've moved specialization for std containers to the internal header file to not pollute downstream. Let me know what you think @lidavidm

src/iceberg/util/formatter.h
wgtmac: Yes :)

lidavidm:
LGTM

zhjwpku (approved):
Impressive, I used chatgpt to help me understanding the C++20 std::views::transform, thanks for working on this.

yingcai-cy (approved):
Good job!

Fokko:
Thanks @wgtmac for working on this, and thanks @lidavidm, @zhjwpku  and @yingcai-cy for the review 🙌

PR #88, opened 2025-04-25
feat: add file reader interface

wgtmac:


wgtmac:
@lidavidm @zhjwpku Could you help review this?

src/iceberg/file_reader.h
```diff
@@ -0,0 +1,142 @@
...
+struct ICEBERG_EXPORT ReaderOptions {
+  /// \brief The path to the file to read.
+  std::string path;
+  /// \brief The total length of the file.
+  std::optional<size_t> length;
+  /// \brief The split to read.
+  std::optional<Split> split;
+  /// \brief The batch size to read. Only applies to implementations that support
+  /// batching.
+  int64_t batch_size;
```
lidavidm, line 108: Is the unit bytes or rows?

(Should this also be optional?)

lidavidm: Ah, presumably bytes?

lidavidm: (Why does Split use size_t but this uses int64_t?)

lidavidm (approved):

src/iceberg/file_reader.h
wgtmac: I was referring to number of rows in a batch (a.k.a. ArrowArray). It seems that we can use `std::optional<size_t>` and define a default batch size if non provided.

lidavidm: Ah, ok. int64_t might be appropriate as that's then consistent with C Data Interface. Clarifying the unit might be helpful.

src/iceberg/file_format.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace iceberg {
+
+/// \brief File format type
+enum class ICEBERG_EXPORT FileFormatType {
+  kParquet,
+  kAvro,
+  kOrc,
+  kPuffin,
+  kMetadata,
```
zhjwpku: Should this be kJson? Metadata is a file format to me ;(

src/iceberg/file_reader.h
```diff
@@ -0,0 +1,142 @@
...
+ public:
+  explicit StructLikeReader(std::unique_ptr<Reader> reader);
+
+  /// \brief Always read data into `StructLike` or monostate if no more data.
+  Result<Data> Next() final;
+
+  DataLayout data_layout() const final { return DataLayout::kStructLike; }
+
+ private:
+  std::unique_ptr<Reader> reader_;
```
zhjwpku, line 70: I’m a bit confused about why keeping a Reader object here.

zhjwpku (approved):

src/iceberg/file_format.h
zhjwpku: ugh, I just checked the java impl, kMetadata should be fine, sorry for the noise.

src/iceberg/file_reader.h
wgtmac: Because this is just an adapter. If the wrapped reader directly returns kStructLike, it does nothing. Otherwise it will try to aggregate them into an ArrowArray.

src/iceberg/file_format.h
Fokko: The `kMetadata` confuses me a bit. I asssume that would mean Iceberg table metadata (which is encoded in Avro).

Wouldn't it make more sense to align this with the `file_format` field:

image

wgtmac: Make sense. I think we don't have to copy what the java impl does. Removed `kMetadata`.

Fokko (approved):

PR #90, opened 2025-05-01
chore: fix a build error due to name conflict

zhjwpku:
FetchContent_MakeAvailable generate a <lowercaseName>_SOURCE_DIR[1], for arrow it's arrow_SOURCE_DIR. arrow use a internal arrow_SOURCE_DIR to set its THIRDPARTY_DIR[2].

The generated arrow_SOURCE_DIR overwrite the arrow's interal arrow_SOURCE_DIR, causing the following build error:

CMake Error at build/_deps/arrow-src/cpp/cmake_modules/ThirdpartyToolchain.cmake:449 (file):
  file STRINGS file
  "/Users/zhjwpku/zhjwpku/iceberg-cpp/build/_deps/arrow-src/thirdparty/versions.txt"
  cannot be read.

Fix this issue by renaming fetchcontent_declare name from Arrow to VendoredArrow.

[1] https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable
[2] https://github.com/apache/arrow/blob/main/cpp/cmake_modules/ThirdpartyToolchain.cmake#L368

### DIFF
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -87,7 +87,7 @@ function(resolve_arrow_dependency)
       "BUNDLED"
       CACHE STRING "" FORCE)
 
-  fetchcontent_declare(Arrow
+  fetchcontent_declare(VendoredArrow
                        ${FC_DECLARE_COMMON_OPTIONS}
                        GIT_REPOSITORY https://github.com/wgtmac/arrow.git
                        GIT_TAG 7d50c4ac803ad983734de5f418b7cd18f25b0dc9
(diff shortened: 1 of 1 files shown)


lidavidm (approved):

wgtmac (approved):
+1

Where did you hit this issue?

zhjwpku:
> +1
> 
> Where did you hit this issue?

I just rebased the main branch and removed the build dir then started a new config/build, using the following:

```
cmake -S . -B build -DCMAKE_INSTALL_PREFIX=/tmp/iceberg/install -DICEBERG_BUILD_STATIC=ON -DICEBERG_BUILD_SHARED=ON
```

I'm wondering why our CI doesn't hit this 😵‍💫

Fokko (approved):

Fokko:
Thanks @zhjwpku for fixing this right away, and thanks @lidavidm and @wgtmac for the quick review

wgtmac:
I tried to reproduce it but unfortunately it built successfully.

zhjwpku:
> I tried to reproduce it but unfortunately it built successfully.

I think this issue may be caused by a bug in CMake, which appears to have been fixed in version 3.30. However, since we need to support CMake versions back to 3.25, this change seems reasonable.

I checked GH runner image[1], the CMake version is 3.31.6, which explains why our CI doesn't trigger the build error.

I bet your CMake version is greater than 3.30 ;)

```
Merge: 6323fdc9cf 86ad7cc886
Author: Brad King <brad.king@kitware.com>
Date:   Mon Sep 23 18:13:34 2024 +0000

    Merge topic 'project-vars' into release-3.30

    86ad7cc886 project: Only define non-cache vars if already defined
    4c152752da Help: State valid scopes for using proj_SOURCE_DIR and proj_BINARY_DIR

    Acked-by: Kitware Robot <kwrobot@kitware.com>
    Acked-by: buildbot <buildbot@kitware.com>
    Merge-request: !9820
```

[1] https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools

wgtmac:
Ah, that makes sense. Perhaps we can add a comment to your fix so we can restore the Arrow project name once bumping the minimum CMake version?

zhjwpku:
> Ah, that makes sense. Perhaps we can add a comment to your fix so we can restore the Arrow project name once bumping the minimum CMake version?

Will add a TODO later, thanks

PR #91, opened 2025-05-05
feat: add manifest related structures

zhjwpku:
Add DataFile, ManifestEntry, ManifestFile, and ManifestList to Iceberg core.
Support for parsing these data structures from Avro file will be added in future PRs.

### DIFF
--- a/src/iceberg/manifest_entry.cc
+++ b/src/iceberg/manifest_entry.cc
@@ -1,0 +1,161 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/manifest_entry.h"
+
+#include <memory>
+#include <optional>
+#include <vector>
+
+#include "iceberg/schema.h"
+#include "iceberg/schema_field.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+const SchemaField DataFile::CONTENT =
+    SchemaField::MakeRequired(134, "content", std::make_shared<IntType>());
+const SchemaField DataFile::FILE_PATH =
+    SchemaField::MakeRequired(100, "file_path", std::make_shared<StringType>());
+const SchemaField DataFile::FILE_FORMAT =
+    SchemaField::MakeRequired(101, "file_format", std::make_shared<IntType>());
+const SchemaField DataFile::RECORD_COUNT =
+    SchemaField::MakeRequired(103, "record_count", std::make_shared<LongType>());
+const SchemaField DataFile::FILE_SIZE =
+    SchemaField::MakeRequired(104, "file_size_in_bytes", std::make_shared<LongType>());
+const SchemaField DataFile::COLUMN_SIZES = SchemaField::MakeOptional(
+    108, "column_sizes",
+    std::make_shared<MapType>(
+        SchemaField::MakeRequired(117, std::string(MapType::kKeyName),
+                                  std::make_shared<IntType>()),
+        SchemaField::MakeRequired(118, std::string(MapType::kValueName),
+                                  std::make_shared<LongType>())));
+const SchemaField DataFile::VALUE_COUNTS = SchemaField::MakeOptional(
+    109, "value_counts",
+    std::make_shared<MapType>(
+        SchemaField::MakeRequired(119, std::string(MapType::kKeyName),
+                                  std::make_shared<IntType>()),
+        SchemaField::MakeRequired(120, std::string(MapType::kValueName),
+                                  std::make_shared<LongType>())));
+const SchemaField DataFile::NULL_VALUE_COUNTS = SchemaField::MakeOptional(
+    110, "null_value_counts",
+    std::make_shared<MapType>(
+        SchemaField::MakeRequired(121, std::string(MapType::kKeyName),
+                                  std::make_shared<IntType>()),
+        SchemaField::MakeRequired(122, std::string(MapType::kValueName),
+                                  std::make_shared<LongType>())));
+const SchemaField DataFile::NAN_VALUE_COUNTS = SchemaField::MakeOptional(
+    137, "nan_value_counts",
+    std::make_shared<MapType>(
+        SchemaField::MakeRequired(138, std::string(MapType::kKeyName),
+                                  std::make_shared<IntType>()),
+        SchemaField::MakeRequired(139, std::string(MapType::kValueName),
+                                  std::make_shared<LongType>())));
+const SchemaField DataFile::LOWER_BOUNDS = SchemaField::MakeOptional(
+    125, "lower_bounds",
+    std::make_shared<MapType>(
+        SchemaField::MakeRequired(126, std::string(MapType::kKeyName),
+                                  std::make_shared<IntType>()),
+        SchemaField::MakeRequired(127, std::string(MapType::kValueName),
+                                  std::make_shared<BinaryType>())));
+const SchemaField DataFile::UPPER_BOUNDS = SchemaField::MakeOptional(
+    128, "upper_bounds",
+    std::make_shared<MapType>(
+        SchemaField::MakeRequired(129, std::string(MapType::kKeyName),
+                                  std::make_shared<IntType>()),
+        SchemaField::MakeRequired(130, std::string(MapType::kValueName),
+                                  std::make_shared<BinaryType>())));
+const SchemaField DataFile::KEY_METADATA =
+    SchemaField::MakeOptional(131, "key_metadata", std::make_shared<BinaryType>());
+const SchemaField DataFile::SPLIT_OFFSETS = SchemaField::MakeOptional(
+    132, "split_offsets",
+    std::make_shared<ListType>(SchemaField::MakeRequired(
+        133, std::string(ListType::kElementName), std::make_shared<LongType>())));
+const SchemaField DataFile::EQUALITY_IDS = SchemaField::MakeOptional(
+    135, "equality_ids",
+    std::make_shared<ListType>(SchemaField::MakeRequired(
+        136, std::string(ListType::kElementName), std::make_shared<IntType>())));
+const SchemaField DataFile::SORT_ORDER_ID =
+    SchemaField::MakeOptional(140, "sort_order_id", std::make_shared<IntType>());
+const SchemaField DataFile::FIRST_ROW_ID =
+    SchemaField::MakeOptional(142, "first_row_id", std::make_shared<LongType>());
+const SchemaField DataFile::REFERENCED_DATA_FILE = SchemaField::MakeOptional(
+    143, "referenced_data_file", std::make_shared<StringType>());
+const SchemaField DataFile::CONTENT_OFFSET =
+    SchemaField::MakeOptional(144, "content_offset", std::make_shared<LongType>());
+const SchemaField DataFile::CONTENT_SIZE =
+    SchemaField::MakeOptional(145, "content_size_in_bytes", std::make_shared<LongType>());
+
+StructType DataFile::GetType(StructType partition_type) {
+  std::vector<SchemaField> fields;
+
+  fields.push_back(CONTENT);
+  fields.push_back(FILE_PATH);
+  fields.push_back(FILE_FORMAT);
+  fields.push_back(SchemaField::MakeRequired(
+      102, "partition", std::make_shared<StructType>(partition_type)));
+  fields.push_back(RECORD_COUNT);
+  fields.push_back(FILE_SIZE);
+  fields.push_back(COLUMN_SIZES);
+  fields.push_back(VALUE_COUNTS);
+  fields.push_back(NULL_VALUE_COUNTS);
+  fields.push_back(NAN_VALUE_COUNTS);
+  fields.push_back(LOWER_BOUNDS);
+  fields.push_back(UPPER_BOUNDS);
+  fields.push_back(KEY_METADATA);
+  fields.push_back(SPLIT_OFFSETS);
+  fields.push_back(EQUALITY_IDS);
+  fields.push_back(SORT_ORDER_ID);
+  fields.push_back(FIRST_ROW_ID);
+  fields.push_back(REFERENCED_DATA_FILE);
+  fields.push_back(CONTENT_OFFSET);
+  fields.push_back(CONTENT_SIZE);
+
+  return StructType(std::move(fields));
+}
+
+const SchemaField ManifestEntry::STATUS =
+    SchemaField::MakeRequired(0, "status", std::make_shared<IntType>());
+const SchemaField ManifestEntry::SNAPSHOT_ID =
+    SchemaField::MakeOptional(1, "snapshot_id", std::make_shared<LongType>());
+const SchemaField ManifestEntry::SEQUENCE_NUMBER =
+    SchemaField::MakeOptional(3, "sequence_number", std::make_shared<LongType>());
+const SchemaField ManifestEntry::FILE_SEQUENCE_NUMBER =
+    SchemaField::MakeOptional(4, "file_sequence_number", std::make_shared<LongType>());
+
+Schema ManifestEntry::GetSchema(StructType partition_type) {
+  return GetSchemaFromDataFileType(DataFile::GetType(partition_type));
+}
+
+Schema ManifestEntry::GetSchemaFromDataFileType(StructType datafile_type) {
+  std::vector<SchemaField> fields;
+
+  fields.push_back(STATUS);
+  fields.push_back(SNAPSHOT_ID);
+  fields.push_back(SEQUENCE_NUMBER);
+  fields.push_back(FILE_SEQUENCE_NUMBER);
+
+  // Add the data file schema
+  auto data_file_type_field = SchemaField::MakeRequired(
+      2, "data_file", std::make_shared<StructType>(DataFile::GetType(datafile_type)));
+  fields.push_back(data_file_type_field);
+
+  return {std::move(fields), /*schema_id=*/std::nullopt};
+}
+
+}  // namespace iceberg
--- a/src/iceberg/manifest_entry.cc
+++ b/src/iceberg/manifest_entry.cc
@@ -20,10 +20,8 @@
 #include "iceberg/manifest_entry.h"
 
 #include <memory>
-#include <optional>
 #include <vector>
 
-#include "iceberg/schema.h"
 #include "iceberg/schema_field.h"
 #include "iceberg/type.h"
 
@@ -138,11 +136,11 @@ const SchemaField ManifestEntry::SEQUENCE_NUMBER =
 const SchemaField ManifestEntry::FILE_SEQUENCE_NUMBER =
     SchemaField::MakeOptional(4, "file_sequence_number", std::make_shared<LongType>());
 
-Schema ManifestEntry::GetSchema(StructType partition_type) {
+StructType ManifestEntry::GetSchema(StructType partition_type) {
   return GetSchemaFromDataFileType(DataFile::GetType(partition_type));
 }
 
-Schema ManifestEntry::GetSchemaFromDataFileType(StructType datafile_type) {
+StructType ManifestEntry::GetSchemaFromDataFileType(StructType datafile_type) {
   std::vector<SchemaField> fields;
 
   fields.push_back(STATUS);
@@ -155,7 +153,7 @@ Schema ManifestEntry::GetSchemaFromDataFileType(StructType datafile_type) {
       2, "data_file", std::make_shared<StructType>(DataFile::GetType(datafile_type)));
   fields.push_back(data_file_type_field);
 
-  return {std::move(fields), /*schema_id=*/std::nullopt};
+  return StructType(std::move(fields));
 }
 
 }  // namespace iceberg
--- a/src/iceberg/manifest_entry.cc
+++ b/src/iceberg/manifest_entry.cc
@@ -26,134 +26,41 @@
 #include "iceberg/type.h"
 
 namespace iceberg {
-const SchemaField DataFile::CONTENT =
-    SchemaField::MakeRequired(134, "content", std::make_shared<IntType>());
-const SchemaField DataFile::FILE_PATH =
-    SchemaField::MakeRequired(100, "file_path", std::make_shared<StringType>());
-const SchemaField DataFile::FILE_FORMAT =
-    SchemaField::MakeRequired(101, "file_format", std::make_shared<IntType>());
-const SchemaField DataFile::RECORD_COUNT =
-    SchemaField::MakeRequired(103, "record_count", std::make_shared<LongType>());
-const SchemaField DataFile::FILE_SIZE =
-    SchemaField::MakeRequired(104, "file_size_in_bytes", std::make_shared<LongType>());
-const SchemaField DataFile::COLUMN_SIZES = SchemaField::MakeOptional(
-    108, "column_sizes",
-    std::make_shared<MapType>(
-        SchemaField::MakeRequired(117, std::string(MapType::kKeyName),
-                                  std::make_shared<IntType>()),
-        SchemaField::MakeRequired(118, std::string(MapType::kValueName),
-                                  std::make_shared<LongType>())));
-const SchemaField DataFile::VALUE_COUNTS = SchemaField::MakeOptional(
-    109, "value_counts",
-    std::make_shared<MapType>(
-        SchemaField::MakeRequired(119, std::string(MapType::kKeyName),
-                                  std::make_shared<IntType>()),
-        SchemaField::MakeRequired(120, std::string(MapType::kValueName),
-                                  std::make_shared<LongType>())));
-const SchemaField DataFile::NULL_VALUE_COUNTS = SchemaField::MakeOptional(
-    110, "null_value_counts",
-    std::make_shared<MapType>(
-        SchemaField::MakeRequired(121, std::string(MapType::kKeyName),
-                                  std::make_shared<IntType>()),
-        SchemaField::MakeRequired(122, std::string(MapType::kValueName),
-                                  std::make_shared<LongType>())));
-const SchemaField DataFile::NAN_VALUE_COUNTS = SchemaField::MakeOptional(
-    137, "nan_value_counts",
-    std::make_shared<MapType>(
-        SchemaField::MakeRequired(138, std::string(MapType::kKeyName),
-                                  std::make_shared<IntType>()),
-        SchemaField::MakeRequired(139, std::string(MapType::kValueName),
-                                  std::make_shared<LongType>())));
-const SchemaField DataFile::LOWER_BOUNDS = SchemaField::MakeOptional(
-    125, "lower_bounds",
-    std::make_shared<MapType>(
-        SchemaField::MakeRequired(126, std::string(MapType::kKeyName),
-                                  std::make_shared<IntType>()),
-        SchemaField::MakeRequired(127, std::string(MapType::kValueName),
-                                  std::make_shared<BinaryType>())));
-const SchemaField DataFile::UPPER_BOUNDS = SchemaField::MakeOptional(
-    128, "upper_bounds",
-    std::make_shared<MapType>(
-        SchemaField::MakeRequired(129, std::string(MapType::kKeyName),
-                                  std::make_shared<IntType>()),
-        SchemaField::MakeRequired(130, std::string(MapType::kValueName),
-                                  std::make_shared<BinaryType>())));
-const SchemaField DataFile::KEY_METADATA =
-    SchemaField::MakeOptional(131, "key_metadata", std::make_shared<BinaryType>());
-const SchemaField DataFile::SPLIT_OFFSETS = SchemaField::MakeOptional(
-    132, "split_offsets",
-    std::make_shared<ListType>(SchemaField::MakeRequired(
-        133, std::string(ListType::kElementName), std::make_shared<LongType>())));
-const SchemaField DataFile::EQUALITY_IDS = SchemaField::MakeOptional(
-    135, "equality_ids",
-    std::make_shared<ListType>(SchemaField::MakeRequired(
-        136, std::string(ListType::kElementName), std::make_shared<IntType>())));
-const SchemaField DataFile::SORT_ORDER_ID =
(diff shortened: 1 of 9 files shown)


src/iceberg/manifest_reader.h
```diff
@@ -26,13 +26,15 @@
 #include <span>
 
 #include "iceberg/file_reader.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/type_fwd.h"
```
wgtmac, line 30: Either remove this inclusion or remove forward declaration below.

src/iceberg/manifest_list.cc
```diff
@@ -0,0 +1,103 @@
...
+    507, "partitions",
+    std::make_shared<ListType>(SchemaField::MakeRequired(
+        508, std::string(ListType::kElementName),
+        std::make_shared<StructType>(FieldSummary::GetType()))));
+const SchemaField ManifestFile::KEY_METADATA =
+    SchemaField::MakeOptional(519, "key_metadata", std::make_shared<BinaryType>());
+const SchemaField ManifestFile::FIRST_ROW_ID =
+    SchemaField::MakeOptional(520, "first_row_id", std::make_shared<LongType>());
+
+Schema ManifestFile::schema() {
```
wgtmac:
```suggestion
Schema ManifestFile::Schema() {
```

It seems to be non-trivial.

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,203 @@
...
+  static const SchemaField EXISTING_FILES_COUNT;
+  static const SchemaField DELETED_FILES_COUNT;
+  static const SchemaField ADDED_ROWS_COUNT;
+  static const SchemaField EXISTING_ROWS_COUNT;
+  static const SchemaField DELETED_ROWS_COUNT;
+  static const SchemaField PARTITIONS;
+  static const SchemaField KEY_METADATA;
+  static const SchemaField FIRST_ROW_ID;
+
+  static Schema schema();
```
wgtmac: Why returning `Schema` but not `StructType`?

wgtmac (commented):
Sorry that I haven't finished the review yet. It is a busy week :(

src/iceberg/manifest_reader.h
zhjwpku: I'm not sure I get this but I don't see forward declaration in this file.

src/iceberg/manifest_list.cc
zhjwpku: Ok, I'll change it in a later commit along with other fixes.

src/iceberg/manifest_list.h
zhjwpku: Seems we don't need schema_id here, I agree return StructType is better, will change.

src/iceberg/manifest_reader.h
```diff
@@ -26,13 +26,15 @@
...
 
 namespace iceberg {
 
 /// \brief Read manifest entries from a manifest file.
 class ICEBERG_EXPORT ManifestReader {
  public:
-  virtual Result<std::span<std::unique_ptr<class ManifestEntry>>> Entries() const = 0;
+  virtual Result<std::span<std::unique_ptr<struct ManifestEntry>>> Entries() const = 0;
```
wgtmac:
```suggestion
  virtual Result<std::span<std::unique_ptr<ManifestEntry>>> Entries() const = 0;
```

src/iceberg/manifest_reader.h
```diff
@@ -41,7 +43,7 @@ class ICEBERG_EXPORT ManifestReader {
 /// \brief Read manifest files from a manifest list file.
 class ICEBERG_EXPORT ManifestListReader {
  public:
-  virtual Result<std::span<std::unique_ptr<class ManifestFile>>> Files() const = 0;
+  virtual Result<std::span<std::unique_ptr<struct ManifestFile>>> Files() const = 0;
```
wgtmac:
```suggestion
  virtual Result<std::span<std::unique_ptr<ManifestFile>>> Files() const = 0;
```

src/iceberg/manifest_reader.h
wgtmac: I mean you may remove the two forward declarations below as the suggested changes.

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,202 @@
...
+
+/// \brief Get the relative manifest content type from name
+ICEBERG_EXPORT constexpr Result<ManifestContent> ManifestContentFromString(
+    std::string_view str) noexcept {
+  if (str == "data") return ManifestContent::kData;
+  if (str == "deletes") return ManifestContent::kDeletes;
+  return InvalidArgument("Invalid manifest content type: {}", str);
+}
+
+struct ICEBERG_EXPORT FieldSummary {
```
wgtmac:
```suggestion
struct ICEBERG_EXPORT PartitionFieldSummary {
```

Is it better to explicitly add `Partition` prefix just in case there will be another kind of field summary in the future?

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,202 @@
...
+  /// Upper bound for the non-null, non-NaN values in the partition field, or null if all
+  /// values are null or NaN
+  std::optional<std::vector<uint8_t>> upper_bound;
+
+  static const SchemaField CONTAINS_NULL;
+  static const SchemaField CONTAINS_NAN;
+  static const SchemaField LOWER_BOUND;
+  static const SchemaField UPPER_BOUND;
+
+  static StructType GetType();
```
wgtmac:
```suggestion
  static const StructType& Type();
```

Should we consider singleton for this?

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,202 @@
...
+  /// Field id: 510
+  /// Lower bound for the non-null, non-NaN values in the partition field, or null if all
+  /// values are null or NaN
+  std::optional<std::vector<uint8_t>> lower_bound;
+  /// Field id: 511
+  /// Upper bound for the non-null, non-NaN values in the partition field, or null if all
+  /// values are null or NaN
+  std::optional<std::vector<uint8_t>> upper_bound;
+
+  static const SchemaField CONTAINS_NULL;
```
wgtmac:
```suggestion
  static const SchemaField kContainsNull;
```

We should still follow the `k`-prefixed camel case, right? @lidavidm

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,202 @@
...
+  static const SchemaField CONTAINS_NAN;
+  static const SchemaField LOWER_BOUND;
+  static const SchemaField UPPER_BOUND;
+
+  static StructType GetType();
+};
+
+/// \brief Entry in a manifest list.
+struct ICEBERG_EXPORT ManifestFile {
```
wgtmac, line 100: This definition is a superset of v1/v2/v3. Have you considered adding an adapter for each version for manifest list writer/reader?

src/iceberg/manifest_list.h
wgtmac: Is it better to use inline static declaration so we have better readability?

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,202 @@
...
+  /// Field id: 519
+  /// Implementation-specific key metadata for encryption
+  std::vector<uint8_t> key_metadata;
+  /// Field id: 520
+  /// The starting _row_id to assign to rows added by ADDED data files
+  int64_t first_row_id;
+
+  /// \brief Checks if this manifest file contains entries with ADDED status.
+  [[nodiscard]] bool has_added_files() const {
+    return added_files_count.has_value() && *added_files_count > 0;
```
wgtmac:
```suggestion
    return added_files_count.valur_or(-1) > 0;
```

Perhaps this is simpler? BTW, we don't need `[[nodiscard]]` for this kind of functions.

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,202 @@
...
+  static const SchemaField EXISTING_FILES_COUNT;
+  static const SchemaField DELETED_FILES_COUNT;
+  static const SchemaField ADDED_ROWS_COUNT;
+  static const SchemaField EXISTING_ROWS_COUNT;
+  static const SchemaField DELETED_ROWS_COUNT;
+  static const SchemaField PARTITIONS;
+  static const SchemaField KEY_METADATA;
+  static const SchemaField FIRST_ROW_ID;
+
+  static StructType Schema();
```
wgtmac:
```suggestion
  static const StructType& Type();
```

Same as the above comment

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,202 @@
...
+/// A new manifest list is written for each attempt to commit a snapshot because the list
+/// of manifests always changes to produce a new snapshot. When a manifest list is
+/// written, the (optimistic) sequence number of the snapshot is written for all new
+/// manifest files tracked by the list.
+///
+/// A manifest list includes summary metadata that can be used to avoid scanning all of
+/// the manifests in a snapshot when planning a table scan. This includes the number of
+/// added, existing, and deleted files, and a summary of values for each field of the
+/// partition spec used to write the manifest.
+struct ManifestList {
```
wgtmac:
```suggestion
struct ICEBERG_EXPORT ManifestList {
```

src/iceberg/manifest_list.cc
```diff
@@ -0,0 +1,105 @@
...
+#include "iceberg/manifest_list.h"
+
+#include <vector>
+
+#include "iceberg/schema_field.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+const SchemaField FieldSummary::CONTAINS_NULL =
```
wgtmac: It seems that we need to add the `doc` field to `SchemaField` to align with the Java impl.

src/iceberg/manifest_list.cc
```diff
@@ -0,0 +1,105 @@
...
+StructType FieldSummary::GetType() {
+  return StructType({
+      CONTAINS_NULL,
+      CONTAINS_NAN,
+      LOWER_BOUND,
+      UPPER_BOUND,
+  });
+}
+
+const SchemaField ManifestFile::MANIFEST_PATH =
```
wgtmac: I'm not sure if using macros is a good idea in modern C++ to replace these repeated lengthy lines. 

@lidavidm WDYT?

src/iceberg/manifest_list.cc
```diff
@@ -0,0 +1,105 @@
...
+  fields.push_back(EXISTING_FILES_COUNT);
+  fields.push_back(DELETED_FILES_COUNT);
+  fields.push_back(ADDED_ROWS_COUNT);
+  fields.push_back(EXISTING_ROWS_COUNT);
+  fields.push_back(DELETED_ROWS_COUNT);
+  fields.push_back(PARTITIONS);
+  fields.push_back(KEY_METADATA);
+  fields.push_back(FIRST_ROW_ID);
+
+  return StructType(std::move(fields));
```
wgtmac: Is it better to use something like

```cpp
  return StructType(std::vector<SchemaField>{
    MANIFEST_PATH,
    MANIFEST_LENGTH,
    PARTITION_SPEC_ID,
    ...
  });
```

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+  /// Full URI for the file with FS scheme
+  std::string file_path;
+  /// Field id: 101
+  /// File format type, avro, orc, parquet, or puffin
+  FileFormatType file_format;
+  /// Field id: 102
+  /// Partition data tuple, schema based on the partition spec output using partition
+  /// field ids for the struct field ids
+  /// TODO(zhjwpku): use StructLike to represent partition data tuple
+  std::map<std::string, std::any> partition;
```
wgtmac: Why `std::map` here?

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+  /// Field id: 104
+  /// Total file size in bytes
+  int64_t file_size_in_bytes = 0;
+  /// Field id: 108
+  /// Key field id: 117
+  /// Value field id: 118
+  /// Map from column id to the total size on disk of all regions that store the column.
+  /// Does not include bytes necessary to read other columns, like footers. Leave null for
+  /// row-oriented formats (Avro)
+  std::unordered_map<int32_t, int64_t> column_sizes;
```
wgtmac: Perhaps it is better to use `std::map`? Sometimes we need the order of field-id. Same for below.

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+  /// Each value must be greater than or equal to all non-null, non-Nan values in the
+  /// column for the file.
+  ///
+  /// Reference:
+  /// - [Binary single-value
+  /// serialization](https://iceberg.apache.org/spec/#binary-single-value-serialization)
+  std::unordered_map<int32_t, std::vector<uint8_t>> upper_bounds;
+  /// Field id: 131
+  /// Implementation-specific key metadata for encryption
+  std::optional<std::vector<uint8_t>> key_metadata;
```
wgtmac: Is `std::vector<uint8_t>` sufficient?

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+  ///
+  /// Reference:
+  /// - [Binary single-value
+  /// serialization](https://iceberg.apache.org/spec/#binary-single-value-serialization)
+  std::unordered_map<int32_t, std::vector<uint8_t>> lower_bounds;
+  /// Field id: 128
+  /// Key field id: 129
+  /// Value field id: 130
+  /// Map from column id to upper bound in the column serialized as binary.
+  /// Each value must be greater than or equal to all non-null, non-Nan values in the
```
wgtmac:
```suggestion
  /// Each value must be greater than or equal to all non-null, non-NaN values in the
```

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+  /// column for the file.
+  ///
+  /// Reference:
+  /// - [Binary single-value
+  /// serialization](https://iceberg.apache.org/spec/#binary-single-value-serialization)
```
wgtmac:
```suggestion
```

src/iceberg/manifest_entry.h
wgtmac: Perhaps we can simply say `Similar to lower_bounds above but for upper bounds`

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+  static const SchemaField KEY_METADATA;
+  static const SchemaField SPLIT_OFFSETS;
+  static const SchemaField EQUALITY_IDS;
+  static const SchemaField SORT_ORDER_ID;
+  static const SchemaField FIRST_ROW_ID;
+  static const SchemaField REFERENCED_DATA_FILE;
+  static const SchemaField CONTENT_OFFSET;
+  static const SchemaField CONTENT_SIZE;
+
+  static StructType GetType(StructType partition_type);
```
wgtmac:
```suggestion
  static StructType Type(StructType partition_type);
```

src/iceberg/manifest_entry.cc
```diff
@@ -0,0 +1,159 @@
...
+    SchemaField::MakeOptional(145, "content_size_in_bytes", std::make_shared<LongType>());
+
+StructType DataFile::GetType(StructType partition_type) {
+  std::vector<SchemaField> fields;
+
+  fields.push_back(CONTENT);
+  fields.push_back(FILE_PATH);
+  fields.push_back(FILE_FORMAT);
+  fields.push_back(SchemaField::MakeRequired(
+      102, "partition", std::make_shared<StructType>(partition_type)));
```
wgtmac: Why not directly accept a `std::shared_ptr<StructType>` as the input parameter?

src/iceberg/manifest_entry.h
wgtmac: What's the trade-off of return value between `StructType` and `std::unique_ptr<StructType>`?

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+  /// Field id: 3
+  /// Data sequence number of the file. Inherited when null and status is 1 (added).
+  std::optional<int64_t> sequence_number;
+  /// Field id: 4
+  /// File sequence number indicating when the file was added. Inherited when null and
+  /// status is 1 (added).
+  std::optional<int64_t> file_sequence_number;
+  /// Field id: 2
+  /// File path, partition tuple, metrics, ...
+  DataFile data_file;
```
wgtmac, line 308: Should we use `std::unique_ptr` to avoid copy or `std::shared_ptr` to make the copy trivial?

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,256 @@
...
+
+  static const SchemaField STATUS;
+  static const SchemaField SNAPSHOT_ID;
+  static const SchemaField SEQUENCE_NUMBER;
+  static const SchemaField FILE_SEQUENCE_NUMBER;
+
+  static StructType GetSchema(StructType partition_type);
```
wgtmac: ditto for the return type and input type

src/iceberg/manifest_reader.h
zhjwpku: I didn't realize those were forward declarations 🤪

zhjwpku: I didn't realize those were forward declarations 🤪

src/iceberg/manifest_list.h
zhjwpku: I see java impl use PartitionFieldSummary, will change as suggested.

src/iceberg/manifest_list.cc
```diff
@@ -0,0 +1,105 @@
...
+      CONTAINS_NULL,
+      CONTAINS_NAN,
+      LOWER_BOUND,
+      UPPER_BOUND,
+  });
+}
+
+const SchemaField ManifestFile::MANIFEST_PATH =
+    SchemaField::MakeRequired(500, "manifest_path", std::make_shared<StringType>());
+const SchemaField ManifestFile::MANIFEST_LENGTH =
```
gty404: How about using this template?

template <typename T>
inline SchemaField MakeRequiredField(int id, std::string name) {
  return SchemaField::MakeRequired(id, std::move(name), std::make_shared<T>());
}

template <typename T>
inline SchemaField MakeOptionalField(int id, std::string name) {
  return SchemaField::MakeOptional(id, std::move(name), std::make_shared<T>());
}

src/iceberg/manifest_entry.h
zhjwpku: I was thinking that map is structure much similar to StructLike, I will change to std::any for now.

src/iceberg/manifest_entry.cc
zhjwpku: agree, will do.

src/iceberg/manifest_entry.h
zhjwpku: I think the `greater than or equal` provides clearer semantics and helps users avoid confusion with C++'s upper_bound, so I'd prefer to keep it as is for now.

src/iceberg/manifest_entry.h
zhjwpku: removed.

src/iceberg/manifest_entry.h
zhjwpku: `key_metadata` is simply a binary blob used for encryption purposes? If so it seems that std::vector<uint8_t> is sufficient; I checked the Rust implementation, they did the same.

src/iceberg/manifest_entry.h
zhjwpku: I will use shared_ptr for both input and return types.

src/iceberg/manifest_entry.h
zhjwpku: The `DataFile` is part of the in-memory representation of `ManifestEntry`, so IMHO there's no need to wrap it in a smart pointer, embedding it directly in `ManifestEntry` should be more memory-efficient.

src/iceberg/manifest_entry.h
zhjwpku: I've converted these to shared_ptr and updated the function names, I'm not sure if the new names are the best fit, feel free to suggest better.

src/iceberg/manifest_list.cc
zhjwpku: done.

src/iceberg/manifest_list.cc
zhjwpku: done.

src/iceberg/manifest_list.h
zhjwpku: done suggested, both `k-prefix` and `inline static`

src/iceberg/manifest_list.h
zhjwpku: done with singleton.

src/iceberg/manifest_list.h
zhjwpku: I haven't considered that, I will do these in later PRs.

src/iceberg/manifest_list.h
zhjwpku: done suggested.

src/iceberg/manifest_list.h
zhjwpku: done.

src/iceberg/manifest_list.h
zhjwpku: good catch, fixed.

src/iceberg/manifest_reader.h
zhjwpku: done.

src/iceberg/manifest_reader.h
zhjwpku: done.

src/iceberg/manifest_entry.h
zhjwpku: I was following the Rust implementation, which uses a HashMap. However, if we need to preserve the order of field IDs, let's switch to using map instead.

src/iceberg/manifest_list.cc
zhjwpku: Ah, good idea, I will add these templates to schema_field.h

src/iceberg/manifest_entry.h
```diff
@@ -0,0 +1,286 @@
...
+  /// blob for direct access to a deletion vector. For deletion vectors, these values are
+  /// required and must exactly match the offset and length stored in the Puffin footer
+  /// for the deletion vector blob.
+  std::optional<int64_t> content_offset;
+  /// Field id: 145
+  /// The length of a referenced content stored in the file; required if content_offset is
+  /// present
+  std::optional<int64_t> content_size_in_bytes;
+
+  inline static const SchemaField kContent = MakeRequiredField<IntType>(134, "content");
```
wgtmac: Should we support adding `doc` to `SchemaField`? In the Java impl, the doc strings are serialized with the schema in the Avro file (manifest list file and manifest file). We can add it later though.

Am I right? @Fokko

src/iceberg/manifest_entry.h
wgtmac: Yes, I believe an empty `key_metadata` is enough.

src/iceberg/schema_field.h
```diff
@@ -86,4 +86,16 @@ class ICEBERG_EXPORT SchemaField : public iceberg::util::Formattable {
   bool optional_;
 };
 
+template <typename T, typename... Args>
+inline SchemaField MakeRequiredField(int id, std::string name, Args&&... args) {
```
wgtmac: Perhaps we can add an extra `doc` parameter but is not actually used before supporting `doc` to `SchemaField`. In this way, we can have full definition of those fields in this PR.

src/iceberg/manifest_list.cc
```diff
@@ -0,0 +1,47 @@
...
+#include "iceberg/manifest_list.h"
+
+#include <vector>
+
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+const StructType& PartitionFieldSummary::Type() {
+  static const std::shared_ptr<StructType> instance{new StructType({
```
wgtmac: Can we use `std::make_unique`? As this is a const variable, we need to use `kType` or something similar.

src/iceberg/manifest_list.cc
```diff
@@ -0,0 +1,47 @@
...
+      PartitionFieldSummary::kConsTainsNull,
+      PartitionFieldSummary::kContainsNaN,
+      PartitionFieldSummary::kLowerBound,
+      PartitionFieldSummary::kUpperBound,
+  })};
+  return *instance;
+}
+
+const StructType& ManifestFile::Type() {
+  static const std::shared_ptr<StructType> instance{new StructType(
```
wgtmac: ditto

src/iceberg/manifest_entry.cc
```diff
@@ -0,0 +1,66 @@
...
+}
+
+std::shared_ptr<StructType> ManifestEntry::TypeFromPartitionType(
+    std::shared_ptr<StructType> partition_type) {
+  return TypeFromDataFileType(DataFile::Type(std::move(partition_type)));
+}
+
+std::shared_ptr<StructType> ManifestEntry::TypeFromDataFileType(
+    std::shared_ptr<StructType> datafile_type) {
+  return std::make_shared<StructType>(std::vector<SchemaField>{
```
wgtmac, line 61: Not related to this PR: is it better to use `std::unique_ptr<Type>` in `SchemaField`? Or is it important to make `SchemaField` copyable? @lidavidm

wgtmac (commented):
Thanks for the update! It generally looks good to me.

src/iceberg/manifest_entry.cc
lidavidm: Copyability is probably more convenient. Either way, I'm not sure schemas are really a bottleneck ever so the distinction probably won't matter

gty404 (approved):

src/iceberg/manifest_entry.cc
wgtmac: Make sense. I used unique_ptr for `MappedField` in the name mapping PR. May switch to shared_ptr if I found it inconvenient.

src/iceberg/schema_field.h
wgtmac: Created https://github.com/apache/iceberg-cpp/pull/96

src/iceberg/manifest_entry.h
zhjwpku: Ok, I'll remove the optional part.

src/iceberg/schema_field.h
wgtmac:
```suggestion
inline SchemaField MakeRequiredField(int32_t id, std::string name, Args&&... args) {
```

src/iceberg/schema_field.h
```diff
@@ -86,4 +86,16 @@ class ICEBERG_EXPORT SchemaField : public iceberg::util::Formattable {
...
+  return SchemaField::MakeRequired(id, std::move(name),
+                                   std::make_shared<T>(std::forward<Args>(args)...));
+}
+
+template <typename T, typename... Args>
+inline SchemaField MakeOptionalField(int id, std::string name, Args&&... args) {
```
wgtmac:
```suggestion
inline SchemaField MakeOptionalField(int32_t id, std::string name, Args&&... args) {
```

src/iceberg/schema_field.h
zhjwpku: I've deleted these helpers because it doesn't fit well with the new added doc field.

src/iceberg/manifest_list.cc
zhjwpku: I just removed the share_ptr, which seem unnecessary.

wgtmac (approved):
+1

Thanks!

lidavidm:
No comments from my side

src/iceberg/manifest_list.h
Fokko: Our perspective from PyIceberg, we project everything into V2 (and V3 in the future). Many of the fields that are missing for V1 can be set (for example, content can always be set to `data`, sequence numbers are set to 0, and more). Therefore, we did not split this out per version to reduce complexity.

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,235 @@
...
+  std::vector<PartitionFieldSummary> partitions;
+  /// Field id: 519
+  /// Implementation-specific key metadata for encryption
+  std::vector<uint8_t> key_metadata;
+  /// Field id: 520
+  /// The starting _row_id to assign to rows added by ADDED data files
+  int64_t first_row_id;
+
+  /// \brief Checks if this manifest file contains entries with ADDED status.
+  bool has_added_files() const { return added_files_count.value_or(-1) > 0; }
```
Fokko: I think this is incorrect. When we don't know, we assume that there are added files.

```suggestion
  bool has_added_files() const { return added_files_count.value_or(1) > 0; }
```

https://github.com/apache/iceberg/blob/4ca3f10832907540eee798236fa401eb07c7e7df/api/src/main/java/org/apache/iceberg/ManifestFile.java#L150

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,235 @@
...
+
+  /// \brief Checks if this manifest file contains entries with EXISTING status.
+  bool has_existing_files() const { return existing_files_count.value_or(-1) > 0; }
```
Fokko: I think this is incorrect. When we don't know, we assume that there are existing files.

```suggestion
  bool has_existing_files() const { return existing_files_count.value_or(1) > 0; }
```

https://github.com/apache/iceberg/blob/4ca3f10832907540eee798236fa401eb07c7e7df/api/src/main/java/org/apache/iceberg/ManifestFile.java#L165

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,235 @@
...
+
+  /// \brief Checks if this manifest file contains entries with DELETED status
+  bool has_deleted_files() const { return deleted_files_count.value_or(-1) > 0; }
```
Fokko: I think this is incorrect. When we don't know, we assume that there are deleted files.

```suggestion
  bool has_deleted_files() const { return deleted_files_count.value_or(-1) > 0; }
```
https://github.com/apache/iceberg/blob/4ca3f10832907540eee798236fa401eb07c7e7df/api/src/main/java/org/apache/iceberg/ManifestFile.java#L180

Fokko (approved):
Three minor things, apart from that, it looks good to me 👍

src/iceberg/manifest_list.h
```diff
@@ -0,0 +1,235 @@
...
+  int64_t first_row_id;
+
+  /// \brief Checks if this manifest file contains entries with ADDED status.
+  bool has_added_files() const { return added_files_count.value_or(1) > 0; }
+
+  /// \brief Checks if this manifest file contains entries with EXISTING status.
+  bool has_existing_files() const { return existing_files_count.value_or(1) > 0; }
+
+  /// \brief Checks if this manifest file contains entries with DELETED status
+  bool has_deleted_files() const { return deleted_files_count.value_or(-1) > 0; }
```
zhjwpku:
```suggestion
  bool has_deleted_files() const { return deleted_files_count.value_or(1) > 0; }
```

src/iceberg/manifest_list.h
zhjwpku: Thanks for sharing the PyIceberg approach, will explore this direction further.

Fokko:
Thanks @zhjwpku for adding this, and thanks @wgtmac and @gty404 for the review 🙌

PR #92, opened 2025-05-06
feat: add base config implementation

yingcai-cy:
introduce a new config class to manage configurable options following design ideas from Velox

### DIFF
--- a/src/iceberg/util/config.h
+++ b/src/iceberg/util/config.h
@@ -23,7 +23,7 @@
 #include <string>
 #include <unordered_map>
 
-#include <iceberg/exception.h>
+#include "iceberg/exception.h"
 
 namespace iceberg {
 namespace internal {
--- a/src/iceberg/util/config.h
+++ b/src/iceberg/util/config.h
@@ -1,0 +1,114 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#pragma once
+
+#include <format>
+#include <functional>
+#include <map>
+#include <string>
+
+namespace iceberg {
+
+template <class ConcreteConfig>
+class ConfigBase {
+ public:
+  template <typename T>
+  class Entry {
+    Entry(
+        std::string key, const T& val,
+        std::function<std::string(const T&)> toStr =
+            [](const T& val) {
+              if constexpr ((std::is_signed_v<T> && std::is_integral_v<T>) ||
+                            std::is_floating_point_v<T>) {
+                return std::to_string(val);
+              } else if constexpr (std::is_same_v<T, bool>) {
+                return val ? "true" : "false";
+              } else if constexpr (std::is_same_v<T, std::string> ||
+                                   std::is_same_v<T, std::string_view>) {
+                return val;
+              } else {
+                throw std::runtime_error(
+                    std::format("Explicit toStr() is required for {}", typeid(T).name()));
+              }
+            },
+        std::function<T(const std::string&)> toT = [](const std::string& val) -> T {
+          if constexpr (std::is_same_v<T, std::string>) {
+            return val;
+          } else if constexpr (std::is_same_v<T, bool>) {
+            return val == "true";
+          } else if constexpr (std::is_signed_v<T> && std::is_integral_v<T>) {
+            return static_cast<T>(std::stoll(val));
+          } else if constexpr (std::is_floating_point_v<T>) {
+            return static_cast<T>(std::stod(val));
+          } else {
+            throw std::runtime_error(
+                std::format("Explicit toT() is required for {}", typeid(T).name()));
+          }
+        })
+        : key_{std::move(key)}, default_{val}, toStr_{toStr}, toT_{toT} {}
+
+    const std::string key_;
+    const T default_;
+    const std::function<std::string(const T&)> toStr_;
+    const std::function<T(const std::string&)> toT_;
+
+    friend ConfigBase;
+    friend ConcreteConfig;
+
+   public:
+    const std::string& key() const { return key_; }
+
+    T value() const { return default_; }
+  };
+
+  template <typename T>
+  ConfigBase& set(const Entry<T>& entry, const T& val) {
+    configs_[entry.key_] = entry.toStr_(val);
+    return *this;
+  }
+
+  template <typename T>
+  ConfigBase& unset(const Entry<T>& entry) {
+    configs_.erase(entry.key_);
+    return *this;
+  }
+
+  ConfigBase& reset() {
+    configs_.clear();
+    return *this;
+  }
+
+  template <typename T>
+  T get(const Entry<T>& entry) const {
+    try {
+      auto iter = configs_.find(entry.key_);
+      return iter != configs_.cend() ? entry.toT_(iter->second) : entry.default_;
+    } catch (std::exception& e) {
+      throw std::runtime_error(
+          std::format("Failed to get config {} with error: {}", entry.key_, e.what()));
+    }
+  }
+
+  std::map<std::string, std::string> const& configs() const { return configs_; }
+
+ protected:
+  std::map<std::string, std::string> configs_;
+};
+
+}  // namespace iceberg
--- a/src/iceberg/util/config.h
+++ b/src/iceberg/util/config.h
@@ -20,53 +20,60 @@
 
 #include <format>
 #include <functional>
-#include <map>
 #include <string>
+#include <unordered_map>
 
 namespace iceberg {
+
+// Default conversion functions
+template <typename U>
+std::string defaultToString(const U& val) {
+  if constexpr ((std::is_signed_v<U> && std::is_integral_v<U>) ||
+                std::is_floating_point_v<U>) {
+    return std::to_string(val);
+  } else if constexpr (std::is_same_v<U, bool>) {
+    return val ? "true" : "false";
+  } else if constexpr (std::is_same_v<U, std::string> ||
+                       std::is_same_v<U, std::string_view>) {
+    return val;
+  } else {
+    throw std::runtime_error(
+        std::format("Explicit toStr() is required for {}", typeid(U).name()));
+  }
+}
+
+template <typename U>
+U defaultFromString(const std::string& val) {
+  if constexpr (std::is_same_v<U, std::string>) {
+    return val;
+  } else if constexpr (std::is_same_v<U, bool>) {
+    return val == "true";
+  } else if constexpr (std::is_signed_v<U> && std::is_integral_v<U>) {
+    return static_cast<U>(std::stoll(val));
+  } else if constexpr (std::is_floating_point_v<U>) {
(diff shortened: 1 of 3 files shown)


src/iceberg/util/config.h
```diff
@@ -0,0 +1,114 @@
...
+namespace iceberg {
+
+template <class ConcreteConfig>
+class ConfigBase {
+ public:
+  template <typename T>
+  class Entry {
+    Entry(
+        std::string key, const T& val,
+        std::function<std::string(const T&)> toStr =
```
wgtmac:
```suggestion
        std::function<std::string(const T&)> to_str =
```

src/iceberg/util/config.h
```diff
@@ -0,0 +1,114 @@
...
+    } catch (std::exception& e) {
+      throw std::runtime_error(
+          std::format("Failed to get config {} with error: {}", entry.key_, e.what()));
+    }
+  }
+
+  std::map<std::string, std::string> const& configs() const { return configs_; }
+
+ protected:
+  std::map<std::string, std::string> configs_;
```
wgtmac:
```suggestion
  std::unordered_map<std::string, std::string> configs_;
```

We don't need the keys to be ordered.

src/iceberg/util/config.h
```diff
@@ -0,0 +1,114 @@
...
+    return *this;
+  }
+
+  ConfigBase& reset() {
+    configs_.clear();
+    return *this;
+  }
+
+  template <typename T>
+  T get(const Entry<T>& entry) const {
```
wgtmac:
```suggestion
  Result<T> get(const Entry<T>& entry) const {
```

We need to return error code instead of exception.

src/iceberg/util/config.h
```diff
@@ -0,0 +1,114 @@
...
+                return val ? "true" : "false";
+              } else if constexpr (std::is_same_v<T, std::string> ||
+                                   std::is_same_v<T, std::string_view>) {
+                return val;
+              } else {
+                throw std::runtime_error(
+                    std::format("Explicit toStr() is required for {}", typeid(T).name()));
+              }
+            },
+        std::function<T(const std::string&)> toT = [](const std::string& val) -> T {
```
wgtmac:
```suggestion
        std::function<T(const std::string&)> to_T = [](const std::string& val) -> T {
```

Or `from_str`?

src/iceberg/util/config.h
```diff
@@ -0,0 +1,114 @@
...
+              if constexpr ((std::is_signed_v<T> && std::is_integral_v<T>) ||
+                            std::is_floating_point_v<T>) {
+                return std::to_string(val);
+              } else if constexpr (std::is_same_v<T, bool>) {
+                return val ? "true" : "false";
+              } else if constexpr (std::is_same_v<T, std::string> ||
+                                   std::is_same_v<T, std::string_view>) {
+                return val;
+              } else {
+                throw std::runtime_error(
```
wgtmac: We can't throw here. Please use `Result`.

src/iceberg/util/config.h
```diff
@@ -0,0 +1,114 @@
...
+          } else {
+            throw std::runtime_error(
+                std::format("Explicit toT() is required for {}", typeid(T).name()));
+          }
+        })
+        : key_{std::move(key)}, default_{val}, toStr_{toStr}, toT_{toT} {}
+
+    const std::string key_;
+    const T default_;
+    const std::function<std::string(const T&)> toStr_;
```
wgtmac:
```suggestion
    const std::function<std::string(const T&)> to_str_;
```

src/iceberg/util/config.h
```diff
@@ -0,0 +1,114 @@
...
+    const std::function<T(const std::string&)> toT_;
+
+    friend ConfigBase;
+    friend ConcreteConfig;
+
+   public:
+    const std::string& key() const { return key_; }
+
+    T value() const { return default_; }
```
wgtmac:
```suggestion
    const T& value() const { return default_; }
```

src/iceberg/util/config.h
lidavidm: Also, should we move these default implementations into named functions instead of inlining them as lambdas?

src/iceberg/util/config.h
yingcai-cy: Since ConfigBase::set returns a reference to enable method chaining, we can’t change the toStr/fromStr return type to Result.

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+#include <format>
+#include <functional>
+#include <string>
+#include <unordered_map>
+
+namespace iceberg {
+
+// Default conversion functions
+template <typename U>
+std::string defaultToString(const U& val) {
```
wgtmac:
```suggestion
std::string DefaultToString(const U& val) {
```

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+                       std::is_same_v<U, std::string_view>) {
+    return val;
+  } else {
+    throw std::runtime_error(
+        std::format("Explicit toStr() is required for {}", typeid(U).name()));
+  }
+}
+
+template <typename U>
+U defaultFromString(const std::string& val) {
```
wgtmac:
```suggestion
U DefaultFromString(const std::string& val) {
```

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+  if constexpr (std::is_same_v<U, std::string>) {
+    return val;
+  } else if constexpr (std::is_same_v<U, bool>) {
+    return val == "true";
+  } else if constexpr (std::is_signed_v<U> && std::is_integral_v<U>) {
+    return static_cast<U>(std::stoll(val));
+  } else if constexpr (std::is_floating_point_v<U>) {
+    return static_cast<U>(std::stod(val));
+  } else {
+    throw std::runtime_error(
```
wgtmac: ditto

src/iceberg/util/config.h
wgtmac: BTW, what about wrapping these functions in the `namespace iceberg::internal`?

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+  if constexpr ((std::is_signed_v<U> && std::is_integral_v<U>) ||
+                std::is_floating_point_v<U>) {
+    return std::to_string(val);
+  } else if constexpr (std::is_same_v<U, bool>) {
+    return val ? "true" : "false";
+  } else if constexpr (std::is_same_v<U, std::string> ||
+                       std::is_same_v<U, std::string_view>) {
+    return val;
+  } else {
+    throw std::runtime_error(
```
wgtmac: Include `iceberg/exception.h` and replace it with `IcebergError` exception

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+    return *this;
+  }
+
+  template <typename T>
+  T get(const Entry<T>& entry) const {
+    auto iter = configs_.find(entry.key_);
+    return iter != configs_.cend() ? entry.from_str_(iter->second) : entry.default_;
+  }
+
+  std::unordered_map<std::string, std::string> const& configs() const { return configs_; }
```
wgtmac:
```suggestion
  const std::unordered_map<std::string, std::string>& configs() const { return configs_; }
```

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+
+   public:
+    const std::string& key() const { return key_; }
+
+    const T& value() const { return default_; }
+  };
+
+  template <typename T>
+  ConfigBase& set(const Entry<T>& entry, const T& val) {
+    configs_[entry.key_] = entry.to_str_(val);
```
wgtmac: Use emplace to avoid copy?

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+    friend ConcreteConfig;
+
+   public:
+    const std::string& key() const { return key_; }
+
+    const T& value() const { return default_; }
+  };
+
+  template <typename T>
+  ConfigBase& set(const Entry<T>& entry, const T& val) {
```
wgtmac:
```suggestion
  ConfigBase& Set(const Entry<T>& entry, const T& val) {
```

Usually functions use capitalized initial. The above `Entry` does not need capitalized initial because they have only trivial getters.

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+    configs_[entry.key_] = entry.to_str_(val);
+    return *this;
+  }
+
+  template <typename T>
+  ConfigBase& unset(const Entry<T>& entry) {
```
wgtmac:
```suggestion
  ConfigBase& Unset(const Entry<T>& entry) {
```

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+    configs_.erase(entry.key_);
+    return *this;
+  }
+
+  ConfigBase& reset() {
```
wgtmac:
```suggestion
  ConfigBase& Reset() {
```

src/iceberg/util/config.h
```diff
@@ -0,0 +1,116 @@
...
+    configs_.clear();
+    return *this;
+  }
+
+  template <typename T>
+  T get(const Entry<T>& entry) const {
```
wgtmac:
```suggestion
  T Get(const Entry<T>& entry) const {
```

src/iceberg/util/config.h
wgtmac: Actually I think the original lambda functions look simpler. @lidavidm

wgtmac:
What about renaming the title to `feat: add base config implementation`? We can mention Velox as a comment to the config class.

src/iceberg/util/config.h
lidavidm: Hmm, okay then. I'm curious how doc generators will handle such a long expression inlined as a default value but I'm not opposed to reverting

src/iceberg/util/config.h
```diff
@@ -23,11 +23,13 @@
 #include <string>
 #include <unordered_map>
 
-namespace iceberg {
+#include <iceberg/exception.h>
```
wgtmac:
```suggestion
#include "iceberg/exception.h"
```

wgtmac (approved):

test/config_test.cc
```diff
@@ -0,0 +1,110 @@
...
+#include <string>
+
+#include <gtest/gtest.h>
+#include <iceberg/util/config.h>
+
+namespace iceberg {
+
+enum class TestEnum { VALUE1, VALUE2, VALUE3 };
+
+std::string enumToString(const TestEnum& val) {
```
lidavidm:
```suggestion
std::string EnumToString(const TestEnum& val) {
```

nit

test/config_test.cc
```diff
@@ -0,0 +1,110 @@
...
+    case TestEnum::VALUE2:
+      return "VALUE2";
+    case TestEnum::VALUE3:
+      return "VALUE3";
+    default:
+      throw std::runtime_error("Invalid enum value");
+  }
+}
+
+TestEnum stringToEnum(const std::string& val) {
```
lidavidm:
```suggestion
TestEnum StringToEnum(const std::string& val) {
```

nit

lidavidm (approved):

test/config_test.cc
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+// Define a concrete config class for testing
+class TestConfig : public ConfigBase<TestConfig> {
+ public:
+  template <typename T>
+  using Entry = const ConfigBase<TestConfig>::Entry<T>;
+
+  inline static const Entry<std::string> STRING_CONFIG{"string_config", "default_value"};
```
zhjwpku: nit: maybe we should follow the k-prefix pattern for static const variables.

zhjwpku (approved):

lidavidm (approved):

Fokko (approved):
Looks good @yingcai-cy Thanks for working on this. Thanks @wgtmac, @lidavidm and @zhjwpku for the review 🙌

rashworld-max (commented):
0297da8eb760effb688862c136e856cb3ae1d49c

PR #93, opened 2025-05-08
feat: add name mapping

wgtmac:


src/iceberg/name_mapping.cc
```diff
@@ -0,0 +1,267 @@
...
+#include <sstream>
+
+#include "iceberg/util/formatter_internal.h"
+
+namespace iceberg {
+
+namespace {
+
+// Helper function to join a list of field names with a dot
+std::string JoinByDot(std::span<const std::string> parts) {
```
lidavidm, line 32: I guess you can use [`join_view`](https://en.cppreference.com/w/cpp/ranges/join_view) for this too

lidavidm (approved):

src/iceberg/util/checked_cast_internal.h
```diff
@@ -0,0 +1,61 @@
+/*
```
zhjwpku: Should we install this header by removing the internal suffix? I don't see it being used in this PR, but in #94, it's included in visit_type.h, which is meant to be installed for type manipulation. If we don't install this header, it might lead to compilation issues.

wgtmac: I was using it for the visitor but not it is not required. Let me remove it for now. I saw that you have added that in your PR.

zhjwpku (approved):

lidavidm (approved):

src/iceberg/name_mapping.cc
wgtmac: I tried it but it seems that Clang does not yet implement this.

src/iceberg/CMakeLists.txt
```diff
@@ -36,7 +37,6 @@ set(ICEBERG_SOURCES
     transform.cc
     transform_function.cc
     type.cc
-    snapshot.cc
```
Fokko, line 39: Not a CPP guy, but why is `snapshot.cc` removed here? 🤔

Fokko:
Thanks @wgtmac for adding this, and @lidavidm @zhjwpku for the reviews 🙌

src/iceberg/CMakeLists.txt
wgtmac: It is duplicate at line 32 :)

PR #94, opened 2025-05-09
feat: add visit type support

zhjwpku:
Add a `TypeId` field to each concrete type to enable macro-based type visitor support.


src/iceberg/util/visit_type.h
```diff
@@ -0,0 +1,121 @@
...
+///
+/// The intent is for this to be called on a generic lambda
+/// that may internally use `if constexpr` or similar constructs.
+template <typename VISITOR, typename... ARGS>
+inline auto VisitType(const Type& type, VISITOR&& visitor, ARGS&&... args)
+    -> decltype(std::forward<VISITOR>(visitor)(type, args...)) {
+  switch (type.type_id()) {
+    ICEBERG_GENERATE_FOR_ALL_TYPES(TYPE_VISIT_INLINE);
+    default:
+      internal::Unreachable("Type not implemented");
```
lidavidm, line 96: The docstring says a fallback is called but the body aborts in the default case?

zhjwpku: TBH I have no idea what that sentence actually means, I don't think these words does any help to understand the template, so I'd rather delete that, or do you have any suggestions?

lidavidm: So this was copied from Arrow? It may be worth noting that in the header just for reference.

Looking at it again, I think what it means is that because the visitor here does not inherit from a base class that has all cases defined, you must make sure to handle all cases. So I guess the implementation is OK. But the wording could be improved.

zhjwpku: > So this was copied from Arrow? It may be worth noting that in the header just for reference.

Yeah, it's adapted from arrow, I'll add the note.

> 
> Looking at it again, I think what it means is that because the visitor here does not inherit from a base class that has all cases defined, you must make sure to handle all cases. So I guess the implementation is OK. But the wording could be improved.

How about we keep this as is for now, and revisit it later, either improve the wording ourselves or sync with Arrow if they update their version.

lidavidm (approved):

src/iceberg/util/unreachable.h
```diff
@@ -0,0 +1,33 @@
...
+
+#pragma once
+
+#include <string_view>
+
+#include "iceberg/iceberg_export.h"
+
+namespace iceberg::internal {
+
+[[noreturn]] ICEBERG_EXPORT void Unreachable(
```
wgtmac, line 28: Why not directly use `std::unreachable`?

src/iceberg/util/checked_cast.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
```
wgtmac, line 20: nit: add the `\file` docstring and mention that it is adapted from Apache Arrow.

wgtmac (approved):
Thanks for adding this! I was thinking of the same thing.

src/iceberg/util/unreachable.h
zhjwpku: I did a quick search and found that std::unreachable was introduced in C++23 [1]. And it doesn't support passing a custom message, which I think might be useful for logging purposes.

[1] https://en.cppreference.com/w/cpp/utility/unreachable

src/iceberg/util/checked_cast.h
zhjwpku: Will add later today, thanks for the suggestion.

src/iceberg/util/unreachable.h
wgtmac: Ah my bad. Please ignore this comment.

src/iceberg/util/checked_cast.h
Fokko: If we take big pieces from other projects, then we also want to add that to the `LICENSE`:

For example in PyIceberg: https://github.com/apache/iceberg-python/blob/260ef54e3920d435ae3b2ccda090e66f9c1ac015/LICENSE#L206-L213

zhjwpku: Sounds good, let's include that in both the LICENSE and NOTICE.

wgtmac (approved):

Xuanwo (approved):
The community has reached a consensus. Let's move!

PR #97, opened 2025-05-13
feat: add json serde to name mapping

wgtmac:
- add roundtrip json conversion to name mapping
- change MappedField to use std::shared_ptr for better usability
- use optional field_id in MappedField

### DIFF
--- a/src/iceberg/name_mapping.h
+++ b/src/iceberg/name_mapping.h
@@ -39,7 +39,6 @@ struct ICEBERG_EXPORT MappedField {
   /// \brief A required list of 0 or more names for a field.
   std::unordered_set<std::string> names;
   /// \brief An optional Iceberg field ID used when a field's name is present in `names`.
-  /// TODO(gangwu): check if we need to make it optional
   std::optional<int32_t> field_id;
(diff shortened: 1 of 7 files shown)


lidavidm (approved):

src/iceberg/name_mapping.h
```diff
@@ -40,10 +40,10 @@ struct ICEBERG_EXPORT MappedField {
   std::unordered_set<std::string> names;
   /// \brief An optional Iceberg field ID used when a field's name is present in `names`.
   /// TODO(gangwu): check if we need to make it optional
```
zhjwpku: The TODO note can be deleted.

zhjwpku (approved):

src/iceberg/name_mapping.h
wgtmac:
```suggestion
```

wgtmac: Good catch!

Fokko (approved):

Fokko:
Looks good @wgtmac, thanks for the review @lidavidm and @zhjwpku 🙌

PR #101, opened 2025-05-19
feat: add zlib dependency

dongxiao1198:
find zlib package in ThirdpartyToolchain.cmake and enable ARROW_WITH_ZLIB


cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -86,6 +86,9 @@ function(resolve_arrow_dependency)
   set(ARROW_DEPENDENCY_SOURCE
       "BUNDLED"
       CACHE STRING "" FORCE)
+  set(ARROW_WITH_ZLIB
```
wgtmac, line 89: We should also add `set(ZLIB_SOURCE "SYSTEM" CACHE STRING "" FORCE)`

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -285,3 +287,20 @@ function(resolve_nlohmann_json_dependency)
...
+  if(ZLIB_FOUND)
+    list(APPEND ICEBERG_SYSTEM_DEPENDENCIES ZLIB)
+    message(STATUS "ZLIB_FOUND")
+    message(STATUS "ZLIB_LIBRARIES: ${ZLIB_LIBRARIES}")
+    message(STATUS "ZLIB_INCLUDE_DIR: ${ZLIB_INCLUDE_DIR}")
+  endif()
+
+endfunction()
+
+resolve_zlib_dependency()
```
wgtmac, line 299: We need to move all of these to the top. `resolve_zlib_dependency()` should be called before `resolve_arrow_dependency()` and `resolve_avro_dependency()`

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -285,3 +287,20 @@ function(resolve_nlohmann_json_dependency)
...
+
+# ----------------------------------------------------------------------
+# zlib
+
+function(resolve_zlib_dependency)
+  # use system zlib, zlib is required by arrow and avro
+  find_package(ZLIB REQUIRED)
+  if(ZLIB_FOUND)
+    list(APPEND ICEBERG_SYSTEM_DEPENDENCIES ZLIB)
+    message(STATUS "ZLIB_FOUND")
```
wgtmac: Can we use a single `message` call?

wgtmac: BTW, we should also check if target `ZLIB::ZLIB` exists

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -285,3 +287,20 @@ function(resolve_nlohmann_json_dependency)
...
+    message(STATUS "ZLIB_LIBRARIES: ${ZLIB_LIBRARIES}")
+    message(STATUS "ZLIB_INCLUDE_DIR: ${ZLIB_INCLUDE_DIR}")
+  endif()
+
```
wgtmac, line 296:
```suggestion
  set(ICEBERG_SYSTEM_DEPENDENCIES
      ${ICEBERG_SYSTEM_DEPENDENCIES}
      PARENT_SCOPE)
```

The CI is failed: https://github.com/apache/iceberg-cpp/actions/runs/15106986625/job/42457669021?pr=101 We need to add this to the parent scope

cmake_modules/IcebergThirdpartyToolchain.cmake
dongxiao1198: DONE

cmake_modules/IcebergThirdpartyToolchain.cmake
dongxiao1198: i found those code in zlib cmake script
    add_library(ZLIB::ZLIB STATIC IMPORTED)
    set(ZLIB_LIBRARIES ${ZLIB_STATIC_LIB})
    set(ZLIB_INCLUDE_DIRS "${ZLIB_PREFIX}/include")
    set_target_properties(ZLIB::ZLIB PROPERTIES IMPORTED_LOCATION ${ZLIB_LIBRARIES})
so i think there is no need to check it.

cmake_modules/IcebergThirdpartyToolchain.cmake
dongxiao1198: I moved them back to the end of file, that will be better.

Fokko (approved):
Looks good @dongxiao1198, thanks for the review @wgtmac

PR #102, opened 2025-05-20
feat: add schema projection support

wgtmac:
Implement a `SchemaProjection` companion data structure to work with expected `Schema` to carry projection result. This is useful to project data returned by different file format reader implementations.

### DIFF
--- a/src/iceberg/schema_util.cc
+++ b/src/iceberg/schema_util.cc
@@ -1,0 +1,213 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/schema_util.h"
+
+#include <format>
+#include <map>
+#include <string_view>
+#include <unordered_map>
+
+#include "iceberg/metadata_columns.h"
+#include "iceberg/schema.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/formatter_internal.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
+
+Status ValidateSchemaEvolution(const Type& expected_type, const Type& source_type) {
+  if (expected_type.is_nested()) {
+    // Nested type requires identical type ids.
+    if (source_type.type_id() != expected_type.type_id()) {
+      return NotSupported("Cannot read {} from {}", expected_type, source_type);
+    }
+    return {};
+  }
+
+  // Short cut for same primitive type.
+  if (expected_type == source_type) {
+    return {};
+  }
+
+  switch (expected_type.type_id()) {
+    case TypeId::kLong: {
+      if (source_type.type_id() == TypeId::kInt) {
+        return {};
+      }
+    } break;
+    case TypeId::kDouble: {
+      if (source_type.type_id() == TypeId::kFloat) {
+        return {};
+      }
+    } break;
+    case TypeId::kDecimal: {
+      if (source_type.type_id() == TypeId::kDecimal) {
+        const auto& expected_decimal =
+            internal::checked_cast<const DecimalType&>(expected_type);
+        const auto& source_decimal =
+            internal::checked_cast<const DecimalType&>(source_type);
+        if (expected_decimal.precision() >= source_decimal.precision() &&
+            expected_decimal.scale() == source_decimal.scale()) {
+          return {};
+        }
+      }
+    } break;
+    default:
+      break;
+  }
+  return NotSupported("Cannot read {} from {}", expected_type, source_type);
+}
+
+// Fix `from` field of `FieldProjection` to use pruned field index.
+void PruneFieldProjection(FieldProjection& field_projection) {
+  std::map<size_t, size_t> local_index_to_pruned_index;
+  for (const auto& child_projection : field_projection.children) {
+    if (child_projection.kind == FieldProjection::Kind::kProjected) {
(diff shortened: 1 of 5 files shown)


src/iceberg/schema_util.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace iceberg {
+
+/// \brief A field schema partner to carry projection information.
+struct ICEBERG_EXPORT FieldProjection {
+  /// \brief How the field is projected.
+  enum class Kind {
+    /// \brief The field is projected from the source with possible conversion for
+    /// supported
+    /// schema evolution.
```
zhjwpku:
```suggestion
    /// supported schema evolution.
```
nit: extra line break

zhjwpku (approved):

lidavidm (approved):

yingcai-cy (approved):

src/iceberg/schema_util.cc
```diff
@@ -0,0 +1,213 @@
...
+#include "iceberg/schema.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/formatter_internal.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
+
+Status ValidateSchemaEvolution(const Type& expected_type, const Type& source_type) {
```
mapleFU, line 37: this lgtm, besides, I found iceberg v3 has more rules, https://iceberg.apache.org/spec/ :
1. unknown -> any types
2. date -> timestamp

src/iceberg/schema_util.cc
```diff
@@ -0,0 +1,213 @@
...
+  if (expected_type.is_nested()) {
+    // Nested type requires identical type ids.
+    if (source_type.type_id() != expected_type.type_id()) {
+      return NotSupported("Cannot read {} from {}", expected_type, source_type);
```
mapleFU, line 42: should we check `struct<a:int, b: int>` and `struct<a:int, c: int>` recursively for map/array/struct? Looks like the caller would not even pass nested_type to this struct

src/iceberg/schema_util.h
```diff
@@ -0,0 +1,97 @@
...
+  enum class Kind {
+    /// \brief The field is projected from the source with possible conversion for
+    /// supported schema evolution.
+    kProjected,
+    /// \brief Metadata column whose value is generated on demand.
+    kMetadata,
+    /// \brief The field is a constant value (e.g. partition field value)
+    kConstant,
+    /// \brief The field is missing in the source and should be filled with default value.
+    kDefault,
```
mapleFU, line 46: So this should be **not null**? Or it can be null like `kNull`?

src/iceberg/schema_util.h
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  /// \brief The field index in the source schema on the same nesting level when
+  /// `kind` is `kProjected`.
+  using SourceFieldIndex = size_t;
+  /// \brief A literal value used when `kind` is `kConstant` or `kDefault`.
+  /// TODO(gangwu): replace it with a specifically defined literal type
+  using Literal = std::any;
+  /// \brief A variant to indicate how to set the value of the field.
+  using From = std::variant<std::monostate, SourceFieldIndex, Literal>;
```
mapleFU, line 58: What does monostate means? From un-exists field?

src/iceberg/schema_util.h
```diff
@@ -0,0 +1,97 @@
...
+struct ICEBERG_EXPORT SchemaProjection {
+  std::vector<FieldProjection> fields;
+};
+
+/// \brief Project the expected schema on top of the source schema.
+///
+/// \param expected_schema The expected schema.
+/// \param source_schema The source schema.
+/// \param prune_source Whether to prune the source schema. If true, the source
+/// schema will be pruned to match the expected schema.
```
mapleFU: Can we explain it more specifically, what would `prune_source` affect `SchemaProjection` result

src/iceberg/schema_util.cc
```diff
@@ -0,0 +1,213 @@
...
+        ICEBERG_RETURN_UNEXPECTED(
+            ValidateSchemaEvolution(*expected_field.type(), *iter->second.field->type()));
+      }
+      // If `prune_source` is false, all fields will be read so the local index
+      // is exactly the position to read data. Otherwise, the local index is computed
+      // by pruning all non-projected fields
+      child_projection.from = iter->second.local_index;
+      child_projection.kind = FieldProjection::Kind::kProjected;
+    } else if (MetadataColumns::IsMetadataColumn(field_id)) {
+      child_projection.kind = FieldProjection::Kind::kMetadata;
```
mapleFU, line 155: Should this be checked before `source_field_map.find(field_id)`?

src/iceberg/schema_util.cc
wgtmac: Yes, our goal is to implement v2 support first. So we don't have unknown type and timestamp with ns precision.

src/iceberg/schema_util.h
wgtmac: In the case of `kNull` or `kDefault`, it is monostate because the value can be deduced.

src/iceberg/schema_util.h
wgtmac: This applies to a field that has set default value, regardless of its nullability. It should be checked before nullability though I didn't implement it since it is a v3 feature.

src/iceberg/schema_util.cc
wgtmac: No, in most cases fields should be found from the source schema. Checking metadata column first may introduces more misses.

src/iceberg/schema_util.h
wgtmac: Added more comment

src/iceberg/schema_util.cc
wgtmac: Good catch! Added more test cases and fixed some issues.

src/iceberg/schema_util.cc
```diff
@@ -0,0 +1,212 @@
...
+#include "iceberg/util/formatter_internal.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
+
+Status ValidateSchemaEvolution(const Type& expected_type, const Type& source_type) {
+  if (expected_type.is_nested()) {
+    // Nested type requires identical type ids.
```
mapleFU: Should we mark this as "subfields would check by children's `ValidateSchemaEvolution` call"?

mapleFU (approved):

src/iceberg/schema_util.cc
wgtmac: Sounds good. Added an explanation.

Fokko (approved):
This looks good, thanks for adding this @wgtmac and especially for the extensive testing 👍 Thanks @lidavidm, @zhjwpku, @mapleFU and @yingcai-cy for the reviews 🚀

PR #103, opened 2025-05-20
fix: bump avro version to enable uuid type test case

wgtmac:


lidavidm (approved):

dongxiao1198:
lgtm

Xuanwo (approved):
The community has reached a consensus; let's proceed.

By the way, what's the process of bumping deps? Is it possible to pin on a tag instead of a commit hash?

wgtmac:
@Xuanwo Thanks for the comment! Ideally it should be downloaded from a released tar ball with checksum verified. However, I have fixed many issues on the avro-cpp side so we have to use a specific commit hash from the main branch before the next 1.13.0 version is released. This is just a transient state to unblock our development.

Fokko (approved):
Nice one @wgtmac! 🙌 Thanks @lidavidm, @Xuanwo and @yingcai-cy for the review 🚀

PR #105, opened 2025-05-20
feat: add avro input&output stream based on arrow stream impl

dongxiao1198:
support basic avro stream based on arrow implement


src/iceberg/CMakeLists.txt
```diff
@@ -79,8 +79,12 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/iceberg_export.h
...
 
 if(ICEBERG_BUILD_BUNDLE)
-  set(ICEBERG_BUNDLE_SOURCES arrow/demo_arrow.cc arrow/arrow_fs_file_io.cc
-                             avro/demo_avro.cc avro/avro_schema_util.cc)
+  set(ICEBERG_BUNDLE_SOURCES
+      arrow/demo_arrow.cc
+      arrow/arrow_fs_file_io.cc
+      avro/demo_avro.cc
+      avro/avro_schema_util.cc
+      avro/avro_stream.cpp)
```
wgtmac: Rename it to avro_stream.cc

test/avro_test.cc
```diff
@@ -44,7 +64,40 @@ TEST(AVROTest, TestDemoAvro) {
   EXPECT_EQ(avro.print(), expected);
 }
 
-TEST(AVROTest, TestDemoAvroReader) {
+TEST_F(AVROTest, TestAvroBasicStream) {
```
wgtmac: What about adding a separate `avro_stream_test.cc` file with more test cases?

src/iceberg/avro/avro_stream.h
```diff
@@ -0,0 +1,118 @@
...
+                           int64_t buffer_size);
+
+  ~AvroInputStream() override;
+
+  /**
+   * Returns some of available data.
+   *
+   * Returns true if some data is available, false if no more data is
+   * available or an error has occurred.
+   */
```
wgtmac:
```suggestion
   /// \brief Returns some of available data.
   ///
   /// Returns true if some data is available, false if no more data is
   /// available or an error has occurred.
```

Let's keep the comment style consistent with this repo.

src/iceberg/avro/avro_stream.cc
```diff
@@ -0,0 +1,148 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "avro_stream.h"
+
+#include <format>
+
+#include <arrow/result.h>
+#include <iceberg/exception.h>
```
wgtmac:
```suggestion

#include "iceberg/exception.h"
```

nit

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <arrow/filesystem/localfs.h>
+#include <arrow/result.h>
+#include <gtest/gtest.h>
+#include <iceberg/arrow/arrow_fs_file_io.h>
+#include <iceberg/avro/avro_stream.h>
```
wgtmac:
```suggestion

#include "iceberg/arrow/arrow_fs_file_io.h"
#include "iceberg/avro/avro_stream.h"
```

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+    TempFileTestBase::SetUp();
+    file_io_ = std::make_shared<iceberg::arrow::ArrowFileSystemFileIO>(
+        std::make_shared<::arrow::fs::LocalFileSystem>());
+    temp_filepath_ = CreateNewTempFilePath();
+    local_fs_ = std::make_shared<::arrow::fs::LocalFileSystem>();
+  }
+
+  std::shared_ptr<AvroOutputStream> CreateOutputStream(const std::string& path,
+                                                       int64_t buffer_size) {
+    std::cout << "CreateOutputStream" << path << std::endl;
```
wgtmac: Can we remove these `std::cout`?

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+
+#include "temp_file_test_base.h"
+
+namespace iceberg::avro {
+
+class AVROStreamTest : public TempFileTestBase {
+ public:
+  void SetUp() override {
+    TempFileTestBase::SetUp();
+    file_io_ = std::make_shared<iceberg::arrow::ArrowFileSystemFileIO>(
```
wgtmac: `file_io_` is not used any where.

src/iceberg/avro/avro_stream.cc
zhjwpku:
```suggestion
#include "iceberg/exception.h"
```

For iceberg headers, it's better to use double quote style since it's in the current working directory.

test/avro_stream_test.cc
zhjwpku:
```suggestion
#include "iceberg/arrow/arrow_fs_file_io.h"
#include "iceberg/avro/avro_stream.h"
```

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+        std::make_shared<::arrow::fs::LocalFileSystem>());
+    temp_filepath_ = CreateNewTempFilePath();
+    local_fs_ = std::make_shared<::arrow::fs::LocalFileSystem>();
+  }
+
+  std::shared_ptr<AvroOutputStream> CreateOutputStream(const std::string& path,
+                                                       int64_t buffer_size) {
+    std::cout << "CreateOutputStream" << path << std::endl;
+    auto arrow_out_ret = local_fs_->OpenOutputStream(path);
+    if (!arrow_out_ret.ok()) {
```
wgtmac, line 40: nit: use `ASSERT_THAT`

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+      throw std::runtime_error("Failed to open output stream: " +
+                               arrow_out_ret.status().message());
+    }
+    return std::make_shared<AvroOutputStream>(std::move(arrow_out_ret.ValueUnsafe()),
+                                              buffer_size);
+  }
+
+  std::shared_ptr<AvroInputStream> CreateInputStream(const std::string& path,
+                                                     int64_t buffer_size) {
+    std::cout << "CreateInputStream" << path << std::endl;
```
wgtmac: ditto

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+};
+
+TEST_F(AVROStreamTest, TestAvroBasicStream) {
+  // Write test data
+  const std::string test_data = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+  auto avro_output_stream = CreateOutputStream(temp_filepath_, buffer_size_);
+  WriteDataToStream(avro_output_stream, test_data);
+
+  auto avro_input_stream = CreateInputStream(temp_filepath_, buffer_size_);
+  {
```
wgtmac: it seems that the block is unnecessary

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  avro_input_stream->backup(backupSize);
+
+  // Check byteCount after backup
+  EXPECT_EQ(avro_input_stream->byteCount(), test_data.size() - backupSize);
+
+  // Read the backed-up data again
+  ASSERT_TRUE(avro_input_stream->next(&data, &len));
+  EXPECT_EQ(len, backupSize);
+  EXPECT_EQ(std::string(reinterpret_cast<const char*>(data), len),
+            test_data.substr(test_data.size() - backupSize));  // NOLINT
```
wgtmac:
```suggestion
            test_data.substr(test_data.size() - backupSize));
```

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  // Check byteCount after skip
+  EXPECT_EQ(avro_input_stream->byteCount(), skipSize);
+
+  // Read the remaining data
+  const uint8_t* data{};
+  size_t len{};
+  ASSERT_TRUE(avro_input_stream->next(&data, &len));
+  EXPECT_EQ(len, test_data.size() - skipSize);
+  EXPECT_EQ(std::string(reinterpret_cast<const char*>(data), len),
+            test_data.substr(skipSize));  // NOLINT
```
wgtmac:
```suggestion
            test_data.substr(skipSize));
```

test/avro_stream_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  // Check byteCount after seek
+  EXPECT_EQ(avro_input_stream->byteCount(), static_cast<size_t>(seekPos));
+
+  // Read the remaining data
+  const uint8_t* data{};
+  size_t len{};
+  ASSERT_TRUE(avro_input_stream->next(&data, &len));
+  EXPECT_EQ(len, test_data.size() - seekPos);
+  EXPECT_EQ(std::string(reinterpret_cast<const char*>(data), len),
+            test_data.substr(seekPos));  // NOLINT
```
wgtmac:
```suggestion
            test_data.substr(seekPos));
```

src/iceberg/avro/avro_stream.cc
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "avro_stream.h"
```
wgtmac: Rename the header file to `"avro_stream_internal.h"` because we don't want to install public headers with symbols from third-party libraries.

wgtmac (approved):
LGTM. Thanks!

zhjwpku (approved):
LGTM

src/iceberg/avro/avro_stream_internal.cc
```diff
@@ -0,0 +1,149 @@
...
+  byte_count_ = position;
+}
+
+AvroOutputStream::AvroOutputStream(std::shared_ptr<arrow::io::OutputStream> output_stream,
+                                   int64_t buffer_size)
+    : output_stream_(std::move(output_stream)),
+      buffer_size_(buffer_size),
+      buffer_(buffer_size) {}
+
+AvroOutputStream::~AvroOutputStream() = default;
```
yingcai-cy, line 109: nit:should we call flush() in destructor?

src/iceberg/avro/avro_stream_internal.cc
```diff
@@ -0,0 +1,149 @@
...
+    *len = available_bytes_ - buffer_pos_;
+    byte_count_ += available_bytes_ - buffer_pos_;
+    buffer_pos_ = available_bytes_;
+    return true;
+  }
+
+  // Read from the input stream when the buffer is empty
+  auto result = input_stream_->Read(buffer_.size(), buffer_.data());
+  if (!result.ok()) {
+    throw IcebergError(
```
yingcai-cy: should use return but not exception?

zhjwpku: I'm afraid not, this is an override method of ::avro::SeekableInputStream, you can see that in the header file.

src/iceberg/avro/avro_stream_internal.cc
dongxiao1198: flush in dtor can not ensure that buffer will be flushed, and it will be difficult to handle flush error, so we advise calling flush() before closing the stream.

src/iceberg/avro/avro_stream_internal.cc
dongxiao1198: I found this doc in avro header:
    /**
     * Returns some of available data.
     *
     * Returns true if some data is available, false if no more data is
     * available or an error has occurred.
     */
    virtual bool next(const uint8_t **data, size_t *len) = 0;

that's right ,i will fix this. 
But since we have no logging system now, just return false will ignore the error msg and nobody can find out the root cause, so i will leave a TODO here to add some error log to debug error.

wgtmac (approved):

yingcai-cy (approved):

src/iceberg/avro/avro_stream_internal.h
```diff
@@ -0,0 +1,98 @@
...
+  ~AvroInputStream() override;
+
+  /// \brief Returns some of available data.
+  /// \return true if some data is available, false if no more data is available or an
+  /// error has occurred.
+  bool next(const uint8_t** data, size_t* len) override;
+
+  /// \brief "Returns" back some of the data to the stream. The returned data must be less
+  /// than what was obtained in the last call to next().
+  void backup(size_t len) override;
```
Fokko, line 41: Could you elaborate why we need a backup method when reading Avro? Does not make a lot of sense to me.

wgtmac: This is a public interface from avro-cpp. IIRC, it internally reads a large chunk into the buffer and then backup some if read too much to reduce I/O.

Fokko (approved):
LGTM @dongxiao1198, thanks for the reviews @yingcai-cy, @zhjwpku and @wgtmac

PR #106, opened 2025-05-21
feat: add field id checker to avro schema

wgtmac:


src/iceberg/avro/avro_schema_util.cc
```diff
@@ -263,4 +265,116 @@ Status ToAvroNodeVisitor::Visit(const SchemaField& field, ::avro::NodePtr* node)
...
+      return InvalidSchema("Unsupported Avro type: {}", static_cast<int>(node->type()));
+  }
+}
+
+Status HasIdVisitor::VisitRecord(const ::avro::NodePtr& node) {
+  static const std::string kFieldIdKey{kFieldIdProp};
+  total_fields_ += node->leaves();
+  for (size_t i = 0; i < node->leaves(); ++i) {
+    if (HasId(node, i, kFieldIdKey)) {
+      fields_with_id_++;
```
Fokko, line 315: Why keep track of the number of fields? Returning a boolean feels much less error-prone to me. On top of that, you can also very nicely `&&` the results:

https://github.com/apache/iceberg-python/blob/f47513b8fc972dd26b99b1a9fbb102c712dd07fe/pyiceberg/io/pyarrow.py#L1132-L1155

wgtmac: This visitor mixes the feature of `HasIds` and `MissingIds` from the Java implementation so I can check them in a single pass. I think an Avro schema should be considered valid when all or none of fields have field-ids, but not in between.

Fokko: Ah, I see, makes sense 👍 

> I think an Avro schema should be considered valid when all or none of fields have field-ids, but not in between.

I couldn't agree more :)

Fokko (approved):

Fokko:
Let's move this forward, thanks @wgtmac 🙌

PR #108, opened 2025-05-23
feat: support decompress gzip metadata

dongxiao1198:
support gzip compressed metadata json

### DIFF
--- a/src/iceberg/result.h
+++ b/src/iceberg/result.h
@@ -31,6 +31,7 @@ namespace iceberg {
 enum class ErrorKind {
   kAlreadyExists,
   kCommitStateUnknown,
+  kDecompressError,
   kInvalidArgument,
   kInvalidExpression,
   kInvalidSchema,
@@ -38,7 +39,6 @@ enum class ErrorKind {
   kJsonParseError,
   kNoSuchNamespace,
   kNoSuchTable,
-  kDecompressError,
   kNotFound,
   kNotImplemented,
   kNotSupported,
@@ -74,6 +74,7 @@ using Status = Result<void>;
 
 DEFINE_ERROR_FUNCTION(AlreadyExists)
 DEFINE_ERROR_FUNCTION(CommitStateUnknown)
+DEFINE_ERROR_FUNCTION(DecompressError)
 DEFINE_ERROR_FUNCTION(InvalidArgument)
 DEFINE_ERROR_FUNCTION(InvalidExpression)
 DEFINE_ERROR_FUNCTION(InvalidSchema)
@@ -81,7 +82,6 @@ DEFINE_ERROR_FUNCTION(IOError)
 DEFINE_ERROR_FUNCTION(JsonParseError)
 DEFINE_ERROR_FUNCTION(NoSuchNamespace)
 DEFINE_ERROR_FUNCTION(NoSuchTable)
-DEFINE_ERROR_FUNCTION(DecompressError)
 DEFINE_ERROR_FUNCTION(NotFound)
 DEFINE_ERROR_FUNCTION(NotImplemented)
 DEFINE_ERROR_FUNCTION(NotSupported)
--- a/src/iceberg/result.h
+++ b/src/iceberg/result.h
@@ -38,6 +38,7 @@ enum class ErrorKind {
   kJsonParseError,
   kNoSuchNamespace,
   kNoSuchTable,
+  kDecompressError,
   kNotFound,
   kNotImplemented,
   kNotSupported,
@@ -80,6 +81,7 @@ DEFINE_ERROR_FUNCTION(IOError)
 DEFINE_ERROR_FUNCTION(JsonParseError)
 DEFINE_ERROR_FUNCTION(NoSuchNamespace)
 DEFINE_ERROR_FUNCTION(NoSuchTable)
+DEFINE_ERROR_FUNCTION(DecompressError)
 DEFINE_ERROR_FUNCTION(NotFound)
 DEFINE_ERROR_FUNCTION(NotImplemented)
 DEFINE_ERROR_FUNCTION(NotSupported)
--- a/src/iceberg/table_metadata.cc
+++ b/src/iceberg/table_metadata.cc
@@ -154,68 +154,40 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
   return MetadataFileCodecType::kNone;
 }
 
-class GZipDecompressor {
- public:
-  GZipDecompressor() : initialized_(false) {}
-
-  ~GZipDecompressor() {
-    if (initialized_) {
-      inflateEnd(&stream_);
-    }
+Result<std::string> DecompressGZIPFile(const std::string& filepath) {
+  gzFile file = gzopen(filepath.c_str(), "rb");
+  if (!file) {
+    return IOError("Failed to open gzip file:{} ", filepath);
   }
 
-  Status Init() {
-    int ret = inflateInit2(&stream_, 15 + 32);
-    if (ret != Z_OK) {
-      return IOError("inflateInit2 failed, result:{}", ret);
-    }
-    initialized_ = true;
-    return {};
+  const int CHUNK_SIZE = 32768;  // 32KB chunks
+  char buffer[CHUNK_SIZE];
+  std::string result;
+  int bytes_read;
+
+  while ((bytes_read = gzread(file, buffer, CHUNK_SIZE)) > 0) {
+    result.append(buffer, bytes_read);
   }
 
-  Result<std::string> Decompress(const std::string& compressed_data) {
-    if (compressed_data.empty()) {
-      return {};
-    }
-    if (!initialized_) {
-      ICEBERG_RETURN_UNEXPECTED(Init());
-    }
-    stream_.avail_in = static_cast<uInt>(compressed_data.size());
-    stream_.next_in = reinterpret_cast<Bytef*>(const_cast<char*>(compressed_data.data()));
-
-    // TODO(xiao.dong) magic buffer 16k, can we get a estimated size from compressed data?
-    std::vector<char> outBuffer(32 * 1024);
-    std::string result;
-    int ret = 0;
-    do {
-      outBuffer.resize(outBuffer.size());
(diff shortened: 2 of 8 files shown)


src/iceberg/table_metadata.cc
```diff
@@ -153,14 +155,42 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
...
+
+  while ((bytes_read = gzread(file, buffer.data(), CHUNK_SIZE)) > 0) {
+    result.append(buffer.data(), bytes_read);
+  }
+
+  int err;
+  const char* error_msg = gzerror(file, &err);
+  if (err != Z_OK) {
+    gzclose(file);
+    return IOError("Error during gzip decompression:{} ", std::string(error_msg));
```
lidavidm: I'm not convinced that `error_msg` is valid after `file` is closed.

lidavidm: In fact, if you peek at the implementation it points into the file structure

https://github.com/madler/zlib/blob/5a82f71ed1dfc0bec044d9702463dbdf84ea3b71/gzlib.c#L489-L505

src/iceberg/table_metadata.cc
```diff
@@ -153,14 +155,42 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
   return MetadataFileCodecType::kNone;
 }
 
+Result<std::string> DecompressGZIPFile(const std::string& filepath) {
+  gzFile file = gzopen(filepath.c_str(), "rb");
```
lidavidm: How is this supposed to work with files that aren't local filesystem files?

dongxiao1198: yes...can not use this here, i will fallback to use stream api,thx

src/iceberg/table_metadata.cc
```diff
@@ -153,14 +154,70 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
...
+  ~GZipDecompressor() {
+    if (initialized_) {
+      inflateEnd(&stream_);
+    }
+  }
+
+  Status Init() {
+    int ret = inflateInit2(&stream_, 15 + 32);
+    if (ret != Z_OK) {
+      return IOError("inflateInit2 failed, result:{}", ret);
```
wgtmac: nit: use a new `DecompressError` or `GZipError`

src/iceberg/table_metadata.cc
```diff
@@ -153,14 +154,70 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
...
+  GZipDecompressor() { memset(&stream_, 0, sizeof(stream_)); }
+
+  ~GZipDecompressor() {
+    if (initialized_) {
+      inflateEnd(&stream_);
+    }
+  }
+
+  Status Init() {
+    int ret = inflateInit2(&stream_, 15 + 32);
```
wgtmac: What does the magic number mean? Should we define it as a constant?

src/iceberg/table_metadata.cc
```diff
@@ -153,14 +154,70 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
   return MetadataFileCodecType::kNone;
 }
 
+class GZipDecompressor {
```
wgtmac: Should we just define it as a function because it is just an one-off thing? Perhaps declare it in `gzip_internal.h` so we can test it individually?

src/iceberg/table_metadata.cc
```diff
@@ -153,14 +154,70 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
...
+      return {};
+    }
+    if (!initialized_) {
+      ICEBERG_RETURN_UNEXPECTED(Init());
+    }
+    stream_.avail_in = static_cast<uInt>(compressed_data.size());
+    stream_.next_in = reinterpret_cast<Bytef*>(const_cast<char*>(compressed_data.data()));
+
+    // TODO(xiao.dong) magic buffer, can we get a estimated size from compressed data?
+    std::vector<char> outBuffer(32 * 1024);
```
```suggestion
    std::vector<char> out_buffer(32 * 1024);
```

src/iceberg/table_metadata.cc
```diff
@@ -153,14 +154,70 @@ Result<MetadataFileCodecType> TableMetadataUtil::CodecFromFileName(
...
+    std::string result;
+    int ret = 0;
+    do {
+      outBuffer.resize(outBuffer.size());
```
wgtmac: Isn't it redundant?

src/iceberg/table_metadata.cc
dongxiao1198: ok, considering of future use, i think move those code to "gzip_internal.cc" to utils dir is much more data.

src/iceberg/util/gzip_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+    if (initialized_) {
+      inflateEnd(&stream_);
+    }
+  }
+
+  Status Init() {
+    // Maximum window size
+    static int WINDOW_BITS = 15;
+    // Determine if this is libz or gzip from header.
+    static int DETECT_CODEC = 32;
```
wgtmac:
```suggestion
    static int kWindowBits = 15;
    // Determine if this is libz or gzip from header.
    static int kDetectCodec = 32;
```

src/iceberg/util/gzip_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+
+    int ret = inflateInit2(&stream_, WINDOW_BITS | DETECT_CODEC);
```
wgtmac:
```suggestion
    int ret = inflateInit2(&stream_, kWindowBits | kDetectCodec);
```

src/iceberg/util/gzip_internal.h
```diff
@@ -0,0 +1,45 @@
...
+  GZipDecompressor();
+
+  ~GZipDecompressor();
+
+  Status Init();
+
+  Result<std::string> Decompress(const std::string& compressed_data);
+
+ private:
+  std::shared_ptr<ZlibImpl> zlib_impl_;
```
wgtmac:
```suggestion
  std::unique_ptr<ZlibImpl> zlib_impl_;
```

wgtmac (approved):

src/iceberg/util/gzip_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+
+#include <zlib.h>
+
+#include <cstring>
+
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+class ZlibImpl {
```
yingcai-cy, line 30: IMHO, if this code is only used for a one-time decompression, introducing those classes might be unnecessarily complex. Maybe one util method is enough.

test/metadata_io_test.cc
```diff
@@ -78,4 +83,66 @@ TEST_F(MetadataIOTest, ReadWriteMetadata) {
   EXPECT_EQ(*metadata_read, metadata);
 }
 
+TEST_F(MetadataIOTest, ReadWriteCompressedMetadata) {
+  std::vector<SchemaField> schema_fields;
+  schema_fields.emplace_back(/*field_id=*/1, "x", std::make_shared<LongType>(),
+                             /*optional=*/false);
+  auto schema = std::make_shared<Schema>(std::move(schema_fields), /*schema_id=*/1);
+
+  TableMetadata metadata{.format_version = 1,
```
yingcai-cy: Extract to a helper method to get TableMetadata?

src/iceberg/util/gzip_internal.cc
dongxiao1198: c interface is better to use dtor to make sure no memleak

src/iceberg/result.h
```diff
@@ -38,6 +38,7 @@ enum class ErrorKind {
   kJsonParseError,
   kNoSuchNamespace,
   kNoSuchTable,
+  kDecompressError,
```
wgtmac: nit: sort alphabetically

src/iceberg/result.h
```diff
@@ -80,6 +81,7 @@ DEFINE_ERROR_FUNCTION(IOError)
 DEFINE_ERROR_FUNCTION(JsonParseError)
 DEFINE_ERROR_FUNCTION(NoSuchNamespace)
 DEFINE_ERROR_FUNCTION(NoSuchTable)
+DEFINE_ERROR_FUNCTION(DecompressError)
```
wgtmac: nit: sort alphabetically

src/iceberg/util/gzip_internal.cc
```diff
@@ -0,0 +1,94 @@
...
+
+  ~ZlibImpl() {
+    if (initialized_) {
+      inflateEnd(&stream_);
+    }
+  }
+
+  Status Init() {
+    // Maximum window size
+    static int kWindowBits = 15;
```
wgtmac:
```suggestion
    constexpr int kWindowBits = 15;
```

src/iceberg/util/gzip_internal.cc
```diff
@@ -0,0 +1,94 @@
...
+    // Determine if this is libz or gzip from header.
+    static int kDetectCodec = 32;
```
wgtmac:
```suggestion
    constexpr int kDetectCodec = 32;
```

src/iceberg/util/gzip_internal.cc
wgtmac: Perhaps we can define a nesting class in that function and the dtor will be called when the function exists. I think @yingcai-cy's point is that we can simply declare `Result<std::string> Decompress(std::string_view input)` in the header file to make it easy to use.

yingcai-cy (approved):
LGTM

lidavidm (commented):
No more comments. I think we will want a bit more abstraction eventually since presumably we will want to support other compression methods but this works for now.

wgtmac:
Iceberg itself only requires gzip for metadata json file. Specific file format libraries we use for Parquet and Avro have already supported different codecs so we don't need to deal with them in both `libiceberg` and `libiceberg-bundle`.

PR #111, opened 2025-05-27
static table metadata access support

lishuxu:


src/iceberg/table.cc
```diff
@@ -0,0 +1,173 @@
...
+BaseTable::BaseTable(std::string name, std::shared_ptr<TableMetadata> metadata)
+    : name_(std::move(name)), metadata_(std::move(metadata)) {
+  if (!metadata_) {
+    throw std::invalid_argument("Table metadata cannot be null");
+  }
+}
+
+void BaseTable::InitSchema() const {
+  for (const auto& schema : metadata_->schemas) {
+    if (schema->schema_id() != std::nullopt) {
```
gty404: if (schema->schema_id())

src/iceberg/table.h
```diff
@@ -108,4 +109,85 @@ class ICEBERG_EXPORT Table {
...
+  std::shared_ptr<TableMetadata> metadata_;
+
+  // once_flags
+  mutable std::once_flag init_schema_once_;
+  mutable std::once_flag init_partition_spec_once_;
+  mutable std::once_flag init_sort_order_once_;
+  mutable std::once_flag init_snapshot_once_;
+};
+
+class ICEBERG_EXPORT StaticTable : public BaseTable {
```
lidavidm: Add docstring?

test/table_test.cc
```diff
@@ -0,0 +1,110 @@
...
+namespace iceberg {
+
+namespace {
+
+class TableTest : public ::testing::Test {
+ protected:
+  void SetUp() override {}
+};
+
+}  // namespace
```
lidavidm: You don't need all of this. You can just declare tests as `TEST(TableTest, FooTest)` instead of declaring this.

src/iceberg/table.h
```diff
@@ -108,4 +109,85 @@ class ICEBERG_EXPORT Table {
   virtual std::unique_ptr<LocationProvider> location_provider() const = 0;
 };
 
+class ICEBERG_EXPORT BaseTable : public Table {
+ public:
+  ~BaseTable() override = default;
+  BaseTable(std::string name, std::shared_ptr<TableMetadata> metadata);
```
wgtmac:
```suggestion
  ~BaseTable() override = default;

  BaseTable(std::string name, std::shared_ptr<TableMetadata> metadata);
```

test/table_test_helper.h
```diff
@@ -0,0 +1,42 @@
...
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+class TableTestHelper {
```
wgtmac: What about renaming this file to `test_common.h`? Functions below do not seem to be related to table. The class also seems to be unnecessary at the moment. We can write short code if we directly define them as functions not class static functions.

src/iceberg/table.h
```diff
@@ -108,4 +109,85 @@ class ICEBERG_EXPORT Table {
...
+  mutable std::once_flag init_schema_once_;
+  mutable std::once_flag init_partition_spec_once_;
+  mutable std::once_flag init_sort_order_once_;
+  mutable std::once_flag init_snapshot_once_;
+};
+
+class ICEBERG_EXPORT StaticTable : public BaseTable {
+ public:
+  ~StaticTable() override = default;
+  StaticTable(std::string name, std::shared_ptr<TableMetadata> metadata)
```
wgtmac:
```suggestion
  ~StaticTable() override = default;

  StaticTable(std::string name, std::shared_ptr<TableMetadata> metadata)
```

src/iceberg/table.cc
```diff
@@ -0,0 +1,173 @@
...
+namespace iceberg {
+
+BaseTable::BaseTable(std::string name, std::shared_ptr<TableMetadata> metadata)
+    : name_(std::move(name)), metadata_(std::move(metadata)) {
+  if (!metadata_) {
+    throw std::invalid_argument("Table metadata cannot be null");
+  }
+}
+
+void BaseTable::InitSchema() const {
```
wgtmac: The `TableMetadata` class has already defined below functions:

```
  /// \brief Get the current schema, return NotFoundError if not found
  Result<std::shared_ptr<Schema>> Schema() const;
  /// \brief Get the current partition spec, return NotFoundError if not found
  Result<std::shared_ptr<PartitionSpec>> PartitionSpec() const;
  /// \brief Get the current sort order, return NotFoundError if not found
  Result<std::shared_ptr<SortOrder>> SortOrder() const;
```

I would suggest to do following things:
- Add and implement `Result<std::share_ptr<Snapshot>> TableMetadata::Snapshot()` just like above functions.
- Modify the `Table` class to return `Result<std::shared_ptr<X>>` for objects of the current snapshot and directly call and return `TableMetadata::XXX()`.

src/iceberg/table.cc
```diff
@@ -0,0 +1,173 @@
...
+  // TODO(lishuxu): Implement history retrieval
+  throw IcebergError("history is not supported for BaseTable now");
+}
+
+Status StaticTable::Refresh() {
+  throw IcebergError("Refresh is not supported for StaticTable");
+}
+
+std::unique_ptr<TableScan> StaticTable::NewScan() const {
+  throw IcebergError("NewScan is not supported for StaticTable");
```
wgtmac: Refactor these functions to return `Result<X>`? We cannot throw exceptions.

wgtmac: BTW, these default implementations can be moved to `BaseTable` by returning `Unsupported` or `Unimplemented` error.

src/iceberg/table.cc
lishuxu: It is reasonable, I have modified it this way.

src/iceberg/table.h
lishuxu: OK, I have added the docstring, thanks!

test/table_test.cc
lishuxu: It is reasonable, I have modified it this way.

test/table_test_helper.h
lishuxu: It is reasonable, I have modified it this way, thanks!

src/iceberg/table.cc
lishuxu: OK, I have changed the API to Result<std::shared_ptr<X>> style. 
However, for performance reasons, the current BaseTable implementation has higher performance, so I did not directly call the functions in TableMetadata.

src/iceberg/table.cc
lishuxu: OK, I have changed the API to Result<X> style. I think we can also keep BaseTable as a virtual class, so that StaticTable returns NotSupported. If the method moves to BaseTable, it should return Unimplemented.

src/iceberg/table.h
```diff
@@ -20,12 +20,16 @@
 #pragma once
 
 #include <memory>
+#include <mutex>
```
wgtmac: What about putting both `BaseTable` and `StaticTable` into a separate header file like `table_impl.h`? The point is to keep `table.h` with minimal inclusion of header files. If we do this, we do not need to include mutex to this file.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,30 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+namespace iceberg {
+
+/// \brief Represents a table scan operation
+class ICEBERG_EXPORT TableScan {
```
wgtmac: Please delete this file. We don't need it in this PR.

test/test_common.h
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "iceberg/type_fwd.h"
+
+namespace iceberg::test {
```
wgtmac:
```suggestion
namespace iceberg {
```

We don't need the `test` namespace to save some keys to type.

src/iceberg/table_metadata.h
```diff
@@ -133,6 +133,8 @@ struct ICEBERG_EXPORT TableMetadata {
   Result<std::shared_ptr<PartitionSpec>> PartitionSpec() const;
   /// \brief Get the current sort order, return NotFoundError if not found
   Result<std::shared_ptr<SortOrder>> SortOrder() const;
+  /// \brief Get the current snapshot, return NotFoundError if not found
+  Result<std::shared_ptr<Snapshot>> Snapshot() const;
```
wgtmac, line 139: Add a test for the new function? You can add one line to an existing test to verify the returned snapshot has expected snapshot_id.

src/iceberg/table.cc
```diff
@@ -0,0 +1,179 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/table.h"
+
+#include <iostream>
```
wgtmac: Remove this

src/iceberg/table.h
```diff
@@ -20,12 +20,16 @@
...
 #include <string>
 #include <unordered_map>
 #include <vector>
 
 #include "iceberg/iceberg_export.h"
+#include "iceberg/location_provider.h"
```
wgtmac: ditto, remove this and below new includes.

src/iceberg/table.cc
```diff
@@ -0,0 +1,179 @@
...
+const std::vector<std::shared_ptr<Snapshot>>& BaseTable::snapshots() const {
+  return metadata_->snapshots;
+}
+
+const std::vector<std::shared_ptr<HistoryEntry>>& BaseTable::history() const {
+  // TODO(lishuxu): Implement history retrieval
+  return history_;
+}
+
+Status StaticTable::Refresh() {
```
wgtmac: Functions here and below should be changed as the default implementations of `BaseTable` so we don't need to reimplement them for other (unsupported) table implementations in the future.

src/iceberg/table.cc
```diff
@@ -0,0 +1,179 @@
...
+#include "iceberg/snapshot.h"
+#include "iceberg/sort_order.h"
+#include "iceberg/table_metadata.h"
+
+namespace iceberg {
+
+BaseTable::BaseTable(std::string name, std::shared_ptr<TableMetadata> metadata)
+    : name_(std::move(name)), metadata_(std::move(metadata)) {
+  if (!metadata_) {
+    throw std::invalid_argument("Table metadata cannot be null");
```
wgtmac:
```suggestion
    throw IcebergError("Table metadata cannot be null to create a table");
```

src/iceberg/table.cc
```diff
@@ -0,0 +1,179 @@
...
+  }
+}
+
+void BaseTable::InitSchema() const {
+  for (const auto& schema : metadata_->schemas) {
+    if (schema->schema_id()) {
+      schemas_map_.emplace(schema->schema_id().value(), schema);
```
wgtmac: This is not efficient if the downstream just need to get the current schema but pays for the time and memory for all schemas. Another drawback is that it duplicates the logic in the `TableMetadata::Schema()`. Same for other versioned objects below.

I think we just need to add a once_flag for metadata_->Schema() and cache the return value. Same for other functions.

src/iceberg/table.cc
```diff
@@ -0,0 +1,179 @@
...
+    return NotFound("Snapshot with ID {} not found", snapshot_id);
+  }
+  return iter->second;
+}
+
+const std::vector<std::shared_ptr<Snapshot>>& BaseTable::snapshots() const {
+  return metadata_->snapshots;
+}
+
+const std::vector<std::shared_ptr<HistoryEntry>>& BaseTable::history() const {
```
wgtmac: Please remove this function from `Table` for now. It is not well-thought. I will add them later.

test/table_test.cc
```diff
@@ -0,0 +1,101 @@
...
+
+#include "iceberg/partition_spec.h"
+#include "iceberg/schema.h"
+#include "iceberg/snapshot.h"
+#include "iceberg/table_metadata.h"
+#include "test_common.h"
+
+namespace iceberg {
+
+TEST(TableTest, TableSchemaV1Test) {
```
wgtmac:
```suggestion
TEST(StaticTable, TableV1) {
```

We don't need `Test` suffix here and the case name need to be more meaningful.

src/iceberg/table.h
```diff
@@ -44,21 +48,21 @@ class ICEBERG_EXPORT Table {
   /// \brief Refresh the current table metadata
   virtual Status Refresh() = 0;
 
-  /// \brief Return the schema for this table
-  virtual const std::shared_ptr<Schema>& schema() const = 0;
+  /// \brief Return the schema for this table, return NotFoundError if not found
+  virtual Result<std::shared_ptr<Schema>> schema() const = 0;
```
wgtmac:
```suggestion
  virtual Result<std::shared_ptr<Schema>> Schema() const = 0;
```

These functions are no longer trivial accessors. We need to rename them with capitalized initial.

src/iceberg/table.h
```diff
@@ -44,21 +48,21 @@ class ICEBERG_EXPORT Table {
...
 
   /// \brief Return a map of schema for this table
   virtual const std::unordered_map<int32_t, std::shared_ptr<Schema>>& schemas() const = 0;
 
-  /// \brief Return the partition spec for this table
-  virtual const std::shared_ptr<PartitionSpec>& spec() const = 0;
+  /// \brief Return the partition spec for this table, return NotFoundError if not found
+  virtual Result<std::shared_ptr<PartitionSpec>> spec() const = 0;
```
wgtmac:
```suggestion
  virtual Result<std::shared_ptr<PartitionSpec>> Spec() const = 0;
```

src/iceberg/table.h
```diff
@@ -44,21 +48,21 @@ class ICEBERG_EXPORT Table {
...
 
   /// \brief Return a map of partition specs for this table
   virtual const std::unordered_map<int32_t, std::shared_ptr<PartitionSpec>>& specs()
       const = 0;
 
-  /// \brief Return the sort order for this table
-  virtual const std::shared_ptr<SortOrder>& sort_order() const = 0;
+  /// \brief Return the sort order for this table, return NotFoundError if not found
+  virtual Result<std::shared_ptr<SortOrder>> sort_order() const = 0;
```
wgtmac:
```suggestion
  virtual Result<std::shared_ptr<SortOrder>> SortOrder() const = 0;
```

src/iceberg/table.h
```diff
@@ -70,8 +74,8 @@ class ICEBERG_EXPORT Table {
   /// \brief Return the table's base location
   virtual const std::string& location() const = 0;
 
-  /// \brief Return the table's current snapshot
-  virtual const std::shared_ptr<Snapshot>& current_snapshot() const = 0;
+  /// \brief Return the table's current snapshot, or NotFoundError if not found
+  virtual Result<std::shared_ptr<Snapshot>> current_snapshot() const = 0;
```
wgtmac:
```suggestion
  virtual Result<std::shared_ptr<Snapshot>> CurrentSnapshot() const = 0;
```

src/iceberg/table.h
lishuxu: I have modified it this format, thanks!

src/iceberg/table.h
lishuxu: I have modified it this format, thanks!

src/iceberg/table.h
lishuxu: OK, I will add new header file table_impl.h

src/iceberg/table_metadata.h
lishuxu: OK, I add test in testcase MetadataSerdeTest

src/iceberg/table.h
lishuxu: The method name schema will conflicts with the Schema object name, and since this is actually a lazy getter, I think it can be treated as an accessor.

src/iceberg/table_impl.cc
```diff
@@ -0,0 +1,171 @@
...
+#include "iceberg/snapshot.h"
+#include "iceberg/sort_order.h"
+#include "iceberg/table_metadata.h"
+
+namespace iceberg {
+
+BaseTable::BaseTable(std::string name, std::shared_ptr<TableMetadata> metadata)
+    : name_(std::move(name)), metadata_(std::move(metadata)) {
+  if (!metadata_) {
+    throw std::invalid_argument("Table metadata cannot be null");
```
wgtmac:
```suggestion
    throw IcebergError("Table metadata cannot be null");
```

src/iceberg/table.cc
wgtmac: This is not resolved yet.

src/iceberg/table.cc
wgtmac: I think these changes are still required

wgtmac (commented):
I think there are still issues to address. Please also rebase this PR on the latest main branch to resolve the conflict.

src/iceberg/type_fwd.h
```diff
@@ -99,6 +99,9 @@ class TransformFunction;
 struct PartitionStatisticsFile;
 struct Snapshot;
 struct SnapshotRef;
+struct SnapshotLogEntry;
+struct MetadataLogEntry;
```
wgtmac, line 103: Please either sort them alphabetically or add a blank line before `SnapshotLogEntry` so we can put these two entries together.

src/iceberg/snapshot.h
```diff
@@ -273,4 +273,20 @@ struct ICEBERG_EXPORT Snapshot {
   bool Equals(const Snapshot& other) const;
 };
 
+/// \brief Represents a snapshot log entry
```
wgtmac: Why moving `SnapshotLogEntry` to this file? I think it is part of the table metadata and should still be there.

src/iceberg/table_metadata.h
```diff
@@ -94,6 +80,8 @@ struct ICEBERG_EXPORT TableMetadata {
   TimePointMs last_updated_ms;
   /// The highest assigned column ID for the table
   int32_t last_column_id;
+  /// The current schema for the table, or null if not set
+  mutable std::shared_ptr<Schema> schema;
```
```suggestion
```

src/iceberg/table_metadata.h
```diff
@@ -129,12 +117,17 @@ struct ICEBERG_EXPORT TableMetadata {
   /// A `long` higher than all assigned row IDs
   int64_t next_row_id;
 
+  /// \brief Used for lazy initialization of schema
+  mutable std::once_flag init_schema_once;
+
```
```suggestion
```

src/iceberg/table_metadata.cc
```diff
@@ -47,13 +47,24 @@ std::string ToString(const MetadataLogEntry& entry) {
 }
 
 Result<std::shared_ptr<Schema>> TableMetadata::Schema() const {
```
wgtmac: Please revert changes to `Schema()` so it is consistent with other similar functions in this file. We can cache these returned values in the `Table` object.

src/iceberg/table_metadata.h
wgtmac: See my comment in `table_metadata.cc` to revert these changes.

src/iceberg/table_metadata.cc
```diff
@@ -76,6 +87,19 @@ Result<std::shared_ptr<SortOrder>> TableMetadata::SortOrder() const {
   return *iter;
 }
 
+Result<std::shared_ptr<Snapshot>> TableMetadata::Snapshot() const {
+  if (current_snapshot_id == Snapshot::kInvalidSnapshotId) {
+    return NotFound("Current snapshot is not defined for this table");
+  }
```
```suggestion
```

We don't need this special check because `kInvalidSnapshotId` can result in `NotFound` too.

src/iceberg/table.h
```diff
@@ -35,77 +36,96 @@ class ICEBERG_EXPORT Table {
...
+  /// \brief Return the table's current snapshot, return null if not found
+  std::shared_ptr<Snapshot> current_snapshot() const;
 
   /// \brief Get the snapshot of this table with the given id, or null if there is no
   /// matching snapshot
   ///
   /// \param snapshot_id the ID of the snapshot to get
   /// \return the Snapshot with the given id
-  virtual Result<std::shared_ptr<Snapshot>> snapshot(int64_t snapshot_id) const = 0;
+  std::shared_ptr<Snapshot> SnapshotById(int64_t snapshot_id) const;
```
wgtmac: What not returning `Result<std::shared_ptr<Snapshot>>`? It might return `NotFound` IMO.

src/iceberg/table.h
```diff
@@ -35,77 +36,96 @@ class ICEBERG_EXPORT Table {
...
   virtual ~Table() = default;
 
-  /// \brief Return the full name for this table
-  virtual const std::string& name() const = 0;
+  /// \brief Construct a table.
+  /// \param[in] identifier The identifier of the table.
+  /// \param[in] metadata The metadata for the table.
+  /// \param[in] metadata_location The location of the table metadata file.
+  /// \param[in] io The FileIO to read and write table data and metadata files.
+  /// \param[in] catalog The catalog that this table belongs to.
```
wgtmac: Should we add a comment to say that the table is read-only if catalog is nullptr?

src/iceberg/table.h
```diff
@@ -35,77 +36,96 @@ class ICEBERG_EXPORT Table {
...
 
-  /// \brief Create a new transaction API to commit multiple table operations at once
-  virtual std::unique_ptr<Transaction> NewTransaction() = 0;
+  mutable std::shared_ptr<SortOrder> sort_order_;
+  mutable std::unordered_map<int32_t, std::shared_ptr<SortOrder>> sort_orders_map_;
 
-  /// TODO(wgtmac): design of FileIO is not finalized yet. We intend to use an
-  /// IO-less design in the core library.
-  // /// \brief Returns a FileIO to read and write table data and metadata files
-  // virtual std::shared_ptr<FileIO> io() const = 0;
```
wgtmac, line 105: I think we need to keep `io()`?

src/iceberg/table.h
```diff
@@ -35,77 +36,96 @@ class ICEBERG_EXPORT Table {
...
+  mutable std::shared_ptr<Snapshot> current_snapshot_;
 
-  /// \brief Returns a LocationProvider to provide locations for new data files
-  virtual std::unique_ptr<LocationProvider> location_provider() const = 0;
+  // once_flags
```
wgtmac: I think a main shortcoming of these once flags is that if later we add a new `Table::Refresh()` function to be in sync with the latest metadata, these flags should also be cleared. I would suggest to remove these flags and solely rely on whether the shared_ptr is null to decide if it is initialized.

src/iceberg/table.cc
```diff
@@ -0,0 +1,130 @@
...
+
+const std::string& Table::uuid() const { return metadata_->table_uuid; }
+
+const std::shared_ptr<Schema>& Table::schema() const {
+  if (!schema_) {
+    const static std::shared_ptr<Schema> kEmptySchema =
+        std::make_shared<Schema>(std::vector<SchemaField>{});
+    auto schema = metadata_->Schema();
+    if (schema.has_value()) {
+      schema_ = schema.value();
```
wgtmac:
```suggestion
      schema_ = std::move(schema.value());
```

src/iceberg/table.cc
```diff
@@ -0,0 +1,130 @@
...
+      if (schema->schema_id()) {
+        schemas_map_.emplace(schema->schema_id().value(), schema);
+      }
+    }
+  });
+  return schemas_map_;
+}
+
+const std::shared_ptr<PartitionSpec>& Table::spec() const {
+  std::call_once(init_partition_spec_once_, [this]() {
```
wgtmac: I'd suggest to follow similar implementation as `Table::schema()`. In this function, we can cache and return `PartitionSpec::Unpartitioned()` when missing.

src/iceberg/table.cc
```diff
@@ -0,0 +1,130 @@
...
+const std::unordered_map<int32_t, std::shared_ptr<PartitionSpec>>& Table::specs() const {
+  std::call_once(init_partition_specs_once_, [this]() {
+    for (const auto& spec : metadata_->partition_specs) {
+      partition_spec_map_[spec->spec_id()] = spec;
+    }
+  });
+  return partition_spec_map_;
+}
+
+const std::shared_ptr<SortOrder>& Table::sort_order() const {
```
wgtmac: Same as above but return `SortOrder::Unsorted()` when missing.

src/iceberg/table.h
```diff
@@ -35,77 +36,96 @@ class ICEBERG_EXPORT Table {
...
+  const std::unordered_map<std::string, std::string>& properties() const;
 
   /// \brief Return the table's base location
-  virtual const std::string& location() const = 0;
+  const std::string& location() const;
 
-  /// \brief Return the table's current snapshot
-  virtual const std::shared_ptr<Snapshot>& current_snapshot() const = 0;
+  /// \brief Return the table's current snapshot, return null if not found
+  std::shared_ptr<Snapshot> current_snapshot() const;
```
wgtmac:
```suggestion
  const std::shared_ptr<Snapshot>& current_snapshot() const;
```

Every table should have at least one snapshot. I think we can safely return a reference.

src/iceberg/table.cc
```diff
@@ -0,0 +1,130 @@
...
+      schema_ = schema.value();
+    } else {
+      schema_ = kEmptySchema;
+    }
+  }
+  return schema_;
+}
+
+const std::unordered_map<int32_t, std::shared_ptr<Schema>>& Table::schemas() const {
+  std::call_once(init_schemas_once_, [this]() {
```
wgtmac: What about removing the once flag and use `std::shared_ptr<std::unordered_map<int32_t, std::shared_ptr<Schema>>>` instead?

wgtmac (changes requested):

src/iceberg/table_metadata.h
```diff
@@ -23,12 +23,14 @@
 /// Table metadata for Iceberg tables.
 
 #include <memory>
+#include <mutex>
```
```suggestion
```

src/iceberg/table_metadata.h
```diff
@@ -23,12 +23,14 @@
...
 #include <string>
 #include <string_view>
 #include <unordered_map>
 #include <vector>
 
 #include "iceberg/iceberg_export.h"
+#include "iceberg/snapshot.h"
```
```suggestion
```

I think we don't need this because of forward declaration.

src/iceberg/table.h
```diff
@@ -35,77 +36,88 @@ class ICEBERG_EXPORT Table {
...
-  /// \brief Returns a LocationProvider to provide locations for new data files
-  virtual std::unique_ptr<LocationProvider> location_provider() const = 0;
+  const std::vector<SnapshotLogEntry>& history() const;
+
+  /// \brief Returns a FileIO to read and write table data and metadata files
+  const std::shared_ptr<FileIO>& io() const;
+
+ private:
+  const TableIdentifier identifier_;
+  const std::shared_ptr<TableMetadata> metadata_;
```
wgtmac:
```suggestion
  std::shared_ptr<TableMetadata> metadata_;
```

I think this will eventually be non-const when `Table::Refresh()` has been added.

src/iceberg/table.h
```diff
@@ -35,77 +36,88 @@ class ICEBERG_EXPORT Table {
...
+  const std::string metadata_location_;
+  std::shared_ptr<FileIO> io_;
+  std::shared_ptr<Catalog> catalog_;
+
+  mutable std::shared_ptr<std::unordered_map<int32_t, std::shared_ptr<Schema>>>
```
wgtmac, line 118:
```suggestion
  // Cache lazy-initialized maps.
  mutable std::shared_ptr<std::unordered_map<int32_t, std::shared_ptr<Schema>>>
```

wgtmac (approved):
+1

src/iceberg/table.cc
```diff
@@ -0,0 +1,110 @@
...
+
+namespace iceberg {
+
+const std::string& Table::uuid() const { return metadata_->table_uuid; }
+
+Result<std::shared_ptr<Schema>> Table::schema() const { return metadata_->Schema(); }
+
+const std::shared_ptr<std::unordered_map<int32_t, std::shared_ptr<Schema>>>&
+Table::schemas() const {
+  if (!schemas_map_) {
```
zhjwpku, line 35: I guess there can be race conditions here and the same pattern below, no?

src/iceberg/table.h
```diff
@@ -19,13 +19,14 @@
 
 #pragma once
 
-#include <memory>
+#include <mutex>
```
zhjwpku: I don't see any usage of mutex, what's the intent here?

gty404 (approved):

Fokko (approved):

Fokko:
Let's move this in, thanks @lishuxu for working on this, and thanks @lidavidm, @zhjwpku, @wgtmac and @gty404 for the review 🚀

PR #110, opened 2025-05-24
refactor: use nesting enum for DataFile and ManifestFile content

wgtmac:

PR #112, opened 2025-05-27
feat: basic table scan planning

gty404:
Introducing basic scan table data interface


src/iceberg/result.h
```diff
@@ -31,6 +31,7 @@ namespace iceberg {
 enum class ErrorKind {
   kAlreadyExists,
   kCommitStateUnknown,
+  kDataInvalid,
```
lidavidm: Shouldn't this be `InvalidData` to be consistent with the other error codes?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+  /// \return A Result containing the reader or an error.
+  Result<std::unique_ptr<ManifestReader>> CreateManifestReader(
+      const std::string& file_path) const;
+
+  std::unique_ptr<ScanContext> context_;
+  std::shared_ptr<FileIO> file_io_;
+};
+
+/// \brief Represents a task to scan a portion of a data file.
+struct ICEBERG_EXPORT FileScanTask {
```
lidavidm: If the fields are meant to be public (because this is a struct), don't use the trailing `_` on fields

src/iceberg/result.h
gty404: It is reasonable, I will make a modification, thanks.

src/iceberg/table_scan.h
gty404: Yes, I have modified it in the subsequent commits.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+  std::optional<int64_t> snapshot_id_;
+  std::shared_ptr<Expression> filter_;
+};
+
+/// \brief Represents a configured scan operation on a table.
+class ICEBERG_EXPORT TableScan {
+ public:
+  /// \brief Scan context holding snapshot and scan-specific metadata.
+  struct ScanContext {
+    std::shared_ptr<Snapshot> snapshot_;  ///< Snapshot to scan.
```
lidavidm: Same here: these are public fields

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+    std::shared_ptr<Schema> schema_;      ///< Projected schema.
+    std::vector<int32_t> field_ids_;      ///< Field IDs of selected columns.
+    std::shared_ptr<Expression> filter_;  ///< Filter expression to apply.
+  };
+
+  /// \brief Constructs a TableScan with the given context and file I/O.
+  /// \param context Scan context including snapshot, schema, and filter.
+  /// \param file_io File I/O instance for reading manifests and data files.
+  TableScan(std::unique_ptr<ScanContext> context, std::shared_ptr<FileIO> file_io);
```
lidavidm: Is there any reason to have unique_ptr vs just the struct itself?

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,122 @@
...
+
+TableScanBuilder& TableScanBuilder::WithFilter(
+    const std::shared_ptr<Expression>& filter) {
+  filter_ = filter;
+  return *this;
+}
+
+Result<std::unique_ptr<TableScan>> TableScanBuilder::Build() {
+  ICEBERG_ASSIGN_OR_RAISE(auto snapshot, snapshot_id_ ? table_.snapshot(*snapshot_id_)
+                                                      : Result<std::shared_ptr<Snapshot>>(
```
lidavidm: Might as well write out the if/else at this point

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,122 @@
...
+      if (auto it = schemas.find(*snapshot->schema_id); it != schemas.end()) {
+        return it->second;
+      }
+      return InvalidData("Schema {} in snapshot {} is not found", *snapshot->schema_id,
+                         snapshot->snapshot_id);
+    }
+    return table_.schema();
+  };
+
+  ICEBERG_ASSIGN_OR_RAISE(auto schema, ResolveSchema());
```
lidavidm: Why are we indirecting through a lambda?

src/iceberg/type_fwd.h
```diff
@@ -91,6 +91,10 @@ class LocationProvider;
 class SortField;
 class SortOrder;
 class Table;
+class TableScan;
```
wgtmac: What about putting all scan related structures into a separate section for better readability?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+/// \brief Represents a task to scan a portion of a data file.
+struct ICEBERG_EXPORT FileScanTask {
+  std::string file_path;                 ///< Path to the data file.
+  uint64_t start;                        ///< Start byte offset.
+  uint64_t length;                       ///< Length in bytes to scan.
+  std::optional<uint64_t> record_count;  ///< Optional number of records.
+  DataFile::Content file_content;        ///< Type of file content.
+  FileFormatType file_format;            ///< Format of the data file.
+  std::shared_ptr<Schema> schema;        ///< Projected schema.
+  std::vector<int32_t> field_ids;        ///< Field IDs to project.
```
wgtmac: Why do we need this? Isn't it included in the `schema` above?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "iceberg/expression/expression.h"
+#include "iceberg/file_io.h"
```
wgtmac:
```suggestion
```

It seems that we don't need them?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+#include "iceberg/manifest_entry.h"
```
wgtmac, line 25: If the only reason to include `manifest_entry.h` is to use `enum DataFile::Content`, what about refactoring it into `enum DataFileContent` and moving it to `type_fwd.h`?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Builder class for creating TableScan instances.
+class ICEBERG_EXPORT TableScanBuilder {
```
wgtmac, line 81: Do we need an abstraction for common parts of other scan builders? We may need more scan types for incremental, append-only, changelog, etc. If there's not much to abstract, I'm fine to keep it as is for now and refactor them later.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,117 @@
...
+  /// \brief Constructs a TableScanBuilder for the given table.
+  /// \param table Reference to the table to scan.
+  explicit TableScanBuilder(const Table& table);
+
+  /// \brief Sets the snapshot ID to scan.
+  /// \param snapshot_id The ID of the snapshot.
+  /// \return Reference to the builder.
+  TableScanBuilder& WithSnapshotId(int64_t snapshot_id);
+
+  /// \brief Selects columns to include in the scan.
```
wgtmac, line 94: Add a comment to explain what will happen if we don't set any column name?

src/iceberg/table_scan.h
wgtmac: Before defining `FileScanTask`, should we abstract it a little bit? I haven't looked at other implementations yet. From the java impl, there's a hierarchy of `ScanTask` and `FileScanTask` is a combination of `ContentScanTask` and `PartitionScanTask`.

I'm not sure if we want to take the complexity to support the java `SplittableScanTask`, `MergableScanTask`, and `ScanTaskGroup`. We can revisit these later.

src/iceberg/table_scan.h
gty404: DataFileContent was modified to be a nested type in [#110](https://github.com/apache/iceberg-cpp/pull/110/) Will there be other issues if it is reverted back?

src/iceberg/table_scan.h
gty404: The forward declaration of Expression did not appear in type_fwd.h, I will add it later and then delete it here.

src/iceberg/table_scan.h
gty404: There is no special reason, I will changed to using struct

src/iceberg/table_scan.h
gty404: The schema includes all columns, and field_ids includes the projected columns.

src/iceberg/table_scan.cc
gty404: I will modify here, thanks.

src/iceberg/table_scan.h
gty404: Yes, in the Java implementation, the hierarchy of Scan and ScanTask has been defined. I will review this part and try to make it more abstract.

src/iceberg/table_scan.h
wgtmac: I made that change because I like shorter names. I'm fine to keep it as is for now.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,116 @@
...
+  /// Defaults to none which means select all columns
+  /// \param column_names A list of column names.
```
wgtmac:
```suggestion
  /// \param column_names A list of column names. If empty, all columns will be selected.
```

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,116 @@
...
+};
+
+/// \brief Represents a configured scan operation on a table.
+class ICEBERG_EXPORT TableScan {
+ public:
+  /// \brief Scan context holding snapshot and scan-specific metadata.
+  struct ScanContext {
+    std::shared_ptr<Snapshot> snapshot;  ///< Snapshot to scan.
+    std::shared_ptr<Schema> schema;      ///< Table schema.
+    std::vector<int32_t> field_ids;      ///< Field IDs of selected columns.
```
wgtmac: Why the builder uses `column_names_` but here use `field_ids`? Should we replace it with `std::shared_ptr<Schema> projected_schema`?

src/iceberg/type_fwd.h
```diff
@@ -120,11 +127,12 @@ class MetadataUpdate;
 class UpdateRequirement;
 
 class AppendFiles;
-class TableScan;
 
 struct DataFile;
 struct ManifestEntry;
 struct ManifestFile;
+class ManifestReader;
```
wgtmac: I was thinking that `struct`s and `class`s should be put in different sections for better alignment.

src/iceberg/table_scan.h
wgtmac: Some thoughts on this function:
- Option for column names is much easier to use but may not be easy to select nested columns.
- The Java impl also provides a method to accept a projected `Schema` and internally these column names are converted to a `Schema` as well. Eventually we will also need a function to accept `std::shared_ptr<Schema>`.
- Similar to the Java impl, if the `filter` has used some columns as the references but are not selected here, we have to add those to projected schema as well.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,116 @@
...
+  /// \brief Selects columns to include in the scan.
+  /// Defaults to none which means select all columns
+  /// \param column_names A list of column names.
+  /// \return Reference to the builder.
+  TableScanBuilder& WithColumnNames(const std::vector<std::string>& column_names);
+
+  /// \brief Applies a filter expression to the scan.
+  /// \param filter Filter expression to use.
+  /// \return Reference to the builder.
+  TableScanBuilder& WithFilter(const std::shared_ptr<Expression>& filter);
```
wgtmac:
```suggestion
  TableScanBuilder& WithFilter(std::shared_ptr<Expression> filter);
```

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,116 @@
...
+  ///
+  /// Returns a list of file scan tasks if successful.
+  /// \return A Result containing scan tasks or an error.
+  Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles() const;
+
+ private:
+  /// \brief Creates a reader for the manifest list.
+  /// \param file_path Path to the manifest list file.
+  /// \return A Result containing the reader or an error.
+  Result<std::unique_ptr<ManifestListReader>> CreateManifestListReader(
```
wgtmac: It seems that `CreateManifestListReader` and `CreateManifestReader` should be utility functions shared by different scan actions. What about moving them to .cc files without depending on the table scan internal state for now?

src/iceberg/table_scan.h
wgtmac: Isn't it too heavy to add `schema` and `field_ids` to every `FileScanTask` which shares the same context? They can be accessed via the TableScan object. I'd suggest to remove them for now until we are clear why it is needed.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,116 @@
...
+  Result<std::unique_ptr<ManifestReader>> CreateManifestReader(
+      const std::string& file_path) const;
+
+  ScanContext context_;
+  std::shared_ptr<FileIO> file_io_;
+};
+
+/// \brief Represents a task to scan a portion of a data file.
+struct ICEBERG_EXPORT FileScanTask {
+  std::string file_path;                 ///< Path to the data file.
```
wgtmac: Should it be replaced by `DataFile`? We might need its metadata like column stats as well. Then we can remove many fields below including record_count, file_content, etc.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,116 @@
...
+  /// \brief Sets the snapshot ID to scan.
+  /// \param snapshot_id The ID of the snapshot.
+  /// \return Reference to the builder.
+  TableScanBuilder& WithSnapshotId(int64_t snapshot_id);
+
+  /// \brief Selects columns to include in the scan.
+  /// Defaults to none which means select all columns
+  /// \param column_names A list of column names.
+  /// \return Reference to the builder.
+  TableScanBuilder& WithColumnNames(const std::vector<std::string>& column_names);
```
yingcai-cy:
```suggestion
  TableScanBuilder& WithColumnNames(std::vector<std::string> column_names);
```
Pass by value and use move in implementation.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,125 @@
...
+  filter_ = filter;
+  return *this;
+}
+
+Result<std::unique_ptr<TableScan>> TableScanBuilder::Build() {
+  std::shared_ptr<Snapshot> snapshot;
+  if (snapshot_id_) {
+    ICEBERG_ASSIGN_OR_RAISE(snapshot, table_.snapshot(*snapshot_id_));
+  } else {
+    snapshot = table_.current_snapshot();
```
yingcai-cy: Can current_snapshot() return nullptr?

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  std::shared_ptr<Schema> schema;
+  if (snapshot->schema_id) {
+    const auto& schemas = table_.schemas();
+    if (auto it = schemas.find(*snapshot->schema_id); it != schemas.end()) {
+      schema = it->second;
+    } else {
+      return InvalidData("Schema {} in snapshot {} is not found", *snapshot->schema_id,
```
yingcai-cy: nit:I think use InvalidArgument error code is more accurate.

src/iceberg/snapshot.h
```diff
@@ -222,6 +222,7 @@ struct ICEBERG_EXPORT DataOperation {
 /// Snapshots are created by table operations.
 struct ICEBERG_EXPORT Snapshot {
   static constexpr int64_t kInvalidSnapshotId = -1;
+  static constexpr int64_t kInitialSequenceNumber = 0;
```
zhjwpku: There is a TableMetadata::kInitialSequenceNumber, which seems to serve the same purpose here?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+
+  /// \brief The residual expression to apply after scanning the data file.
+  const std::shared_ptr<Expression>& residual() const;
+
+  int64_t size_bytes() const override;
+  int32_t files_count() const override;
+  int64_t estimated_row_count() const override;
+
+ private:
+  std::shared_ptr<DataFile> data_file_;                  ///< Data file metadata.
```
zhjwpku: nit: Perhaps keep the comments on a separate line for consistency?

src/iceberg/manifest_reader.h
```diff
@@ -34,6 +34,7 @@ namespace iceberg {
 /// \brief Read manifest entries from a manifest file.
 class ICEBERG_EXPORT ManifestReader {
  public:
+  virtual ~ManifestReader() = default;
```
wgtmac, line 37: Why do we need to explicitly define it?

src/iceberg/manifest_reader.h
```diff
@@ -43,10 +44,27 @@ class ICEBERG_EXPORT ManifestReader {
...
 
  private:
   std::unique_ptr<StructLikeReader> reader_;
 };
 
+/// \brief Creates a reader for the manifest list.
+/// \param file_path Path to the manifest list file.
+/// \return A Result containing the reader or an error.
+Result<std::unique_ptr<ManifestListReader>> CreateManifestListReader(
+    const std::string& file_path) {
```
wgtmac: I think this is not enough. At least we need extra parameters like `table_format_version` and `file_io`.

src/iceberg/snapshot.h
wgtmac: I'm not sure which one is the best place to define this. Attribute like sequence number belongs to both TableMetadata and Snapshot. Perhaps we can define a `iceberg/constants.h` to define those common constants?

src/iceberg/manifest_reader.h
wgtmac: BTW, should we use `std::string_view` for path?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+
+#include <string>
+#include <vector>
+
+#include "iceberg/manifest_entry.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+class ICEBERG_EXPORT ScanTask {
```
wgtmac, line 31: Add a docstring for this class?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+  /// \brief The residual expression to apply after scanning the data file.
+  const std::shared_ptr<Expression>& residual() const;
+
+  int64_t size_bytes() const override;
+  int32_t files_count() const override;
+  int64_t estimated_row_count() const override;
+
+ private:
+  std::shared_ptr<DataFile> data_file_;                  ///< Data file metadata.
+  std::vector<std::shared_ptr<DataFile>> delete_files_;  ///< Delete files metadata.
```
wgtmac: Should this be `DeleteFile`?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+  std::optional<int64_t> limit;  ///< Optional limit on the number of rows to scan.
+};
+
+/// \brief Builder class for creating TableScan instances.
+class ICEBERG_EXPORT TableScanBuilder {
+ public:
+  /// \brief Constructs a TableScanBuilder for the given table.
+  /// \param table The table to scan.
+  /// \param table_metadata The metadata of the table to scan.
+  explicit TableScanBuilder(const Table& table,
```
wgtmac: Why do we need both `table` and `table_metadata`?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+
+/// \brief Represents a configured scan operation on a table.
+class ICEBERG_EXPORT TableScan {
+ public:
+  virtual ~TableScan() = default;
+
+  /// \brief Constructs a TableScan with the given context and file I/O.
+  /// \param context Scan context including snapshot, schema, and filter.
+  /// \param file_io File I/O instance for reading manifests and data files.
+  TableScan(TableScanContext context, std::shared_ptr<FileIO> file_io);
```
wgtmac, line 148: What is the reason that `file_io` is not included in the `context`?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+  /// \return Reference to the builder.
+  TableScanBuilder& WithLimit(std::optional<int64_t> limit);
+
+  /// \brief Builds and returns a TableScan instance.
+  /// \return A Result containing the TableScan or an error.
+  Result<std::unique_ptr<TableScan>> Build();
+
+ private:
+  const Table& table_;  ///< Reference to the table to scan.
+  std::vector<std::string> column_names_;
```
wgtmac, line 133: Why there is no `std::shared_ptr<Schema>` for `WithSchema`? Same question for filter.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+ public:
+  DataScan(TableScanContext context, std::shared_ptr<FileIO> file_io);
+
+  /// \brief Plans the scan tasks by resolving manifests and data files.
+  /// \return A Result containing scan tasks or an error.
+  Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles() const override;
+
+ private:
+  // Use indexed data structures for efficient lookups
+  struct DeleteFileIndex {
```
wgtmac: Should we just move them to source file since they are private and static?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,204 @@
...
+  /// \brief Plans the scan tasks by resolving manifests and data files.
+  /// \return A Result containing scan tasks or an error.
+  virtual Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles() const = 0;
+
+ protected:
+  const TableScanContext context_;
+  std::shared_ptr<FileIO> file_io_;
+};
+
+class ICEBERG_EXPORT DataScan : public TableScan {
```
lishuxu: Add a docstring for this class?

src/iceberg/snapshot.h
gty404: I didn't notice TableMetadata::kInitialSequenceNumber, maybe I should use it directly instead of redefining it. In iceberg-java, kInitialSequenceNumber is defined in TableMetadata.

src/iceberg/manifest_reader.h
gty404: Do you need to provide a ManifestListReaderBuilder/ManifestReaderBuilder?

src/iceberg/manifest_reader.h
gty404: The IDE prompted it, so I made it up

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,273 @@
...
+        return InvalidArgument("Schema {} in snapshot {} is not found",
+                               *snapshot->schema_id, snapshot->snapshot_id);
+      }
+    } else {
+      schema = table_.schema();
+    }
+
+    // TODO(gty404): collect touched columns from filter expression
+    std::vector<SchemaField> projected_fields;
+    projected_fields.reserve(column_names_.size());
```
lishuxu: What should be the behavior when column_names_ is empty?

src/iceberg/table_scan.h
gty404: We have not explicitly defined DeleteFile, and from the definition of DataFile, it seems to include DeleteFile?

src/iceberg/table_scan.h
gty404: I cannot get the table_metadata from the Table interface, and the table interface cannot get some fields in the table_metadata.

src/iceberg/table_scan.h
gty404: I stored them directly in the TableScanContext

src/iceberg/table_scan.h
gty404: I just want to make a distinction between table-related and non-table data.

src/iceberg/table_scan.cc
gty404: If empty, all columns will be selected

src/iceberg/manifest_reader.h
wgtmac: I don't know yet. It depends on how we will use them. BTW @dongxiao1198 will work on manifest reading.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,210 @@
...
+  virtual int64_t size_bytes() const = 0;
+
+  /// \brief The number of files that should be read by this scan task.
+  virtual int32_t files_count() const = 0;
+
+  /// \brief The number of rows that should be read by this scan task.
+  virtual int64_t estimated_row_count() const = 0;
+};
+
+/// \brief Represents a task to scan a portion of a data file.
```
wgtmac:
```suggestion
/// \brief Task representing a data file and its corresponding delete files.
```

Copied from `iceberg-python`

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,210 @@
...
+
+#include <string>
+#include <vector>
+
+#include "iceberg/manifest_entry.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Represents a task to scan a table or a portion of it.
```
wgtmac:
```suggestion
/// \brief An abstract scan task.
```

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,210 @@
...
+
+  /// \brief The number of files that should be read by this scan task.
+  virtual int32_t files_count() const = 0;
+
+  /// \brief The number of rows that should be read by this scan task.
+  virtual int64_t estimated_row_count() const = 0;
+};
+
+/// \brief Represents a task to scan a portion of a data file.
+class ICEBERG_EXPORT FileScanTask : public ScanTask {
```
wgtmac, line 46: Some thoughts about `FileScanTask`:

1. Should we remove `ScanTask` abstraction above? If we remove the abstraction, we can directly use aggregate initialization to create a task. Otherwise we may need to expand the constructor every time a new parameter is required.
2. If we do (1) above, is it possible also to make it a simple struct by removing all functions (as they are all trivial accessors).
3. Should we add fields  (a.k.a. spec and partition_value) from Java `PartitionScanTask` to support partitioning? We can add them later but a TODO comment is desirable.
4. Should we combine `start` and `length`, and wrap them by `std::optional`? I believe they are not required at all times.

src/iceberg/table_scan.h
wgtmac: `Table` is just a wrapper around `TableMetadata`. I think we could either (1) expose `TableMetadata` via `Table` (and thus remove `table_metadata` from ctor), or (2) remove `table` and add `file_io` to ctor. I think passing a reference of table may cause lifetime issue. I prefer (2) here. WDYT?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,210 @@
...
+
+  /// \brief Selects columns to include in the scan.
+  /// \param column_names A list of column names. If empty, all columns will be selected.
+  /// \return Reference to the builder.
+  TableScanBuilder& WithColumnNames(std::vector<std::string> column_names);
+
+  /// \brief Sets the schema to use for the scan.
+  /// \param schema The schema to use.
+  /// \return Reference to the builder.
+  TableScanBuilder& WithSchema(std::shared_ptr<Schema> schema);
```
wgtmac: I think we don't need this. We just need schema of a specific snapshot id which can be obtained via table_metadata. Did I miss something?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,210 @@
...
+
+ private:
+  /// \brief Reference to the table to scan.
+  const Table& table_;
+  /// \brief column names to project in the scan.
+  std::vector<std::string> column_names_;
+  /// \brief snapshot ID to scan, if specified.
+  std::optional<int64_t> snapshot_id_;
+  /// \brief Context for the scan, including snapshot, schema, and filter.
+  TableScanContext context_;
```
wgtmac, line 137: In the Java version of `TableScanContext`, `column_names_` and `snapshot_id_` are also stored in it. Should we follow the same pattern? If we do this, it seems that `TableScanBuilder` is indeed a `TableScanContextBuilder`.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+  /// \brief Find delete files that match the sequence number of a data entry.
+  std::vector<ManifestEntry*> FindRelevantEntries(const ManifestEntry& data_entry) const {
+    std::vector<ManifestEntry*> relevant_deletes;
+
+    // Use lower_bound for efficient range search
+    auto data_sequence_number =
+        data_entry.sequence_number.value_or(TableMetadata::kInitialSequenceNumber);
+    for (auto it = sequence_index.lower_bound(data_sequence_number);
+         it != sequence_index.end(); ++it) {
+      // Additional filtering logic here
```
wgtmac: What is the additional filtering logic? Did you mean to further check if the delete files can be filtered?

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+
+namespace {
+/// \brief Use indexed data structures for efficient lookups
+struct DeleteFileIndex {
+  /// \brief Index by sequence number for quick filtering
+  std::multimap<int64_t, ManifestEntry*> sequence_index;
+
+  /// \brief Build the index from a list of manifest entries.
+  void BuildIndex(const std::vector<std::unique_ptr<ManifestEntry>>& entries) {
+    sequence_index.clear();
```
wgtmac: Is `DeleteFileIndex` supposed to be reused? Does it make sense to make `BuildIndex` private and call it from the ctor?

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+
+const std::vector<std::shared_ptr<DataFile>>& FileScanTask::delete_files() const {
+  return delete_files_;
+}
+
+int64_t FileScanTask::start() const { return start_; }
+
+int64_t FileScanTask::length() const { return length_; }
+
+int64_t FileScanTask::size_bytes() const {
```
```suggestion
int64_t FileScanTask::SizeBytes() const {
```

This is not trivial.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+    sizeInBytes += delete_file->file_size_in_bytes;
+  });
+  return sizeInBytes;
+}
+
+int32_t FileScanTask::files_count() const {
+  return static_cast<int32_t>(delete_files_.size() + 1);
+}
+
+int64_t FileScanTask::estimated_row_count() const {
```
wgtmac: ditto

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+
+int64_t FileScanTask::size_bytes() const {
+  int64_t sizeInBytes = length_;
+  std::ranges::for_each(delete_files_, [&sizeInBytes](const auto& delete_file) {
+    sizeInBytes += delete_file->file_size_in_bytes;
+  });
+  return sizeInBytes;
+}
+
+int32_t FileScanTask::files_count() const {
```
wgtmac: I'm not sure if we need to rename it to `FilesCount()`. @lidavidm suggestion?

src/iceberg/table_scan.h
gty404: TableMetadata should not be exposed through Table, I'll remove table then.

src/iceberg/table_scan.h
gty404: This is used to specify the projected schema without specifying the column name, and I have modified it to WithProjectedSchema.

src/iceberg/table_scan.cc
gty404: DataFiles only need to retain DeleteFiles with a sequence greater than their own?

src/iceberg/table_scan.h
gty404: 1. I initially expected it to just be a struct, but since the previous comments suggested doing an abstraction, I referred to the design in iceberg-java/iceberg-python.
3. Partition spec and value can be obtained from DataFile and Snapshot, and we can add these interfaces when needed for subsequent PR
4. Sure, I will modify it to optional, thanks.

src/iceberg/table_scan.h
gty404: I originally expected that TableScanContext would be context information retained after converting various input parameters, and that what was no longer needed in the subsequent file scanning process would be removed.

src/iceberg/table_scan.h
wgtmac: > Partition spec and value can be obtained from DataFile and Snapshot

That's a good point

src/iceberg/table_scan.h
wgtmac: Do you mean you will remove `TableScanContext`? It depends on whether you will reuse it to plan files for metadata tables.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,210 @@
...
+  /// \return A reference to the TableScanContext.
+  const TableScanContext& context() const;
+
+  /// \brief Returns the file I/O instance used for reading manifests and data files.
+  /// \return A shared pointer to the FileIO instance.
+  const std::shared_ptr<FileIO>& io() const;
+
+  /// \brief Plans the scan tasks by resolving manifests and data files.
+  /// \return A Result containing scan tasks or an error.
+  virtual Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles() const = 0;
```
wgtmac, line 168:
```suggestion
  virtual Result<std::vector<std::shared_ptr<ScanTask>>> PlanTasks() const = 0;
```

I remember that @lishuxu has commented to rename it to `PlanTasks`. Do we need to modify the signature as above?

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,210 @@
...
+
+ protected:
+  /// \brief context for the scan, including snapshot, schema, and filter.
+  const TableScanContext context_;
+  /// \brief File I/O instance for reading manifests and data files.
+  std::shared_ptr<FileIO> file_io_;
+};
+
+/// \brief A scan that reads data files and applies delete files to filter rows.
+class ICEBERG_EXPORT DataScan : public TableScan {
```
wgtmac: I'm a little bit confused about the name of Scan and ScanTask across different implementations. Should this be `DataTableScan` which produces `FileScanTask`? For `DataScan`, I think it should produce a group of `DataTask` which contains rows of `FileScanTask`.

Simply put:
Scan -> ScanTask
TableScan -> FileScanTask
DataScan -> DataTask

DataScan inherits TableScan inherits Scan
DataTask inherits FileScanTask inherits ScanTask

WDYT? @gty404 @Fokko

wgtmac: BTW, I think we can constantly evolve this design because APIs can be unstable before the 1.0.0 release.

src/iceberg/table_scan.h
gty404: TableScanContext will be relied on during subsequent plan files

src/iceberg/table_scan.h
gty404: In iceberg-java, the result of planFiles is a one-to-one correspondence between the entire file and FileScanTask. planTask is the result of splitting the result of planFiles. Each scanTask may contain multiple files, or only a part of one file. I am not sure if the split feature is needed now, so I did not use PlanTask.

src/iceberg/table_scan.h
Fokko: At PyIceberg we (tried to) copied the Java structure, but in the end I think it is too much OOP for Python. Maybe good to start small in C++ as well. While we can change APIs until 1.0.0, I think it is important to get this one right pretty early on, since this is the main integration point for query engines.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,294 @@
...
+// implement FileScanTask
+FileScanTask::FileScanTask(std::shared_ptr<DataFile> file,
+                           std::vector<std::shared_ptr<DataFile>> delete_files,
+                           int64_t start, int64_t length,
+                           std::shared_ptr<Expression> residual)
+    : data_file_(std::move(file)),
+      delete_files_(std::move(delete_files)),
+      start_(start),
+      length_(length),
+      residual_(std::move(residual)) {}
```
Fokko: It looks like this `FileScanTask` is inspired by PyIceberg, while I think it might be better to follow Java: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/FileScanTask.java

The name `FileScanTask` implies to me, that it will read a full file, and then the `start` and `end` do not make sense. This is in the case you want to split up the work by row-group, rather than file.

gty404: Yes, the start and end are not required, I will remove them from FileScanTask.

src/iceberg/manifest_reader.h
```diff
@@ -43,10 +44,27 @@ class ICEBERG_EXPORT ManifestReader {
...
 
  private:
   std::unique_ptr<StructLikeReader> reader_;
 };
 
+/// \brief Creates a reader for the manifest list.
+/// \param file_path Path to the manifest list file.
+/// \return A Result containing the reader or an error.
+Result<std::unique_ptr<ManifestListReader>> CreateManifestListReader(
+    const std::string_view& file_path) {
```
mapleFU: Just `std::string_view` is ok. `std::string_view` is just 16B and is lightweight to construct. psABI can pass 16B in registers, so just std::string_view by value is ok

src/iceberg/manifest_reader.h
```diff
@@ -43,10 +44,27 @@ class ICEBERG_EXPORT ManifestReader {
...
+  return NotImplemented("CreateManifestListReader is not implemented yet.");
+}
+
+/// \brief Creates a reader for a manifest file.
+/// \param file_path Path to the manifest file.
+/// \return A Result containing the reader or an error.
+Result<std::unique_ptr<ManifestReader>> CreateManifestReader(
+    const std::string_view& file_path) {
```
mapleFU: ditto

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+      // Additional filtering logic here
+      relevant_deletes.push_back(it->second);
+    }
+
+    return relevant_deletes;
+  }
+
+ private:
+  /// \brief Index by sequence number for quick filtering
+  std::multimap<int64_t, ManifestEntry*> sequence_index;
```
mapleFU:
```suggestion
  std::multimap<int64_t, ManifestEntry*> sequence_index_;
```

For private variables? Besides, why not `std::map<int64_t, std::vector<const ManifestEntry*>>`?

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+
+  /// \brief Find delete files that match the sequence number of a data entry.
+  std::vector<ManifestEntry*> FindRelevantEntries(const ManifestEntry& data_entry) const {
+    std::vector<ManifestEntry*> relevant_deletes;
+
+    // Use lower_bound for efficient range search
+    auto data_sequence_number =
+        data_entry.sequence_number.value_or(TableMetadata::kInitialSequenceNumber);
+    for (auto it = sequence_index.lower_bound(data_sequence_number);
+         it != sequence_index.end(); ++it) {
```
mapleFU: Is this incorrect? Since it find the lowerbound and traverse all the sequence numbers above `data_sequence_number`

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+TableScanBuilder::TableScanBuilder(std::shared_ptr<TableMetadata> table_metadata,
+                                   std::shared_ptr<FileIO> file_io)
+    : file_io_(std::move(file_io)) {
+  context_.table_metadata = std::move(table_metadata);
+}
+
+TableScanBuilder& TableScanBuilder::WithColumnNames(
+    std::vector<std::string> column_names) {
+  column_names_.reserve(column_names.size());
+  column_names_ = std::move(column_names);
```
mapleFU: We don't need reserve before move. `std::move` usally exchange memory between `column_names_` and `column_names`.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+
+    const auto& schemas = table_metadata->schemas;
+    const auto it = std::ranges::find_if(schemas, [&schema_id](const auto& schema) {
+      return schema->schema_id() == *schema_id;
+    });
+    if (it == schemas.end()) {
+      return InvalidArgument("Schema {} in snapshot {} is not found",
+                             *snapshot->schema_id, snapshot->snapshot_id);
+    }
+    auto schema = *it;
```
mapleFU:
```suggestion
    const auto& schema = *it;
```

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,286 @@
...
+
+  // TODO(gty404): build residual expression from filter
+  std::shared_ptr<Expression> residual;
+  std::vector<std::shared_ptr<FileScanTask>> tasks;
+  DeleteFileIndex delete_file_index(positional_delete_entries);
+  for (const auto& data_entry : data_entries) {
+    auto matched_deletes = GetMatchedDeletes(*data_entry, delete_file_index);
+    const auto& data_file = data_entry->data_file;
+    tasks.emplace_back(std::make_shared<FileScanTask>(
+        data_file, std::move(matched_deletes), std::move(residual)));
```
mapleFU:
```suggestion
        data_file, std::move(matched_deletes), residual));
```

Seems that we cannot move `residual` because it's used multiple times in the loop?

src/iceberg/table_scan.cc
gty404: Yes, the meaning here is to find all DeleteFiles corresponding to this DataFile. Only those with a sequence number higher than the DataFile need to be read.

src/iceberg/table_scan.cc
gty404: It is for the convenience of searching by sequence as mentioned earlier.

src/iceberg/table_scan.cc
gty404: Yes, this has not been implemented yet, and it is expected that this residual will be shared among all tasks.

src/iceberg/table_scan.h
```diff
@@ -0,0 +1,200 @@
...
```
wgtmac:
```suggestion
class ICEBERG_EXPORT DataTableScan : public TableScan {
```

I think `DataScan` is ambiguous and people may assume that it will return rows of data instead of the file list. `DataTable` is better to differentiate from `MetadataTable`.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+#include "iceberg/schema_field.h"
+#include "iceberg/snapshot.h"
+#include "iceberg/table_metadata.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
+/// \brief Use indexed data structures for efficient lookups
+class DeleteFileIndex {
```
mapleFU: In intuition this can build once ( push all element and sort one), then when querying, return a slice for it. But this also looks good to me

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+      residual_(std::move(residual)) {}
+
+const std::shared_ptr<DataFile>& FileScanTask::data_file() const { return data_file_; }
+
+const std::vector<std::shared_ptr<DataFile>>& FileScanTask::delete_files() const {
+  return delete_files_;
+}
+
+int64_t FileScanTask::SizeBytes() const {
+  int64_t sizeInBytes = data_file_->file_size_in_bytes;
```
mapleFU:
```suggestion
  int64_t size_in_bytes = data_file_->file_size_in_bytes;
```

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+
+Result<std::unique_ptr<TableScan>> TableScanBuilder::Build() {
+  const auto& table_metadata = context_.table_metadata;
+  auto snapshot_id = snapshot_id_ ? snapshot_id_ : table_metadata->current_snapshot_id;
+  if (!snapshot_id) {
+    return InvalidArgument("No snapshot ID specified for table {}",
+                           table_metadata->table_uuid);
+  }
+  auto iter = std::ranges::find_if(table_metadata->snapshots,
+                                   [&snapshot_id](const auto& snapshot) {
```
mapleFU: Use the way below to eval `*snapshot_id` only once

```suggestion
                                   [id = *snapshot_id](const auto& snapshot) {
```

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+    const auto& snapshot = context_.snapshot;
+    auto schema_id =
+        snapshot->schema_id ? snapshot->schema_id : table_metadata->current_schema_id;
+    if (!schema_id) {
+      return InvalidArgument("No schema ID found in snapshot {} for table {}",
+                             snapshot->snapshot_id, table_metadata->table_uuid);
+    }
+
+    const auto& schemas = table_metadata->schemas;
+    const auto it = std::ranges::find_if(schemas, [&schema_id](const auto& schema) {
```
mapleFU: ditto

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+      context_.projected_schema = schema;
+    } else {
+      // TODO(gty404): collect touched columns from filter expression
+      std::vector<SchemaField> projected_fields;
+      projected_fields.reserve(column_names_.size());
+      for (const auto& column_name : column_names_) {
+        // TODO(gty404): support case-insensitive column names
+        auto field_opt = schema->GetFieldByName(column_name);
+        if (!field_opt) {
+          return InvalidArgument("Column {} not found in schema", column_name);
```
mapleFU: should we add schema to output?

mapleFU (approved):

src/iceberg/table.cc
```diff
@@ -107,4 +108,8 @@ const std::vector<SnapshotLogEntry>& Table::history() const {
 
 const std::shared_ptr<FileIO>& Table::io() const { return io_; }
 
+std::unique_ptr<TableScanBuilder> Table::NewScan() const {
+  return std::make_unique<TableScanBuilder>(metadata_, io_);
```
Fokko, line 112: How about passing in the `Table` instead? That has all the metadata, and also the `io`

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
+/// \brief Use indexed data structures for efficient lookups
+class DeleteFileIndex {
+ public:
+  /// \brief Build the index from a list of manifest entries.
+  explicit DeleteFileIndex(const std::vector<std::unique_ptr<ManifestEntry>>& entries) {
```
Fokko: Do we want to add this right away, or defer this to a later PR? Previously at PyIceberg we threw an exception when we encountered a positional or equality delete.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+  const auto relevant_entries = delete_file_index.FindRelevantEntries(data_entry);
+  std::vector<std::shared_ptr<DataFile>> matched_deletes;
+  if (relevant_entries.empty()) {
+    return matched_deletes;
+  }
+
+  matched_deletes.reserve(relevant_entries.size());
+  for (const auto& delete_entry : relevant_entries) {
+    // TODO(gty404): check if the delete entry contains the data entry's file path
+    matched_deletes.emplace_back(delete_entry->data_file);
```
Fokko: Without this filter, it will likely explode the number of relevant entries for each of the data files.

src/iceberg/table_scan.cc
Fokko: This differs per equality and positional deletes. I think there is a pretty good overview here: https://iceberg.apache.org/spec/#scan-planning

src/iceberg/table_scan.cc
Fokko: Higher or equal for positional deletes: https://iceberg.apache.org/spec/#scan-planning

src/iceberg/table_scan.cc
gty404: Sure, it will cause this issue. I will implement the filter in the later PR, thanks.

src/iceberg/table_scan.cc
gty404: In the PyIceberg [DataScan](https://github.com/apache/iceberg-python/blob/045dd10a45cbf0b82b095b5bde47f4731a3bf88b/pyiceberg/table/__init__.py#L1926), I see that an exception is only thrown when there is an equality delete.

src/iceberg/table.cc
gty404: If I pass "Table" to the "TableScanBuilder," I cannot pass it further to "DataTableScan," as "Table" can only be passed by reference to the "TableScanBuilder."

wgtmac: @Fokko There are some (outdated) comments on this: https://github.com/apache/iceberg-cpp/pull/112#discussion_r2173635835

src/iceberg/table_scan.cc
wgtmac: I think for the MVP version we don't need to support delete files. Perhaps we should remove it from the current PR. In the future we can add them back with some good test cases.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+}
+
+int64_t FileScanTask::EstimatedRowCount() const {
+  if (data_file_->file_size_in_bytes == 0) {
+    return 0;
+  }
+  const auto size_in_bytes = data_file_->file_size_in_bytes;
+  const double scannedFileFraction =
+      static_cast<double>(size_in_bytes) / data_file_->file_size_in_bytes;
+  return static_cast<int64_t>(scannedFileFraction * data_file_->record_count);
```
zhjwpku: I don't understand the math here, shouldn't the result be `data_file_->record_count - sum(delete_file.record_count)`?

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,285 @@
...
+  const auto& table_metadata = context_.table_metadata;
+  auto snapshot_id = snapshot_id_ ? snapshot_id_ : table_metadata->current_snapshot_id;
+  if (!snapshot_id) {
+    return InvalidArgument("No snapshot ID specified for table {}",
+                           table_metadata->table_uuid);
+  }
+  auto iter = std::ranges::find_if(
+      table_metadata->snapshots,
+      [id = *snapshot_id](const auto& snapshot) { return snapshot->snapshot_id == id; });
+  if (iter == table_metadata->snapshots.end() || *iter == nullptr) {
```
zhjwpku: Why can iter be nullptr?

```
  /// A list of valid snapshots
  std::vector<std::shared_ptr<Snapshot>> snapshots;
```

src/iceberg/table_scan.cc
gty404: Previously, FileScanTask might correspond to a part of DataFile's data, so we used the ratio of the data read to estimate the number of records. Now, there is indeed a problem with this approach, so I have made the necessary modifications.

src/iceberg/table_scan.cc
gty404: Alright, I have removed the logic related to deleting files from the current PR.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,192 @@
...
+#include "iceberg/schema_field.h"
+#include "iceberg/snapshot.h"
+#include "iceberg/table_metadata.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+// implement FileScanTask
+FileScanTask::FileScanTask(std::shared_ptr<DataFile> file)
+    : data_file_(std::move(file)) {}
```
wgtmac, line 38:
```suggestion
FileScanTask::FileScanTask(std::shared_ptr<DataFile> data_file)
    : data_file_(std::move(data_file)) {}
```

nit: I think eventually we need to rename it to `data_file` once we will add delete files.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,192 @@
...
+
+TableScanBuilder::TableScanBuilder(std::shared_ptr<TableMetadata> table_metadata,
+                                   std::shared_ptr<FileIO> file_io)
+    : file_io_(std::move(file_io)) {
+  context_.table_metadata = std::move(table_metadata);
+}
+
+TableScanBuilder& TableScanBuilder::WithColumnNames(
+    std::vector<std::string> column_names) {
+  column_names_ = std::move(column_names);
```
wgtmac, line 56: nit: make sure `context_.projected_schema` is not set.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,192 @@
...
+  return *this;
+}
+
+TableScanBuilder& TableScanBuilder::WithProjectedSchema(std::shared_ptr<Schema> schema) {
+  context_.projected_schema = std::move(schema);
```
wgtmac, line 61: nit: make sure `column_names_` is not set.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,192 @@
...
+}
+
+Result<std::unique_ptr<TableScan>> TableScanBuilder::Build() {
+  const auto& table_metadata = context_.table_metadata;
+  auto snapshot_id = snapshot_id_ ? snapshot_id_ : table_metadata->current_snapshot_id;
+  if (!snapshot_id) {
+    return InvalidArgument("No snapshot ID specified for table {}",
+                           table_metadata->table_uuid);
+  }
+  auto iter = std::ranges::find_if(
```
wgtmac, line 97: nit: add `Result<std::shared_ptr<Snapshot>> TableMetadata::Snapshot(int64_t snapshot_id) const` and move the logic below to it.

src/iceberg/table_scan.cc
```diff
@@ -0,0 +1,192 @@
...
+    const auto& snapshot = context_.snapshot;
+    auto schema_id =
+        snapshot->schema_id ? snapshot->schema_id : table_metadata->current_schema_id;
+    if (!schema_id) {
+      return InvalidArgument("No schema ID found in snapshot {} for table {}",
+                             snapshot->snapshot_id, table_metadata->table_uuid);
+    }
+
+    const auto& schemas = table_metadata->schemas;
+    const auto it = std::ranges::find_if(schemas, [id = *schema_id](const auto& schema) {
```
wgtmac, line 115: ditto, we can add `TableMetadata::Schema(int64_t schema_id)` for this.

src/iceberg/table_scan.cc
gty404: In a conflict situation, it's not very convenient to throw an exception here, I want to put the check in the build.

PR #113, opened 2025-05-28
feat: implement avro file reader

wgtmac:
- Refactor a little bit of Reader and ReaderFactory interfaces.
- Implement the skeleton of Avro reader to read data into ArrowArray.
- `AppendDatumToBuilder` is not implemented yet.

### DIFF
--- a/src/iceberg/arrow/arrow_fs_file_io.h
+++ b/src/iceberg/arrow/arrow_fs_file_io.h
@@ -46,6 +46,9 @@ class ICEBERG_BUNDLE_EXPORT ArrowFileSystemFileIO : public FileIO {
   /// \brief Delete a file at the given location.
   Status DeleteFile(const std::string& file_location) override;
 
+  /// \brief Get the Arrow file system.
+  const std::shared_ptr<::arrow::fs::FileSystem>& fs() const { return arrow_fs_; }
+
  private:
   std::shared_ptr<::arrow::fs::FileSystem> arrow_fs_;
 };
--- a/src/iceberg/avro/avro_data_util.cc
+++ b/src/iceberg/avro/avro_data_util.cc
@@ -17,31 +17,16 @@
  * under the License.
  */
 
-#pragma once
-
-#include <string>
-
-#include "iceberg/avro.h"
-#include "iceberg/file_reader.h"
-#include "iceberg/iceberg_bundle_export.h"
+#include "iceberg/avro/avro_data_util_internal.h"
 
 namespace iceberg::avro {
 
-class ICEBERG_BUNDLE_EXPORT DemoAvro : public Avro {
- public:
-  DemoAvro() = default;
-  ~DemoAvro() override = default;
-  std::string print() const override;
-};
-
-class ICEBERG_BUNDLE_EXPORT DemoAvroReader : public Reader {
- public:
-  DemoAvroReader() = default;
-  ~DemoAvroReader() override = default;
-  Status Open(const ReaderOptions& options) override;
-  Status Close() override;
-  Result<Data> Next() override;
-  DataLayout data_layout() const override;
-};
+Status AppendDatumToBuilder(const ::avro::NodePtr& avro_node,
+                            const ::avro::GenericDatum& avro_datum,
+                            const SchemaProjection& projection,
+                            const Schema& arrow_schema,
+                            ::arrow::ArrayBuilder* array_builder) {
+  return NotImplemented("AppendDatumToBuilder is not yet implemented");
+}
 
 }  // namespace iceberg::avro
--- a/src/iceberg/avro/avro_reader.cc
+++ b/src/iceberg/avro/avro_reader.cc
@@ -1,0 +1,222 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/avro/avro_reader.h"
+
+#include <memory>
+
+#include <arrow/array/builder_base.h>
+#include <arrow/c/bridge.h>
+#include <arrow/filesystem/filesystem.h>
+#include <arrow/record_batch.h>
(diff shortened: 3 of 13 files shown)


src/iceberg/avro/avro_reader.cc
```diff
@@ -0,0 +1,222 @@
...
+    if (!import_result.ok()) {
+      return InvalidSchema("Failed to import the arrow schema: {}",
+                           import_result.status().message());
+    }
+    context_->arrow_schema_ = import_result.MoveValueUnsafe();
+
+    auto arrow_struct_type =
+        std::make_shared<::arrow::StructType>(context_->arrow_schema_->fields());
+    auto builder_result = ::arrow::MakeBuilder(arrow_struct_type);
+    if (!builder_result.ok()) {
```
lidavidm, line 168: Not required for this PR but maybe we should consider a way to convert between Arrow and Iceberg Result?

lidavidm (approved):

src/iceberg/avro/avro_data_util.cc
```diff
@@ -0,0 +1,32 @@
...
+#include "iceberg/avro/avro_data_util_internal.h"
+
+namespace iceberg::avro {
+
+Status AppendDatumToBuilder(const ::avro::NodePtr& avro_node,
+                            const ::avro::GenericDatum& avro_datum,
+                            const SchemaProjection& projection,
+                            const Schema& arrow_schema,
+                            ::arrow::ArrayBuilder* array_builder) {
+  return NotImplemented("AppendDatumToBuilder is not yet implemented");
```
mapleFU, line 29: Would this implement in another patch?

src/iceberg/avro/avro_reader.cc
```diff
@@ -0,0 +1,222 @@
...
+      }
+      ICEBERG_RETURN_UNEXPECTED(
+          AppendDatumToBuilder(reader_->readerSchema().root(), *context_->datum_,
+                               projection_, *read_schema_, context_->builder_.get()));
+    }
+
+    return ConvertBuilderToArrowArray();
+  }
+
+  Status Close() {
```
mapleFU, line 142: should we mark close and checkClosed here?

src/iceberg/avro/avro_reader.cc
```diff
@@ -0,0 +1,222 @@
...
+      return {};
+    }
+
+    auto builder_result = context_->builder_->Finish();
+    if (!builder_result.ok()) {
+      return InvalidArrowData("Failed to finish the arrow array builder: {}",
+                              builder_result.status().message());
+    }
+
+    auto array = builder_result.MoveValueUnsafe();
```
mapleFU, line 188:
```suggestion
    auto array = std::move(builder_result).MoveValueUnsafe();
```

?

src/iceberg/avro/avro_reader.cc
wgtmac: Yes, I've thought about this.

src/iceberg/avro/avro_data_util.cc
wgtmac: Yes

src/iceberg/avro/avro_reader.cc
wgtmac: I think we don't need this because the state is cleared.

src/iceberg/avro/avro_reader.cc
wgtmac: Why need this move?

mapleFU: Oh my, just MoveValueUnsafe is ok

src/iceberg/expected.h
```diff
@@ -122,7 +122,7 @@ inline constexpr bool is_error_type_valid_v = is_error_type_valid<T>::value;
 }  // namespace expected_detail
 
 template <class E>
-class ICEBERG_EXPORT [[nodiscard]] unexpected {
```
wgtmac, line 125: I have to remove these to fix https://github.com/apache/iceberg-cpp/actions/runs/15342734208/job/43172206905. I think adding `ICEBERG_EXPORT` here makes all `Result<E>` defined in the `libiceberg-bundle` as  import symbols and the linker cannot find them anywhere. @zhjwpku @lidavidm

lidavidm: Hmm, if it works it works. I really don't get how the Windows model works.

src/iceberg/avro/avro_reader.cc
```diff
@@ -0,0 +1,222 @@
...
+// 3. collect basic reader metrics
+class AvroBatchReader::Impl {
+ public:
+  Status Open(const ReaderOptions& options) {
+    batch_size_ = options.batch_size;
+    read_schema_ = options.projection;
+
+    // Open the input stream and adapt to the avro interface.
+    // TODO(gangwu): make this configurable
+    constexpr int64_t kDefaultBufferSize = 1024 * 1024;
```
Fokko, line 87: Looking at the TODO, we can also defer this to a later PR. Implementations aim to create manifest files of ~8MB, and performance-wise wise it is best to read them all the way directly. The manifest list can be unbounded (theoretically).

src/iceberg/avro/avro_reader.cc
```diff
@@ -0,0 +1,222 @@
...
+    // Create a base reader without setting reader schema to enable projection.
+    auto base_reader =
+        std::make_unique<::avro::DataFileReaderBase>(std::move(input_stream));
+    const ::avro::ValidSchema& file_schema = base_reader->dataSchema();
+
+    // Validate field ids in the file schema.
+    HasIdVisitor has_id_visitor;
+    ICEBERG_RETURN_UNEXPECTED(has_id_visitor.Visit(file_schema));
+    if (has_id_visitor.HasNoIds()) {
+      // TODO(gangwu): support applying field-ids based on name mapping
```
Fokko, line 100: Keep in mind that name-mapping only applies to data-files

wgtmac: Yes, I think we just need to pass `NameMapping` object (if available) via ReaderOptions to read data files. For manifest and manifest list files, it should error out when field ids are missing.

src/iceberg/avro/avro_reader.cc
wgtmac: Sounds good to me. I just blindly chose a default value for now.

src/iceberg/arrow/arrow_fs_file_io.h
```diff
@@ -46,6 +46,9 @@ class ICEBERG_BUNDLE_EXPORT ArrowFileSystemFileIO : public FileIO {
   /// \brief Delete a file at the given location.
   Status DeleteFile(const std::string& file_location) override;
 
+  /// \brief Get the Arrow file system.
+  const std::shared_ptr<::arrow::fs::FileSystem>& fs() const { return arrow_fs_; }
```
wgtmac, line 50: @Fokko We have two libraries: `libiceberg` and `libiceberg-bundle`:
- libiceberg: it only uses interfaces of `FileIO` and `FileReader`, downstream projects should provide their own implementations.
- libiceberg-bundle: it uses `ArrowFileSystemFileIO` as the implementation and both Avro and Parquet reader implementations assume that arrow::FileSystem is available.

PR #117, opened 2025-06-04
feat: implement Primitive type Literal

mapleFU:
Before implement expression, Primitive should be implemented. The spec follows: https://iceberg.apache.org/spec/#schemas-and-data-types

In this patch, only few types of Literals is supported, including int, long, float, double, boolean, string and binary.

TODO:
* [x] Implement detail logics about comparing 
* [x] Implement detail logics about castTo
* [x] ToString and printing logics
* [x] Tests

Would in other patch:
* More primitive type basic supports ( maybe without decimal now )
* Implement detail logics about serialize/deserialize


src/iceberg/datum.h
```diff
@@ -0,0 +1,108 @@
...
+  std::strong_ordering operator<=>(const AboveMax&) const = default;
+};
+
+using PrimitiveLiteralValue =
+    std::variant<bool,                  // for boolean
+                 int32_t,               // for int, date
+                 int64_t,               // for long, timestamp, timestamp_tz, time
+                 float,                 // for float
+                 double,                // for double
+                 std::string,           // for string
```
mapleFU: C++ string not requires utf8, so I wonder whether `std::vector<uint8_t>` with string is possible

src/iceberg/datum.h
```diff
@@ -0,0 +1,108 @@
...
+                 std::vector<uint8_t>,  // for binary, fixed, decimal and uuid
```
mapleFU: uuid and decimal are fixed 16 bytes, maybe tools needs added

src/iceberg/datum.h
```diff
@@ -0,0 +1,108 @@
...
+  Result<PrimitiveLiteral> CastTo(
+      const std::shared_ptr<PrimitiveType>& target_type) const;
+
+  std::partial_ordering operator<=>(const PrimitiveLiteral& other) const;
+
+  std::string ToString() const;
+
+ private:
+  PrimitiveLiteralValue value_;
+  std::shared_ptr<PrimitiveType> type_;
```
mapleFU: don't know `std::shared_ptr<PrimitiveType>` to heavy for this type

src/iceberg/datum.h
```diff
@@ -0,0 +1,108 @@
...
+                 int32_t,               // for int, date
+                 int64_t,               // for long, timestamp, timestamp_tz, time
+                 float,                 // for float
+                 double,                // for double
+                 std::string,           // for string
+                 std::vector<uint8_t>,  // for binary, fixed, decimal and uuid
+                 BelowMin, AboveMax>;
+
+/// \brief PrimitiveLiteral is owned literal of a primitive type.
+class PrimitiveLiteral {
```
zhjwpku: Rename to Datum since you are using datum.h/cc as the filename?
Personally I prefer the following:
```
using PrimitiveLiteral = std::variant<bool, ...
class Datum {
    PrimitiveLiteral literal_,
    std::shared_ptr<PrimitiveType>
};
```

src/iceberg/datum.h
zhjwpku: Does it make sense to add a std::array<uint8_t, 16> for decimal and uuid?

src/iceberg/datum.h
zhjwpku: Small String Optimization (SSO) will make PrimitiveLiteralValue 40 bytes, may be use a unique_ptr to own the string?

src/iceberg/datum.h
mapleFU: Looks ok to me

src/iceberg/datum.h
mapleFU: LGTM

lidavidm:
The idea seems reasonable to me.

src/iceberg/datum.h
mapleFU: Just a nit, Literal might be used as vector<Literal>, so more bytes is not critical?

src/iceberg/literal.cc
```diff
@@ -0,0 +1,318 @@
...
+                            static_cast<int>(target_type_id));
+  }
+}
+
+Result<PrimitiveLiteral> PrimitiveLiteral::CastFromDouble(TypeId target_type_id) const {
+  auto double_val = std::get<double>(value_);
+
+  switch (target_type_id) {
+    case TypeId::kFloat: {
+      if (double_val > std::numeric_limits<float>::max()) {
```
mapleFU: this might lost precision, don't know the design considerations here

src/iceberg/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set(ICEBERG_SOURCES
     expression/expression.cc
     file_reader.cc
     json_internal.cc
+    literal.cc
```
wgtmac: Should it be put into the `expression` subdir?

src/iceberg/literal.h
```diff
@@ -0,0 +1,137 @@
...
+    std::strong_ordering operator<=>(const AboveMax&) const = default;
+  };
+
+  using PrimitiveLiteralValue =
+      std::variant<bool,                     // for boolean
+                   int32_t,                  // for int, date
+                   int64_t,                  // for long, timestamp, timestamp_tz, time
+                   float,                    // for float
+                   double,                   // for double
+                   std::string,              // for string
```
wgtmac: Do we need a `string_view` variant?

src/iceberg/datum.h
zhjwpku: I think the data part of vector will not be in Literal's layout?
If we are not going to use something like std::vector<Datum>, I think it's fine with the current design.

mapleFU: Emmm, a tiny wrapper like this should be defined if unique_ptr<string> should be added. I'm also ok for the case

```c++
  struct StringLiteralValue {
    StringLiteralValue(std::string value);
    StringLiteralValue(const StringLiteralValue&);
    StringLiteralValue(StringLiteralValue&&) noexcept;
    const std::string& str() const {
      return *value_;
    }
    bool operator==(const StringLiteralValue&) const;
    std::strong_ordering operator<=>(const StringLiteralValue&) const;
  private:
    std::unique_ptr<std::string> value_;
  };
```

In libc++ version, `sizeof(PrimitiveLiteralValue) == 48`, since `std::vector` is large, it's still 48b after change to `StringLiteralValue`.

After change to below:

```c++
  template <typename T>
  struct LiteralValueWrapper {
    LiteralValueWrapper(T value);
    LiteralValueWrapper(const LiteralValueWrapper&);
    LiteralValueWrapper(LiteralValueWrapper&&) noexcept = default;
    const std::string& str() const {
      return *value_;
    }
    bool operator==(const LiteralValueWrapper<T>&) const;
    std::strong_ordering operator<=>(const LiteralValueWrapper<T>&) const;
  private:
    std::unique_ptr<T> value_;
  };

  using PrimitiveLiteralValue =
      std::variant<bool,                     // for boolean
                   int32_t,                  // for int, date
                   int64_t,                  // for long, timestamp, timestamp_tz, time
                   float,                    // for float
                   double,                   // for double
                   LiteralValueWrapper<std::string>,              // for string
                   LiteralValueWrapper<std::vector<uint8_t>>,     // for binary, fixed
                   std::array<uint8_t, 16>,  // for uuid and decimal
                   BelowMin, AboveMax>;
```

The sizeof `sizeof(PrimitiveLiteralValue) == 48` becomes 40B, just eliminate 8B 😅. I think it's due to C++ doesn't do any niche optimization for std::variant... If the largest element in `PrimitiveLiteralValue` is 8B, `sizeof(PrimitiveLiteralValue)` can be 32B. So I think currently std::string is enough here?

src/iceberg/literal.h
```diff
@@ -0,0 +1,137 @@
...
+  static PrimitiveLiteral Double(double value);
+  static PrimitiveLiteral String(std::string value);
+  static PrimitiveLiteral Binary(std::vector<uint8_t> value);
+
+  /// Create iceberg value from bytes.
+  ///
+  /// See [this spec](https://iceberg.apache.org/spec/#binary-single-value-serialization)
+  /// for reference.
+  static Result<PrimitiveLiteral> Deserialize(std::span<const uint8_t> data);
+  /// Serialize iceberg value to bytes.
```
zhjwpku: nit: add an empty line after a method?

src/iceberg/literal.h
```diff
@@ -0,0 +1,137 @@
...
+  /// Factory methods for primitive types
+  static PrimitiveLiteral Boolean(bool value);
+  static PrimitiveLiteral Int(int32_t value);
+  static PrimitiveLiteral Long(int64_t value);
+  static PrimitiveLiteral Float(float value);
+  static PrimitiveLiteral Double(double value);
+  static PrimitiveLiteral String(std::string value);
+  static PrimitiveLiteral Binary(std::vector<uint8_t> value);
+
+  /// Create iceberg value from bytes.
```
zhjwpku:
```suggestion
  /// Create iceberg Literal from bytes.
```

src/iceberg/literal.h
```diff
@@ -0,0 +1,137 @@
...
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "iceberg/result.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+/// \brief PrimitiveLiteral is owned literal of a primitive type.
```
zhjwpku: I had some trouble understanding this sentence, how about:

```
PrimitiveLiteral is a literal value that is associated with a primitive type.
```

src/iceberg/literal.h
```diff
@@ -0,0 +1,137 @@
...
+class ICEBERG_EXPORT PrimitiveLiteral {
+ private:
+  /// \brief Exception type for values that are below the minimum allowed value for a
+  /// primitive type.
+  ///
+  /// When casting a value to a narrow primitive type, if the value exceeds the maximum of
+  /// dest type, it might be above the maximum allowed value for that type.
```
zhjwpku:
```suggestion
  /// target type, it might be above the maximum allowed value for that type.
```

src/iceberg/literal.h
```diff
@@ -0,0 +1,137 @@
...
+  struct BelowMin {
+    bool operator==(const BelowMin&) const = default;
+    std::strong_ordering operator<=>(const BelowMin&) const = default;
+  };
+
+  /// \brief Exception type for values that are above the maximum allowed value for a
+  /// primitive type.
+  ///
+  /// When casting a value to a narrow primitive type, if the value exceeds the maximum of
+  /// dest type, it might be above the maximum allowed value for that type.
```
zhjwpku:
```suggestion
  /// target type, it might be above the maximum allowed value for that type.
```

src/iceberg/literal.h
```diff
@@ -0,0 +1,137 @@
...
+  ///
+  /// If conversion to a target type is not supported, this method returns an error.
+  ///
+  /// This method may return BelowMin or AboveMax when the target type is not as wide as
+  /// the original type. These values indicate that the containing predicate can be
+  /// simplified. For example, Integer.MAX_VALUE+1 converted to an int will result in
+  /// AboveMax and can simplify a < Integer.MAX_VALUE+1 to always true.
+  ///
+  /// @param target_type A primitive PrimitiveType
+  /// @return A Result containing a literal of the given type or an error if conversion
```
zhjwpku:
```suggestion
  /// \param target_type A primitive PrimitiveType
  /// \return A Result containing a literal of the given type or an error if conversion
```

src/iceberg/datum.h
zhjwpku: Thanks for the analysis, let's keep std::string :)

zhjwpku (approved):
LGTM

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+
+ public:
+  /// Factory methods for primitive types
+  static PrimitiveLiteral Boolean(bool value);
+  static PrimitiveLiteral Int(int32_t value);
+  static PrimitiveLiteral Long(int64_t value);
+  static PrimitiveLiteral Float(float value);
+  static PrimitiveLiteral Double(double value);
+  static PrimitiveLiteral String(std::string value);
+  static PrimitiveLiteral Binary(std::vector<uint8_t> value);
```
yingcai-cy: Can UUIDLiteral and DecimalLiteral also be created via factory methods?

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+                   double,                   // for double
+                   std::string,              // for string
+                   std::vector<uint8_t>,     // for binary, fixed
+                   std::array<uint8_t, 16>,  // for uuid and decimal
+                   BelowMin, AboveMax>;
+
+ public:
+  /// Factory methods for primitive types
+  static PrimitiveLiteral Boolean(bool value);
+  static PrimitiveLiteral Int(int32_t value);
```
yingcai-cy: Not sure if we should treat Timestamp and Date differently from long/int in expressions or elsewhere.

src/iceberg/literal.cc
```diff
@@ -0,0 +1,323 @@
...
+    case TypeId::kDouble:
+    case TypeId::kBoolean:
+    case TypeId::kString:
+    case TypeId::kBinary:
+      break;
+    default:
+      break;
+  }
+
+  return NotSupported("Cast from {} to {} is not implemented", type_->ToString(),
```
yingcai-cy: not implemented -> not supported?

src/iceberg/literal.h
mapleFU: Sure, we can uses more factory here. This patch only supports int, float, string, binary. We can first make this in so we can then write the (1) impl logic for other literals (2) basic implement for predicate expressions.

src/iceberg/literal.h
mapleFU: Yes, we can delay it into future patches.

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+#include <variant>
+#include <vector>
+
+#include "iceberg/result.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+/// \brief PrimitiveLiteral is a literal value that is associated with a primitive type.
+class ICEBERG_EXPORT PrimitiveLiteral {
```
wgtmac: Should we simply use `Literal` instead of `PrimitiveLiteral` which looks lengthy?

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+  static PrimitiveLiteral Float(float value);
+  static PrimitiveLiteral Double(double value);
+  static PrimitiveLiteral String(std::string value);
+  static PrimitiveLiteral Binary(std::vector<uint8_t> value);
+
+  /// Create iceberg literal from bytes.
+  ///
+  /// See [this spec](https://iceberg.apache.org/spec/#binary-single-value-serialization)
+  /// for reference.
+  static Result<PrimitiveLiteral> Deserialize(std::span<const uint8_t> data);
```
wgtmac: Don't we need the type as an input argument?

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+  /// primitive type.
+  ///
+  /// When casting a value to a narrow primitive type, if the value exceeds the maximum of
+  /// target type, it might be above the maximum allowed value for that type.
+  struct AboveMax {
+    bool operator==(const AboveMax&) const = default;
+    std::strong_ordering operator<=>(const AboveMax&) const = default;
+  };
+
+  using PrimitiveLiteralValue =
```
wgtmac:
```suggestion
  using Value =
```

This is equivalent to `Literal::Value` so we don't need to duplicate them.

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+  /// the original type. These values indicate that the containing predicate can be
+  /// simplified. For example, std::numeric_limits<int>::max()+1 converted to an int will
+  /// result in AboveMax and can simplify a < std::numeric_limits<int>::max()+1 to always
+  /// true.
+  ///
+  /// \param target_type A primitive PrimitiveType
+  /// \return A Result containing a literal of the given type or an error if conversion
+  /// was not valid
+  Result<PrimitiveLiteral> CastTo(
+      const std::shared_ptr<PrimitiveType>& target_type) const;
```
wgtmac:
```suggestion
  Result<PrimitiveLiteral> CastTo(
      const PrimitiveType& target_type) const;
```

std::shared_ptr has a narrower use case

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+
+  /// Compare two PrimitiveLiterals. Both literals must have the same type
+  /// and should not be AboveMax or BelowMin.
+  std::partial_ordering operator<=>(const PrimitiveLiteral& other) const;
+
+  bool isAboveMax() const;
+  bool isBelowMin() const;
```
wgtmac:
```suggestion
  bool IsAboveMax() const;
  bool IsBelowMin() const;
```

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  using PrimitiveLiteralValue =
+      std::variant<bool,                     // for boolean
+                   int32_t,                  // for int, date
+                   int64_t,                  // for long, timestamp, timestamp_tz, time
+                   float,                    // for float
+                   double,                   // for double
+                   std::string,              // for string
+                   std::vector<uint8_t>,     // for binary, fixed
```
wgtmac: Why not reusing `std::string`?

src/iceberg/literal.h
wgtmac: I think we need. Otherwise we don't know the exact type of the literal.

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+  std::string ToString() const;
+
+ private:
+  PrimitiveLiteral(PrimitiveLiteralValue value, std::shared_ptr<PrimitiveType> type);
+
+  static PrimitiveLiteral BelowMinLiteral(std::shared_ptr<PrimitiveType> type);
+  static PrimitiveLiteral AboveMaxLiteral(std::shared_ptr<PrimitiveType> type);
+
+  // Helper methods for type casting
+  Result<PrimitiveLiteral> CastFromInt(TypeId target_type_id) const;
```
wgtmac: What about moving these functions to anonymous namespace in the source file? Or make them as public non-member functions if you want to add test cases for them. Besides, why choosing `TypeId` instead of `Type`? Current functions look weird to me.

src/iceberg/literal.h
```diff
@@ -0,0 +1,138 @@
...
+
+  bool isAboveMax() const;
+  bool isBelowMin() const;
+
+  std::string ToString() const;
+
+ private:
+  PrimitiveLiteral(PrimitiveLiteralValue value, std::shared_ptr<PrimitiveType> type);
+
+  static PrimitiveLiteral BelowMinLiteral(std::shared_ptr<PrimitiveType> type);
```
wgtmac: I don't think we need a type here, right? If your intention is to make them comparable, I think we can add `ToAboveMax()` and `ToBelowMin()` to replace these two static functions to keep the original type.

wgtmac (changes requested):

src/iceberg/literal.h
mapleFU: PrimitiveLiteral is "owned", I think using string_view is high risk and most of time doesn't bring up performance

src/iceberg/literal.h
mapleFU: I afraid in the future there would be MapLiteral and StructLiteral. But all is ok to me

src/iceberg/literal.h
mapleFU: You're right, I missed this

src/iceberg/literal.h
mapleFU: It's a good idea. C++ doesn't have a concept for "utf8 encoded string", here I would like to use `std::string` to repr this. But `std::vector<uint8_t>` or `std::string` all looks ok to me

src/iceberg/literal.h
mapleFU: We don't have a `Clone` for `PrimitiveType`?

src/iceberg/literal.h
mapleFU: > I don't think we need a type here, right?

`BelowMin` and `AboveMax` should be typed, a BelowMin for int32 and other types have different meaning

mapleFU: sample: `Literal(type:int64, value: int32_t::max() + 1)` cast to int32_t becomes `Literal(type:int32, value: AboveMax)`

src/iceberg/literal.h
wgtmac: MapLiteral and StructLiteral may be combination of Literals and the majority use case is still for primitive types. So I believe making them shorter looks better. @Fokko WDYT?

src/iceberg/literal.h
wgtmac: No, actually I believe most primitive types should use singleton.

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+namespace iceberg {
+
+/// \brief Literal is a literal value that is associated with a primitive type.
+class ICEBERG_EXPORT Literal {
+ private:
+  /// \brief Exception type for values that are below the minimum allowed value for a
+  /// primitive type.
+  ///
+  /// When casting a value to a narrow primitive type, if the value exceeds the maximum of
+  /// target type, it might be above the maximum allowed value for that type.
```
```suggestion
  /// \brief Sentinel value to indicate that the literal value is below the valid range
  /// of a specific primitive type. It can happen when casting a literal to a narrower
  /// primitive type.
```

I assume you mean `exceptional` instead of `exception`?

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  struct BelowMin {
+    bool operator==(const BelowMin&) const = default;
+    std::strong_ordering operator<=>(const BelowMin&) const = default;
+  };
+
+  /// \brief Exception type for values that are above the maximum allowed value for a
+  /// primitive type.
+  ///
+  /// When casting a value to a narrow primitive type, if the value exceeds the maximum of
+  /// target type, it might be above the maximum allowed value for that type.
```
```suggestion
  /// \brief Sentinel value to indicate that the literal value is above the valid range
  /// of a specific primitive type. It can happen when casting a literal to a narrower
  /// primitive type.
```

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  struct AboveMax {
+    bool operator==(const AboveMax&) const = default;
+    std::strong_ordering operator<=>(const AboveMax&) const = default;
+  };
+
+  using Value = std::variant<bool,         // for boolean
```
wgtmac, line 52: Should we make it public so users are able to get the value?

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+                             int64_t,      // for long, timestamp, timestamp_tz, time
+                             float,        // for float
+                             double,       // for double
+                             std::string,  // for string
+                             std::vector<uint8_t>,     // for binary, fixed
+                             std::array<uint8_t, 16>,  // for uuid and decimal
+                             BelowMin, AboveMax>;
+
+ public:
+  /// Factory methods for primitive types
```
```suggestion
  /// \brief Factory methods for primitive types
```

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  static Literal Boolean(bool value);
+  static Literal Int(int32_t value);
+  static Literal Long(int64_t value);
+  static Literal Float(float value);
+  static Literal Double(double value);
+  static Literal String(std::string value);
+  static Literal Binary(std::vector<uint8_t> value);
+
+  /// Create iceberg literal from bytes.
```
```suggestion
  /// \brief Restore a literal from single-value serialization.
```

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  ///
+  /// See [this spec](https://iceberg.apache.org/spec/#binary-single-value-serialization)
+  /// for reference.
+  static Result<Literal> Deserialize(std::span<const uint8_t> data,
+                                     std::shared_ptr<PrimitiveType> type);
+
+  /// Serialize iceberg literal to bytes.
```
```suggestion
  /// \brief Perform single-value serialization.
```

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  ///
+  /// See [this spec](https://iceberg.apache.org/spec/#binary-single-value-serialization)
+  /// for reference.
+  Result<std::vector<uint8_t>> Serialize() const;
+
+  /// Get the Iceberg Type of the literal.
```
```suggestion
  /// \brief Get the literal type.
```

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  const std::shared_ptr<PrimitiveType>& type() const;
+
+  /// Converts this literal to a literal of the given type.
```
```suggestion
  /// \brief Converts this literal to a literal of the given type.
```

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  /// simplified. For example, std::numeric_limits<int>::max()+1 converted to an int will
+  /// result in AboveMax and can simplify a < std::numeric_limits<int>::max()+1 to always
+  /// true.
+  ///
+  /// \param target_type A primitive PrimitiveType
+  /// \return A Result containing a literal of the given type or an error if conversion
+  /// was not valid
+  Result<Literal> CastTo(const std::shared_ptr<PrimitiveType>& target_type) const;
+
+  /// Compare two PrimitiveLiterals. Both literals must have the same type
```
```suggestion
  /// \brief Compare two PrimitiveLiterals. Both literals must have the same type
```

What if they don't have same type? Return unordered?

test/literal_test.cc
```diff
@@ -0,0 +1,386 @@
...
+
+#include <gtest/gtest.h>
+
+#include "iceberg/type.h"
+#include "matchers.h"
+
+namespace iceberg {
+
+// Boolean type tests
+TEST(PrimitiveLiteralTest, BooleanBasics) {
```
wgtmac:
```suggestion
TEST(LiteralTest, BooleanBasics) {
```

Same for below

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  /// and should not be AboveMax or BelowMin.
+  std::partial_ordering operator<=>(const Literal& other) const;
+
+  bool IsAboveMax() const;
```
wgtmac, line 116: Please add docstring to all public functions.

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+ * under the License.
+ */
+
+#include "iceberg/expression/literal.h"
+
+#include <cmath>
+#include <concepts>
+#include <sstream>
+
+#include "iceberg/exception.h"
```
wgtmac, line 26: Please do not throw. It can only be used in a constructor.

src/iceberg/expression/literal.h
```diff
@@ -0,0 +1,131 @@
...
+  bool IsBelowMin() const;
+
+  std::string ToString() const;
+
+ private:
+  Literal(Value value, std::shared_ptr<PrimitiveType> type);
+
+  friend class LiteralCaster;
```
wgtmac, line 129: What is the trade-off between a friend class as this and exposing `value_` via a public function?

test/CMakeLists.txt
```diff
@@ -69,6 +69,11 @@ target_sources(util_test PRIVATE expected_test.cc formatter_test.cc config_test.
 target_link_libraries(util_test PRIVATE iceberg_static GTest::gtest_main GTest::gmock)
 add_test(NAME util_test COMMAND util_test)
 
+add_executable(literal_test)
```
wgtmac: Can we reuse `expression_test` executable?

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+
+  switch (target_type_id) {
+    case TypeId::kLong:
+      return Literal::Long(static_cast<int64_t>(int_val));
+    case TypeId::kFloat:
+      return Literal::Float(static_cast<float>(int_val));
+    case TypeId::kDouble:
+      return Literal::Double(static_cast<double>(int_val));
+    default:
+      return NotSupported("Cast from Int to {} is not implemented",
```
wgtmac, line 79: It is better to call `ToString` on the `target_type` in the error msg.

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+        return BelowMinLiteral(target_type);
+      }
+      return Literal::Int(static_cast<int32_t>(long_val));
+    }
+    case TypeId::kFloat:
+      return Literal::Float(static_cast<float>(long_val));
+    case TypeId::kDouble:
+      return Literal::Double(static_cast<double>(long_val));
+    default:
+      return NotSupported("Cast from Long to {} is not supported",
```
wgtmac, line 105: ditto

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  auto float_val = std::get<float>(literal.value_);
+  auto target_type_id = target_type->type_id();
+
+  switch (target_type_id) {
+    case TypeId::kDouble:
+      return Literal::Double(static_cast<double>(float_val));
+    default:
+      return NotSupported("Cast from Float to {} is not supported",
+                          static_cast<int>(target_type_id));
```
wgtmac: ditto

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+      return result;
+    }
+    case TypeId::kDecimal:
+    case TypeId::kUuid:
+    case TypeId::kFixed:
+    case TypeId::kDate:
+    case TypeId::kTime:
+    case TypeId::kTimestamp:
+    case TypeId::kTimestampTz: {
+      throw IcebergError("Not implemented: ToString for " + type_->ToString());
```
wgtmac, line 300: Please do not throw

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+    }
+    case TypeId::kString: {
+      return std::get<std::string>(value_);
+    }
+    case TypeId::kBinary: {
+      const auto& binary_data = std::get<std::vector<uint8_t>>(value_);
+      std::string result;
+      result.reserve(binary_data.size() * 2);  // 2 chars per byte
+      for (const auto& byte : binary_data) {
+        result += std::format("{:02X}", byte);
```
wgtmac: nit: use `std::format_to`

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+
+    default:
+      // For unsupported types, return unordered
+      return std::partial_ordering::unordered;
+  }
+}
+
+std::string Literal::ToString() const {
+  if (std::holds_alternative<BelowMin>(value_)) {
+    return "BelowMin";
```
```suggestion
    return "belowMin";
```

To be consistent with java impl

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+  }
+  if (std::holds_alternative<AboveMax>(value_)) {
+    return "AboveMax";
```
```suggestion
    return "aboveMax";
```

src/iceberg/expression/literal.cc
```diff
@@ -0,0 +1,358 @@
...
+
+// Cast method
+Result<Literal> Literal::CastTo(const std::shared_ptr<PrimitiveType>& target_type) const {
+  return LiteralCaster::CastTo(*this, target_type);
+}
+
+// Template function for floating point comparison following Iceberg rules:
+// -NaN < NaN, but all NaN values (qNaN, sNaN) are treated as equivalent within their sign
+template <std::floating_point T>
+std::partial_ordering iceberg_float_compare(T lhs, T rhs) {
```
```suggestion
std::partial_ordering CompareFloat(T lhs, T rhs) {
```

wgtmac: Isn't this a strong order?

src/iceberg/expression/literal.h
mapleFU: We can expose this in the future, but currently I think encapsulate in here is ok

src/iceberg/expression/literal.h
mapleFU: > What if they don't have same type? Return unordered?

Yes. This expects a:
1. `CastTo()` 
2. check BelowMin and AboveMax
3. Compare

Ordering.

src/iceberg/expression/literal.h
mapleFU: Emm currently I think user don't know what's in the `value_`, and might construct some weird `value_` with unexpected type.

src/iceberg/expression/literal.cc
mapleFU: Emmm what should we do here? Here I want to first mark it as unimplemented, when all literal is implemented, this should be change to `std::string`. Should this first as:

```
return fmt::format("Not implemented: ToString for {}", type_->ToString());
```

Or

```
Result<std::string> .ToString()
```

?

src/iceberg/expression/literal.cc
mapleFU: Nice catch! It's a strong_order

src/iceberg/literal.h
Fokko: I think `Literal` is more appropriate. Also, there is always an option to introduce a `PrimitiveLiteral` in the class hierarchy. PyIceberg and Iceberg-Java don't have complex types of literals, so I think we're safe there (until it happens).

src/iceberg/expression/literal.cc
Fokko: Maybe something similar to Java, where you get the Object type with the pointer?

wgtmac: I think eventually we will implement this for all types. I'm fine to throw at this moment (with a comment to explain the current state) or apply suggestion from @Fokko to print anything to prevent it from failing.

test/literal_test.cc
wgtmac: This is not resolved yet. Please rename these test cases.

mapleFU: oh I forgot to push this..

mapleFU:
Can we merge this firstly? My following patch might based on this 😂

Fokko:
Sure thing @mapleFU, thanks @wgtmac, @zhjwpku and @yingcai-cy for the review!

Issue #118, opened 2025-06-05
[Fix warning] I found some warning when build code fix them？


MisterRaindrop:
hi @wgtmac 
I’ve noticed some compilation warnings. I understand that it’s normal to have warnings during the ongoing development of the code. However, they’re quite annoying to look at. If they haven’t been fixed, I can take care of them right away. 
like it:
```
[  1%] Built target nanoarrow
[  1%] Building CXX object src/iceberg/CMakeFiles/iceberg_shared.dir/catalog/in_memory_catalog.cc.o
[  1%] Building CXX object src/iceberg/CMakeFiles/iceberg_shared.dir/demo.cc.o
[  2%] Building CXX object src/iceberg/CMakeFiles/iceberg_shared.dir/expression/expression.cc.o
[  2%] Building CXX object src/iceberg/CMakeFiles/iceberg_shared.dir/file_reader.cc.o
[  2%] Building CXX object src/iceberg/CMakeFiles/iceberg_shared.dir/json_internal.cc.o
/code/src/iceberg/json_internal.cc: In function 'nlohmann::json_abi_v3_11_3::json iceberg::ToJson(const Type&)':
/code/src/iceberg/json_internal.cc:480:1: warning: control reaches end of non-void function [-Wreturn-type]
  480 | }
      | ^
```

MisterRaindrop:
I have see warning like it, It seem code run normally. however, there are compile on warnings. All warnings are of this kind.
Do we need to handle this warning?  If we need to handle it, it might be better to throw an exception.
like it:
```
nlohmann::json ToJson(const Type& type) {
  switch (type.type_id()) {
     case xx:
           xxxx
  }
}
```

wgtmac:
Yes, I think we need to fix them. I also noticed that Windows CI has reported similar issues. Do you want to submit a PR?

MisterRaindrop:
Sure. Let’s discuss how to fix it. I understand that an exception should be thrown because this is not the expected behavior.

wgtmac:
TBH, I do think this is a bug from the compiler because we have exhausted all cases for the switch statement (please correct me if I was wrong). To fix this, we can simply use `Unreachable` defined in iceberg/util/unreachable.h in the default branch.

wgtmac:
I prefer the latter one. Thanks!

PR #119, opened 2025-06-06
fix: fix build warning in switch case (#118)

MisterRaindrop:
fix some compilation warnings. see issue (https://github.com/apache/iceberg-cpp/issues/118)

### DIFF
--- a/src/iceberg/json_internal.cc
+++ b/src/iceberg/json_internal.cc
@@ -477,10 +477,9 @@ nlohmann::json ToJson(const Type& type) {
     }
     case TypeId::kUuid:
       return "uuid";
-    default:
-      internal::Unreachable(std::format("Unknown type id: {}",
-                            static_cast<int>(type.type_id())));
   }
+  internal::Unreachable(
+      std::format("Unknown type id: {}", static_cast<int>(type.type_id())));
 }
 
 nlohmann::json ToJson(const Schema& schema) {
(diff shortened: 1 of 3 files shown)


mapleFU:
can we do like below to avoid default in switch case?

```diff
switch () {
 ...
}
+ Unreachable();
```

MisterRaindrop:
If there is no break statement in a certain case, the program may execute the Unreachable() function, which is not what we want (what we want is for each case to handle the situation and then exit, so we must be careful with the break statements). Additionally, when the compiler does not support exhaustive checking, we cannot know at compile - time if there are any missing cases.


Of course, if it is indeed placed later, I can also make modifications.

MisterRaindrop:
Which way do you think is better?

```
switch () {
 ...
}
+ Unreachable();
```

or

```
switch () {
 ...
default:
+ Unreachable();
}

```

mapleFU:
Isn't all this pattern in this case is `switch (...) { case1: return case2: return ...}`? If a new enum is added, without default, compiler would compile error here

MisterRaindrop:
Okay. If you’re sure it will cause an error, then I’ll modify it to put it outside.

mapleFU (approved):

Fokko (approved):
Nice, thanks for fixing this @MisterRaindrop

PR #120, opened 2025-06-10
feat: add or expression

yingcai-cy:
- or expression
 - change negate return type form Result to Expression
 - add some tests

### DIFF
--- a/src/iceberg/expression/expression.h
+++ b/src/iceberg/expression/expression.h
@@ -25,9 +25,8 @@
 #include <memory>
 #include <string>
 
-#include "iceberg/expected.h"
+#include "iceberg/exception.h"
 #include "iceberg/iceberg_export.h"
-#include "iceberg/result.h"
 
 namespace iceberg {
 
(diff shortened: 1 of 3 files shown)


src/iceberg/expression/expression.h
```diff
@@ -67,8 +66,8 @@ class ICEBERG_EXPORT Expression {
   virtual Operation op() const = 0;
 
   /// \brief Returns the negation of this expression, equivalent to not(this).
-  virtual Result<std::shared_ptr<Expression>> Negate() const {
-    return InvalidExpression("Expression cannot be negated");
+  virtual std::shared_ptr<Expression> Negate() const {
+    throw IcebergError("Expression cannot be negated");
```
wgtmac, line 70: Why changing the return type? We need to stick to `Result` instead of exception.

yingcai-cy: It’s been a couple of days, let me recall. I believe the reason is that all expression implementations have a corresponding Negate form, so none of them would return incorrect results in this context.

Fokko: Everything can be negated in theory, and this is also the case in practice.

Fokko (approved):

wgtmac (approved):
+1

Fokko:
Thanks @yingcai-cy for working on this, and @wgtmac for the review 🚀

PR #121, opened 2025-06-10
For iceberg v1 PartitionSpec's PartitionField id can be optional

Smith-Cruise:
Just looks like forget to use `allow_field_id_missing` in `PartitionFieldFromJson(const nlohmann::json& json, bool allow_field_id_missing)`

In v1 iceberg, we should allow field_id is missing.

### DIFF
--- a/src/iceberg/json_internal.cc
+++ b/src/iceberg/json_internal.cc
@@ -1053,7 +1053,7 @@ Status ParsePartitionSpecs(const nlohmann::json& json, int8_t format_version,
     int32_t next_partition_field_id = PartitionSpec::kLegacyPartitionDataIdStart;
     std::vector<PartitionField> fields;
     for (const auto& entry_json : partition_spec_json) {
-      ICEBERG_ASSIGN_OR_RAISE(auto field, PartitionFieldFromJson(entry_json));
+      ICEBERG_ASSIGN_OR_RAISE(auto field, PartitionFieldFromJson(entry_json, true));
       int32_t field_id = field->field_id();
       if (field_id == SchemaField::kInvalidFieldId) {
         // If the field ID is not set, we need to assign a new one


src/iceberg/json_internal.cc
```diff
@@ -1053,7 +1053,7 @@ Status ParsePartitionSpecs(const nlohmann::json& json, int8_t format_version,
     int32_t next_partition_field_id = PartitionSpec::kLegacyPartitionDataIdStart;
     std::vector<PartitionField> fields;
     for (const auto& entry_json : partition_spec_json) {
-      ICEBERG_ASSIGN_OR_RAISE(auto field, PartitionFieldFromJson(entry_json));
+      ICEBERG_ASSIGN_OR_RAISE(auto field, PartitionFieldFromJson(entry_json, true));
```
mapleFU: ( can we use something like code below?)

```c++
ICEBERG_ASSIGN_OR_RAISE(auto field, PartitionFieldFromJson(entry_json, /*allow_field_id_missing=*/format_version==1));
```

Fokko:
I think we should improve this in another way. I don't think we should set it to `-1` when it is missing:

https://github.com/apache/iceberg-cpp/blob/ed49d1e8f0dc31d99d4fff437a6c1583a0dfced5/src/iceberg/json_internal.cc#L675

Instead, I would follow the same pattern as in Java. Where we assign them incrementally from 1000:

https://github.com/apache/iceberg/blob/7b8bd29ce80bc78ed882f0613f7570e78e325988/core/src/main/java/org/apache/iceberg/PartitionSpecParser.java#L137-L144

This is also suggested by the spec:

image

In V1, it is not allowed to drop or re-order fields, so this will always be consistent:

image

src/iceberg/json_internal.cc
```diff
@@ -1053,14 +1052,10 @@ Status ParsePartitionSpecs(const nlohmann::json& json, int8_t format_version,
...
-      int32_t field_id = field->field_id();
-      if (field_id == SchemaField::kInvalidFieldId) {
-        // If the field ID is not set, we need to assign a new one
-        field_id = next_partition_field_id++;
-      }
-      fields.emplace_back(field->source_id(), field_id, std::string(field->name()),
-                          std::move(field->transform()));
+      ICEBERG_ASSIGN_OR_RAISE(
+          auto field, PartitionFieldFromJson(entry_json, next_partition_field_id));
+      next_partition_field_id++;
```
mapleFU: ++next_partition_field_id;

Smith-Cruise: changed.

Smith-Cruise:
> Generally I think the logic with `PartitionFieldFromJson(json, /*allow_field_id_missing=*/true)` and current logic is equal, but this LGTM. Also cc @wgtmac

Just ensure that `field_id` is always valid when PartitionField is created.

wgtmac:
I believe @Fokko's idea has already been implemented as in https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/json_internal.cc#L1058-L1061. Actually my intention is that `json_internal.h/.cc` are not public and `-1` is used as an internal indicator for this special logic. So I agree with @mapleFU that the current PR is literally the same thing.

Smith-Cruise:
> I believe @Fokko's idea has already been implemented as in https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/json_internal.cc#L1058-L1061. Actually my intention is that `json_internal.h/.cc` are not public and `-1` is used as an internal indicator for this special logic. So I agree with @mapleFU that the current PR is literally the same thing.

Or just using `ICEBERG_ASSIGN_OR_RAISE(auto field, PartitionFieldFromJson(entry_json, /*allow_field_id_missing=*/format_version==1));`.

Fokko:
> I believe @Fokko's idea has already been implemented

Nice, that's great to hear! @Smith-Cruise Could you fix the CI? :)

Fokko (approved):

wgtmac (approved):
LGTM. Thanks @Smith-Cruise!

Sorry I don't have permission to rerun CIs. You may just push again to re-trigger or wait for @Fokko to rerun it.

BTW, could you modify the PR title to apply conventional commit message? For example, `fix: allow PartitionField's field_id to be missing in Iceberg v1`

PR #124, opened 2025-06-20
feat: add support for avro to arrow data conversion

wgtmac:


wgtmac:
@lidavidm @zhjwpku Could you please take a look?

src/iceberg/avro/avro_data_util.cc
```diff
@@ -17,16 +17,383 @@
...
+            "Expected Avro long with TIMESTAMP_MICROS for timestamp field, got: {}",
+            ToString(avro_node));
+      }
+      auto* builder = internal::checked_cast<::arrow::TimestampBuilder*>(array_builder);
+      ICEBERG_ARROW_RETURN_NOT_OK(builder->Append(avro_datum.value<int64_t>()));
+      return {};
+    }
+
+    default:
+      return InvalidArgument("Unsupported primitive type: {}",
```
mapleFU: can we add this is unsupport just in avro, like `"Unsupported primitive type for avro:` ?

src/iceberg/avro/avro_data_util.cc
```diff
@@ -17,16 +17,383 @@
...
+      }
+      return {};
+    }
+
+    case TypeId::kString: {
+      if (avro_node->type() != ::avro::AVRO_STRING) {
+        return InvalidArgument("Expected Avro string for string field, got: {}",
+                               ToString(avro_node));
+      }
+      auto* builder = internal::checked_cast<::arrow::StringBuilder*>(array_builder);
```
mapleFU, line 298: So we don't uses large string and string view, aha

src/iceberg/avro/avro_data_util.cc
```diff
@@ -17,16 +17,383 @@
...
+
+    case TypeId::kDouble: {
+      auto* builder = internal::checked_cast<::arrow::DoubleBuilder*>(array_builder);
+      if (avro_node->type() == ::avro::AVRO_DOUBLE) {
+        ICEBERG_ARROW_RETURN_NOT_OK(builder->Append(avro_datum.value<double>()));
+      } else if (avro_node->type() == ::avro::AVRO_FLOAT) {
+        ICEBERG_ARROW_RETURN_NOT_OK(
+            builder->Append(static_cast<double>(avro_datum.value<float>())));
+      } else {
+        return InvalidArgument("Expected Avro double for double field, got: {}",
```
mapleFU:
```suggestion
        return InvalidArgument("Expected Avro double or float for double field, got: {}",
```
?

src/iceberg/avro/avro_data_util.cc
```diff
@@ -17,16 +17,383 @@
...
+
+    case TypeId::kLong: {
+      auto* builder = internal::checked_cast<::arrow::Int64Builder*>(array_builder);
+      if (avro_node->type() == ::avro::AVRO_LONG) {
+        ICEBERG_ARROW_RETURN_NOT_OK(builder->Append(avro_datum.value<int64_t>()));
+      } else if (avro_node->type() == ::avro::AVRO_INT) {
+        ICEBERG_ARROW_RETURN_NOT_OK(
+            builder->Append(static_cast<int64_t>(avro_datum.value<int32_t>())));
+      } else {
+        return InvalidArgument("Expected Avro long for long field, got: {}",
```
mapleFU: long or int?

src/iceberg/avro/avro_data_util.cc
wgtmac: Yes, let's make it simple for now. We can use reader options to indicate whether large variants are needed in the future.

src/iceberg/avro/avro_data_util.cc
```diff
@@ -17,16 +17,383 @@
...
+Status AppendListToBuilder(const ::avro::NodePtr& avro_node,
+                           const ::avro::GenericDatum& avro_datum,
+                           const FieldProjection& projection, const ListType& list_type,
+                           ::arrow::ArrayBuilder* array_builder) {
+  if (avro_node->type() != ::avro::AVRO_ARRAY) {
+    return InvalidArgument("Expected Avro array, got type: {}", ToString(avro_node));
+  }
+  const auto& avro_array = avro_datum.value<::avro::GenericArray>();
+
+  auto* list_builder = internal::checked_cast<::arrow::ListBuilder*>(array_builder);
```
zhjwpku, line 106: Is there any chance `list_builder` could be a nullptr, should we add a check for that?
If the user provides the same schema for ToArrowSchema and Project, it shouldn't be an issue. However, should we return an error if the user make mistakes? Not so sure.

test/avro_data_test.cc
```diff
@@ -0,0 +1,365 @@
...
+
+#include <arrow/c/bridge.h>
+#include <arrow/json/from_string.h>
+#include <avro/Compiler.hh>
+#include <avro/Generic.hh>
+#include <avro/Node.hh>
+#include <avro/Types.hh>
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <iceberg/arrow/arrow_error_transform_internal.h>
```
zhjwpku: nit: use "" for iceberg headers?

zhjwpku (approved):

src/iceberg/avro/avro_data_util.cc
wgtmac: I don't think so. This is not supposed to be called by users.

zhjwpku: ok, that should be fine.

src/iceberg/avro/avro_data_util_internal.h
```diff
@@ -26,10 +26,21 @@
...
+/// \brief Append an Avro datum to an Arrow array builder.
+///
+/// This function handles schema evolution by using the provided projection to map
+/// fields from the Avro data to the expected Arrow schema.
+///
+/// \param avro_node The Avro schema node (must be a record at root level)
+/// \param avro_datum The Avro data to append
+/// \param projection Schema projection from `projected_schema` to `avro_node`
+/// \param projected_schema The projected schema
+/// \param array_builder The Arrow array builder to append to (must be a struct builder)
```
mapleFU, line 38: So it's a undefined behavior to pass-in a different type builder? Can the argument here just a StructBuilder?

src/iceberg/avro/avro_data_util.cc
mapleFU: Seems like underlying typeof `::arrow::ArrayBuilder` is owned by user, and would have ub when type is wrong. But it's ok to me here

src/iceberg/avro/avro_data_util.cc
```diff
@@ -17,16 +17,440 @@
...
+    const auto& array_entries = avro_array.value();
+
+    const auto& key_field = map_type.key();
+    const auto& value_field = map_type.value();
+
+    ICEBERG_ARROW_RETURN_NOT_OK(map_builder->Append());
+    auto* key_builder = map_builder->key_builder();
+    auto* item_builder = map_builder->item_builder();
+
+    const auto& record_node = avro_node->leafAt(0);
```
mapleFU, line 163: should check leafCount == 1?

mapleFU (approved):

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -95,8 +98,8 @@ function(resolve_arrow_dependency)
 
   fetchcontent_declare(VendoredArrow
                        ${FC_DECLARE_COMMON_OPTIONS}
-                       GIT_REPOSITORY https://github.com/wgtmac/arrow.git
-                       GIT_TAG 7d50c4ac803ad983734de5f418b7cd18f25b0dc9
+                       GIT_REPOSITORY https://github.com/apache/arrow.git
+                       GIT_TAG 5f0aeb5de53fb25b59a52661a80071faef99a4a4
```
Fokko, line 102: Nice!

src/iceberg/avro/avro_data_util.cc
wgtmac: These are all internal functions and array builders are owned by the AvroReader.

src/iceberg/avro/avro_data_util.cc
wgtmac: `::avro::AVRO_ARRAY` type should guarantee this so I don't think we need to over-protect it.

src/iceberg/avro/avro_data_util_internal.h
wgtmac: That requires additional cast before using this function.

mapleFU (approved):

Fokko (approved):

Fokko:
Thanks @wgtmac for working on this, and @zhjwpku and @mapleFU for the review 💪

PR #127, opened 2025-06-22
feat: avro support applying field-ids based on name mapping

MisterRaindrop:
see issue https://github.com/apache/iceberg-cpp/issues/128

### DIFF
--- a/src/iceberg/avro/avro_schema_util.cc
+++ b/src/iceberg/avro/avro_schema_util.cc
@@ -777,6 +777,13 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
 
 namespace {
 
+void CopyCustomAttributes(const ::avro::CustomAttributes& source,
+                          ::avro::CustomAttributes& target) {
+  for (const auto& attr_pair : source.attributes()) {
+    target.addAttribute(attr_pair.first, attr_pair.second, /*addQuote=*/false);
+  }
+}
+
 Result<::avro::NodePtr> CreateRecordNodeWithFieldIds(const ::avro::NodePtr& original_node,
                                                      const MappedField& field) {
   auto new_record_node = std::make_shared<::avro::NodeRecord>();
@@ -857,14 +864,7 @@ Result<::avro::NodePtr> CreateArrayNodeWithFieldIds(const ::avro::NodePtr& origi
     if (original_node->customAttributes() > 0) {
       ::avro::CustomAttributes merged_attributes;
       const auto& original_attrs = original_node->customAttributesAt(0);
-      const auto& existing_attrs = original_attrs.attributes();
-      for (const auto& attr_pair : existing_attrs) {
-        // Skip element-id as we might set it differently
-        if (attr_pair.first != kElementIdProp) {
-          merged_attributes.addAttribute(attr_pair.first, attr_pair.second,
-                                         /*addQuote=*/false);
-        }
-      }
+      CopyCustomAttributes(original_attrs, merged_attributes);
       // Add merged attributes if we found any
       if (merged_attributes.attributes().size() > 0) {
         new_array_node->addCustomAttributesForField(merged_attributes);
@@ -898,17 +898,10 @@ Result<::avro::NodePtr> CreateArrayNodeWithFieldIds(const ::avro::NodePtr& origi
                                  std::to_string(*element_field.field_id),
                                  /*addQuote=*/false);
 
-  // Then merge any custom attributes from original node (except element-id)
+  // Then merge any custom attributes from original node
   if (original_node->customAttributes() > 0) {
     const auto& original_attrs = original_node->customAttributesAt(0);
-    const auto& existing_attrs = original_attrs.attributes();
-    for (const auto& attr_pair : existing_attrs) {
-      // Skip element-id as we've already set it above
-      if (attr_pair.first != kElementIdProp) {
-        merged_attributes.addAttribute(attr_pair.first, attr_pair.second,
-                                       /*addQuote=*/false);
-      }
-    }
+    CopyCustomAttributes(original_attrs, merged_attributes);
   }
 
   // Add all attributes at once
@@ -969,26 +962,12 @@ Result<::avro::NodePtr> CreateMapNodeWithFieldIds(const ::avro::NodePtr& origina
   if (original_node->customAttributes() > 0) {
     // Merge attributes for key (index 0)
     const auto& original_key_attrs = original_node->customAttributesAt(0);
-    const auto& existing_key_attrs = original_key_attrs.attributes();
-    for (const auto& attr_pair : existing_key_attrs) {
-      // Skip if it's the key ID property we're already setting
-      if (attr_pair.first != kKeyIdProp) {
-        key_attributes.addAttribute(attr_pair.first, attr_pair.second,
-                                    /*addQuote=*/false);
-      }
-    }
+    CopyCustomAttributes(original_key_attrs, key_attributes);
 
     // Merge attributes for value (index 1)
     if (original_node->customAttributes() > 1) {
       const auto& original_value_attrs = original_node->customAttributesAt(1);
-      const auto& existing_value_attrs = original_value_attrs.attributes();
-      for (const auto& attr_pair : existing_value_attrs) {
-        // Skip if it's the value ID property we're already setting
-        if (attr_pair.first != kValueIdProp) {
-          value_attributes.addAttribute(attr_pair.first, attr_pair.second,
-                                        /*addQuote=*/false);
-        }
-      }
+      CopyCustomAttributes(original_value_attrs, value_attributes);
     }
   }
 


src/iceberg/CMakeLists.txt
```diff
@@ -98,7 +98,8 @@ if(ICEBERG_BUILD_BUNDLE)
       avro/avro_data_util.cc
       avro/avro_reader.cc
       avro/avro_schema_util.cc
-      avro/avro_stream_internal.cc)
+      avro/avro_stream_internal.cc
+      json_internal.cc)
```
wgtmac: I believe we shouldn't add this

MisterRaindrop: I parse the name_mapping with the NameMappingFromJson, this function in json_internal.cc

MisterRaindrop: I'm confused. Parsing name_mapping requires json function. But the method is defined in the json_internal.cc

src/iceberg/avro/avro_reader.cc
```diff
@@ -96,11 +99,22 @@ class AvroBatchReader::Impl {
...
     HasIdVisitor has_id_visitor;
     ICEBERG_RETURN_UNEXPECTED(has_id_visitor.Visit(file_schema));
+
     if (has_id_visitor.HasNoIds()) {
-      // TODO(gangwu): support applying field-ids based on name mapping
-      return NotImplemented("Avro file schema has no field IDs");
-    }
-    if (!has_id_visitor.AllHaveIds()) {
+      // Apply field IDs based on name mapping if available
+      auto name_mapping_iter = options.properties.find("name_mapping");
```
mapleFU: Can put put these string_view literals to a common places? Or just write like this is ok? @wgtmac

src/iceberg/avro/avro_reader.cc
```diff
@@ -96,11 +99,22 @@ class AvroBatchReader::Impl {
...
+      if (name_mapping_iter != options.properties.end()) {
```
mapleFU:
```
if (auto name_mapping_iter = options.properties.find(); name_mapping_iter != options.properties.end())
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -195,6 +209,147 @@ class AvroBatchReader::Impl {
     return arrow_array;
   }
 
+  // Apply field IDs to Avro schema nodes based on name mapping
+  Status ApplyFieldIdsFromNameMapping(const ::avro::NodePtr& node,
+                                      const NameMapping& name_mapping) {
```
mapleFU: Can the arguments be like:
```c++
  Status ApplyFieldIdsFromNameMapping(const NameMapping& name_mapping,
                                      ::avro::Node* node) {
```

`const NodePtr&` is a bit confusing here since we don't need rebind to other field?

src/iceberg/avro/avro_reader.cc
```diff
@@ -195,6 +209,147 @@ class AvroBatchReader::Impl {
...
+      case ::avro::AVRO_DOUBLE:
+      case ::avro::AVRO_STRING:
+      case ::avro::AVRO_BYTES:
+      case ::avro::AVRO_FIXED:
+        return {};
+      case ::avro::AVRO_NULL:
+      case ::avro::AVRO_ENUM:
+      default:
+        return InvalidSchema("Unsupported Avro type for field ID application: {}",
+                             static_cast<int>(node->type()));
```
mapleFU: @wgtmac so here needs your `ToString`?

src/iceberg/avro/avro_reader.cc
```diff
@@ -195,6 +209,147 @@ class AvroBatchReader::Impl {
...
+    }
+  }
+
+  Status ApplyFieldIdsToRecord(const ::avro::NodePtr& node,
+                               const NameMapping& name_mapping) {
+    for (size_t i = 0; i < node->leaves(); ++i) {
+      const std::string& field_name = node->nameAt(i);
+      ::avro::NodePtr field_node = node->leafAt(i);
```
mapleFU:
```suggestion
      const ::avro::NodePtr& field_node = node->leafAt(i);
```

?

src/iceberg/avro/avro_reader.cc
```diff
@@ -195,6 +209,147 @@ class AvroBatchReader::Impl {
...
+    const auto& branch_1 = node->leafAt(1);
+
+    if (branch_0->type() == ::avro::AVRO_NULL) {
+      return ApplyFieldIdsFromNameMapping(branch_1, name_mapping);
+    }
+    if (branch_1->type() == ::avro::AVRO_NULL) {
+      return ApplyFieldIdsFromNameMapping(branch_0, name_mapping);
+    }
+
+    return InvalidSchema("Union type must have exactly one null branch");
```
mapleFU: The branch above doesn't check "exactly once", isn't they?

src/iceberg/avro/avro_reader.cc
```diff
@@ -195,6 +209,147 @@ class AvroBatchReader::Impl {
...
+        ICEBERG_RETURN_UNEXPECTED(ApplyFieldIdsFromNameMapping(field_node, name_mapping));
+      }
+    }
+    return {};
+  }
+
+  Status ApplyFieldIdsToArray(const ::avro::NodePtr& node,
+                              const NameMapping& name_mapping) {
+    if (node->leaves() != 1) {
+      return InvalidSchema("Array type must have exactly one leaf");
```
mapleFU: Maybe we can add a Todo, and print the `node` in the future

src/iceberg/CMakeLists.txt
wgtmac: Symbols of `json_internal.cc` are included in the `libiceberg` library. The `libiceberg-bundle` library links to `libiceberg` so you don't need to compile the source again.

src/iceberg/avro/avro_reader.cc
MisterRaindrop: It’s great to put them in common places. Do we have a file to put them in, or do I need to create a file for common places?

wgtmac: I agree it is better to have a common place. However I do have a concern on the current approach.

Is the key `"name_mapping"` documented somewhere or is it an implicit behavior from other implementations? I was thinking it can be passed via `ReaderOptions` in a deserialized form rather than deserializing it by every reader.

MisterRaindrop: Let me understand this.

name_mapping in ReaderOptions
```
struct ReaderOptions {
	...
	std::shared_ptr<class NameMapping> name_mapping;
}
```
then user create reader options with name mapping, and if `options.name_mapping` have 

 call `ApplyFieldIdsFromNameMapping(*options.name_mapping, file_schema.root()))`  to parse?

Do you mean this? @wgtmac

wgtmac: Yes, I mean this exactly. Does it make sense? In this way, perhaps you don't need anything from `json_internal.h/cc`.

src/iceberg/avro/avro_reader.cc
MisterRaindrop: Do we need to make the changes? I noticed that the other code in the file also follows this style.

src/iceberg/avro/avro_constants.h
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <string_view>
+
+namespace iceberg::avro {
```
wgtmac: Thanks for adding this! Can we rename it to `iceberg/avro/constants.h` to have a shorter name?

src/iceberg/file_reader.h
```diff
@@ -29,6 +29,7 @@
 
 #include "iceberg/arrow_c_data.h"
 #include "iceberg/file_format.h"
+#include "iceberg/name_mapping.h"
```
wgtmac:
```suggestion
```

We don't need this line because you've already used forward declaration below.

src/iceberg/avro/avro_reader.cc
```diff
@@ -30,11 +30,14 @@
 #include <avro/DataFile.hh>
 #include <avro/Generic.hh>
 #include <avro/GenericDatum.hh>
+#include <nlohmann/json.hpp>
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -96,11 +99,17 @@ class AvroBatchReader::Impl {
...
-      // TODO(gangwu): support applying field-ids based on name mapping
-      return NotImplemented("Avro file schema has no field IDs");
-    }
-    if (!has_id_visitor.AllHaveIds()) {
+      // Apply field IDs based on name mapping if available
+      if (options.name_mapping) {
+        ICEBERG_RETURN_UNEXPECTED(ApplyFieldIdsFromNameMapping(*options.name_mapping,
+                                                               file_schema.root().get()));
+      } else {
+        return NotImplemented(
```
```suggestion
        return InvalidSchema(
```

You've implemented this so it should be a new error now.

src/iceberg/avro/avro_reader.cc
```diff
@@ -195,6 +204,156 @@ class AvroBatchReader::Impl {
     return arrow_array;
   }
 
+  // Apply field IDs to Avro schema nodes based on name mapping
+  Status ApplyFieldIdsFromNameMapping(const NameMapping& name_mapping,
```
wgtmac: Can we move this to `avro_schema_util_internal.h` and `avro_schema_util.cc` so we can have better test coverage?

src/iceberg/avro/avro_reader.cc
wgtmac: There is a `std::string ToString(const ::avro::NodePtr& node)` in the `avro_schema_util_internal.h` to make it easier for better error message.

test/avro_reader_test.cc
```diff
@@ -0,0 +1,197 @@
...
+
+#include "iceberg/json_internal.h"
+#include "iceberg/name_mapping.h"
+#include "iceberg/result.h"
+#include "iceberg/schema.h"
+#include "matchers.h"
+
+namespace iceberg::avro {
+
+class AvroReaderTest : public ::testing::Test {
```
wgtmac: These test cases are unrelated to this change or avro reader? If yes, please remove them.

src/iceberg/avro/avro_reader.cc
```diff
@@ -96,11 +99,17 @@ class AvroBatchReader::Impl {
...
     ICEBERG_RETURN_UNEXPECTED(has_id_visitor.Visit(file_schema));
+
     if (has_id_visitor.HasNoIds()) {
-      // TODO(gangwu): support applying field-ids based on name mapping
-      return NotImplemented("Avro file schema has no field IDs");
-    }
-    if (!has_id_visitor.AllHaveIds()) {
+      // Apply field IDs based on name mapping if available
+      if (options.name_mapping) {
+        ICEBERG_RETURN_UNEXPECTED(ApplyFieldIdsFromNameMapping(*options.name_mapping,
```
wgtmac: Do we need to further check that all fields have applied field-ids?

src/iceberg/avro/avro_reader.cc
```diff
@@ -195,6 +204,156 @@ class AvroBatchReader::Impl {
...
+
+      // Try to find field ID by name in the name mapping
+      if (auto field_ref = name_mapping.Find(field_name)) {
+        if (field_ref->get().field_id.has_value()) {
+          // Add field ID attribute to the node
+          ::avro::CustomAttributes attributes;
+          attributes.addAttribute(std::string(kFieldId),
+                                  std::to_string(field_ref->get().field_id.value()),
+                                  false);
+          node->addCustomAttributesForField(attributes);
```
wgtmac: I think `addCustomAttributesForField` is not supposed to work like this. IIUC, `addCustomAttributesForField` is called for each sub-field of a parent node. It is impossible to in-place modify attributes of an `avro::Node`. We might need to create a new avro schema by merging original attributes with new ones.

src/iceberg/avro/avro_reader.cc
MisterRaindrop: When the verification of verify_visitor.AllHaveIds() fails, should we print detailed information such as…
```
return InvalidSchema("Not all fields have field IDs after applying name mapping. {}", 
                              verify_visitor.GetMissingFieldIdsInfo());
```
get detailed information or simple 
```
return InvalidSchema("Not all fields have field IDs after applying name mapping")
```

src/iceberg/avro/avro_reader.cc
MisterRaindrop: I understand.

if name mapping available I will call the `CreateAvroNodeWithFieldIds`, will create new schema and update avro file_schema 

like this
```
ICEBERG_ASSIGN_OR_RAISE(auto new_root_node,
                                CreateAvroNodeWithFieldIds(file_schema.root(), *options.name_mapping));
        
// Create a new schema with the updated root node
auto new_schema = ::avro::ValidSchema(new_root_node);
// Update the file schema to use the new schema with field IDs
file_schema = new_schema;
```

src/iceberg/avro/avro_reader.cc
wgtmac: I think we can simply provide the shorter message. We need to add a logger interface to print detail error message.

MisterRaindrop: add a logger interface? you mean add a log system like glog or spdlog?

wgtmac: I mean a pluggable logger interface. For example, we can use pure interface in `libiceberg` and link `libiceberg-bundle` with `spdlog` so users can choose their own logger bindings if they do not use `libiceberg-bundle`

MisterRaindrop: I understand. in this commit, I will report a simple error message, and add todo print detail error message in future, how do you feel about?

wgtmac: SGTM. Thanks!

MisterRaindrop:
@wgtmac hi please help me check it again when you’re free.

src/iceberg/avro/avro_schema_util_internal.h
```diff
@@ -144,4 +145,39 @@ std::string ToString(const ::avro::LogicalType::Type& logical_type);
 /// \return True if the node has a map logical type, false otherwise.
 bool HasMapLogicalType(const ::avro::NodePtr& node);
 
+/// \brief Create a new Avro node with field IDs from name mapping.
+/// \param original_node The original Avro node to copy.
+/// \param name_mapping The name mapping to apply field IDs from.
+/// \return A new Avro node with field IDs applied, or an error.
+Result<::avro::NodePtr> CreateAvroNodeWithFieldIds(const ::avro::NodePtr& original_node,
```
wgtmac: Can we add only the following function to this header file?

```
Result<::avro::NodePtr> ApplyNameMapping(const ::avro::NodePtr& original_node,
                                         const NameMapping& name_mapping);
```

Other functions are only internal and do not have to be in the header files.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,201 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
   return SchemaProjection{std::move(field_projection.children)};
 }
 
+// Helper function to create a new Avro node with field IDs from name mapping
+Result<::avro::NodePtr> CreateAvroNodeWithFieldIds(const ::avro::NodePtr& original_node,
+                                                   const NameMapping& name_mapping) {
```
wgtmac:
```suggestion
                                                   const MappedField& mapped_field) {
```

Two suggestions here:
1. These utility functions can be wrapped in the anonymous namespace.
2. They can accept `MappedField` instead of `NameMapping` to avoid creating NameMapping over and over again.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,201 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  }
+}
+
+Result<::avro::NodePtr> CreateRecordNodeWithFieldIds(const ::avro::NodePtr& original_node,
+                                                     const NameMapping& name_mapping) {
+  auto new_record_node = std::make_shared<::avro::NodeRecord>();
+  new_record_node->setName(original_node->name());
+
+  for (size_t i = 0; i < original_node->leaves(); ++i) {
+    const std::string& field_name = original_node->nameAt(i);
```
wgtmac, line 797: We need to check `i < original_node->names()` and return error if fails before calling `original_node->nameAt(i)` just in case of a malformed node. Same for other `xxxAt(i)` variants.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,201 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+    ::avro::NodePtr field_node = original_node->leafAt(i);
+
+    // Try to find field ID by name in the name mapping
+    if (auto field_ref = name_mapping.Find(field_name)) {
```
wgtmac: How do we support case sensitivity? The NameMapping class does not yet support it. We can leave a TODO comment and support this as a followup.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,201 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+      if (field_ref->get().field_id.has_value()) {
+        // Add field ID attribute to the new node
+        ::avro::CustomAttributes attributes;
+        attributes.addAttribute(std::string(kFieldIdProp),
+                                std::to_string(field_ref->get().field_id.value()), false);
+        new_record_node->addCustomAttributesForField(attributes);
```
wgtmac: We need to preserve existing attributes of original leaves.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,201 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+      } else {
+        // Recursively apply field IDs to child nodes
+        ICEBERG_ASSIGN_OR_RAISE(auto new_field_node,
+                                CreateAvroNodeWithFieldIds(field_node, name_mapping));
+        new_record_node->addName(field_name);
+        new_record_node->addLeaf(new_field_node);
+      }
+    } else {
+      // Recursively apply field IDs to child nodes even if no mapping found
+      ICEBERG_ASSIGN_OR_RAISE(auto new_field_node,
```
wgtmac: Should we return error status if field_id is missing? Then we don't need to validate this again in the Avro reader after applying name mapping.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,201 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  if (original_node->leaves() != 1) {
+    return InvalidSchema("Array type must have exactly one leaf");
+  }
+
+  auto new_array_node = std::make_shared<::avro::NodeArray>();
+  new_array_node->setName(original_node->name());
+  new_array_node->setLogicalType(original_node->logicalType());
+
+  // Check if this is a map represented as array
+  if (original_node->logicalType().type() == ::avro::LogicalType::CUSTOM &&
```
wgtmac: Please reuse `bool HasMapLogicalType(const ::avro::NodePtr& node)` in the `avro_schema_util_internal.h`

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,201 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+      if (field_ref->get().field_id.has_value()) {
+        // Add field ID attribute to the new node
+        ::avro::CustomAttributes attributes;
+        attributes.addAttribute(std::string(kFieldIdProp),
+                                std::to_string(field_ref->get().field_id.value()), false);
+        new_record_node->addCustomAttributesForField(attributes);
+      }
+
+      // Recursively apply field IDs to nested fields if they exist
+      if (field_ref->get().nested_mapping && field_node->type() == ::avro::AVRO_RECORD) {
```
wgtmac: This if/else block looks incorrect. We should simply pass `const MappedField&` to `CreateAvroNodeWithFieldIds` and let it decide how to dispatch the call based on the actual node type.

src/iceberg/avro/avro_schema_util.cc
wgtmac: Similarly, we may need to preserve default value of each field.

test/avro_schema_test.cc
```diff
@@ -1070,3 +1075,135 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
 }
 
 }  // namespace iceberg::avro
+
+// NameMapping tests for Avro schema context
+namespace iceberg::avro {
+
+std::unique_ptr<NameMapping> CreateTestNameMapping() {
```
wgtmac: Could you please revert changes in this file? Or we can move them to `name_mapping_test.cc`.

src/iceberg/avro/avro_schema_util_internal.h
wgtmac: Please add test cases for applying name mapping to different Avro schemas (record/map/array/union). I think the current implementation has some obvious bugs for nested types. It would be helpful if we have them.

wgtmac (changes requested):

src/iceberg/avro/avro_schema_util_internal.h
MisterRaindrop: ApplyNameMapping is delete,  now I use CreateAvroNodeWithFieldIds to create new record node, because ApplyNameMapping have an ambiguity, we need to create new avro node, not apply in original avro record node.

I will move other function internal.

test/avro_schema_test.cc
```diff
@@ -22,10 +22,15 @@
 #include <avro/Compiler.hh>
 #include <avro/NodeImpl.hh>
 #include <avro/Types.hh>
+#include <gmock/gmock.h>
```
wgtmac: Please revert (unrelated) changes in this file.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  auto new_record_node = std::make_shared<::avro::NodeRecord>();
+  new_record_node->setName(original_node->name());
+
+  if (original_node->leaves() > original_node->names()) {
+    return InvalidSchema("Node has {} leaves but only {} names", original_node->leaves(),
+                         original_node->names());
+  }
+
+  for (size_t i = 0; i < original_node->leaves(); ++i) {
+    const std::string& field_name = original_node->nameAt(i);
```
wgtmac, line 797:
```suggestion
    if (i >= original_node->names()) {
      return InvalidSchema(...);
    }
    const std::string& field_name = original_node->nameAt(i);
```

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+    ::avro::NodePtr field_node = original_node->leafAt(i);
```
wgtmac, line 798:
```suggestion
    if (i >= original_node->leaves()) {
      return InvalidSchema(...);
    }
    ::avro::NodePtr field_node = original_node->leafAt(i);
```

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+Result<::avro::NodePtr> CreateRecordNodeWithFieldIds(const ::avro::NodePtr& original_node,
+                                                     const MappedField& field) {
+  auto new_record_node = std::make_shared<::avro::NodeRecord>();
+  new_record_node->setName(original_node->name());
+
+  if (original_node->leaves() > original_node->names()) {
+    return InvalidSchema("Node has {} leaves but only {} names", original_node->leaves(),
+                         original_node->names());
+  }
+
```
wgtmac:
```suggestion
```

I don't think we need to check this. I meant to add suggested changes below.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+        }
+      }
+
+      // Add field ID attribute to the new node (preserving existing attributes)
+      attributes.addAttribute(std::string(kFieldIdProp),
+                              std::to_string(nested_field->field_id.value()), false);
+
+      if (!attributes.attributes().empty()) {
+        new_record_node->addCustomAttributesForField(attributes);
+      }
```
wgtmac:
```suggestion
      new_record_node->addCustomAttributesForField(attributes);
```

This check is unnecessary.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+                             field_name);
+      }
+
+      // Preserve existing custom attributes for this field
+      ::avro::CustomAttributes attributes;
+      if (i < original_node->customAttributes()) {
+        // Copy all existing attributes from the original node
+        const auto& original_attrs = original_node->customAttributesAt(i);
+        const auto& existing_attrs = original_attrs.attributes();
+        for (const auto& attr_pair : existing_attrs) {
```
wgtmac:
```suggestion
        for (const auto& attr_pair : original_node->customAttributesAt(i).attributes()) {
```

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+      }
+
+      // Recursively apply field IDs to nested fields
+      ICEBERG_ASSIGN_OR_RAISE(auto new_nested_node,
+                              CreateAvroNodeWithFieldIds(field_node, *nested_field));
+      new_record_node->addName(field_name);
+      new_record_node->addLeaf(new_nested_node);
+    } else {
+      // If no nested field found, this is an error
+      return InvalidSchema("Field '{}' not found in nested mapping", field_name);
```
wgtmac: It is much better to return error immediately after we fail to find the field id. Then we don't need the large if branch at line 817.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+      }
+
+      // Preserve existing custom attributes for this field
+      ::avro::CustomAttributes attributes;
+      if (i < original_node->customAttributes()) {
+        // Copy all existing attributes from the original node
+        const auto& original_attrs = original_node->customAttributesAt(i);
+        const auto& existing_attrs = original_attrs.attributes();
+        for (const auto& attr_pair : existing_attrs) {
+          attributes.addAttribute(attr_pair.first, attr_pair.second, false);
```
wgtmac:
```suggestion
          attributes.addAttribute(attr_pair.first, attr_pair.second, /*addQuote=*/false);
```

wgtmac: For better readability.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+    new_array_node->addLeaf(new_element_node);
+    return new_array_node;
+  }
+
+  // For regular arrays, try to find element field ID from nested mapping
+  const MappedField* element_field = nullptr;
+  if (field.nested_mapping) {
+    auto fields_span = field.nested_mapping->fields();
+    for (const auto& f : fields_span) {
+      if (f.names.find(std::string(kElement)) != f.names.end()) {
```
wgtmac: Do we actually need to find the field name of a map or array? IIUC, the field name mapping should only apply to struct fields. We can blindly use the child nested_mapping.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+      return InvalidSchema("Field ID is missing for element field in array");
+    }
+
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto new_element_node,
+        CreateAvroNodeWithFieldIds(original_node->leafAt(0), *element_field));
+    new_array_node->addLeaf(new_element_node);
+  } else {
+    // If no element field found, this is an error
+    return InvalidSchema("Element field not found in nested mapping for array");
```
wgtmac: Ditto

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  if (element_field) {
+    // Check if field_id is present
+    if (!element_field->field_id.has_value()) {
+      return InvalidSchema("Field ID is missing for element field in array");
+    }
+
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto new_element_node,
+        CreateAvroNodeWithFieldIds(original_node->leafAt(0), *element_field));
+    new_array_node->addLeaf(new_element_node);
```
wgtmac: Shouldn't we add `element_field->field_id` to the custom attributes?

test/name_mapping_test.cc
```diff
@@ -24,8 +24,18 @@
 #include <unordered_set>
 #include <vector>
 
+#include <avro/Compiler.hh>
+#include <avro/NodeImpl.hh>
+#include <avro/Types.hh>
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
```
wgtmac: We don't need this, isn't it?

test/name_mapping_test.cc
```diff
@@ -299,3 +361,436 @@ TEST(CreateMappingTest, ListSchemaToMapping) {
 }
 
 }  // namespace iceberg
+
+// NameMapping tests for Avro schema context
+namespace iceberg::avro {
```
wgtmac: Ah, test cases below this line should be in the avro_schema_test.cc.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,237 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  if (original_node->leaves() != 2) {
+    return InvalidSchema("Map type must have exactly two leaves");
+  }
+
+  auto new_map_node = std::make_shared<::avro::NodeMap>();
+
+  // Try to find key and value fields from nested mapping
+  const MappedField* key_field = nullptr;
+  const MappedField* value_field = nullptr;
+  if (field.nested_mapping) {
```
wgtmac: ditto, I think we can blindly get field ids for key and value fields.

wgtmac: BTW, these ids are not added to the attributes below.

test/name_mapping_test.cc
```diff
@@ -299,3 +361,436 @@ TEST(CreateMappingTest, ListSchemaToMapping) {
...
+    const auto& array_node = new_node->leafAt(1);
+    EXPECT_EQ(array_node->type(), ::avro::AVRO_ARRAY);
+    EXPECT_EQ(array_node->leaves(), 1);
+
+    // Note: Array nodes don't support custom attributes in Avro C++
+    // We only verify the structure is correct, not the attributes
+  } catch (const std::exception& e) {
+    // If we get an exception about attributes not being supported, that's expected
+    // for array nodes in Avro C++
+    EXPECT_TRUE(std::string(e.what()).find("This type does not have attribute") !=
```
wgtmac: Please fix test cases like this.

src/iceberg/avro/avro_reader.cc
```diff
@@ -91,16 +92,43 @@ class AvroBatchReader::Impl {
...
+            auto new_root_node,
+            CreateAvroNodeWithFieldIds(file_schema.root(), mapped_field));
+
+        // Create a new schema with the updated root node
+        auto new_schema = ::avro::ValidSchema(new_root_node);
+
+        // Verify that all fields now have IDs after applying the name mapping
+        HasIdVisitor verify_visitor;
+        ICEBERG_RETURN_UNEXPECTED(verify_visitor.Visit(new_schema));
+        if (!verify_visitor.AllHaveIds()) {
```
wgtmac: I think we can remove this extra check now. Your implementation can guarantee all fields are processed.

wgtmac (changes requested):
Thanks for the update! I have left some comments. Let me know when all test cases are good.

MisterRaindrop:
@wgtmac hi, I update code and all test is ok

test/CMakeLists.txt
```diff
@@ -30,8 +30,7 @@ configure_file("${CMAKE_SOURCE_DIR}/test/test_config.h.in"
 
 add_executable(schema_test)
 target_sources(schema_test
-               PRIVATE name_mapping_test.cc
-                       schema_test.cc
+               PRIVATE schema_test.cc
```
wgtmac: Please revert changes in this file.

test/name_mapping_test.cc
```diff
@@ -24,9 +24,18 @@
 #include <unordered_set>
 #include <vector>
 
+#include <avro/Compiler.hh>
+#include <avro/NodeImpl.hh>
+#include <avro/Types.hh>
```
wgtmac:
```suggestion
```

test/name_mapping_test.cc
```diff
@@ -24,9 +24,18 @@
...
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "iceberg/avro/avro_reader.h"
+#include "iceberg/avro/avro_schema_util_internal.h"
+#include "iceberg/file_reader.h"
+#include "iceberg/json_internal.h"
+#include "matchers.h"
+
```
wgtmac:
```suggestion
```

test/avro_schema_test.cc
```diff
@@ -23,9 +23,13 @@
 #include <avro/NodeImpl.hh>
 #include <avro/Types.hh>
 #include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
```
wgtmac:
```suggestion
```

test/avro_schema_test.cc
```diff
@@ -23,9 +23,13 @@
...
 
+#include "iceberg/avro/avro_reader.h"
```
wgtmac:
```suggestion
```

test/avro_schema_test.cc
```diff
@@ -23,9 +23,13 @@
...
 #include "iceberg/avro/avro_schema_util_internal.h"
+#include "iceberg/json_internal.h"
```
wgtmac:
```suggestion
```

test/avro_schema_test.cc
```diff
@@ -46,6 +50,24 @@ void CheckFieldIdAt(const ::avro::NodePtr& node, size_t index, int32_t field_id,
...
+  nested_fields.emplace_back(MappedField{.names = {"value"}, .field_id = 3});
+  nested_fields.emplace_back(MappedField{.names = {"description"}, .field_id = 4});
+  auto nested_mapping = MappedFields::Make(std::move(nested_fields));
+
+  fields.emplace_back(MappedField{
+      .names = {"data"}, .field_id = 5, .nested_mapping = std::move(nested_mapping)});
+
+  return NameMapping::Make(std::move(fields));
+}
+
```
wgtmac:
```suggestion
```
I don't see it is used anywhere.

test/avro_schema_test.cc
```diff
@@ -1057,4 +1079,383 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+  ASSERT_THAT(result, IsOk());
+
+  const auto& new_node = *result;
+  EXPECT_EQ(new_node->type(), ::avro::AVRO_RECORD);
+  EXPECT_EQ(new_node->names(), 3);
+  EXPECT_EQ(new_node->leaves(), 3);
+
+  // Check that field IDs are properly applied
+  ASSERT_EQ(new_node->customAttributes(), 3);
+  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(new_node, 0, 1));
```
wgtmac:
```suggestion
  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(node, /*index=*/0, /*field_id=*/1));
```

For better readability. Same for all calls to `CheckFieldIdAt` below.

src/iceberg/avro/avro_schema_util_internal.h
```diff
@@ -144,4 +145,11 @@ std::string ToString(const ::avro::LogicalType::Type& logical_type);
 /// \return True if the node has a map logical type, false otherwise.
 bool HasMapLogicalType(const ::avro::NodePtr& node);
 
+/// \brief Create a new Avro node with field IDs from name mapping.
+/// \param original_node The original Avro node to copy.
+/// \param mapped_field The mapped field to apply field IDs from.
+/// \return A new Avro node with field IDs applied, or an error.
+Result<::avro::NodePtr> CreateAvroNodeWithFieldIds(const ::avro::NodePtr& original_node,
```
wgtmac: nit: we can add the following convenience function so callers (e.g. AvroReader and test cases) don't need to convert from NameMapping to MappedField every time:

```
Result<::avro::NodePtr> CreateAvroNodeWithFieldIds(const ::avro::NodePtr& original_node,
                                                   const NameMapping& mapping);
```

test/avro_schema_test.cc
```diff
@@ -1057,4 +1079,383 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+
+  auto result = CreateAvroNodeWithFieldIds(avro_schema.root(), mapped_field);
+  ASSERT_THAT(result, IsOk());
+
+  const auto& new_node = *result;
+  EXPECT_EQ(new_node->type(), ::avro::AVRO_RECORD);
+  EXPECT_EQ(new_node->names(), 2);
+  EXPECT_EQ(new_node->leaves(), 2);
+
+  // Check array field structure only - don't access any attributes
```
wgtmac: Why not checking the ids?

test/avro_schema_test.cc
```diff
@@ -1057,4 +1079,383 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+
+  auto result = CreateAvroNodeWithFieldIds(avro_schema.root(), mapped_field);
+  ASSERT_THAT(result, IsOk());
+
+  const auto& new_node = *result;
+  EXPECT_EQ(new_node->type(), ::avro::AVRO_RECORD);
+  EXPECT_EQ(new_node->names(), 2);
+  EXPECT_EQ(new_node->leaves(), 2);
+
+  // Check map field structure only - don't access any attributes
```
wgtmac: Same questions here.

test/avro_schema_test.cc
MisterRaindrop: because array and map types do not have attributes, this comment was added. I will remove this comment to avoid misunderstanding.

test/avro_schema_test.cc
MisterRaindrop: the same

test/avro_schema_test.cc
wgtmac: I was asking why `CheckFieldIdAt` was not called for any field in the list and map types. Otherwise we haven't checked whether field ids are correctly set.

test/avro_schema_test.cc
```diff
@@ -1057,4 +1059,366 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(new_node, /*index=*/1, /*field_id=*/2));
+
+  // Check array field structure and element field ID
+  const auto& array_node = new_node->leafAt(1);
+  EXPECT_EQ(array_node->type(), ::avro::AVRO_ARRAY);
+  EXPECT_EQ(array_node->leaves(), 1);
+
+  // Check that array element has field ID applied
+  const auto& element_node = array_node->leafAt(0);
+  EXPECT_EQ(element_node->type(), ::avro::AVRO_STRING);
```
wgtmac, line 1275:
```suggestion
  EXPECT_EQ(element_node->type(), ::avro::AVRO_STRING);
  ASSERT_EQ(array_node->customAttributes(), 1);
  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(array_node, /*index=*/0, /*field_id=*/20));
```

test/avro_schema_test.cc
```diff
@@ -1057,4 +1059,366 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+  ASSERT_EQ(new_node->customAttributes(), 2);
+  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(new_node, /*index=*/0, /*field_id=*/1));
+  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(new_node, /*index=*/1, /*field_id=*/2));
+
+  // Check map field structure and key-value field IDs
+  const auto& map_node = new_node->leafAt(1);
+  EXPECT_EQ(map_node->type(), ::avro::AVRO_MAP);
+  ASSERT_GE(map_node->leaves(), 2);
+  EXPECT_EQ(map_node->leafAt(0)->type(), ::avro::AVRO_STRING);
+  EXPECT_EQ(map_node->leafAt(1)->type(), ::avro::AVRO_STRING);
```
wgtmac, line 1315:
```suggestion
  EXPECT_EQ(map_node->leafAt(1)->type(), ::avro::AVRO_STRING);
  ASSERT_EQ(map_node->customAttributes(), 2);
  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(map_node, /*index=*/0, /*field_id=*/30));
  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(map_node, /*index=*/1, /*field_id=*/31));
```

test/avro_schema_test.cc
```diff
@@ -1057,4 +1059,366 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+
+TEST_F(NameMappingAvroSchemaTest, ApplyNameMappingToMap) {
+  // Create an Avro map schema without field IDs
+  std::string avro_schema_json = R"({
+    "type": "record",
+    "name": "test_record",
+    "fields": [
+      {"name": "id", "type": "int"},
+      {"name": "properties", "type": {
+        "type": "map",
```
wgtmac, line 1288: Could you also add a test case for map type whose key is not a string type? We need to cover the case where map is an `array<struct<key,value>>`

src/iceberg/avro/avro_reader.cc
```diff
@@ -96,16 +97,29 @@ class AvroReader::Impl {
...
+      if (options.name_mapping) {
+        ICEBERG_ASSIGN_OR_RAISE(
+            auto new_root_node,
+            CreateAvroNodeWithFieldIds(file_schema.root(), *options.name_mapping));
+
+        // Create a new schema with the updated root node
+        auto new_schema = ::avro::ValidSchema(new_root_node);
+
+        // Update the file schema to use the new schema with field IDs
+        file_schema = new_schema;
```
```suggestion
        file_schema = ::avro::ValidSchema(new_root_node);
```

We don't need a temporary variable.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -783,4 +785,232 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  for (size_t i = 0; i < original_node->leaves(); ++i) {
+    if (i >= original_node->names()) {
+      return InvalidSchema("Index {} is out of bounds for names (size: {})", i,
+                           original_node->names());
+    }
+    const std::string& field_name = original_node->nameAt(i);
+    if (i >= original_node->leaves()) {
+      return InvalidSchema("Index {} is out of bounds for leaves (size: {})", i,
+                           original_node->leaves());
+    }
```
wgtmac:
```suggestion
```

This is checked already in the for loop.

src/iceberg/avro/avro_schema_util.cc
wgtmac: Please add this comment.

src/iceberg/avro/avro_schema_util_internal.h
```diff
@@ -144,4 +145,18 @@ std::string ToString(const ::avro::LogicalType::Type& logical_type);
...
```
wgtmac:
```suggestion
Result<::avro::NodePtr> MakeAvroNodeWithFieldIds(const ::avro::NodePtr& original_node,
```

src/iceberg/avro/avro_schema_util_internal.h
```diff
@@ -144,4 +145,18 @@ std::string ToString(const ::avro::LogicalType::Type& logical_type);
...
+                                                   const MappedField& mapped_field);
+
+/// \brief Create a new Avro node with field IDs from name mapping.
+/// \param original_node The original Avro node to copy.
+/// \param mapping The name mapping to apply field IDs from.
+/// \return A new Avro node with field IDs applied, or an error.
+Result<::avro::NodePtr> CreateAvroNodeWithFieldIds(const ::avro::NodePtr& original_node,
```
wgtmac:
```suggestion
Result<::avro::NodePtr> MakeAvroNodeWithFieldIds(const ::avro::NodePtr& original_node,
```

src/iceberg/avro/avro_schema_util_internal.h
wgtmac: We usually use `Make` instead of `Create` as the convention...

test/avro_schema_test.cc
MisterRaindrop: I don’t understand this test case.  Do we need to test failure scenarios?
`CheckFieldIdAt(array_node, /*index=*/0, /*field_id=*/20)`

wgtmac: No, it should succeed. Otherwise this PR is incomplete because child fields of map and list types do not have correct field ids. I think I have left some comments like https://github.com/apache/iceberg-cpp/pull/127/files#r2203809525 to address this but they are still unresolved. That's why I have asked for completely passing test cases again and again during the review.

MisterRaindrop: thanks review, You’re right. I’ll try to fix it. Maybe you still need to help me check it later.

MisterRaindrop:
Thanks review PR, I have to fix field id, and test case is ok, please check if there are any other suggestions? @wgtmac

test/avro_schema_test.cc
```diff
@@ -1057,4 +1059,454 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+  ASSERT_EQ(new_node->customAttributes(), 2);
+  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(new_node, /*index=*/0, /*field_id=*/1));
+  ASSERT_NO_FATAL_FAILURE(CheckFieldIdAt(new_node, /*index=*/1, /*field_id=*/2));
+
+  // Check union field
+  const auto& union_node = new_node->leafAt(1);
+  EXPECT_EQ(union_node->type(), ::avro::AVRO_UNION);
+  EXPECT_EQ(union_node->leaves(), 2);
+
+  // Check that the non-null branch has field ID applied
```
wgtmac:
```suggestion
```

wgtmac: Remove this as it is confusing.

src/iceberg/avro/avro_schema_util.cc
wgtmac: Could you please resolve this?

src/iceberg/avro/avro_schema_util.cc
wgtmac: I'd prefer to apply this suggestion to make it shorter.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,278 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+        const auto& existing_attrs = original_attrs.attributes();
+        for (const auto& attr_pair : existing_attrs) {
+          // Copy each existing attribute to preserve original metadata
+          attributes.addAttribute(attr_pair.first, attr_pair.second, false);
+        }
+      }
+
+      // Add field ID attribute to the new node (preserving existing attributes)
+      attributes.addAttribute(std::string(kFieldIdProp),
+                              std::to_string(nested_field->field_id.value()), false);
```
wgtmac:
```suggestion
                              std::to_string(nested_field->field_id.value()), /*addQuote=*/false);
```

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,278 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto new_element_node,
+        MakeAvroNodeWithFieldIds(original_node->leafAt(0), *element_field));
+    new_array_node->addLeaf(new_element_node);
+
+    // Add element field ID as custom attribute
+    ::avro::CustomAttributes element_attributes;
+    element_attributes.addAttribute(std::string(kFieldIdProp),
+                                    std::to_string(*element_field->field_id), false);
```
wgtmac:
```suggestion
                                    std::to_string(*element_field->field_id), /*addQuote=*/false);
```

test/avro_schema_test.cc
```diff
@@ -1057,4 +1059,454 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+  })";
+  auto avro_schema = ::avro::compileJsonSchemaFromString(avro_schema_json);
+
+  auto result = MakeAvroNodeWithFieldIds(avro_schema.root(), *name_mapping);
+  ASSERT_THAT(result, IsError(ErrorKind::kInvalidSchema));
+  ASSERT_THAT(result,
+              HasErrorMessage("Field 'unknown_field' not found in nested mapping"));
+}
+
+TEST_F(NameMappingAvroSchemaTest, MissingArrayElementError) {
```
wgtmac: Should we remove `MissingArrayElementError` and `MissingMapKeyValueError`? They are actually testing the top-level fields instead of specific array or map type. So I believe these cases are duplicates of the above one.

test/avro_schema_test.cc
```diff
@@ -1057,4 +1059,454 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+      {"name": "name", "type": "string"},
+      {"name": "unknown_field", "type": "int"}
+    ]
+  })";
+  auto avro_schema = ::avro::compileJsonSchemaFromString(avro_schema_json);
+
+  auto result = MakeAvroNodeWithFieldIds(avro_schema.root(), *name_mapping);
+  ASSERT_THAT(result, IsError(ErrorKind::kInvalidSchema));
+  ASSERT_THAT(result,
+              HasErrorMessage("Field 'unknown_field' not found in nested mapping"));
```
wgtmac: Same as above.

test/avro_schema_test.cc
```diff
@@ -1057,4 +1059,454 @@ TEST(AvroSchemaProjectionTest, ProjectDecimalIncompatible) {
...
+      {"name": "id", "type": "int"},
+      {"name": "name", "type": "string"}
+    ]
+  })";
+  auto avro_schema = ::avro::compileJsonSchemaFromString(avro_schema_json);
+
+  auto result = MakeAvroNodeWithFieldIds(avro_schema.root(), *name_mapping);
+  ASSERT_THAT(result, IsError(ErrorKind::kInvalidSchema));
+  ASSERT_THAT(result,
+              HasErrorMessage("Field ID is missing for field 'name' in nested mapping"));
```
wgtmac: TBH, I don't think this is error because it is acceptable to have some fields without field-ids after applying name mappings.

MisterRaindrop: Ok, I will fix

src/iceberg/avro/avro_schema_util.cc
MisterRaindrop: It's adding comment, right ? review too many submission, sometimes I forget. so sorry, 😂

MisterRaindrop:
please review it thanks

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,247 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+    return InvalidSchema("Array type must have exactly one leaf");
+  }
+
+  auto new_array_node = std::make_shared<::avro::NodeArray>();
+
+  // Check if this is a map represented as array
+  if (HasMapLogicalType(original_node)) {
+    ICEBERG_ASSIGN_OR_RAISE(auto new_element_node,
+                            MakeAvroNodeWithFieldIds(original_node->leafAt(0), field));
+    new_array_node->addLeaf(new_element_node);
```
wgtmac, line 861: We need to check `if (original_node->customAttributes() > 0)` and add it to `new_array_node`.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,247 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto new_element_node,
+      MakeAvroNodeWithFieldIds(original_node->leafAt(0), element_field));
+  new_array_node->addLeaf(new_element_node);
+
+  // Add element field ID as custom attribute
+  ::avro::CustomAttributes element_attributes;
+  element_attributes.addAttribute(std::string(kElementIdProp),
```
wgtmac: Ditto, we may need to merge attributes from the original node if exist.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,247 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+  // Preserve existing custom attributes from the original node and add field ID
+  // attributes Copy existing attributes from the original node (if any)
+  if (original_node->customAttributes() > 0) {
+    const auto& original_attrs = original_node->customAttributesAt(0);
+    const auto& existing_attrs = original_attrs.attributes();
+    for (const auto& attr_pair : existing_attrs) {
+      // Copy each existing attribute to preserve original metadata
+      ::avro::CustomAttributes attributes;
+      attributes.addAttribute(attr_pair.first, attr_pair.second, /*addQuote=*/false);
+      new_map_node->addCustomAttributesForField(attributes);
```
wgtmac: I think this is wrong. Perhaps existing test cases do not cover this so they are not revealed. We need to merge `original_node->customAttributesAt(0)` to `key_attributes`, and merge `original_node->customAttributesAt(1)` to `value_attributes`, if they exist.

MisterRaindrop:
@wgtmac you can start review it, thanks

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,301 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+    new_array_node->addLeaf(new_element_node);
+
+    // Check and add custom attributes
+    if (original_node->customAttributes() > 0) {
+      ::avro::CustomAttributes merged_attributes;
+      const auto& original_attrs = original_node->customAttributesAt(0);
+      const auto& existing_attrs = original_attrs.attributes();
+      for (const auto& attr_pair : existing_attrs) {
+        // Skip element-id as we might set it differently
+        if (attr_pair.first != kElementIdProp) {
```
wgtmac: Why do we need this check? It shouldn't happen.

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,301 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+                                 std::to_string(*element_field.field_id),
+                                 /*addQuote=*/false);
+
+  // Then merge any custom attributes from original node (except element-id)
+  if (original_node->customAttributes() > 0) {
+    const auto& original_attrs = original_node->customAttributesAt(0);
+    const auto& existing_attrs = original_attrs.attributes();
+    for (const auto& attr_pair : existing_attrs) {
+      // Skip element-id as we've already set it above
+      if (attr_pair.first != kElementIdProp) {
```
wgtmac: ditto, this check shouldn't happen

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,301 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+                                /*addQuote=*/false);
+
+  // Merge custom attributes from original node if they exist
+  if (original_node->customAttributes() > 0) {
+    // Merge attributes for key (index 0)
+    const auto& original_key_attrs = original_node->customAttributesAt(0);
+    const auto& existing_key_attrs = original_key_attrs.attributes();
+    for (const auto& attr_pair : existing_key_attrs) {
+      // Skip if it's the key ID property we're already setting
+      if (attr_pair.first != kKeyIdProp) {
```
wgtmac: ditto

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -773,4 +775,301 @@ Result<SchemaProjection> Project(const Schema& expected_schema,
...
+      }
+    }
+
+    // Merge attributes for value (index 1)
+    if (original_node->customAttributes() > 1) {
+      const auto& original_value_attrs = original_node->customAttributesAt(1);
+      const auto& existing_value_attrs = original_value_attrs.attributes();
+      for (const auto& attr_pair : existing_value_attrs) {
+        // Skip if it's the value ID property we're already setting
+        if (attr_pair.first != kValueIdProp) {
```
wgtmac: ditto.

BTW, do you think it worths adding a function to copy all custom attributes to a new one to avoid these repeated lines?

MisterRaindrop: @wgtmac  I see there are multiple similar parts in the code. I think it's worth adding a function. What about you? If it's okay, I will add a function.

wgtmac: Yes, I think it makes sense. Thanks!

MisterRaindrop:
@wgtmac It's ready to review

wgtmac (approved):
LGTM. Thanks @MisterRaindrop!

wgtmac:
@Fokko @zeroshade Could you please take a look? I think this is ready to merge.

zeroshade (approved):

PR #130, opened 2025-06-27
refactor: replace std::any placeholder with Literal

wgtmac:
Literal can be used for partition values and default values

### DIFF
--- a/src/iceberg/manifest_entry.h
+++ b/src/iceberg/manifest_entry.h
@@ -19,7 +19,6 @@
 
 #pragma once
 
-#include <any>
 #include <cstdint>
 #include <map>
 #include <memory>
@@ -27,6 +26,7 @@
 #include <string>
 #include <vector>
 
+#include "iceberg/expression/literal.h"
 #include "iceberg/file_format.h"
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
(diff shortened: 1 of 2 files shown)


wgtmac:
I've removed `std::any` and replace them with Literal. Let me know what you think. @zhjwpku @mapleFU

src/iceberg/manifest_entry.h
```diff
@@ -77,9 +77,8 @@ struct ICEBERG_EXPORT DataFile {
   FileFormatType file_format;
   /// Field id: 102
   /// Partition data tuple, schema based on the partition spec output using partition
-  /// field ids for the struct field ids
-  /// TODO(zhjwpku): use StructLike to represent partition data tuple
-  std::any partition;
+  /// field ids
+  std::vector<Literal> partition;
```
mapleFU, line 81: This is ok to me, and should we extract a `Literals` or `StructLike` type using `Vector<Literal>`, or just leave it here?

mapleFU (approved):

src/iceberg/manifest_entry.h
wgtmac: I think we can add an optimized data structure for single value partition later or simply add something like https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/small_vector.h

zhjwpku (approved):
LGTM, thanks

Fokko (approved):

Fokko:
Looks good @wgtmac, thanks for the review @lidavidm, @zhjwpku and @mapleFU

PR #132, opened 2025-06-29
feat: implement transform ResultType

zhjwpku:


wgtmac:
```
  Connection #0 to host www.apache.org left intact

  Issue another request to this URL:
  'https://archive.apache.org/dist/thrift/0.20.0/thrift-0.20.0.tar.gz'

  timeout on name lookup is not supported

  Host archive.apache.org:443 was resolved.
```

We have seen quite a lot timeout from downloading thrift-0.20.0. Should we fix this in the Arrow repo to use more stable URL? @lidavidm @raulcd

wgtmac:
cc @gty404

src/iceberg/transform_function.cc
```diff
@@ -48,7 +48,27 @@ Result<ArrowArray> BucketTransform::Transform(const ArrowArray& input) {
...
+    case TypeId::kDecimal:
+    case TypeId::kDate:
+    case TypeId::kTime:
+    case TypeId::kTimestamp:
+    case TypeId::kTimestampTz:
+    case TypeId::kString:
+    case TypeId::kUuid:
+    case TypeId::kFixed:
+    case TypeId::kBinary:
+      return std::make_shared<IntType>();
```
gty404: Do you need to define a static variable to avoid creating a new object every time, or add a singleton to the corresponding type? The following return values are also similar.

zhjwpku: I've always wanted to add a singleton for each primitive type. If we reach a consensus, I can create a separate PR to implement that. WDYT @wgtmac @lidavidm @gty404 @mapleFU @Fokko @raulcd

lidavidm: +1 Sorry - I had also meant to but as you can see recently I don't have the time.

zhjwpku: Changed to the singleton way, please take a look. @gty404

gty404 (approved):
The type objects in test can also be modified to singleton.

zhjwpku:
> The type objects in test can also be modified to singleton.

Yeah, that's an oversight, fixed.

src/iceberg/transform_function.cc
```diff
@@ -48,7 +48,27 @@ Result<ArrowArray> BucketTransform::Transform(const ArrowArray& input) {
 }
 
 Result<std::shared_ptr<Type>> BucketTransform::ResultType() const {
-  return NotImplemented("BucketTransform::result_type");
+  auto src_type = source_type();
+  if (!src_type) {
+    return NotSupported("null is not a valid input type for bucket transform");
```
wgtmac, line 61: Will this ever happen? cc @gty404

wgtmac: Should we use `Missing source type for xx transform` as the error message?

gty404: I think it will not happen, and for Transform, the original columns must exist.

zhjwpku: I'm not sure, the `Transform::Bind` doesn't check the `source_type`, so it seems possible that `TransformFunction` can has a nullptr as it's source type.

Should we error out in `Transform::Bind`?

wgtmac: I agree. I think it makes sense to fix `Transform::Bind` by not directly calling `std::make_unique<XYZTransform>(source_type)`. Instead we can use `XYZTransform::Make(source_type)` and return error for invalid `source_type`.

src/iceberg/transform_function.cc
```diff
@@ -48,7 +52,32 @@ Result<ArrowArray> BucketTransform::Transform(const ArrowArray& input) {
 }
 
 Result<std::shared_ptr<Type>> BucketTransform::ResultType() const {
-  return NotImplemented("BucketTransform::result_type");
+  auto src_type = source_type();
+  switch (src_type->type_id()) {
```
wgtmac: Shouldn't we check this in the Make function and blindly return `int32()` here? We can add `static bool BucketTransform::Accepts(const std::shared_ptr<Type>& source_type)`. I'm open to discuss this but not a blocker for this PR.

zhjwpku: True, changed to suggested, I didn't not add `Accepts` though.

zeroshade (approved):

PR #139, opened 2025-07-05
chore: bump C++ standard to 23

zhjwpku:


zhjwpku:
Waiting https://github.com/apache/arrow/issues/46998 to be resolved, but should be fine for review.

lidavidm (approved):

zhjwpku:
> This is no longer a draft, right?

Yeah, I just make it ready for review.

zhjwpku:
> So should `Result<>` be removed after this is introduced?

Personally I think Result/Status are well-structured wrappers, but I'm glad to hear others' opinions.

> 
> Besides, C++23 has resize_and_overwrite, allowing zero-cost buffer initialization

Ah, I wasn't aware of this feature, could you suggest where it might be applied?

I think we can adopt more C++23 features in future PRs, assuming compiler support. I tried using `views::join_with` to replace `JoinByDot`, but it's only supported starting from Clang 21 (see [1]), so I gave it up.

[1] https://libcxx.llvm.org/Status/Cxx23.html

wgtmac:
> I just changed to a more recent arrow commit which contains the Apache Thrift update[1], which should fix our frequent CI failure.
> 
> [1] [apache/arrow#46912](https://github.com/apache/arrow/pull/46912)

I think we need to update it to use Apache Arrow 21.0.0 once it is released. It is only a temporary state to pin to an unreleased version.

zhjwpku:
> > I just changed to a more recent arrow commit which contains the Apache Thrift update[1], which should fix our frequent CI failure.
> > [1] [apache/arrow#46912](https://github.com/apache/arrow/pull/46912)
> 
> I think we need to update it to use Apache Arrow 21.0.0 once it is released. It is only a temporary state to pin to an unreleased version.

Sure, will keep an eye on it.

Xuanwo (approved):
Thank you for working on this, let's move!

PR #140, opened 2025-07-05
feat: add spdlog dependency to implement logging

wgtmac:
Added `spdlog` as a third-party dependency to implement logger support. My plan is to add a logger interface so we don't expose `spdlog` symbols externally (just like what we did with `nanoarrow` and `nlohman-json`).

### DIFF
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -283,6 +283,55 @@ function(resolve_nlohmann_json_dependency)
 endfunction()
 
 # ----------------------------------------------------------------------
+# spdlog
+
+function(resolve_spdlog_dependency)
+  prepare_fetchcontent()
+
+  find_package(Threads REQUIRED)
+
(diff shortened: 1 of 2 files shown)


Kelvinyu1117:
I would like to suggest another alternative:
https://github.com/odygrd/quill

I used this logger professionally for ultra low latency application, maybe we can have a look.

wgtmac:
Perhaps we can design a pluggable logger interface so that downstream projects can plug in their own logger of choice.

lidavidm (approved):
Adding the logger interface makes sense. I'm not sure this library will log so much that logger performance will be a concern?

zhjwpku:
+1 for adding the logger interface. 
It seems easier for iceberg, but for iceberg-bundle, the logger needs to handle not just iceberg but also libraries like Arrow, Avro, and others that have their own logging interfaces.

mapleFU (approved):

zhjwpku (approved):

yingcai-cy (approved):

PR #143, opened 2025-07-08
feat: add manifest list reader

dongxiao1198:
- Add manifest list reader
- Integrate with avro reader
- Add simple ut


src/iceberg/manifest_reader.h
```diff
@@ -34,19 +34,17 @@ namespace iceberg {
 /// \brief Read manifest entries from a manifest file.
 class ICEBERG_EXPORT ManifestReader {
  public:
-  virtual Result<std::span<std::unique_ptr<ManifestEntry>>> Entries() const = 0;
+  virtual Result<std::vector<std::unique_ptr<ManifestEntry>>> Entries() const = 0;
 
- private:
-  std::unique_ptr<Reader> reader_;
+  static std::shared_ptr<ManifestReader> NewReader(std::unique_ptr<Reader> reader);
```
wgtmac:
```suggestion
  static std::unique_ptr<ManifestReader> MakeReader(std::unique_ptr<Reader> reader);
```

It is a convention to use Make instead of New, Create, etc.

wgtmac: BTW, should we create the reader internally using ReaderFactory? We know the manifest schema and file format. Sometimes it is useful to directly return Arrow arrays of manifest entries. Therefore I propose the following interface and make them all static functions:

```cpp
public:
  static Result<std::unique_ptr<Reader>> MakeReader(
      const std::string& manifest_list_location,
      FileIO& file_io,
      std::shared_ptr<Schema> schema = kManifestSchema);

  static Result<std::vector<std::unique_ptr<ManifestEntry>>> ReadEntries(
      const std::string& manifest_list_location,
      FileIO& file_io);
};
```

src/iceberg/manifest_reader.h
```diff
@@ -34,19 +34,17 @@ namespace iceberg {
...
 
 /// \brief Read manifest files from a manifest list file.
 class ICEBERG_EXPORT ManifestListReader {
  public:
-  virtual Result<std::span<std::unique_ptr<ManifestFile>>> Files() const = 0;
+  virtual Result<std::vector<std::unique_ptr<ManifestFile>>> Files() const = 0;
 
- private:
-  std::unique_ptr<Reader> reader_;
+  static std::shared_ptr<ManifestListReader> NewReader(std::unique_ptr<Reader> reader);
```
wgtmac: ditto

src/iceberg/manifest_reader.h
wgtmac: For ReadEntries, I also think `std::vector<ManifestEntry>` looks better than `std::vector<std::unique_ptr<ManifestEntry>>`.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,244 @@
...
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+#define ARROW_RETURN_IF_NOT_OK(status, error)                      \
+  if (status != NANOARROW_OK) {                                    \
+    return InvalidArrowData("NanoArrow error: {}", error.message); \
```
wgtmac:
```suggestion
    return InvalidArrowData("Nanoarrow error: {}", error.message); \
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,244 @@
...
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+#define ARROW_RETURN_IF_NOT_OK(status, error)                      \
+  if (status != NANOARROW_OK) {                                    \
```
wgtmac:
```suggestion
  if (status != NANOARROW_OK) [[unlikely]] {                                    \
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,244 @@
...
+  return {};
+}
+
+Result<std::vector<std::unique_ptr<ManifestFile>>> ManifestListReaderImpl::Files() const {
+  std::vector<std::unique_ptr<ManifestFile>> manifest_files;
+  auto arrow_schema = reader_->Schema();
+  if (!arrow_schema.has_value()) {
+    return InvalidArgument("Get schema failed in reader:{}",
+                           arrow_schema.error().message);
+  }
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(auto arrow_schema, reader_->Schema());
```

nit: you may simply write the above

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,244 @@
...
+  internal::ArrowSchemaGuard schema_guard(&arrow_schema.value());
+  auto schema = FromArrowSchema(arrow_schema.value(), std::nullopt);
```
wgtmac: We should pass this schema explicitly when creating the reader instead of converting it back here.

src/iceberg/manifest_reader.h
dongxiao1198: Agree, user should not care of creating the file reader or stream. I will make a refactor of those code.

dongxiao1198: > For ReadEntries, I also think `std::vector<ManifestEntry>` looks better than `std::vector<std::unique_ptr<ManifestEntry>>`.

Manifest list schema is almost a fixed schema, i think there is no need to add it to interface
  static Result<std::shared_ptr<ManifestListReader>> MakeReader(
      const std::string& manifest_list_location, std::shared_ptr<FileIO> file_io);


But for manifest EnteryReader , the schema was based on partition type,  i think the way like this is much better:
  static Result<std::shared_ptr<ManifestReader>> MakeReader(
      const std::string& manifest_location, std::shared_ptr<FileIO> file_io,
      std::shared_ptr<Schema> partition_schema);

src/iceberg/manifest_reader.h
```diff
@@ -34,19 +34,20 @@ namespace iceberg {
 /// \brief Read manifest entries from a manifest file.
 class ICEBERG_EXPORT ManifestReader {
  public:
-  virtual Result<std::span<std::unique_ptr<ManifestEntry>>> Entries() const = 0;
+  virtual Result<std::vector<ManifestEntry>> Entries() const = 0;
 
- private:
-  std::unique_ptr<Reader> reader_;
+  static Result<std::shared_ptr<ManifestReader>> MakeReader(
```
wgtmac:
```suggestion
  static Result<std::unique_ptr<ManifestReader>> MakeReader(
```

Users can cast to shared_ptr from unique_ptr but cannot from the inverse direction.

src/iceberg/manifest_reader.h
```diff
@@ -34,19 +34,20 @@ namespace iceberg {
...
 
 /// \brief Read manifest files from a manifest list file.
 class ICEBERG_EXPORT ManifestListReader {
  public:
-  virtual Result<std::span<std::unique_ptr<ManifestFile>>> Files() const = 0;
+  virtual Result<std::vector<ManifestFile>> Files() const = 0;
 
- private:
-  std::unique_ptr<Reader> reader_;
+  static Result<std::shared_ptr<ManifestListReader>> MakeReader(
```
wgtmac:
```suggestion
  static Result<std::unique_ptr<ManifestListReader>> MakeReader(
```

src/iceberg/manifest_reader_internal.h
```diff
@@ -0,0 +1,57 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/internal/manifest_reader_internal.h
+/// Reader implement for manifest files.
```
wgtmac:
```suggestion
/// Reader implementation for manifest list files and manifest files.
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/type.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+#define ARROW_RETURN_IF_NOT_OK(status, error)                      \
```
wgtmac:
```suggestion
#define NANOARROW_RETURN_IF_NOT_OK(status, error)                      \
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+  if (status != NANOARROW_OK) [[unlikely]] {                       \
+    return InvalidArrowData("Nanoarrow error: {}", error.message); \
```
wgtmac, line 38: Please also include the error code in the message.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+      internal::ArrowArrayGuard array_guard(&result.value().value());
+      auto parse_result =
+          ParseManifestListEntry(&arrow_schema, &result.value().value(), *schema_);
+      if (!parse_result.has_value()) {
+        return InvalidArgument("Failed to parse manifest list entry:{}",
+                               parse_result.error().message);
+      }
+      manifest_files.insert(manifest_files.end(),
+                            std::make_move_iterator(parse_result.value().begin()),
+                            std::make_move_iterator(parse_result.value().end()));
```
wgtmac:
```suggestion
    ICEBERG_ASSIGN_OR_RAISE(auto next_data, reader_->Next());
    if (next_data.has_value()) {
      auto arrow_data = std::move(next_data.value());
      internal::ArrowArrayGuard array_guard(&arrow_data);
      ICEBERG_ASSIGN_OR_RAISE(auto entries,
          ParseManifestListEntry(&arrow_schema, &arrow_data, *schema_));
      manifest_files.insert(manifest_files.end(),
                            std::make_move_iterator(entries.begin()),
                            std::make_move_iterator(entries.end()));
```

I don't these errors should be wrapped as `InvalidArgument`. Let's use `ICEBERG_ASSIGN_OR_RAISE` to simplify them.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+  status = ArrowArrayViewValidate(&array_view, NANOARROW_VALIDATION_LEVEL_FULL, &error);
+  ARROW_RETURN_IF_NOT_OK(status, error);
+
+  std::vector<ManifestFile> manifest_files;
+  manifest_files.resize(array_in->length);
+
+  for (int64_t idx = 0; idx < array_in->n_children; idx++) {
+    const auto& field = iceberg_schema.GetFieldByIndex(idx);
+    if (!field.has_value()) {
+      return InvalidArgument("Field not found in schema: {}", idx);
```
wgtmac:
```suggestion
      return InvalidSchema("Field index {} is not found in schema", idx);
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+    }
+    auto field_name = field.value().get().name();
+    auto view_of_column = array_view.children[idx];
+
+#define PARSE_PRIMITIVE_FIELD(field_name, type)                           \
```
wgtmac: Can we use template (together with concepts) instead of marco to be type safe? This marco only applies to integer values due to `ArrowArrayViewGetIntUnsafe`.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+  for (size_t row_idx = 0; row_idx < view_of_column->length; row_idx++) { \
+    if (!ArrowArrayViewIsNull(view_of_column, row_idx)) {                 \
+      auto value = ArrowArrayViewGetIntUnsafe(view_of_column, row_idx);   \
+      manifest_files[row_idx].field_name = static_cast<type>(value);      \
+    }                                                                     \
+  }
+
+    if (field_name == ManifestFile::kManifestPath.name()) {
```
wgtmac: String comparison would be slow. Can we simply use switch case like below:

```
switch (idx) {
  case 0: 
    // deal with ManifestFile::kManifestPath
  case 1: 
    // deal with ManifestFile::kManifestLength
  ...
}
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+      PARSE_PRIMITIVE_FIELD(added_rows_count, int64_t);
+    } else if (field_name == ManifestFile::kExistingRowsCount.name()) {
+      PARSE_PRIMITIVE_FIELD(existing_rows_count, int64_t);
+    } else if (field_name == ManifestFile::kDeletedRowsCount.name()) {
+      PARSE_PRIMITIVE_FIELD(deleted_rows_count, int64_t);
+    } else if (field_name == ManifestFile::kPartitions.name()) {
+      // view_of_column is list<struct<PartitionFieldSummary>>
+      auto manifest_count = view_of_column->length;
+      if (view_of_column->storage_type != ArrowType::NANOARROW_TYPE_LIST) {
+        return InvalidArgument("partitions field should be a list.");
```
wgtmac: `InvalidArgument` usually deals with invalid user input to a function. We might need to use `InvalidSchema` or even add a new `InvalidManifest` to be more specific. Same for other places.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+  for (size_t row_idx = 0; row_idx < view_of_column->length; row_idx++) { \
+    if (!ArrowArrayViewIsNull(view_of_column, row_idx)) {                 \
+      auto value = ArrowArrayViewGetIntUnsafe(view_of_column, row_idx);   \
+      manifest_files[row_idx].field_name = static_cast<type>(value);      \
+    }                                                                     \
+  }
+
+    if (field_name == ManifestFile::kManifestPath.name()) {
+      for (size_t row_idx = 0; row_idx < view_of_column->length; row_idx++) {
+        if (!ArrowArrayViewIsNull(view_of_column, row_idx)) {
```
wgtmac: We may need to return error if a required field is null. Same for other places.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,233 @@
...
+      PARSE_PRIMITIVE_FIELD(existing_files_count, int32_t);
+    } else if (field_name == ManifestFile::kDeletedFilesCount.name()) {
+      PARSE_PRIMITIVE_FIELD(deleted_files_count, int32_t);
+    } else if (field_name == ManifestFile::kAddedRowsCount.name()) {
+      PARSE_PRIMITIVE_FIELD(added_rows_count, int64_t);
+    } else if (field_name == ManifestFile::kExistingRowsCount.name()) {
+      PARSE_PRIMITIVE_FIELD(existing_rows_count, int64_t);
+    } else if (field_name == ManifestFile::kDeletedRowsCount.name()) {
+      PARSE_PRIMITIVE_FIELD(deleted_rows_count, int64_t);
+    } else if (field_name == ManifestFile::kPartitions.name()) {
```
wgtmac: What about adding a separate function for the logic in this branch?

test/manifest_list_reader_test.cc
```diff
@@ -0,0 +1,128 @@
...
+    TempFileTestBase::SetUp();
+    local_fs_ = std::make_shared<::arrow::fs::LocalFileSystem>();
+    file_io_ = std::make_shared<iceberg::arrow::ArrowFileSystemFileIO>(local_fs_);
+  }
+
+  std::shared_ptr<::arrow::fs::LocalFileSystem> local_fs_;
+  std::shared_ptr<FileIO> file_io_;
+};
+
+TEST_F(ManifestListReaderTest, BasicTest) {
```
wgtmac, line 49: We need to add `operator==` overloads to `ManifestFile` and `ManifestEntry` to make test cases of reading manifests simpler and shorter. These can be in a separate PR.

wgtmac (changes requested):

wgtmac:
Could you help review this? @lidavidm @zhjwpku

src/iceberg/manifest_reader_internal.cc
dongxiao1198: there is no code in ArrowError

src/iceberg/manifest_reader_internal.cc
dongxiao1198: we got only 3 getters for primitive types from nanoarrow
ArrowArrayViewGetIntUnsafe
ArrowArrayViewGetUIntUnsafe
ArrowArrayViewGetDoubleUnsafe
since manifest has no float and uint, it is ok here

nanoarrow cast all int type(int32 int64 int8) to int64..

src/iceberg/manifest_reader_internal.cc
wgtmac: I mean `status` here.

test/manifest_list_reader_test.cc
dongxiao1198: got it

src/iceberg/manifest_reader_internal.cc
dongxiao1198: logical to parse std::vector<PartitionFieldSummary> partitions is really, i will make a new function for it.

src/iceberg/arrow_c_data_guard_internal.h
lidavidm, line 1: Maybe we want to evaluate Nanoarrow eventually? As it has these helpers

test/manifest_list_reader_test.cc
```diff
@@ -0,0 +1,128 @@
...
+  auto manifest_reader_result = ManifestListReader::MakeReader(path, file_io_);
+  ASSERT_EQ(manifest_reader_result.has_value(), true);
+  auto manifest_reader = std::move(manifest_reader_result.value());
+  auto read_result = manifest_reader->Files();
+  ASSERT_EQ(read_result.has_value(), true);
+  ASSERT_EQ(read_result.value().size(), 4);
+  std::string test_dir_prefix = "/tmp/db/db/iceberg_test/metadata/";
+  for (const auto& file : read_result.value()) {
+    auto manifest_path = file.manifest_path.substr(test_dir_prefix.size());
+    if (manifest_path == "2bccd69e-d642-4816-bba0-261cd9bd0d93-m0.avro") {
```
lidavidm, line 61: Maybe we also need to assert that we actually see each of these manifests exactly once?

src/iceberg/manifest_reader_internal.cc
lidavidm: Has this been addressed? Nanoarrow uses errno error codes so you can get the message that way

lidavidm (approved):
No major comments

test/manifest_list_reader_test.cc
dongxiao1198: I will implement the operator== and validator of manifest list in my next pr to check the file

src/iceberg/manifest_reader.h
```diff
@@ -35,35 +35,32 @@ namespace iceberg {
 class ICEBERG_EXPORT ManifestReader {
  public:
   virtual ~ManifestReader() = default;
-  virtual Result<std::span<std::unique_ptr<ManifestEntry>>> Entries() const = 0;
+  virtual Result<std::vector<ManifestEntry>> Entries() const = 0;
 
- private:
-  std::unique_ptr<Reader> reader_;
+  static Result<std::unique_ptr<ManifestReader>> MakeReader(
+      const std::string& manifest_location, std::shared_ptr<FileIO> file_io,
```
wgtmac:
```suggestion
      std::string_view manifest_location, std::shared_ptr<FileIO> file_io,
```

src/iceberg/manifest_reader.h
```diff
@@ -35,35 +35,32 @@ namespace iceberg {
...
 class ICEBERG_EXPORT ManifestListReader {
  public:
   virtual ~ManifestListReader() = default;
-  virtual Result<std::span<std::unique_ptr<ManifestFile>>> Files() const = 0;
+  virtual Result<std::vector<ManifestFile>> Files() const = 0;
 
- private:
-  std::unique_ptr<Reader> reader_;
+  static Result<std::unique_ptr<ManifestListReader>> MakeReader(
+      const std::string& manifest_list_location, std::shared_ptr<FileIO> file_io);
```
wgtmac:
```suggestion
      std::string_view manifest_list_location, std::shared_ptr<FileIO> file_io);
```

src/iceberg/manifest_reader.h
```diff
@@ -35,35 +35,32 @@ namespace iceberg {
...
 };
 
 /// \brief Creates a reader for the manifest list.
 /// \param file_path Path to the manifest list file.
 /// \return A Result containing the reader or an error.
 Result<std::unique_ptr<ManifestListReader>> CreateManifestListReader(
-    std::string_view file_path) {
-  return NotImplemented("CreateManifestListReader is not implemented yet.");
-}
+    std::string_view file_path);
```
wgtmac: We need to remove `CreateManifestListReader` and `CreateManifestReader`, then use above static functions in the table_scan.cc as they are used there.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,251 @@
...
+        break;
+      case 15:
+        PARSE_PRIMITIVE_FIELD(first_row_id, int64_t);
+        break;
+      default:
+        return InvalidManifestList("Unsupported type: {}", field_name);
+    }
+  }
+  return manifest_files;
+}  // namespace iceberg
```
wgtmac:
```suggestion
}
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -0,0 +1,251 @@
...
+    if (!field.has_value()) {
+      return InvalidSchema("Field index {} is not found in schema", idx);
+    }
+    auto field_name = field.value().get().name();
+    bool required = !field.value().get().optional();
+    auto view_of_column = array_view.children[idx];
+
+#define PARSE_PRIMITIVE_FIELD(item, type)                                               \
+  for (size_t row_idx = 0; row_idx < view_of_column->length; row_idx++) {               \
+    if (!ArrowArrayViewIsNull(view_of_column, row_idx)) {                               \
```
wgtmac, line 140: There is an issue with regard to the definition of `ManifestFile`. If a field is optional, we need to choose from the below options:

1. Define it as `std::optional<T>` in the `ManifestFile`, then we don't need to do anything if the read value is null. For example, `added_files_count` is defined in this way.
2. Define it as `T`, then we need to assign a default value depending on its meaning when the read value is null. For example, `content` and `sequence_number` are defined in this way.

I think we need to address this to avoid any potential headache in the future. We can fix this in a followup PR.

Any suggestion? @lidavidm @zhjwpku @gty404 @mapleFU

wgtmac (approved):

src/iceberg/manifest_reader_internal.cc
dongxiao1198: maybe specific the version of manifest&manifest list in v1|v2|v3 is better to valid the schema of file

gty404: To address the issue of compatibility between versions, I understand that there are two approaches:

1. The io layer is only responsible for executing deserialization accurately without making excessive assumptions. Compatibility is handled by the upper layer, as it has more information to assist in interpreting missing fields, such as the table format version existing in the table metadata.

2. During the underlying deserialization, supplement missing fields to be transparent to the upper layer. This requires that new fields all have reasonable default values, so that the upper layer always sees the latest version of the meta.

Both approaches require that the supplementation of missing values be as converged as possible to the same place, to avoid each logic that consumes the field needing to handle compatibility issues.

zhjwpku (approved):

gty404 (approved):

Xuanwo (approved):
Thank you @dongxiao1198 for working on this!

PR #144, opened 2025-07-14
refactor: Add SchemaById and SnapshotById to TableMetadata

gty404:


src/iceberg/table_metadata.h
```diff
@@ -123,12 +123,17 @@ struct ICEBERG_EXPORT TableMetadata {
 
   /// \brief Get the current schema, return NotFoundError if not found
   Result<std::shared_ptr<Schema>> Schema() const;
+  /// \brief Get the current schema by ID, return NotFoundError if not found
```
wgtmac, line 126: What about adding some test cases for these new functions?

gty404: Okay, if there are no other questions, I will start adding test cases

PR #145, opened 2025-07-15
feat: basic implement GetFieldByNameCaseInsensitive

mapleFU:


src/iceberg/type.cc
```diff
@@ -28,6 +28,16 @@
...
 namespace iceberg {
 
+namespace {
+bool StringEqualsCaseInsensitive(std::string_view lhs, std::string_view rhs) {
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+  return std::equal(lhs.begin(), lhs.end(), rhs.begin(),
+                    [](char a, char b) { return std::tolower(a) == std::tolower(b); });
+}
```
mapleFU, line 38: the bad thing is that this could only uses ascii, I suggest import icu or utf8 libs ( or even boost::locale?) for this scenerio

PR #150, opened 2025-07-22
feat: support manifest reader

dongxiao1198:
- add manifest reader
- refactor of arrow array parser
- add basic case


src/iceberg/util/string_utils.h
```diff
@@ -0,0 +1,47 @@
...
+
+#pragma once
+
+#include <algorithm>
+#include <ranges>
+#include <string>
+
+namespace iceberg::internal {
+
+class StringUtils {
```
wgtmac: Either add `ICEBERG_EXPORT` or rename to `string_utils_internal.h`. As you have included it in the public file_format.h, I think we need to add `ICEBERG_EXPORT`.

src/iceberg/avro/avro_schema_util_internal.h
```diff
@@ -144,4 +144,6 @@ std::string ToString(const ::avro::LogicalType::Type& logical_type);
 /// \return True if the node has a map logical type, false otherwise.
 bool HasMapLogicalType(const ::avro::NodePtr& node);
 
+void RegisterLogicalTypes();
```
wgtmac: I think eventually we can add a new file `iceberg/avro/register.h` to register logical type, avro reader and avro writer in a single call.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -38,6 +38,121 @@ namespace iceberg {
...
+    if (!ArrowArrayViewIsNull(array_view, row_idx)) {                                   \
+      auto value = ArrowArrayViewGetIntUnsafe(array_view, row_idx);                     \
+      item = static_cast<type>(value);                                                  \
+    } else if (required) {                                                              \
+      return InvalidManifestList("Field {} is required but null at row {}", field_name, \
+                                 row_idx);                                              \
+    }                                                                                   \
+  }
+
+#define PARSE_ENUM_FIELD(item, array_view, type)                                        \
```
wgtmac: Any difference with `PARSE_PRIMITIVE_FIELD` above?

src/iceberg/manifest_reader_internal.cc
```diff
@@ -38,6 +38,121 @@ namespace iceberg {
...
+                                 row_idx);                                              \
+    }                                                                                   \
+  }
+
+#define PARSE_STRING_FIELD(item, array_view)                                            \
+  for (size_t row_idx = 0; row_idx < array_view->length; row_idx++) {                   \
+    if (!ArrowArrayViewIsNull(array_view, row_idx)) {                                   \
+      auto value = ArrowArrayViewGetStringUnsafe(array_view, row_idx);                  \
+      std::string path_str(value.data, value.size_bytes);                               \
+      item = path_str;                                                                  \
```
wgtmac:
```suggestion
      item = std::string(value.data, value.size_bytes);                                                                  \
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -38,6 +38,121 @@ namespace iceberg {
     return InvalidArrowData("Nanoarrow error: {}", error.message); \
   }
 
+#define PARSE_PRIMITIVE_FIELD(item, array_view, type)                                   \
```
wgtmac, line 41: Should we rename it to `PARSE_INTEGER_FIELD`?

src/iceberg/manifest_reader_internal.cc
```diff
@@ -38,6 +38,121 @@ namespace iceberg {
...
+      auto buffer = ArrowArrayViewGetBytesUnsafe(array_view, row_idx);                  \
+      item = std::vector<uint8_t>(buffer.data.as_char,                                  \
+                                  buffer.data.as_char + buffer.size_bytes);             \
+    } else if (required) {                                                              \
+      return InvalidManifestList("Field {} is required but null at row {}", field_name, \
+                                 row_idx);                                              \
+    }                                                                                   \
+  }
+
+#define PARSE_PRIMITIVE_VECTOR_FIELD(item, count, array_view, type)                 \
```
wgtmac: ditto, should we rename it to `PARSE_INTEGER_VECTOR_FIELD`?

src/iceberg/manifest_reader_internal.cc
```diff
@@ -38,6 +38,121 @@ namespace iceberg {
...
+  for (int64_t manifest_idx = 0; manifest_idx < count; manifest_idx++) {            \
+    auto offset = ArrowArrayViewListChildOffset(array_view, manifest_idx);          \
+    auto next_offset = ArrowArrayViewListChildOffset(array_view, manifest_idx + 1); \
+    for (int64_t offset_idx = offset; offset_idx < next_offset; offset_idx++) {     \
+      item.emplace_back(static_cast<type>(                                          \
+          ArrowArrayViewGetIntUnsafe(array_view->children[0], offset_idx)));        \
+    }                                                                               \
+  }
+
+#define PARSE_PRIMITIVE_MAP_FIELD(item, count, array_view)                           \
```
wgtmac: This is restricted to `map<int, long>`? Perhaps use a clearer name?

src/iceberg/manifest_reader_internal.cc
```diff
@@ -38,6 +38,121 @@ namespace iceberg {
...
+      auto next_offset = array_view->buffer_views[1].data.as_int32[row_idx + 1];     \
+      for (int32_t offset_idx = offset; offset_idx < next_offset; offset_idx++) {    \
+        auto key = ArrowArrayViewGetIntUnsafe(view_of_map_key, offset_idx);          \
+        auto value = ArrowArrayViewGetIntUnsafe(view_of_map_value, offset_idx);      \
+        item[key] = value;                                                           \
+      }                                                                              \
+    }                                                                                \
+  } while (0)
+
+#define PARSE_BINARY_MAP_FIELD(item, count, array_view)                              \
```
wgtmac: Same question here.

wgtmac: BTW, I think `PARSE_PRIMITIVE_MAP_FIELD` and `PARSE_BINARY_MAP_FIELD` can be combined into a single macro/template if we make key/value parsing and `item[key]` assignment as function arguments.

src/iceberg/util/string_utils.h
dongxiao1198: it was under internal ns， so i renamed it.

src/iceberg/manifest_reader_internal.cc
dongxiao1198: merged into one macros

src/iceberg/util/string_utils.h
wgtmac: It is included by file_format.h so we cannot rename it to be internal. Otherwise the example build ci will fail due to missing the header file.

src/iceberg/avro/avro_register.h
```diff
@@ -0,0 +1,32 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <avro/LogicalType.hh>
```
wgtmac: We should not expose avro headers. Perhaps define `MapLogicalType` in the avro_schema_util_internal.h?

src/iceberg/avro/avro_register.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace iceberg::avro {
+
+struct MapLogicalType : public ::avro::CustomLogicalType {
+  MapLogicalType() : ::avro::CustomLogicalType("map") {}
+};
+
+void RegisterLogicalTypes();
```
wgtmac: Do we need to add `ICEBERG_BUNDLE_EXPORT`?

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -49,17 +49,12 @@ constexpr std::string_view kValueIdProp = "value-id";
...
-};
-
 ::avro::LogicalType GetMapLogicalType() {
   static std::once_flag flag{};
   std::call_once(flag, []() {
     // Register the map logical type with the avro custom logical type registry.
     // See https://github.com/apache/avro/pull/3326 for details.
-    ::avro::CustomLogicalTypeRegistry::instance().registerType(
-        "map", [](const std::string&) { return std::make_shared<MapLogicalType>(); });
+    RegisterLogicalTypes();
```
wgtmac: Should we move once_flag and call_once into RegisterLogicalTypes?

wgtmac (approved):

src/iceberg/avro/avro_register.cc
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "avro_register.h"
```
gty404:
```suggestion
#include "iceberg/avro/avro_register.h"
```

src/iceberg/manifest_reader_internal.cc
```diff
@@ -134,112 +227,336 @@ Result<std::vector<ManifestFile>> ParseManifestListEntry(ArrowSchema* schema,
...
   while (true) {
-    auto result = reader_->Next();
-    if (!result.has_value()) {
-      return InvalidManifestList("Failed to read manifest list entry:{}",
-                                 result.error().message);
-    }
-    if (result.value().has_value()) {
-      internal::ArrowArrayGuard array_guard(&result.value().value());
+    ICEBERG_ASSIGN_OR_RAISE(auto result, reader_->Next());
+    if (result.has_value()) {
```
gty404, line 553: Here, `result.has_value()` is definitely true.

dongxiao1198: since it was defined in optional<xx> we should check it anyway. When reading at the end of file, it will be nullotp.

PR #153, opened 2025-07-25
feat: add a plugable logging system.

yingcai-cy:


src/iceberg/util/logger.h
```diff
@@ -0,0 +1,285 @@
...
+  static LoggerRegistry& Instance();
+
+  /// \brief Set the default logger implementation
+  ///
+  /// \tparam LoggerImpl The logger implementation type
+  /// \param logger The logger instance to set as default
+  template <typename LoggerImpl>
+  void SetDefaultLogger(std::shared_ptr<LoggerImpl> logger) {
+    static_assert(std::is_base_of_v<LoggerInterface<LoggerImpl>, LoggerImpl>,
+                  "LoggerImpl must inherit from LoggerInterface");
```
lidavidm: This can be a concept right, since we have C++20?

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,285 @@
...
+    default_logger_ = std::static_pointer_cast<void>(logger);
+    log_func_ = [logger](LogLevel level, std::string_view format_str,
+                         const std::string& formatted_message) {
+      if (logger->ShouldLog(level)) {
+        logger->LogImpl(level, "{}", formatted_message);
+      }
+    };
```
lidavidm: Ultimately we still do a dynamic call, does it save us anything versus a virtual call? It is significantly more complicated

yingcai-cy: Yes, function pointers are generally faster than virtual calls, and in some scenarios, the compiler is able to inline them, which can further improve performance.

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,286 @@
...
+  template <typename... Args>
+  void LogWithLocationImpl(LogLevel, std::string_view, const std::source_location&,
+                           Args&&...) const noexcept {}
+
+  void SetLevelImpl(LogLevel) noexcept {}
+  LogLevel GetLevelImpl() const noexcept { return LogLevel::kOff; }
+};
+
+/// \brief Global logger registry for managing logger instances
+class LoggerRegistry {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT LoggerRegistry {
```

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/// \brief Logger interface that uses CRTP to avoid virtual function overhead
+/// \tparam Derived The concrete logger implementation
+template <typename Derived>
+class LoggerInterface {
+ public:
+  /// \brief Check if a log level is enabled
+  bool ShouldLog(LogLevel level) const noexcept {
+    return static_cast<const Derived*>(this)->ShouldLogImpl(level);
```
wgtmac: Should we wrap `static_cast<const Derived*>(this)` into a separate function?

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,286 @@
...
+  kDebug = 1,
+  kInfo = 2,
+  kWarn = 3,
+  kError = 4,
+  kCritical = 5,
+  kOff = 6
+};
+
+/// \brief Convert log level to string representation
+constexpr std::string_view LogLevelToString(LogLevel level) {
```
wgtmac: Add `ICEBERG_EXPORT` if it is supposed to be public.

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,286 @@
...
+class LoggerInterface {
+ public:
+  /// \brief Check if a log level is enabled
+  bool ShouldLog(LogLevel level) const noexcept {
+    return static_cast<const Derived*>(this)->ShouldLogImpl(level);
+  }
+
+  /// \brief Log a message with the specified level
+  template <typename... Args>
+  void Log(LogLevel level, std::string_view format_str, Args&&... args) const {
```
wgtmac: Should we merge it with `LogWithLocation`? IMHO, we need location at all times.

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,286 @@
...
+  void SetLevel(LogLevel level) { static_cast<Derived*>(this)->SetLevelImpl(level); }
+
+  /// \brief Get the current minimum log level
+  LogLevel GetLevel() const noexcept {
+    return static_cast<const Derived*>(this)->GetLevelImpl();
+  }
+
+  // Convenience methods for different log levels
+  template <typename... Args>
+  void Trace(std::string_view format_str, Args&&... args) const {
```
wgtmac:
```suggestion
  void Trace(std::string_view format_str, Args&&... args) {
```

The extensibility will be limited if it and its friends below are const.

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,286 @@
...
+    case LogLevel::kOff:
+      return "OFF";
+    default:
+      return "UNKNOWN";
+  }
+}
+
+/// \brief Logger interface that uses CRTP to avoid virtual function overhead
+/// \tparam Derived The concrete logger implementation
+template <typename Derived>
```
wgtmac, line 69: nit: we can add a concept for Derived like below:

```cpp
template <typename T>
concept LoggerImpl = requires(T& logger, LogLevel level, std::string_view format_str,
                              std::source_location location) {
  { logger.ShouldLog(level) } -> std::convertible_to<bool>;
  { logger.Log(level, format_str, location) } -> std::same_as<void>;
  { logger.SetLevel(level) } -> std::same_as<void>;
  { logger.GetLevel() } -> std::convertible_to<LogLevel>;
};

template <LoggerImpl Derived>
class LoggerInterface {
 ...
};
```

src/iceberg/util/logger.h
```diff
@@ -0,0 +1,286 @@
...
+  LoggerInterface() = default;
+  ~LoggerInterface() = default;
+};
+
+/// \brief Concept to constrain types that implement the Logger interface
+template <typename T>
+concept Logger = std::is_base_of_v<LoggerInterface<T>, T>;
+
+/// \brief Default spdlog-based logger implementation
+class SpdlogLogger : public LoggerInterface<SpdlogLogger> {
```
wgtmac: Can we move `SpdlogLogger` and others to the source files?

evindj:
Hi folks, is this change ready for yet if not I am happy to take a stab at it as it is blocking the loggingmetricsreporter. let  me know if there is more to do here, happy to have a look. @wgtmac @lidavidm

wgtmac:
@evindj Thanks for your interest! I believe @kamcheungting-db is currently working on this so it is better to avoid conflict. We can postpone loggingmetricsreporter once logging has been added.

PR #154, opened 2025-07-25
feat: add scaffolding work for parquet reader

wgtmac:


cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -65,8 +65,9 @@ function(resolve_arrow_dependency)
   set(ARROW_BUILD_STATIC
       ON
       CACHE BOOL "" FORCE)
+  # Workaround undefined symbol: arrow::ipc::ReadSchema(arrow::io::InputStream*, arrow::ipc::DictionaryMemo*)
```
lidavidm: Ah, parquet-cpp can embed Arrow schemas in the Parquet metadata, so we need to be able to read IPC schemas. Possibly upstream should enable/require ARROW_IPC when Parquet is being built.

src/iceberg/parquet/parquet_data_util_internal.h
```diff
@@ -0,0 +1,43 @@
...
+namespace iceberg::parquet {
+
+/// \brief Convert record batch read from a Parquet file to projected Iceberg Schema.
+///
+/// \param record_batch The record batch to convert.
+/// \param output_arrow_schema The Arrow schema to convert to.
+/// \param projected_schema The projected Iceberg schema.
+/// \param projection The projection from projected Iceberg schema to the record batch.
+/// \return The converted record batch.
+Result<std::shared_ptr<::arrow::RecordBatch>> ConvertRecordBatch(
```
lidavidm: nit, but if the job is to project a record batch, maybe just call it `ProjectRecordBatch`?

src/iceberg/result.h
```diff
@@ -93,6 +94,7 @@ DEFINE_ERROR_FUNCTION(NotAllowed)
 DEFINE_ERROR_FUNCTION(NotFound)
 DEFINE_ERROR_FUNCTION(NotImplemented)
 DEFINE_ERROR_FUNCTION(NotSupported)
+DEFINE_ERROR_FUNCTION(ParquetError)
```
lidavidm: Hmm, Parquet itself has some exception classes too; would it be possible to translate errors into IOError or some other error instead of requiring a Parquet-specific class?

That said there's also InvalidArrowData. Should we call this InvalidParquetData to match? (Maybe that's too specific, but perhaps InvalidParquetData + IOError covers everything?)

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -0,0 +1,300 @@
...
+  Result<ArrowSchema> Schema() {
+    if (!context_) {
+      ICEBERG_RETURN_UNEXPECTED(InitReadContext());
+    }
+
+    ArrowSchema arrow_schema;
+    auto export_result =
+        ::arrow::ExportSchema(*context_->output_arrow_schema_, &arrow_schema);
+    if (!export_result.ok()) {
+      return ParquetError("Failed to export Arrow schema: {}", export_result.message());
```
lidavidm: Is this really a ParquetError? (Ditto for ExportRecordBatch above)

lidavidm (approved):
just nits, feel free to ignore

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -0,0 +1,300 @@
...
+#include "iceberg/parquet/parquet_reader.h"
+
+#include <memory>
+
+#include <arrow/c/bridge.h>
+#include <arrow/memory_pool.h>
+#include <arrow/record_batch.h>
+#include <arrow/result.h>
+#include <arrow/type.h>
+#include <iceberg/result.h>
```
zhjwpku: nit: use "" for iceberg headers

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -0,0 +1,300 @@
...
+  if (!has_field_ids) {
+    // TODO(gangwu): apply name mapping to Parquet schema
+    return NotImplemented("Applying name mapping to Parquet schema is not implemented");
+  }
+
+  ::parquet::arrow::SchemaManifest schema_manifest;
+  auto schema_manifest_result = ::parquet::arrow::SchemaManifest::Make(
+      metadata->schema(), metadata->key_value_metadata(), reader->properties(),
+      &schema_manifest);
+  if (!schema_manifest_result.ok()) {
```
zhjwpku: Can we use ICEBERG_ARROW_ASSIGN_OR_RETURN here?

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -0,0 +1,300 @@
...
+      if (row_group_indices.empty()) {
+        // None of the row groups are selected, return an empty record batch reader
+        context_->record_batch_reader_ = std::make_unique<EmptyRecordBatchReader>();
+        return {};
+      }
+    }
+
+    // Create the record batch reader
+    ICEBERG_ASSIGN_OR_RAISE(auto column_indices, SelectedColumnIndices(projection_));
+    auto reader_result = reader_->GetRecordBatchReader(row_group_indices, column_indices);
```
zhjwpku: Use ICEBERG_ARROW_ASSIGN_OR_RETURN to transform arrow/parquet error to iceberg error?(Which might make the code shorter, but I don't insist)

zhjwpku (approved):
Looking good, thanks for working on this.

src/iceberg/result.h
wgtmac: I have thought of similar things before adding this. `InvalidParquetData` sounds better. Let me change this.

zeroshade (approved):

PR #156, opened 2025-07-27
feat: implement Literal Transform

zhjwpku:


src/iceberg/transform_function.cc
```diff
@@ -51,6 +60,57 @@ Result<ArrowArray> BucketTransform::Transform(const ArrowArray& input) {
...
 
+Result<std::optional<Literal>> BucketTransform::Transform(const Literal& literal) {
+  assert(literal.type() == source_type());
+  if (literal.IsBelowMin() || literal.IsAboveMax()) {
+    return InvalidArgument(
+        "Cannot apply bucket transform to literal with value {} of type {}",
+        literal.ToString(), source_type()->ToString());
+  }
+  int32_t hash_value = 0;
+  switch (source_type()->type_id()) {
```
lidavidm: Since we don't actually care about the type, you could `std::visit` the `value()` instead?

src/iceberg/transform_function.cc
```diff
@@ -124,6 +224,34 @@ Result<ArrowArray> YearTransform::Transform(const ArrowArray& input) {
...
+    case TypeId::kDate: {
+      auto value = std::get<int32_t>(literal.value());
+      auto epoch = sys_days(year{1970} / January / 1);
+      auto ymd = year_month_day(epoch + days{value});
+      return Literal::Int(static_cast<int32_t>(ymd.year()));
+    }
+    case TypeId::kTimestamp:
+    case TypeId::kTimestampTz: {
+      auto value = std::get<int64_t>(literal.value());
+      // Convert milliseconds-since-epoch into a `year_month_day` object
```
lidavidm: Aren't timestamps in microseconds?

test/transform_test.cc
```diff
@@ -193,4 +193,296 @@ TEST(TransformResultTypeTest, NegativeCases) {
...
+      {.source_type = iceberg::float64(), .source = Literal::Double(2.71828)},
+      {.source_type = iceberg::string(), .source = Literal::String("Hello, World!")},
+      {.source_type = iceberg::binary(), .source = Literal::Binary({0x01, 0x02, 0x03})},
+  };
+
+  for (const auto& c : cases) {
+    auto transformPtr = transform->Bind(c.source_type);
+    ASSERT_TRUE(transformPtr.has_value()) << "Failed to bind void transform";
+    auto result = transformPtr.value()->Transform(c.source);
+    ASSERT_TRUE(result == std::nullopt)
```
lidavidm: nit, but gtest gives better failure messages if you use `ASSERT_EQ(std::nullopt, result);` for this kind of thing

src/iceberg/transform_function.cc
mapleFU: Seems `literal.type() == source_type()` is enforced?

lidavidm: Right, but in this particular case, the "logical" type doesn't matter, right? The only thing that matters is the "physical" data type, so you could use std::visit + static_assert to enforce that all branches are taken without having to list out each Iceberg type

src/iceberg/transform_function.cc
zhjwpku: You are right, fixed.

test/transform_test.cc
zhjwpku: fixed.

src/iceberg/transform_function.cc
zhjwpku: Ok, I will try this later, thanks for the input.

src/iceberg/transform.h
```diff
@@ -172,6 +174,8 @@ class ICEBERG_EXPORT TransformFunction {
   TransformFunction(TransformType transform_type, std::shared_ptr<Type> source_type);
   /// \brief Transform an input array to a new array
   virtual Result<ArrowArray> Transform(const ArrowArray& data) = 0;
```
wgtmac: Perhaps we can remove this now? cc @gty404

src/iceberg/expression/literal.h
```diff
@@ -63,7 +63,10 @@ class ICEBERG_EXPORT Literal {
   /// \brief Factory methods for primitive types
   static Literal Boolean(bool value);
   static Literal Int(int32_t value);
+  static Literal Date(int32_t value);
   static Literal Long(int64_t value);
+  static Literal Timestamp(int64_t value);
```
wgtmac, line 68: Do we need to add `Time`?

src/iceberg/expression/literal.h
```diff
@@ -85,6 +88,9 @@ class ICEBERG_EXPORT Literal {
   /// \brief Get the literal type.
   const std::shared_ptr<PrimitiveType>& type() const;
 
+  /// \brief Get the literal value.
+  const Value& value() const { return value_; }
```
wgtmac, line 97: `Value` is still private. We might need to make it public.

src/iceberg/expression/literal.cc
```diff
@@ -208,12 +216,30 @@ std::partial_ordering Literal::operator<=>(const Literal& other) const {
       return this_val <=> other_val;
     }
 
+    case TypeId::kDate: {
```
wgtmac, line 212: Should we merge it with the above case? We can add [[fallthrough]] to make it explicit.

src/iceberg/util/macros.h
```diff
@@ -36,3 +36,8 @@
 #define ICEBERG_ASSIGN_OR_RAISE(lhs, rexpr)                                             \
   ICEBERG_ASSIGN_OR_RAISE_IMPL(ICEBERG_ASSIGN_OR_RAISE_NAME(result_, __COUNTER__), lhs, \
                                rexpr)
+
+#define NANOARROW_RETURN_IF_NOT_OK(status, error)                  \
+  if (status != NANOARROW_OK) [[unlikely]] {                       \
```
wgtmac: We shouldn't expose nanoarrow publicly.

src/iceberg/transform.h
zhjwpku: I'm not sure this is a good API, since it requires constructing ArrowSchema on the fly. I think multi-args transforms could be handled in other ways, I'll leave it as is for now.

src/iceberg/expression/literal.h
zhjwpku: done

src/iceberg/expression/literal.h
zhjwpku: Reasonable, done.

src/iceberg/util/macros.h
zhjwpku: Yeah, I move it back :)

src/iceberg/expression/literal.cc
zhjwpku: Good catch, done that way.

src/iceberg/transform_function.cc
zhjwpku: I've implemented this idea in `BucketTransform::Transform`. Since we're not handling all of the variant's alternative types, I added a branch with std::unreachable(). I'm not sure if this is the best approach, please take another look when you have a chance. @lidavidm @mapleFU @wgtmac

lidavidm (approved):

src/iceberg/transform.h
wgtmac: I have to admit that `Result<ArrowArray> Transform(const ArrowArray& data)` is a bad design. It was added at the time when we don't have `Literal` yet. We can safely remove it since it is no longer needed.

src/iceberg/transform_function.cc
```diff
@@ -51,6 +61,45 @@ Result<ArrowArray> BucketTransform::Transform(const ArrowArray& input) {
   return NotImplemented("BucketTransform::Transform");
 }
 
+Result<std::optional<Literal>> BucketTransform::Transform(const Literal& literal) {
+  assert(literal.type() == source_type());
```
wgtmac, line 56: Not related to this PR: we should introduce something like `ARROW_DCHECK` to add some explicit debug-level checks.

src/iceberg/transform_function.cc
```diff
@@ -91,6 +140,46 @@ Result<ArrowArray> TruncateTransform::Transform(const ArrowArray& input) {
...
+      return Literal::Long(value % width_);
+    }
+    case TypeId::kDecimal: {
+      // TODO(zhjwpku): Handle decimal truncation logic here
+      return NotImplemented("Truncate for Decimal is not implemented yet");
+    }
+    case TypeId::kString: {
+      auto value = std::get<std::string>(literal.value());
+      if (value.size() > static_cast<size_t>(width_)) {
+        value.resize(width_);
```
wgtmac: This is not correct. We need to handle UTF8 code point.

src/iceberg/transform_function.cc
```diff
@@ -91,6 +140,46 @@ Result<ArrowArray> TruncateTransform::Transform(const ArrowArray& input) {
...
+  if (literal.IsBelowMin() || literal.IsAboveMax()) {
+    return InvalidArgument(
+        "Cannot apply truncate transform to literal with value {} of type {}",
+        literal.ToString(), source_type()->ToString());
+  }
+
+  switch (source_type()->type_id()) {
+    case TypeId::kInt: {
+      auto value = std::get<int32_t>(literal.value());
+      return Literal::Int(value % width_);
```
wgtmac: This looks incorrect because it is `v - (v % W)` per the spec. We should be consistent with the Java impl:

```java
  public static int truncateInt(int width, int value) {
    return value - (((value % width) + width) % width);
  }

  public static long truncateLong(int width, long value) {
    return value - (((value % width) + width) % width);
  }
```

test/transform_test.cc
```diff
@@ -193,4 +193,297 @@ TEST(TransformResultTypeTest, NegativeCases) {
...
+    Literal source;
+    Literal expected;
+  };
+
+  const std::vector<Case> cases = {
+      {.source_type = iceberg::int32(),
+       .source = Literal::Int(123456),
+       .expected = Literal::Int(1)},
+      {.source_type = iceberg::string(),
+       .source = Literal::String("Hello, World!"),
```
wgtmac, line 302: Please add some non-ASCII UTF8 strings.

test/transform_test.cc
```diff
@@ -193,4 +193,297 @@ TEST(TransformResultTypeTest, NegativeCases) {
...
+
+  struct Case {
+    std::shared_ptr<Type> source_type;
+    Literal source;
+    Literal expected;
+  };
+
+  const std::vector<Case> cases = {
+      {.source_type = iceberg::timestamp(),
+       .source = Literal::Timestamp(1622547800000000),
```
wgtmac, line 344: nit: add a comment to print human readable timestamp values.

test/transform_test.cc
```diff
@@ -193,4 +193,297 @@ TEST(TransformResultTypeTest, NegativeCases) {
   }
 }
 
+TEST(TransformFunctionTransformTest, IdentityTransform) {
```
wgtmac: `TransformFunctionTransformTest` -> `TransformLiteralTest` ?

Let's make it short...

src/iceberg/transform_function.cc
zhjwpku: Agreed, we may incorporate this alongside our logging infrastructure.

src/iceberg/transform_function.cc
zhjwpku: Good catch, I didn't realize that before, please check again if I'm doing it right now.

src/iceberg/transform_function.cc
zhjwpku: Double checked rust impl, what you proposed is the right way, changed.

test/transform_test.cc
zhjwpku: Add some emoji cases.

test/transform_test.cc
zhjwpku: done.

test/transform_test.cc
zhjwpku: done

src/iceberg/transform.h
```diff
@@ -170,8 +172,8 @@ class ICEBERG_EXPORT TransformFunction {
  public:
   virtual ~TransformFunction() = default;
   TransformFunction(TransformType transform_type, std::shared_ptr<Type> source_type);
-  /// \brief Transform an input array to a new array
-  virtual Result<ArrowArray> Transform(const ArrowArray& data) = 0;
+  /// \brief Transform an input Literal to a new Literal
+  virtual Result<std::optional<Literal>> Transform(const Literal& literal) = 0;
```
wgtmac: Does it make sense to introduce a `void` variant to Literal? It does not deserve to pay for an extra optional value check every time when the void transform is a rare case.

test/transform_test.cc
```diff
@@ -193,4 +187,316 @@ TEST(TransformResultTypeTest, NegativeCases) {
...
+       .source = Literal::Int(123456),
+       .expected = Literal::Int(123455)},
+      {.source_type = iceberg::string(),
+       .width = 5,
+       .source = Literal::String("Hello, World!"),
+       .expected = Literal::String("Hello")},
+      {.source_type = iceberg::string(),
+       .width = 5,
+       .source = Literal::String("😜🧐🤔🤪🥳"),
+       // Truncate to 5 bytes, the safe point should be four bytes which fits the first
```
wgtmac: No, truncation should be based on the number of unicode codepoints (a.k.a. emoji symbols) not the number of characters. Please take a look at https://github.com/apache/iceberg/blob/acecf3cb697a7fa6bfcb1b58fc5d7d70914436e7/api/src/main/java/org/apache/iceberg/util/UnicodeUtil.java#L41-L55

lidavidm: I think this will be annoying to do in C++ without pulling in a library like utf8proc (I suppose assuming UTF-8 you can implement just the bit logic to detect a codepoint)

(FWIW: one emoji can be multiple codepoints, IIRC (due to e.g. flags, skin color selectors) An emoji would be a grapheme cluster?)

src/iceberg/transform.h
zhjwpku: I add a NullType and Literal::Null(), I use `Null` because spec says void transform `Always produces null`.
NullType is inherited from PrimitiveType because Literal holds a PrimitiveType shared ptr.

test/transform_test.cc
zhjwpku: I checked https://en.wikipedia.org/wiki/UTF-8, UTF-8 encodes code points in one to four bytes, the first byte of code point always follow `(value[i] & 0xC0) != 0x80`, so I just do some simple math to resize the string.

test/transform_test.cc
```diff
@@ -24,6 +24,7 @@
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
+#include <iceberg/expression/literal.h>
```
wgtmac:
```suggestion
#include "iceberg/expression/literal.h"
```

src/iceberg/type.h
```diff
@@ -191,6 +191,23 @@ class ICEBERG_EXPORT MapType : public NestedType {
 /// Primitive types do not have nested fields.
 /// @{
 
+/// \brief A data type that has no physical storage.
+/// Technically, this is a primitive type, we treat it as a primitive type for
+/// convenience.
+class ICEBERG_EXPORT NullType : public PrimitiveType {
```
wgtmac: I don't think adding `NullType` and `TypeId::kNull` is the right path forward. It is confusing since it does not exist in the spec and may conflict with the purpose of `UnknownType` in v3. I'd suggest to revert changes related to this new type.

src/iceberg/type_fwd.h
```diff
@@ -36,6 +36,8 @@ enum class TypeId {
   kStruct,
   kList,
   kMap,
+  kNull,  // Note: A type having no physical storage. This is not an iceberg type, we add
```
wgtmac: See my comment above.

src/iceberg/transform_function.cc
```diff
@@ -150,17 +150,25 @@ Result<std::optional<Literal>> TruncateTransform::Transform(const Literal& liter
...
-        size_t safe_point = width_;
-        while (safe_point > 0 && (value[safe_point] & 0xC0) == 0x80) {
-          // Find the last valid UTF-8 character boundary before or at width_
-          safe_point--;
+      size_t code_point_count = 0;
+      size_t safe_point = 0;
+
+      for (size_t i = 0; i < value.size(); ++i) {
+        // Start of a new UTF-8 code point
+        if ((value[i] & 0xC0) != 0x80) {
```
wgtmac: nit: add a `iceberg/util/truncate_util.h` so we can test individual functions (utf8, integer, decimal, etc.) better without the boilerplate to create literal and transform.

src/iceberg/transform_function.cc
```diff
@@ -420,8 +428,8 @@ Result<std::unique_ptr<TransformFunction>> HourTransform::Make(
 VoidTransform::VoidTransform(std::shared_ptr<Type> const& source_type)
     : TransformFunction(TransformType::kVoid, source_type) {}
 
-Result<std::optional<Literal>> VoidTransform::Transform(const Literal& literal) {
-  return std::nullopt;
+Result<Literal> VoidTransform::Transform(const Literal& literal) {
+  return Literal::Null();
```
wgtmac: Per the spec, `All transforms must return `null` for a `null` input value`. We need to check if the input literal is a null for every kind of transform.

src/iceberg/type.h
zhjwpku: Yeah, more sense, removed.

src/iceberg/transform_function.cc
zhjwpku: Yeah, I read the transform section again and made some changes accordingly.

src/iceberg/transform_function.cc
zhjwpku: done.

src/iceberg/expression/literal.cc
```diff
@@ -126,12 +126,23 @@ Literal::Literal(Value value, std::shared_ptr<PrimitiveType> type)
     : value_(std::move(value)), type_(std::move(type)) {}
 
 // Factory methods
+
```
wgtmac, line 129:
```suggestion
```

src/iceberg/expression/literal.cc
```diff
@@ -126,12 +126,23 @@ Literal::Literal(Value value, std::shared_ptr<PrimitiveType> type)
...
 Literal Literal::Boolean(bool value) { return {Value{value}, iceberg::boolean()}; }
 
 Literal Literal::Int(int32_t value) { return {Value{value}, iceberg::int32()}; }
 
+Literal Literal::Date(int32_t value) { return {Value{value}, iceberg::date()}; }
```
wgtmac: nit: we don't need `iceberg::` namespace prefix.

src/iceberg/transform_function.h
```diff
@@ -97,11 +97,11 @@ class YearTransform : public TransformFunction {
   /// \param source_type Must be a timestamp type.
   explicit YearTransform(std::shared_ptr<Type> const& source_type);
 
-  /// \brief Extracts the year from each timestamp in the input array.
-  Result<ArrowArray> Transform(const ArrowArray& input) override;
+  /// \brief Extract a date or timestamp year, as years from 1970.
```
wgtmac, line 100: The comment is incorrect.

src/iceberg/transform_function.h
```diff
@@ -116,11 +116,11 @@ class MonthTransform : public TransformFunction {
   /// \param source_type Must be a timestamp type.
   explicit MonthTransform(std::shared_ptr<Type> const& source_type);
 
-  /// \brief Extracts the month (1-12) from each timestamp in the input array.
-  Result<ArrowArray> Transform(const ArrowArray& input) override;
+  /// \brief Extract a date or timestamp month, as months from 1970-01-01.
```
wgtmac, line 119: ditto

src/iceberg/transform_function.h
wgtmac: Also uses `Extracts` to be consistent. Same for below.

src/iceberg/util/truncate_utils.h
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    return std::move(source);
+  }
+
+  /// \brief Truncate an integer v, either int32_t or int64_t, to v - (v % W).
+  ///
+  /// The remainder, v % W, must be positive. For languages where % can produce negative
+  /// values, the correct truncate function is: v - (((v % W) + W) % W)
+  template <typename T>
```
wgtmac, line 65: nit: use concepts to limit T to int32_t and int64_t

src/iceberg/transform_function.h
zhjwpku: I agree the comment is not clear, but that's from the spec, see [1]. We might also want to change the spec if it's not clear.

[1] https://iceberg.apache.org/spec/#partition-transforms

zhjwpku: iceberg-rust use the same description[1], so I'm inclined to keep it consistent.

[1] https://github.com/apache/iceberg-rust/blob/main/crates/iceberg/src/transform/temporal.rs#L44

src/iceberg/transform_function.cc
```diff
@@ -148,14 +253,51 @@ Result<std::unique_ptr<TransformFunction>> YearTransform::Make(
...
+  if (literal.IsNull()) [[unlikely]] {
+    return Literal::Null(iceberg::int32());
+  }
 
-Result<std::shared_ptr<Type>> MonthTransform::ResultType() const {
-  return iceberg::int32();
+  using namespace std::chrono;  // NOLINT
+  switch (source_type()->type_id()) {
+    case TypeId::kDate: {
+      auto value = std::get<int32_t>(literal.value());
```
wgtmac, line 271: nit: I think we need also move all of these transform functions to a dedicated utility file and add exhaustive test cases. Could you create an issue for this?

wgtmac (approved):
+1 Thanks!

My only concern is that we need to have better test coverage of those transform utility functions. We can improve them in followup PRs as this one is getting large enough.

src/iceberg/transform_function.cc
zhjwpku: Good idea.
https://github.com/apache/iceberg-cpp/issues/164

src/iceberg/expression/literal.h
```diff
@@ -48,27 +48,35 @@ class ICEBERG_EXPORT Literal {
...
     std::strong_ordering operator<=>(const AboveMax&) const = default;
   };
-
-  using Value = std::variant<bool,         // for boolean
-                             int32_t,      // for int, date
-                             int64_t,      // for long, timestamp, timestamp_tz, time
-                             float,        // for float
-                             double,       // for double
-                             std::string,  // for string
+  using Value = std::variant<std::monostate,  // for null
```
mapleFU, line 51: Previously I may think this only present the literal value, but this is also ok to me

src/iceberg/util/truncate_utils.h
```diff
@@ -0,0 +1,72 @@
...
+ICEBERG_EXPORT class TruncateUtils {
+ public:
+  /// \brief Truncate a UTF-8 string to a specified number of code points.
+  ///
+  /// \param source The input string to truncate.
+  /// \param L The maximum number of code points allowed in the output string.
+  /// \return A valid UTF-8 string truncated to L code points.
+  /// If the input string is already valid and has fewer than L code points, it is
+  /// returned unchanged.
+  static std::string TruncateUTF8(std::string&& source, size_t L) {
```
mapleFU: Why not 

```
static std::string TruncateUTF8(std::string source, size_t L) {
```

src/iceberg/transform_function.cc
```diff
@@ -87,14 +129,52 @@ TruncateTransform::TruncateTransform(std::shared_ptr<Type> const& source_type,
...
+      return Literal::String(TruncateUtils::TruncateUTF8(std::move(value), width_));
+    }
+    case TypeId::kBinary: {
+      /// In contrast to strings, binary values do not have an assumed encoding and are
+      /// truncated to L bytes.
+      auto value = std::get<std::vector<uint8_t>>(literal.value());
+      if (value.size() > static_cast<size_t>(width_)) {
+        value.resize(width_);
+      }
+      return Literal::Binary(value);
```
mapleFU:
```suggestion
      return Literal::Binary(std::move(value));
```

src/iceberg/transform_function.cc
```diff
@@ -176,11 +318,39 @@ Result<std::unique_ptr<TransformFunction>> MonthTransform::Make(
...
-  return NotImplemented("DayTransform::Transform");
+Result<Literal> DayTransform::Transform(const Literal& literal) {
+  assert(literal.type() == source_type());
+  if (literal.IsBelowMin() || literal.IsAboveMax()) {
+    return InvalidArgument(
+        "Cannot apply day transform to literal with value {} of type {}",
+        literal.ToString(), source_type()->ToString());
+  }
+  if (literal.IsNull()) [[unlikely]] {
+    return Literal::Null(iceberg::int32());
```
mapleFU: why not date here?

zhjwpku: Because the return type should be `int` per the spec [1].

[1] https://iceberg.apache.org/spec/#partition-transforms

mapleFU: What about L345? Is that in-consistent?

zhjwpku: Ah, my oversight, updated to `int` and adjusted the corresponding test case.

mapleFU (commented):
Perhaps we can assert return literal type is equal to result type?

zhjwpku:
> Perhaps we can assert return literal type is equal to result type?

That would require introducing a temporary variable before the return statement. I think if we have thorough test cases, the assert might not be necessary.

src/iceberg/util/truncate_utils.h
```diff
@@ -0,0 +1,72 @@
...
+#pragma once
+
+#include <string>
+#include <utility>
+
+#include "iceberg/iceberg_export.h"
+
+namespace iceberg {
+
+ICEBERG_EXPORT class TruncateUtils {
```
mapleFU:
```suggestion
class ICEBERG_EXPORT TruncateUtils {
```

mapleFU (approved):

src/iceberg/util/truncate_utils.h
zhjwpku: I initially used this approach, but the pre-commit clang-format hook misformatted the file, so I followed the StringUtils approach instead.

zhjwpku: Ok, I found that upgrade https://github.com/pre-commit/mirrors-clang-format to v20.1.8 works, so changed to the suggested approach, thanks.

PR #158, opened 2025-08-03
feat(parquet): add HasFieldIds check

wgtmac:


src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -30,8 +33,21 @@ Result<std::vector<int>> SelectedColumnIndices(const SchemaProjection& projectio
...
+    return true;
+  }
+
+  if (node->is_group()) {
+    auto group_node = internal::checked_pointer_cast<::parquet::schema::GroupNode>(node);
+    for (int i = 0; i < group_node->field_count(); i++) {
+      if (HasFieldIds(group_node->field(i))) {
+        return true;
+      }
+    }
```
mapleFU, line 47: What does this means? Any leaf has field-id?

wgtmac: Just simply check if any field (whether leaf or non-leaf) has field-id. This is in sync with the Java impl.

wgtmac: I think we cannot blindly check if all fields have ids because it is valid to have some fields that do not have ids, especially when variant type is supported.

mapleFU: I knew that for 3-levels complex type, fieldId might not exists, but here I don't fully understand the purpose, so any leaf has field-id means writer is field-id aware?

mapleFU (approved):

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: I think Iceberg only supports three-level list encoding per the spec: https://iceberg.apache.org/spec/#parquet

> Lists must use the [3-level representation](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists).

wgtmac:
@Fokko @zeroshade Could you help review this? Thanks!

Fokko (approved):
LGTM

Fokko:
Thanks @wgtmac for adding this, and thanks @mapleFU  and @lidavidm for the review 🚀

PR #159, opened 2025-08-04
feat(parquet): add schema projection for parquet

wgtmac:


lidavidm:
I'm guessing one of the GTest or GMock macros expands to something weird.

https://github.com/google/googletest/issues/4531

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,392 @@
...
+    return std::nullopt;
+  }
+  int key = metadata->FindKey(kParquetFieldIdKey.data());
+  if (key < 0) {
+    return std::nullopt;
+  }
+  std::string field_id_str = metadata->value(key);
+  int32_t field_id = -1;
+  try {
+    field_id = std::stoi(field_id_str);
```
mapleFU: Can we use `std::from_chars` here?

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,392 @@
...
+  return field_id < 0 ? std::nullopt : std::make_optional(field_id);
+}
+
+std::optional<int32_t> GetFieldId(const ::parquet::arrow::SchemaField& parquet_field) {
+  return FieldIdFromMetadata(parquet_field.field->metadata());
+}
+
+Status ValidateParquetSchemaEvolution(
+    const Type& expected_type, const ::parquet::arrow::SchemaField& parquet_field) {
+  const auto& arrow_type = parquet_field.field->type();
```
mapleFU, line 69: Can any field evolve from null?

test/parquet_schema_test.cc
```diff
@@ -17,50 +17,494 @@
...
 TEST(HasFieldIds, PrimitiveNode) {
   EXPECT_FALSE(HasFieldIds(MakeInt32Node("test_field")));
   EXPECT_TRUE(HasFieldIds(MakeInt32Node("test_field", /*field_id=*/1)));
 }
 
-TEST(HasFieldIds, GroupNode) {
-  auto group_node_without_field_id =
-      MakeGroupNode("test_group", {MakeInt32Node("c1"), MakeInt32Node("c2")});
-  EXPECT_FALSE(HasFieldIds(group_node_without_field_id));
+// TEST(HasFieldIds, GroupNode) {
```
wgtmac: The linter is satisfied if I comment out these lines. @lidavidm

lidavidm: Weird, I'm not sure what's going on here...NodePtr is a shared_ptr and this seems to work elsewhere...

https://github.com/llvm/llvm-project/issues/122405 also seems to be related - maybe this check is a bit flaky. We can uncomment and use NOLINT on the specific line?

test/parquet_schema_test.cc
```diff
@@ -50,17 +136,375 @@ TEST(HasFieldIds, PrimitiveNode) {
...
       "test_group",
       {MakeInt32Node("c1", /*field_id=*/2), MakeInt32Node("c2", /*field_id=*/3)},
       /*field_id=*/1);
-  EXPECT_TRUE(HasFieldIds(group_node_with_full_field_id));
+  EXPECT_TRUE(HasFieldIds(group_node_with_full_field_id));  // NOLINT
 
   auto group_node_with_partial_field_id = MakeGroupNode(
       "test_group", {MakeInt32Node("c1", /*field_id=*/1), MakeInt32Node("c2")});
-  EXPECT_TRUE(HasFieldIds(group_node_with_partial_field_id));
+  EXPECT_TRUE(HasFieldIds(group_node_with_partial_field_id));  // NOLINT
```
lidavidm: nits: could we add a comment saying why there's a nolint (and/or use `NOLINT(name-of-lint)` to be specific), and maybe NOLINTBEGIN/NOLINTEND would be cleaner (up to you)?

wgtmac: Agreed. I'm still investigating so it is not finalized yet.

FTR: bumping clang-tidy version from 19 to 22 does eliminate errors from 8 to 1.

lidavidm: Ah, nice. So it does seem like a false positive then

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,392 @@
...
+    const ::parquet::arrow::SchemaField& parquet_field = parquet_fields[i];
+    auto field_id = GetFieldId(parquet_field);
+    if (!field_id) {
+      continue;
+    }
+    if (const auto [iter, inserted] = field_context_map.emplace(
+            std::piecewise_construct, std::forward_as_tuple(field_id.value()),
+            std::forward_as_tuple(i, parquet_field));
+        !inserted) [[unlikely]] {
+      return InvalidSchema("Duplicate field id found in Parquet schema: {}",
```
mapleFU:
```suggestion
      return InvalidSchema("Duplicate field id '{}' found in Parquet schema",
```
And add schema as string?

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,392 @@
...
+                         parquet_fields.size());
+  }
+
+  const auto& parquet_field = parquet_fields[0];
+  auto element_field_id = GetFieldId(parquet_field);
+  if (!element_field_id) {
+    return InvalidSchema("List element field missing field id");
+  }
+
+  const auto& expected_element_field = list_type.fields().back();
```
mapleFU: why this variable is named as "expected_element_field" rather than "element_field"?

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,392 @@
...
+    return InvalidSchema("Map key field id mismatch, expected {}, got {}",
+                         expected_key_field.field_id(), key_field_id.value());
+  }
+  if (expected_value_field.field_id() != value_field_id.value()) {
+    return InvalidSchema("Map value field id mismatch, expected {}, got {}",
+                         expected_value_field.field_id(), value_field_id.value());
+  }
+
+  FieldProjection result;
+
```
mapleFU: reserve children to 2?

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,392 @@
...
+      } else {
+        child_projection.attributes =
+            std::make_shared<ParquetExtraAttributes>(parquet_field.column_index);
+      }
+      child_projection.from = iter->second.local_index;
+      child_projection.kind = FieldProjection::Kind::kProjected;
+    } else if (MetadataColumns::IsMetadataColumn(field_id)) {
+      child_projection.kind = FieldProjection::Kind::kMetadata;
+    } else if (expected_field.optional()) {
+      child_projection.kind = FieldProjection::Kind::kNull;
```
mapleFU: (So only struct can contain these `kMetadata` and `kNull` projections. List entry and map value cannot?

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,392 @@
...
+                        parquet_fields[0].children);
+    default:
+      return InvalidSchema("Unsupported nested type: {}", expected_type);
+  }
+}
+
+void CollectColumnIds(const FieldProjection& field_projection,
+                      std::vector<int32_t>* column_ids) {
+  if (field_projection.attributes) {
+    auto parquet_attributes = internal::checked_cast<const ParquetExtraAttributes&>(
```
mapleFU:
```suggestion
    const auto& parquet_attributes = internal::checked_cast<const ParquetExtraAttributes&>(
```

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: I'm not sure. I think we can delay this support until we implement V3 unknown type.

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: Unfortunately we don't have full schema here and we don't even know the field names.

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: Right. I think direct children of map and list types do not support schema evolution per the spec (but if the child is a struct, its fields support schema evolution).

src/iceberg/parquet/parquet_schema_util.cc
mapleFU: We can leave a little todo(v3) or something here?

test/parquet_schema_test.cc
```diff
@@ -17,50 +17,494 @@
...
+TEST(HasFieldIdsTest, PrimitiveNode) {
   EXPECT_FALSE(HasFieldIds(MakeInt32Node("test_field")));
   EXPECT_TRUE(HasFieldIds(MakeInt32Node("test_field", /*field_id=*/1)));
 }
 
-TEST(HasFieldIds, GroupNode) {
-  auto group_node_without_field_id =
-      MakeGroupNode("test_group", {MakeInt32Node("c1"), MakeInt32Node("c2")});
-  EXPECT_FALSE(HasFieldIds(group_node_without_field_id));
+TEST(HasFieldIdsTest, GroupNode) {
```
wgtmac, line 142: I guess the original test name `HasFieldIds` conflicts with the function name `HasFieldIds` so the `TEST` expansion may have confused the linter. @lidavidm @mapleFU

wgtmac: Never mind. I don't know why it did succeed once but keep failing after that.

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,393 @@
...
+namespace {
+
+constexpr std::string_view kParquetFieldIdKey = "PARQUET:field_id";
+
+std::optional<int32_t> FieldIdFromMetadata(
+    const std::shared_ptr<const ::arrow::KeyValueMetadata>& metadata) {
+  if (!metadata) {
+    return std::nullopt;
+  }
+  int key = metadata->FindKey(kParquetFieldIdKey.data());
```
zhjwpku: Below is the FindKey signature, can we just pass kParquetFieldIdKey without the .data() part?
```
  /// \brief Perform linear search for key, returning -1 if not found
  int FindKey(std::string_view key) const;
```

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,393 @@
...
+
+Result<FieldProjection> ProjectList(
+    const ListType& list_type,
+    const std::vector<::parquet::arrow::SchemaField>& parquet_fields) {
+  if (parquet_fields.size() != 1) {
+    return InvalidSchema("List type must have exactly one field, got {}",
+                         parquet_fields.size());
+  }
+
+  const auto& parquet_field = parquet_fields[0];
```
zhjwpku: nit: use `parquet_fields.back()`? Since you use `list_type.fields().back()` below.

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: Nice catch!

mapleFU (approved):

zhjwpku (approved):
LGTM, thanks for working on this.

wgtmac:
@Fokko @zeroshade Could you help review this? Thanks!

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,393 @@
...
+        return {};
+      }
+      break;
+    case TypeId::kDate:
+      if (arrow_type->id() == ::arrow::Type::DATE32) {
+        return {};
+      }
+      break;
+    case TypeId::kTime:
+      if (arrow_type->id() == ::arrow::Type::TIME64) {
```
Fokko, line 104: Should we also check for `::arrow::TimeUnit::MICRO` here?

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,393 @@
...
+        return {};
+      }
+      break;
+    case TypeId::kBinary:
+      if (arrow_type->id() == ::arrow::Type::BINARY) {
+        return {};
+      }
+      break;
+    case TypeId::kDecimal:
+      if (arrow_type->id() == ::arrow::Type::DECIMAL128) {
```
Fokko, line 143: Do we also accept the other types?

image

wgtmac: This is still an open PR: https://github.com/apache/arrow/pull/45351

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: Good catch! I have added an exhaustive test case to make sure I don't miss any primitive type.

dongxiao1198 (approved):
lgtm

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,397 @@
...
+      }
+      break;
+    case TypeId::kUuid:
+      if (arrow_type->id() == ::arrow::Type::FIXED_SIZE_BINARY) {
+        const auto& fixed_binary =
+            internal::checked_cast<const ::arrow::FixedSizeBinaryType&>(*arrow_type);
+        if (fixed_binary.byte_width() == 16) {
+          return {};
+        }
+      }
```
zeroshade: We should probably also allow https://github.com/apache/arrow/blob/main/cpp/src/arrow/extension/uuid.h#L35

You can validate via `arrow_type->id() == ::arrow::Type::EXTENSION` and the `extension_name()` == `"arrow.uuid"`

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,397 @@
...
+    case TypeId::kString:
+      if (arrow_type->id() == ::arrow::Type::STRING) {
+        return {};
+      }
+      break;
+    case TypeId::kBinary:
+      if (arrow_type->id() == ::arrow::Type::BINARY) {
+        return {};
+      }
+      break;
```
zeroshade, line 141: What about LargeString, LargeBinary, StringView and BinaryView?

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,397 @@
...
+}
+
+std::optional<int32_t> GetFieldId(const ::parquet::arrow::SchemaField& parquet_field) {
+  return FieldIdFromMetadata(parquet_field.field->metadata());
+}
+
+// TODO(gangwu): support v3 unknown type
+Status ValidateParquetSchemaEvolution(
+    const Type& expected_type, const ::parquet::arrow::SchemaField& parquet_field) {
+  const auto& arrow_type = parquet_field.field->type();
```
zeroshade, line 69: I forget offhand if this will return a `DictionaryType` for dictionary encoded columns, if so then you need to check for the `DictionaryType` and then switch on the `ValueType` of it.

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -17,20 +17,397 @@
...
+        }
+      }
+      break;
+    case TypeId::kStruct:
+      if (arrow_type->id() == ::arrow::Type::STRUCT) {
+        return {};
+      }
+      break;
+    case TypeId::kList:
+      if (arrow_type->id() == ::arrow::Type::LIST) {
```
zeroshade, line 179: What about LargeList and ListView?

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: I don't think parquet-cpp has supported these types.

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: No it won't. Reading dictionary is supported via an option to create a RecordReader: https://github.com/apache/arrow/blob/2dd3ccda6437f79aa34641bd3197dd7392ae4aec/cpp/src/parquet/column_reader.h#L266

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: ListView is not supported by parquet-cpp yet. I think we should just support simple list and binary type variants in the early versions of iceberg-cpp. Once parquet-cpp has full support, we can leverage them later.

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: Good catch! Fixed.

Xuanwo (approved):
Thank you for working on this, let's move!

PR #160, opened 2025-08-06
feat: add file writer and manifest writer interface definition

dongxiao1198:
[Define of file writer and manifest writer](feat: add file writer and manifest writer interface definition)

### DIFF
--- a/src/iceberg/file_writer.h
+++ b/src/iceberg/file_writer.h
@@ -64,7 +64,7 @@ class ICEBERG_EXPORT Writer {
   /// \brief Write arrow data to the file.
   ///
   /// \return Status of write results.
-  virtual Status Write(const ArrowArray& data) = 0;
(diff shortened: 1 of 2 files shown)


src/iceberg/file_writer.h
```diff
@@ -0,0 +1,86 @@
...
+  /// \brief Open the writer.
+  virtual Status Open(const struct WriterOptions& options) = 0;
+
+  /// \brief Close the writer.
+  virtual Status Close() = 0;
+
+  /// \brief Write arrow data to the file.
+  ///
+  /// \return Status of write results.
+  virtual Status Write(const ArrowArray& data) = 0;
```
wgtmac:
```suggestion
  virtual Status Write(ArrowArray data) = 0;
```

Usually array data is moved to the callee.

Fokko (approved):

PR #161, opened 2025-08-06
fix: correct return type of ArrowFileSystemFileIO functions( MakeMockFileIO, MakeLocalFileIO )

nullccxsy:
change the return type of MakeMockFileIO, MakeLocalFileIO to FileIO


wgtmac (approved):
nit: the PR title is actually a `fix` instead of `feat`

Fokko (approved):
Makes sense, thanks @nullccxsy

PR #172, opened 2025-08-13
fix: content in manifest entry should be optional

nullccxsy:
kContent in manifest entry should be optional, not required


wgtmac (approved):
Make sense. Thanks!

zeroshade (approved):
Should there be a corresponding test for this?

PR #175, opened 2025-08-14
fix: correct partition field handling for non-partitioned tables and add test for manifest file reader

HeartLinked:
- Fix ManifestEntry::operator== logic with proper parentheses grouping
- Add null pointer safety in DataFile::Type() for non-partitioned tables
- Change partition field type validation from list to struct in manifest reader
- Support empty partition structs for non-partitioned tables (n_children == 0)
- Update unit test for manifest file: separate v1 and v2 test, and add another v2 test for non-partitioned  tables


test/manifest_reader_test.cc
```diff
@@ -45,7 +46,7 @@ class ManifestReaderTest : public TempFileTestBase {
     avro::RegisterLogicalTypes();
   }
 
-  std::vector<ManifestEntry> prepare_manifest_entries() {
+  std::vector<ManifestEntry> prepareV1ManifestEntries() {
```
wgtmac:
```suggestion
  std::vector<ManifestEntry> PrepareV1ManifestEntries() {
```

Please fix it by using capitalized initial.

src/iceberg/manifest_entry.cc
```diff
@@ -29,13 +29,16 @@ namespace iceberg {
...
-         (!data_file && !other.data_file);
+         sequence_number == other.sequence_number &&
+         file_sequence_number == other.file_sequence_number &&
+         ((data_file && other.data_file && *data_file == *other.data_file) ||
+          (!data_file && !other.data_file));
 }
 
 std::shared_ptr<StructType> DataFile::Type(std::shared_ptr<StructType> partition_type) {
+  if (!partition_type) {
+    partition_type = std::make_shared<StructType>(std::vector<SchemaField>{});
```
wgtmac:
```suggestion
    partition_type = PartitionSpec::Unpartitioned()->schema();
```

A more elegant way to fix this is like above. However, we need to fix that as you can see that it is initialized with a nullptr too: 
https://github.com/apache/iceberg-cpp/blob/a54f116211258f4199b5b1665494e9eb4bec45b2/src/iceberg/partition_spec.cc#L46-L52

test/manifest_reader_test.cc
```diff
@@ -115,7 +116,88 @@ TEST_F(ManifestReaderTest, BasicTest) {
...
+
+  void SetUp() override {
+    TempFileTestBase::SetUp();
+    local_fs_ = std::make_shared<::arrow::fs::LocalFileSystem>();
+    file_io_ = std::make_shared<iceberg::arrow::ArrowFileSystemFileIO>(local_fs_);
+
+    avro::RegisterLogicalTypes();
+  }
+
+  std::vector<ManifestEntry> prepareV2NonPartitionedManifestEntries() {
```
wgtmac:
```suggestion
  std::vector<ManifestEntry> PrepareV2NonPartitionedManifestEntries() {
```

test/manifest_reader_test.cc
```diff
@@ -115,7 +116,88 @@ TEST_F(ManifestReaderTest, BasicTest) {
...
+    std::vector<ManifestEntry> manifest_entries;
+    std::string test_dir_prefix = "/tmp/db/db/v2_manifest_non_partitioned/data/";
```
wgtmac, line 137: The two `db`s in the prefix looks weird. Is this something that can be fixed?

wgtmac (changes requested):

test/manifest_reader_test.cc
HeartLinked: The path in the previous test code was exactly like this (as seen in this commit: [Commit 043701c
](https://github.com/apache/iceberg-cpp/commit/043701c14d6015ffe3ec682b6fec886ba810b4d8#diff-6432ba44184f2025b3037c028b74b589b3ee8ccebffd5a003a037827cf7f70fa) Therefore, I decided to maintain consistency with it and adopted the same path specification when generating the test file. I don't consider this to be an explicit error—for the purpose of testing, it meets the expected requirements.

test/manifest_reader_test.cc
HeartLinked: fixed

src/iceberg/manifest_entry.cc
HeartLinked: fixed

wgtmac (approved):

Xuanwo (approved):
Thank you for working on this!

PR #176, opened 2025-08-14
feat: add manifest&manifest list writer

dongxiao1198:
1 add v1v2v3 writer definition
2 add v1v2v3 metadata wrapper


src/iceberg/manifest_writer.h
```diff
@@ -35,14 +35,24 @@ namespace iceberg {
...
+  virtual Status Close() = 0;
 
   /// \brief Creates a writer for a manifest file.
+  /// \param format_version Format version of the manifest.
+  /// \param snapshot_id ID of the snapshot.
+  /// \param first_row_id First row ID of the snapshot.
   /// \param manifest_location Path to the manifest file.
   /// \param file_io File IO implementation to use.
   /// \return A Result containing the writer or an error.
   static Result<std::unique_ptr<ManifestWriter>> MakeWriter(
```
wgtmac:
```suggestion
  static Result<std::unique_ptr<ManifestWriter>> Make(
```

src/iceberg/manifest_writer.h
```diff
@@ -51,13 +61,27 @@ class ICEBERG_EXPORT ManifestWriter {
...
   /// \brief Creates a writer for the manifest list.
+  /// \param format_version Format version of the manifest list.
+  /// \param snapshot_id ID of the snapshot.
+  /// \param parent_snapshot_id ID of the parent snapshot.
+  /// \param sequence_number Sequence number of the snapshot.
+  /// \param first_row_id First row ID of the snapshot.
   /// \param manifest_list_location Path to the manifest list file.
   /// \param file_io File IO implementation to use.
   /// \return A Result containing the writer or an error.
   static Result<std::unique_ptr<ManifestListWriter>> MakeWriter(
```
wgtmac:
```suggestion
  static Result<std::unique_ptr<ManifestListWriter>> Make(
```

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,91 @@
...
+                                                std::move(schema));
+    case 2:
+      return std::make_unique<ManifestWriterV2>(snapshot_id, std::move(writer),
+                                                std::move(schema));
+    case 3:
+      return std::make_unique<ManifestWriterV3>(snapshot_id, first_row_id,
+                                                std::move(writer), std::move(schema));
+
+    default:
+      return InvalidArgument("Unsupported manifest format version: {}", format_version);
```
wgtmac:
```suggestion
      return NotSupported("Unsupported manifest format version: {}", format_version);
```

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,91 @@
...
+      return std::make_unique<ManifestListWriterV2>(snapshot_id, parent_snapshot_id,
+                                                    sequence_number, std::move(writer),
+                                                    std::move(schema));
+    case 3:
+      return std::make_unique<ManifestListWriterV3>(snapshot_id, parent_snapshot_id,
+                                                    sequence_number, first_row_id,
+                                                    std::move(writer), std::move(schema));
+
+    default:
+      return InvalidArgument("Unsupported manifest list format version: {}",
```
wgtmac:
```suggestion
      return NotSupported("Unsupported manifest list format version: {}",
```

src/iceberg/manifest_writer.h
```diff
@@ -35,14 +35,24 @@ namespace iceberg {
...
 
   /// \brief Creates a writer for a manifest file.
+  /// \param format_version Format version of the manifest.
+  /// \param snapshot_id ID of the snapshot.
+  /// \param first_row_id First row ID of the snapshot.
   /// \param manifest_location Path to the manifest file.
   /// \param file_io File IO implementation to use.
   /// \return A Result containing the writer or an error.
   static Result<std::unique_ptr<ManifestWriter>> MakeWriter(
+      int32_t format_version, int64_t snapshot_id, int64_t first_row_id,
```
wgtmac: Should we make some parameters optional, like first_row_id? Same question for ManifestListWriter.

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,91 @@
...
+#include "iceberg/schema.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+Result<std::unique_ptr<ManifestWriter>> ManifestWriter::MakeWriter(
+    int32_t format_version, int64_t snapshot_id, int64_t first_row_id,
+    std::string_view manifest_location, std::shared_ptr<FileIO> file_io,
+    std::shared_ptr<Schema> partition_schema) {
+  auto manifest_entry_schema = ManifestEntry::TypeFromPartitionType(partition_schema);
```
wgtmac:
```suggestion
  auto manifest_entry_schema = ManifestEntry::TypeFromPartitionType(std::move(partition_schema));
```

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,91 @@
...
+  auto fields_span = manifest_entry_schema->fields();
+  std::vector<SchemaField> fields(fields_span.begin(), fields_span.end());
+  auto schema = std::make_shared<Schema>(fields);
```
wgtmac, line 106: nit: you may use `FromStructType` in the src/iceberg/schema_internal.h

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+Result<std::unique_ptr<ManifestListWriter>> ManifestListWriter::MakeWriter(
+    int32_t format_version, int64_t snapshot_id, int64_t parent_snapshot_id,
+    int64_t sequence_number, int64_t first_row_id,
+    std::string_view manifest_list_location, std::shared_ptr<FileIO> file_io) {
+  std::vector<SchemaField> fields(ManifestFile::Type().fields().begin(),
+                                  ManifestFile::Type().fields().end());
+  auto schema = std::make_shared<Schema>(fields);
```
wgtmac, line 174: ditto, you may use FromStructType.

src/iceberg/v1_metadata.h
```diff
@@ -0,0 +1,53 @@
...
+/// \brief v1 metadata wrapper.
+///
+/// Wrapper for v1 manifest list and manifest entry.
+class V1MetaData {
+ public:
+  /// \brief v1 manifest file wrapper.
+  struct ManifestFileWrapper : public ManifestFile {
+    ManifestFileWrapper() = default;
+
+    ManifestFile wrap(ManifestFile file, int64_t first_row_id) { return *this; }
```
wgtmac: We don't need first_row_id for V1. It is used only by V3.

wgtmac:
```suggestion
    ManifestFile Wrap(ManifestFile file, int64_t first_row_id) { return *this; }
```

src/iceberg/manifest_writer.h
```diff
@@ -51,13 +61,27 @@ class ICEBERG_EXPORT ManifestWriter {
 class ICEBERG_EXPORT ManifestListWriter {
  public:
   virtual ~ManifestListWriter() = default;
-  virtual Status WriteManifestFiles(const std::vector<ManifestFile>& files) const = 0;
+
+  /// \brief Write manifest file list to manifest list file.
+  /// \param file Manifest file to write.
+  /// \return Status::OK() if all file was written successfully
+  virtual Status WriteManifestFile(const ManifestFile& file) const = 0;
```
wgtmac: Add an overload to accept `const std::vector<ManifestFile>&`?

wgtmac: We may also change the function name to `Add` and `AddAll` respectively.

src/iceberg/manifest_writer_internal.h
```diff
@@ -0,0 +1,160 @@
...
+  Status WriteManifestEntry(const ManifestEntry& entry) const override;
+
+  Status Close() override;
+
+ private:
+  V3MetaData::ManifestEntryWrapper wrapper_;
+};
+
+/// \brief Write manifest files to a manifest list file.
+class ManifestListWriterImpl : public ManifestListWriter {
```
wgtmac: The Java-equivalent has `protected abstract ManifestFile prepare(ManifestFile manifest)`, we need to add it too.

src/iceberg/v2_metadata.h
```diff
@@ -0,0 +1,58 @@
...
+/// \brief v2 metadata wrapper.
+///
+/// Wrapper for v2 manifest list and manifest entry.
+class V2MetaData {
+ public:
+  /// \brief v2 manifest file wrapper.
+  struct ManifestFileWrapper : public ManifestFile {
+    explicit ManifestFileWrapper(int64_t commit_snapshotId, int64_t sequence_number) {}
+
+    ManifestFile wrap(ManifestFile file, int64_t first_row_id) { return *this; }
```
wgtmac: ditto for first_row_id

src/iceberg/v1_metadata.h
wgtmac: Do we want to make a copy for every ManifestFile?

src/iceberg/manifest_writer.h
```diff
@@ -35,14 +35,24 @@ namespace iceberg {
 class ICEBERG_EXPORT ManifestWriter {
  public:
   virtual ~ManifestWriter() = default;
-  virtual Status WriteManifestEntries(
-      const std::vector<ManifestEntry>& entries) const = 0;
+
+  /// \brief Write manifest entry to file
```
HeartLinked:
```suggestion
  /// \brief Write manifest entry to file.
```

src/iceberg/manifest_writer.h
```diff
@@ -51,13 +61,27 @@ class ICEBERG_EXPORT ManifestWriter {
 class ICEBERG_EXPORT ManifestListWriter {
  public:
   virtual ~ManifestListWriter() = default;
-  virtual Status WriteManifestFiles(const std::vector<ManifestFile>& files) const = 0;
+
+  /// \brief Write manifest file list to manifest list file.
```
HeartLinked: The comment "Write manifest file list" here is inconsistent with the single-add function.

src/iceberg/manifest_writer.h
HeartLinked: > Should we make some parameters optional, like first_row_id? Same question for ManifestListWriter.

I agree, in the Java implementation, different versions from v1 to v4 are achieved by having subclasses inherit from a base writer class, which conveniently allows different construction parameters for each version. I find this approach more intuitive. If we don't implement it this way, should we at least create multiple constructors with different parameters to explicitly distinguish between versions?

src/iceberg/v1_metadata.h
dongxiao1198: That's a problem, java implement just overwrite all getters and keep a reference of manifest, wrapp ths file by getter in wrapper. But in cpp, our manifest is a public struct.
It  seems impossible to avoid this copy without refactor out manifest and list into class and make all members private with a public virtual getter.

src/iceberg/manifest_writer.h
dongxiao1198: Yes we can change those params to optional. 
This "ManifestListWriter::Make" builder function is a unified entry like org.apache.iceberg.ManifestLists#write in java implement.

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,101 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto writer, WriterFactoryRegistry::Open(
+                                           FileFormatType::kAvro,
+                                           {.path = std::string(manifest_list_location),
+                                            .schema = schema,
+                                            .io = std::move(file_io)}));
+  switch (format_version) {
+    case 1:
+      return std::make_unique<ManifestListWriterV1>(snapshot_id, parent_snapshot_id,
+                                                    std::move(writer), std::move(schema));
+    case 2:
```
HeartLinked: sequence_number is required for V2 manifest list, so here needs check validation for v2?

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,101 @@
...
+      return std::make_unique<ManifestWriterV3>(snapshot_id, first_row_id.value(),
+                                                std::move(writer), std::move(schema));
+
+    default:
+      return NotSupported("Unsupported manifest format version: {}", format_version);
+  }
+}
+
+Result<std::unique_ptr<ManifestListWriter>> ManifestListWriter::Make(
+    int32_t format_version, int64_t snapshot_id, int64_t parent_snapshot_id,
```
HeartLinked: parent_snapshot_id may be optional?

HeartLinked: Would it be a better choice to split the make function into separate v1~v3 versions? I feel that as the number of parameters increases, having to check various optional values in a single function might make the logic overly complex.

dongxiao1198:
> What about this? `ManifestListWriter` does not need to be an abstract class. Only `ManifestFileAdapter` should have subclasses from v1 ~ v4 to deal with different schemas and value-filling logics. @dongxiao1198 @HeartLinked
> 
> ```c++
> // Implemented by different versions with different schemas to
> // append a list of `ManifestFile`s to an `ArrowArray`.
> class ManifestFileAdapter {
>  public:
>   virtual Status StartAppending() const = 0;
>   virtual Status Append(const ManifestFile& file) const = 0;
>   virtual Result<ArrowArray> FinishAppending() const = 0;
>   int64_t size() const { return size_; }
> 
>  private:
>   std::shared_ptr<Schema> manifest_list_schema_;
>   ArrowSchema schema_;  // converted from manifest_list_schema_
>   ArrowArray array_;    // array to append `ManifestFile`s
>   int64_t size_;
> };
> 
> class ICEBERG_EXPORT ManifestListWriter {
>  public:
>   static Result<std::unique_ptr<ManifestListWriter>> Make(
>       std::string_view manifest_list_location, std::shared_ptr<FileIO> file_io,
>       const std::unordered_map<std::string, std::string>& meta);
> 
>   Status AddAll(const std::vector<ManifestFile>& files) const {
>     for (const auto& file : files) {
>       Add(file);
>     }
>     return {};
>   }
> 
>   Status Add(const ManifestFile& file) const {
>     if (adapter_->size() >= kBatchSize) {
>       ICEBERG_ASSIGN_OR_RAISE(auto array, adapter_->FinishAppending());
>       ICEBERG_RETURN_UNEXPECTED(writer_->Write(array));
>       ICEBERG_RETURN_UNEXPECTED(adapter_->StartAppending());
>     }
>     return adapter_->Append(file);
>   }
> 
>   Status Close() const;
> 
>  private:
>   static constexpr int64_t kBatchSize = 1024;
>   std::unique_ptr<Writer> writer_;
>   std::unique_ptr<ManifestFileAdapter> adapter_;
> };
> ```

The latest patch add a new header for this adapter at src/iceberg/metadata_adapter.h.
Split writer maker function into 4 independent builder function.

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,225 @@
...
+      WriterFactoryRegistry::Open(
+          FileFormatType::kAvro,
+          {.path = std::string(location), .schema = schema, .io = std::move(file_io)}));
+  return writer;
+}
+
+Result<std::unique_ptr<ManifestWriter>> ManifestWriter::MakeV1Writer(
+    std::optional<int64_t> snapshot_id, std::string_view manifest_location,
+    std::shared_ptr<FileIO> file_io, std::shared_ptr<Schema> partition_schema) {
+  auto schema = ParseSchema(partition_schema);
```
wgtmac: Schemas may vary among versions so we cannot use a common function to create it.

src/iceberg/metadata_adapter.h
```diff
@@ -0,0 +1,66 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/metadata_adapter.h
```
wgtmac: Should we rename it to `manifest_adapter.h`?

src/iceberg/manifest_writer.h
```diff
@@ -35,14 +35,57 @@ namespace iceberg {
 class ICEBERG_EXPORT ManifestWriter {
  public:
   virtual ~ManifestWriter() = default;
-  virtual Status WriteManifestEntries(
-      const std::vector<ManifestEntry>& entries) const = 0;
+
+  /// \brief Write manifest entry to file.
+  /// \param entry Manifest entry to write.
+  /// \return Status::OK() if entry was written successfully
+  virtual Status Add(const ManifestEntry& entry) = 0;
```
wgtmac: Do we really need to make `ManifestWriter` and `ManifestListWriter` pure virtual if they only have a single `XXXImpl` subclass?

src/iceberg/v4_metadata.h
```diff
@@ -0,0 +1,64 @@
...
+/// \file iceberg/v4_metadata.h
+
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/metadata_adapter.h"
+
+namespace iceberg {
+
+/// \brief Adapter to convert V4 ManifestEntry to `ArrowArray`.
+class ManifestEntryAdapterV4 : public ManifestEntryAdapter {
```
wgtmac: Remove this since V4 is far from being finalized.

src/iceberg/manifest_writer.h
dongxiao1198: 1 this way can hide adapter and constructor inside
2 We may create gmock class for test in the future

wgtmac: I don't think we will mock these writers. The constructor only sets the pointers of adaptor and writer, then does nothing. All the logics are in the adaptor code.

If we keep `Add` as a virtual function, adding a single `ManifestEntry` may require two virtual functions calls, one from here and one from the adapter. The cost may be non-trivial if we call them repeatedly for many times. So I'm inclined to change classes here to remove virtual functions.

src/iceberg/manifest_adapter.h
```diff
@@ -0,0 +1,66 @@
...
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+// \brief Implemented by different versions with different schemas to
+// append a list of `ManifestEntry`s to an `ArrowArray`.
+class ManifestEntryAdapter {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT ManifestEntryAdapter {
```

src/iceberg/manifest_adapter.h
```diff
@@ -0,0 +1,66 @@
...
+  int64_t size() const { return size_; }
+
+ protected:
+  ArrowArray array_;  // array to append `ManifestEntry`s
+  int64_t size_ = 0;
+};
+
+// \brief Implemented by different versions with different schemas to
+// append a list of `ManifestFile`s to an `ArrowArray`.
+class ManifestFileAdapter {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT ManifestFileAdapter {
```

src/iceberg/manifest_adapter.h
```diff
@@ -0,0 +1,66 @@
...
+namespace iceberg {
+
+// \brief Implemented by different versions with different schemas to
+// append a list of `ManifestEntry`s to an `ArrowArray`.
+class ManifestEntryAdapter {
+ public:
+  ManifestEntryAdapter() = default;
+  virtual ~ManifestEntryAdapter() = default;
+
+  virtual Status StartAppending() = 0;
```
wgtmac, line 37: Both adaptors have StartAppending and FinishAppending, should be abstract a common base class for appending to nanoarrow arrays?

dongxiao1198: of cause yes

src/iceberg/manifest_adapter.h
```diff
@@ -0,0 +1,67 @@
...
+/// Base class of adapter for v1v2v3v4 metadata.
+
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+// \brief Base class of adapter for v1v2v3v4 metadata.
```
wgtmac:
```suggestion
// \brief Base class to append manifest metadata to Arrow array.
```

src/iceberg/manifest_adapter.h
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+#pragma once
+
+/// \file iceberg/metadata_adapter.h
+/// Base class of adapter for v1v2v3v4 metadata.
+
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
```
wgtmac:
```suggestion
#include "iceberg/type_fwd.h"
```

src/iceberg/v1_metadata.h
```diff
@@ -0,0 +1,62 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/v1_metadata.h
+
+#include "iceberg/manifest_adapter.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
```
wgtmac:
```suggestion
```

Let's use forward declaration.

src/iceberg/v2_metadata.h
```diff
@@ -0,0 +1,62 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/v2_metadata.h
+
+#include "iceberg/manifest_adapter.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
```
wgtmac:
```suggestion
```

src/iceberg/v3_metadata.h
```diff
@@ -0,0 +1,64 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/v3_metadata.h
+
+#include "iceberg/manifest_adapter.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
```
wgtmac:
```suggestion
```

src/iceberg/manifest_writer.cc
```diff
@@ -0,0 +1,182 @@
...
+Result<std::unique_ptr<Writer>> OpenFileWriter(std::string_view location,
+                                               const std::shared_ptr<Schema> schema,
+                                               std::shared_ptr<FileIO> file_io) {
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto writer,
+      WriterFactoryRegistry::Open(
+          FileFormatType::kAvro,
+          {.path = std::string(location), .schema = schema, .io = std::move(file_io)}));
+  return writer;
+}
```
wgtmac:
```suggestion
Result<std::unique_ptr<Writer>> OpenFileWriter(std::string_view location,
                                               std::shared_ptr<Schema> schema,
                                               std::shared_ptr<FileIO> file_io) {
  ICEBERG_ASSIGN_OR_RAISE(
      auto writer,
      WriterFactoryRegistry::Open(
          FileFormatType::kAvro,
          {.path = std::string(location), .schema = std::move(schema), .io = std::move(file_io)}));
  return writer;
}
```

wgtmac (approved):

HeartLinked (approved):

zeroshade (approved):

PR #177, opened 2025-08-15
feat: scaffolding work for expression

wgtmac:
- Add interface for term, unbound, bound, reference, predicate, etc.
- Add factory to create expressions.

### DIFF
--- a/src/iceberg/expression/expression.cc
+++ b/src/iceberg/expression/expression.cc
@@ -174,13 +174,17 @@ Result<Expression::Operation> Negate(Expression::Operation op) {
     case Expression::Operation::kNotStartsWith:
       return Expression::Operation::kStartsWith;
     case Expression::Operation::kTrue:
+      return Expression::Operation::kFalse;
     case Expression::Operation::kFalse:
+      return Expression::Operation::kTrue;
+    case Expression::Operation::kAnd:
+      return Expression::Operation::kOr;
+    case Expression::Operation::kOr:
+      return Expression::Operation::kAnd;
     case Expression::Operation::kNot:
     case Expression::Operation::kCountStar:
     case Expression::Operation::kMax:
     case Expression::Operation::kMin:
-    case Expression::Operation::kAnd:
-    case Expression::Operation::kOr:
     case Expression::Operation::kCount:
       return InvalidArgument("No negation for operation: {}", op);
   }


src/iceberg/expression/term.h
```diff
@@ -0,0 +1,258 @@
...
+
+  const SchemaField& field() const { return *field_; }
+
+  std::string_view name() const override { return field_->name(); }
+
+  std::string ToString() const override;
+
+  Result<Literal::Value> Evaluate(const StructLike& data) const override;
+
+  Result<std::vector<Literal::Value>> Evaluate(const ArrowArray& data) const override;
```
mapleFU: Why output is not an ArrowArray?

src/iceberg/expression/term.h
```diff
@@ -0,0 +1,258 @@
...
+  /// \brief Create a named reference to a field.
+  ///
+  /// \param field_name The name of the field to reference
+  explicit NamedReference(std::string field_name);
+
+  ~NamedReference() override = default;
+
+  std::string_view name() const override { return field_name_; }
+
+  Result<std::unique_ptr<BoundReference>> Bind(const Schema& schema,
```
mapleFU: Not a severe problem but a `std::enable_shared_from_this<>` type generate a `unique_ptr` is a bit weird

src/iceberg/expression/term.h
wgtmac: Isn't it over-complicated to return an ArrowArray?

src/iceberg/expression/term.h
wgtmac: I agree. It would be much easier to exclusively use shared_ptr for expressions because they might need to be rewritten for severals times (e.g. bind, eliminate not, constant folding, rebalance, etc.)

src/iceberg/expression/term.h
mapleFU: The input is an arrow array, but returns `vec<literal>`...

src/iceberg/expression/predicate.h
```diff
@@ -0,0 +1,192 @@
...
+  Expression::Operation operation_;
+  std::shared_ptr<TermType> term_;
+};
+
+/// \brief Unbound predicates contain unbound terms and must be bound to a concrete schema
+/// before they can be evaluated.
+///
+/// \tparam B The bound type this predicate produces when binding is successful
+template <typename B>
+class ICEBERG_EXPORT UnboundPredicate : public Predicate<UnboundTerm<B>>,
```
mapleFU, line 64: Waht can be `B` be here? Generally `UnboundTerm<BoundReference>`?

src/iceberg/expression/expressions.cc
```diff
@@ -0,0 +1,344 @@
...
+
+  if (left->op() == Expression::Operation::kTrue) {
+    return right;
+  }
+
+  if (right->op() == Expression::Operation::kTrue) {
+    return left;
+  }
+
+  return std::make_shared<::iceberg::And>(left, right);
```
mapleFU:
```suggestion
  return std::make_shared<::iceberg::And>(std::move(left), std::move(right));
```

src/iceberg/expression/predicate.h
wgtmac: All kinds of bound term, including BoundReference, BoundTransform and BoundExtract.

src/iceberg/expression/term.h
wgtmac: Building an `ArrowArray` is way too complicated compared with a Literal. @gty404 has explored this already when adding the transform function.

wgtmac: We can remove the batch evaluate function and add it back later if the requirement is clear.

wgtmac:
This PR is ready for review. It looks pretty similar to the Java equivalent. Let me know what you think. @dongxiao1198 @lidavidm @mapleFU @yingcai-cy @zhjwpku

zhjwpku (approved):

dongxiao1198 (approved):

src/iceberg/expression/expressions.h
```diff
@@ -0,0 +1,263 @@
...
+      std::string name, std::vector<Literal> values);
+
+  /// \brief Create an IN predicate for an unbound term.
+  template <typename B>
+  static std::shared_ptr<UnboundPredicate<B>> In(std::shared_ptr<UnboundTerm<B>> expr,
+                                                 std::vector<Literal> values);
+
+  /// \brief Create an IN predicate for a field name with initializer list.
+  static std::shared_ptr<UnboundPredicate<BoundReference>> In(
+      std::string name, std::initializer_list<Literal> values);
```
mapleFU, line 241: curiously why somewhere initializer list, other places vector?

src/iceberg/expression/term.h
```diff
@@ -0,0 +1,267 @@
...
+  using BoundType = B;
+};
+
+/// \brief Base class for bound terms.
+class ICEBERG_EXPORT BoundTerm : public Bound, public Term {
+ public:
+  ~BoundTerm() override;
+
+  /// \brief Returns the type produced by this term.
+  virtual std::shared_ptr<Type> type() const = 0;
```
mapleFU, line 103: would `const std::shared_ptr<Type>&` a better choice?

src/iceberg/expression/term.h
mapleFU: Ok, this is ok to me

src/iceberg/expression/predicate.h
```diff
@@ -0,0 +1,210 @@
...
+
+#include "iceberg/expression/expression.h"
+#include "iceberg/expression/term.h"
+
+namespace iceberg {
+
+/// \brief A predicate is a boolean expression that tests a term against some criteria.
+///
+/// \tparam TermType The type of the term being tested
+template <typename TermType>
```
mapleFU: I'm thinking a bit that whether these can be concepts, but this is not a blocking issue

src/iceberg/expression/predicate.h
```diff
@@ -0,0 +1,210 @@
...
+  /// \return true if the predicate passes, false otherwise
+  virtual Result<bool> Test(const Literal::Value& value) const = 0;
+
+  enum class Kind : int8_t {
+    // A unary predicate (tests for null, not-null, etc.).
+    kUnary = 0,
+    // A literal predicate (compares against a literal).
+    kLiteral,
+    // A set predicate (tests membership in a set).
+    kIn,
```
mapleFU: kSet?

src/iceberg/expression/term.h
wgtmac: No, we can't due to following code where Transform::ResultType() cannot return const ref.

```cpp
std::shared_ptr<Type> BoundTransform::type() const {
  return transform_func_->ResultType();
}
```

src/iceberg/expression/expressions.h
wgtmac: Ah, I missed some overloads. Let me fix it.

src/iceberg/expression/predicate.h
wgtmac: Changed

src/iceberg/expression/expression.cc
```diff
@@ -87,4 +91,91 @@ bool Or::Equals(const Expression& expr) const {
...
+      return Expression::Operation::kEq;
+    case Expression::Operation::kIn:
+      return Expression::Operation::kNotIn;
+    case Expression::Operation::kNotIn:
+      return Expression::Operation::kIn;
+    case Expression::Operation::kStartsWith:
+      return Expression::Operation::kNotStartsWith;
+    case Expression::Operation::kNotStartsWith:
+      return Expression::Operation::kStartsWith;
+    default:
```
mapleFU: Avoid default and list all operations here?

src/iceberg/expression/expression.h
```diff
@@ -193,4 +194,10 @@ class ICEBERG_EXPORT Or : public Expression {
   std::shared_ptr<Expression> right_;
 };
 
+/// \brief Returns a string representation of an expression operation.
+ICEBERG_EXPORT std::string_view ToString(Expression::Operation op);
+
+/// \brief Returns the negated operation.
+Result<Expression::Operation> Negate(Expression::Operation op);
```
mapleFU: do not export?

src/iceberg/expression/expressions.cc
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  if (right->op() == Expression::Operation::kFalse) {
+    return left;
+  }
+
+  return std::make_shared<::iceberg::Or>(std::move(left), std::move(right));
+}
+
+// Transform functions
```
mapleFU, line 28: Can these be member of transform? A root expression might be too flexible?

src/iceberg/expression/predicate.h
mapleFU: Can we give `UnboundPredicate<BoundReference>` an alias for help understanding?

src/iceberg/expression/expression.cc
```diff
@@ -87,4 +91,91 @@ bool Or::Equals(const Expression& expr) const {
   return false;
 }
 
+std::string_view ToString(Expression::Operation op) {
+  switch (op) {
+    case Expression::Operation::kAnd:
```
mapleFU, line 96: Suddenly I found isn't this a Predicate::Operation rather than Expression::Operation?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,398 @@
...
+}
+
+template <typename B>
+Result<std::shared_ptr<Expression>> UnboundPredicate<B>::Negate() const {
+  ICEBERG_ASSIGN_OR_RAISE(auto negated_op, ::iceberg::Negate(BASE::op()));
+  return std::make_shared<UnboundPredicate>(negated_op, BASE::term(), values_);
+}
+
+template <typename B>
+Result<std::shared_ptr<Expression>> UnboundPredicate<B>::Bind(const Schema& schema,
```
mapleFU, line 125: Would a `getType -> OperationType` helper better in this case rather than `BASE::op() == Expression::Operation::kIn || BASE::op() == Expression::Operation::kNotIn`?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,398 @@
...
+        return Expressions::AlwaysTrue();
+      }
+      return std::make_shared<BoundUnaryPredicate>(Expression::Operation::kNotNull,
+                                                   std::move(bound_term));
+    case Expression::Operation::kIsNan:
+    case Expression::Operation::kNotNan:
+      if (!IsFloatingType(bound_term->type()->type_id())) {
+        return InvalidExpression("{} cannot be used with a non-floating-point column",
+                                 BASE::op());
+      }
```
mapleFU, line 171: Would this hardly happens since base class might already handle this problem?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,398 @@
...
+      BASE::op() == Expression::Operation::kNotStartsWith) {
+    if (bound_term->type()->type_id() != TypeId::kString) {
+      return InvalidExpression(
+          "Term for STARTS_WITH or NOT_STARTS_WITH must produce a string: {}: {}",
+          *bound_term, *bound_term->type());
+    }
+  }
+
+  if (values_.size() != 1) {
+    return InvalidExpression("Literal operation requires a single value");
```
mapleFU: Print the values_ size?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,398 @@
...
+  return std::make_shared<BoundLiteralPredicate>(BASE::op(), std::move(bound_term),
+                                                 literal);
+}
+
+template <typename B>
+Result<std::shared_ptr<Expression>> UnboundPredicate<B>::BindInOperation(
+    std::shared_ptr<B> bound_term) const {
+  std::vector<Literal> converted_literals;
+  for (const auto& literal : values_) {
+    auto primitive_type =
```
mapleFU, line 240: check literal not null before casting?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,398 @@
...
+        internal::checked_pointer_cast<PrimitiveType>(bound_term->type());
+    ICEBERG_ASSIGN_OR_RAISE(auto converted, literal.CastTo(primitive_type));
+    if (converted.IsNull()) {
+      return InvalidExpression("Invalid value for conversion to type {}: {} ({})",
+                               *bound_term->type(), literal.ToString(), *literal.type());
+    }
+    if (!converted.IsBelowMin() && !converted.IsAboveMax()) {
```
mapleFU, line 248: Add a comment for reason here?

src/iceberg/expression/term.cc
```diff
@@ -0,0 +1,161 @@
...
+}
+
+Result<Literal::Value> BoundTransform::Evaluate(const StructLike& data) const {
+  throw IcebergError("BoundTransform::Evaluate(StructLike) not implemented");
+}
+
+bool BoundTransform::MayProduceNull() const {
+  // transforms must produce null for null input values
+  // transforms may produce null for non-null inputs when not order-preserving
+  // FIXME: add Transform::is_order_preserving()
```
mapleFU, line 129: fixme with naming?

wgtmac: FIXME does not need a name. I just think that adding a name does not promise anything. :)

src/iceberg/expression/predicate.cc
wgtmac: I just found a inconsistency between Java and C++ impls about `Literal::CastTo`. The Java impl returns null for non-supported casting but C++ returns `NotSupported` error. Why is that?

wgtmac: The current impl will fail anyway if the input is a null literal because the `Literal::CastTo` has enforced this.

src/iceberg/expression/predicate.cc
wgtmac: No, the unbound form does not know its type but only the ref name.

src/iceberg/expression/predicate.cc
wgtmac: Since operation is already a member variable, we don't want an extra indirection from type to operation type.

src/iceberg/expression/expression.cc
wgtmac: I don't understand what you meant. Predicate is a subclass of Expression so it naturally has operation type. Non-predicate subclasses such as aggregate functions also have operation type.

src/iceberg/expression/predicate.h
wgtmac: I cannot come up with a good name. IMHO, an alias just adds another layer of indirection which may lead to more confusion.

src/iceberg/expression/expressions.cc
wgtmac: These are different concepts and abstractions:

- `Transform`: the simplest form of a transform function with a name and an optional parameter.
- `TransformFunction`: A bound version of `Transform` with concrete source and result types. It is like a compute kernel function to transform values.
- `UnboundTransform`: `Transform` with a name reference.
- `BoundTransform`: A bound version of `UnboundTransform` with bound reference.

In short, `Transform` and `TransformFunction` are abstractions for transform function metadata and kernels, and should have nothing to do with the expression/predicate. `UnboundTransform` and `BoundTransform` are expressions and are built on top of `Transform` and `TransformFunction`. It is an anti-pattern to create expression from non-expression class.

mapleFU (approved):

src/iceberg/expression/expression.cc
```diff
@@ -87,4 +91,100 @@ bool Or::Equals(const Expression& expr) const {
...
+    case Expression::Operation::kCount:
+      return "COUNT";
+    case Expression::Operation::kNot:
+      return "NOT";
+    case Expression::Operation::kCountStar:
+      return "COUNT_STAR";
+    case Expression::Operation::kMax:
+      return "MAX";
+    case Expression::Operation::kMin:
+      return "MIN";
```
zeroshade, line 141: I don't think these are in the spec, are they?

src/iceberg/expression/expression.cc
```diff
@@ -87,4 +91,100 @@ bool Or::Equals(const Expression& expr) const {
...
+      return Expression::Operation::kNotIn;
+    case Expression::Operation::kNotIn:
+      return Expression::Operation::kIn;
+    case Expression::Operation::kStartsWith:
+      return Expression::Operation::kNotStartsWith;
+    case Expression::Operation::kNotStartsWith:
+      return Expression::Operation::kStartsWith;
+    case Expression::Operation::kTrue:
+    case Expression::Operation::kFalse:
+    case Expression::Operation::kNot:
```
zeroshade, line 184: shouldn't there be negations for these?

src/iceberg/expression/expressions.cc
```diff
@@ -0,0 +1,356 @@
...
+#include "iceberg/exception.h"
+#include "iceberg/transform.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+// Logical operations
+
+std::shared_ptr<Expression> Expressions::And(std::shared_ptr<Expression> left,
+                                             std::shared_ptr<Expression> right) {
```
zeroshade: It's not necessary, but would be convenient (and come in handy) if you have a version of this which takes an arbitrary number of arguments and then collapses them into groups of `And(And(And(arg1, arg2), arg3), ....)`

src/iceberg/expression/expressions.cc
```diff
@@ -0,0 +1,356 @@
...
+
+  if (right->op() == Expression::Operation::kTrue) {
+    return left;
+  }
+
+  return std::make_shared<::iceberg::And>(std::move(left), std::move(right));
+}
+
+std::shared_ptr<Expression> Expressions::Or(std::shared_ptr<Expression> left,
+                                            std::shared_ptr<Expression> right) {
```
zeroshade: Same as above

src/iceberg/expression/expressions.cc
```diff
@@ -0,0 +1,356 @@
...
+std::shared_ptr<UnboundTransform> Expressions::Transform(
+    std::string name, std::shared_ptr<::iceberg::Transform> transform) {
+  return std::make_shared<UnboundTransform>(Ref(std::move(name)), std::move(transform));
+}
+
+// Template implementations for unary predicates
+
+std::shared_ptr<UnboundPredicate<BoundReference>> Expressions::IsNull(std::string name) {
+  return IsNull<BoundReference>(Ref(std::move(name)));
+}
```
zeroshade, line 66: this isn't a bound reference, shouldn't this be UnboundReference?

src/iceberg/expression/expressions.cc
```diff
@@ -0,0 +1,356 @@
...
+template <typename B>
+std::shared_ptr<UnboundPredicate<B>> Expressions::IsNull(
+    std::shared_ptr<UnboundTerm<B>> expr) {
+  return std::make_shared<UnboundPredicate<B>>(Expression::Operation::kIsNull,
+                                               std::move(expr));
+}
+
+std::shared_ptr<UnboundPredicate<BoundReference>> Expressions::NotNull(std::string name) {
+  return NotNull<BoundReference>(Ref(std::move(name)));
+}
```
zeroshade, line 77: same as above

src/iceberg/expression/expressions.cc
```diff
@@ -0,0 +1,356 @@
...
+template <typename B>
+std::shared_ptr<UnboundPredicate<B>> Expressions::NotNull(
+    std::shared_ptr<UnboundTerm<B>> expr) {
+  return std::make_shared<UnboundPredicate<B>>(Expression::Operation::kNotNull,
+                                               std::move(expr));
+}
+
+std::shared_ptr<UnboundPredicate<BoundReference>> Expressions::IsNaN(std::string name) {
+  return IsNaN<BoundReference>(Ref(std::move(name)));
+}
```
zeroshade, line 88: and so on....

src/iceberg/expression/expressions.h
```diff
@@ -0,0 +1,273 @@
...
+  /// \brief Create an IS NULL predicate for a field name.
+  static std::shared_ptr<UnboundPredicate<BoundReference>> IsNull(std::string name);
+
+  /// \brief Create an IS NULL predicate for an unbound term.
+  template <typename B>
+  static std::shared_ptr<UnboundPredicate<B>> IsNull(
+      std::shared_ptr<UnboundTerm<B>> expr);
+
+  /// \brief Create a NOT NULL predicate for a field name.
+  static std::shared_ptr<UnboundPredicate<BoundReference>> NotNull(std::string name);
```
zeroshade, line 130: A field name in and of itself is not a bound reference until you bind it to a schema and have the bound type, etc.

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,400 @@
...
+      if (!bound_term->MayProduceNull()) {
+        return Expressions::AlwaysTrue();
+      }
+      return std::make_shared<BoundUnaryPredicate>(Expression::Operation::kNotNull,
+                                                   std::move(bound_term));
+    case Expression::Operation::kIsNan:
+    case Expression::Operation::kNotNan:
+      if (!IsFloatingType(bound_term->type()->type_id())) {
+        return InvalidExpression("{} cannot be used with a non-floating-point column",
+                                 BASE::op());
```
zeroshade, line 170: I believe other implementations just return Null in this case rather than error

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,400 @@
...
+  if (values_.size() != 1) {
+    return InvalidExpression("Literal operation requires a single value but got {}",
+                             values_.size());
+  }
+
+  const auto& literal = values_[0];
+
+  if (literal.IsNull()) {
+    return InvalidExpression("Invalid value for conversion to type {}: {} ({})",
+                             *bound_term->type(), literal.ToString(), *literal.type());
```
zeroshade, line 201: why is it invalid to create a null for a type?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,400 @@
...
+          *bound_term, *bound_term->type());
+    }
+  }
+
+  if (values_.size() != 1) {
+    return InvalidExpression("Literal operation requires a single value but got {}",
+                             values_.size());
+  }
+
+  const auto& literal = values_[0];
```
zeroshade: should we cast the literal to the bound_term's type?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,400 @@
...
+    std::shared_ptr<B> bound_term) const {
+  std::vector<Literal> converted_literals;
+  for (const auto& literal : values_) {
+    auto primitive_type =
+        internal::checked_pointer_cast<PrimitiveType>(bound_term->type());
+    ICEBERG_ASSIGN_OR_RAISE(auto converted, literal.CastTo(primitive_type));
+    if (converted.IsNull()) {
+      return InvalidExpression("Invalid value for conversion to type {}: {} ({})",
+                               *bound_term->type(), literal.ToString(), *literal.type());
+    }
```
zeroshade, line 246: why filter out BelowMin and AboveMax but error on the cast?

src/iceberg/expression/predicate.cc
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// BoundSetPredicate implementation
+BoundSetPredicate::BoundSetPredicate(Expression::Operation op,
+                                     std::shared_ptr<BoundTerm> term,
+                                     std::span<const Literal> literals)
+    : BoundPredicate(op, std::move(term)) {
+  for (const auto& literal : literals) {
+    value_set_.push_back(literal.value());
+  }
```
zeroshade, line 379: If we're saying this is bound, shouldn't we validate that the literals are all the correct type and cast if they aren't?

src/iceberg/expression/term.cc
```diff
@@ -0,0 +1,161 @@
...
+
+Bound::~Bound() = default;
+
+BoundTerm::~BoundTerm() = default;
+
+Reference::~Reference() = default;
+
+template <typename B>
+Result<std::shared_ptr<B>> Unbound<B>::Bind(const Schema& schema) const {
+  return Bind(schema, /*case_sensitive=*/true);
```
zeroshade, line 41: shouldn't case_sensitive be a parameter rather than always be true?

src/iceberg/expression/term.cc
```diff
@@ -0,0 +1,161 @@
...
+}
+
+std::shared_ptr<Type> BoundTransform::type() const {
+  return transform_func_->ResultType();
+}
+
+bool BoundTransform::Equals(const BoundTerm& other) const {
+  if (other.kind() == Term::Kind::kTransform) {
+    const auto& other_transform = internal::checked_cast<const BoundTransform&>(other);
+    return *ref_ == *other_transform.ref_ && *transform_ == *other_transform.transform_;
```
zeroshade, line 140: can we settle on either overloading `operator==` or having an `Equals` method and not mix them?

src/iceberg/expression/term.cc
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  if (transform_->transform_type() == TransformType::kIdentity &&
+      other.kind() == Term::Kind::kReference) {
+    return ref_->Equals(other);
```
zeroshade: either this should be `*ref_ == other` or above should be `ref_->Equals(*other_transform.ref_)` right?

src/iceberg/expression/expression.cc
wgtmac: They do not exist in the rest spec but are supported expressions by the Java code.

src/iceberg/expression/expression.cc
wgtmac: I'm following the Java impl: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/Expression.java#L64-L97

src/iceberg/expression/expressions.cc
wgtmac: This is indeed a unbound reference. It is calling IsNull function with template parameter `BoundReference` to indicate its bound type.

src/iceberg/expression/expressions.h
wgtmac: Right, so the return type is a `UnboundPredicate` whose bound type is a `BoundReference `.

src/iceberg/expression/predicate.cc
wgtmac: The Java impl throws... 

https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java#L141-L152

src/iceberg/expression/predicate.cc
wgtmac: See https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java#L173-L176

For equality check with null values, we use Operation::kIsNull and Operation::kNotNull. In SQL spec, comparing with null always produces null and sometimes the semantics are not always respected by different engines.

src/iceberg/expression/predicate.cc
wgtmac: Good catch!

src/iceberg/expression/predicate.cc
wgtmac: Because those values are always evaluated to true but unsupported/invalid casts are unacceptable: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/UnboundPredicate.java#L222

src/iceberg/expression/predicate.cc
wgtmac: It has been guaranteed by UnboundPredicate<B>::BindInOperation. Perhaps I can add a DCHECK here.

src/iceberg/expression/term.cc
wgtmac: It is already supported by the interface:

```
class ICEBERG_EXPORT Unbound {
 public:
  /// \brief Bind this expression to a concrete schema.
  ///
  /// \param schema The schema to bind against
  /// \param case_sensitive Whether field name matching should be case sensitive
  /// \return A bound expression or an error if binding fails
  virtual Result<std::shared_ptr<B>> Bind(const Schema& schema,
                                          bool case_sensitive) const = 0;

  /// \brief Overloaded Bind method that uses case-sensitive matching by default.
  Result<std::shared_ptr<B>> Bind(const Schema& schema) const;
}
```

src/iceberg/expression/term.cc
wgtmac: There are other places using the same pattern. Perhaps I can create dedicated PR to clean them up.

wgtmac: I think the problem is that Equals can have virtual inheritance but `operator==` does not. So we depend on `Equals` to implement type-specific logic and also implement `operator==` to make it easier to use.

src/iceberg/expression/term.cc
wgtmac: Good catch!

src/iceberg/expression/expression.cc
```diff
@@ -87,4 +91,100 @@ bool Or::Equals(const Expression& expr) const {
...
+    case Expression::Operation::kCountStar:
+    case Expression::Operation::kMax:
+    case Expression::Operation::kMin:
+    case Expression::Operation::kAnd:
+    case Expression::Operation::kOr:
```
Fokko: This is interesting, we can negate `and` and `or` through the Morgan's law

src/iceberg/expression/predicate.cc
Fokko: I think throwing is reasonable

src/iceberg/expression/expression.cc
wgtmac: Done!

PR #181, opened 2025-08-16
feat: add row-based immutable data structure

wgtmac:
- Add StructLike, MapLike, and ArrayLike interfaces
- Add wrapper for ManifestFile and ArrowArray

### DIFF
--- a/src/iceberg/manifest_reader_internal.cc
+++ b/src/iceberg/manifest_reader_internal.cc
@@ -574,7 +574,7 @@ Result<std::vector<ManifestFile>> ManifestListReaderImpl::Files() const {
 }
 
 Result<ManifestFileField> ManifestFileFieldFromIndex(int32_t index) {
-  if (index >= 0 && index < static_cast<int32_t>(ManifestFileField::kNextId)) {
+  if (index >= 0 && index < static_cast<int32_t>(ManifestFileField::kNextUnusedId)) {
     return static_cast<ManifestFileField>(index);
   }
   return InvalidArgument("Invalid manifest file field index: {}", index);
--- a/src/iceberg/manifest_reader_internal.cc
+++ b/src/iceberg/manifest_reader_internal.cc
@@ -27,6 +27,7 @@
 #include "iceberg/manifest_list.h"
 #include "iceberg/schema.h"
 #include "iceberg/type.h"
+#include "iceberg/util/checked_cast.h"
 #include "iceberg/util/macros.h"
 
 namespace iceberg {
@@ -37,7 +38,7 @@ namespace iceberg {
   }
 
 #define PARSE_PRIMITIVE_FIELD(item, array_view, type)                                   \
-  for (size_t row_idx = 0; row_idx < array_view->length; row_idx++) {                   \
+  for (int64_t row_idx = 0; row_idx < array_view->length; row_idx++) {                  \
     if (!ArrowArrayViewIsNull(array_view, row_idx)) {                                   \
       auto value = ArrowArrayViewGetIntUnsafe(array_view, row_idx);                     \
       item = static_cast<type>(value);                                                  \
@@ -48,7 +49,7 @@ namespace iceberg {
   }
 
 #define PARSE_STRING_FIELD(item, array_view)                                            \
-  for (size_t row_idx = 0; row_idx < array_view->length; row_idx++) {                   \
+  for (int64_t row_idx = 0; row_idx < array_view->length; row_idx++) {                  \
     if (!ArrowArrayViewIsNull(array_view, row_idx)) {                                   \
       auto value = ArrowArrayViewGetStringUnsafe(array_view, row_idx);                  \
       item = std::string(value.data, value.size_bytes);                                 \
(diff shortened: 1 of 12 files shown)


src/iceberg/manifest_reader_internal.cc
```diff
@@ -567,4 +567,11 @@ Result<std::vector<ManifestFile>> ManifestListReaderImpl::Files() const {
   return manifest_files;
 }
 
+Result<ManifestFileField> ManifestFileFieldFromIndex(int32_t index) {
+  if (index >= 0 && index < static_cast<int32_t>(ManifestFileField::kNextId)) {
```
HeartLinked: Would it be better to standardize the code format using index >= static_cast<int32_t>(ManifestFileField::kManifestPath) ?

src/iceberg/manifest_reader_internal.h
```diff
@@ -54,4 +54,26 @@ class ManifestListReaderImpl : public ManifestListReader {
   std::unique_ptr<Reader> reader_;
 };
 
+enum class ManifestFileField : int32_t {
+  kManifestPath = 0,
+  kManifestLength,
```
HeartLinked:
```suggestion
  kManifestLength = 1,
```

src/iceberg/manifest_reader_internal.cc
wgtmac: Why? The 1st value is always at index 0 regardless of its name. Perhaps `kNextId` is confusing.

src/iceberg/row/struct_like.h
```diff
@@ -0,0 +1,96 @@
...
+/// Note that all string and binary values are stored as non-owned string_view
+/// and their lifetime is managed by the wrapped object.
+using Scalar = std::variant<std::monostate,  // for null
+                            bool,            // for boolean
+                            int32_t,         // for int, date
+                            int64_t,  // for long, timestamp, timestamp_tz, and time
+                            float,    // for float
+                            double,   // for double
+                            std::string_view,  // for non-owned string, binary and fixed
+                            std::shared_ptr<StructLike>,  // for struct
```
mapleFU, line 52: It's a bit confusing, if we want ownership, it might be:
```c++
Scalar = variant<Literal, std::shared_ptr<StructLike>, ...>
```

If we don't want ownership, it would be:

```
<...
  std::string_view,
  const StructLike*,
  const ArrayLike*,
  ...
>
```

Mixing string_view and sptr is so weird...

mapleFU: And scalar seems re-invent some member in literal?

wgtmac: These structures are all views. I don't want any ownership for string/binary/fixed/uuid, so use std::string_view and do not reuse Literal. But we need ownership for nested struct/map/list views, so sptr is used.

src/iceberg/row/manifest_wrapper.cc
```diff
@@ -0,0 +1,166 @@
...
+    case ManifestFileField::kContent:
+      return static_cast<int32_t>(manifest_file.content);
+    case ManifestFileField::kSequenceNumber:
+      return manifest_file.sequence_number;
+    case ManifestFileField::kMinSequenceNumber:
+      return manifest_file.min_sequence_number;
+    case ManifestFileField::kAddedSnapshotId:
+      return manifest_file.added_snapshot_id;
+    case ManifestFileField::kAddedFilesCount: {
+      if (manifest_file.added_files_count.has_value()) {
```
mapleFU: Wraps a FromOptional?

src/iceberg/row/manifest_wrapper.h
```diff
@@ -0,0 +1,100 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/manifest_wrapper.h
+/// Wrapper classes for manifest-related data structures that implement
+/// StructLike, ArrayLike, and MapLike interfaces for unified data access.
+
+#include <functional>
```
mapleFU, line 26: Why this is included?

wgtmac: Because `std::reference_wrapper` is used in the header file.

src/iceberg/row/struct_like.h
```diff
@@ -0,0 +1,99 @@
...
+#include "iceberg/type_fwd.h"
+#include "iceberg/util/decimal.h"
+
+namespace iceberg {
+
+/// \brief A scalar value depending on its data type.
+///
+/// Note that all string and binary values are stored as non-owned string_view
+/// and their lifetime is managed by the wrapped object.
+using Scalar = std::variant<std::monostate,  // for null
```
HuaHuaY, line 44: Can `Scalar` be merged with `Literal::Value`?

wgtmac: They are slightly different:
- scalar employs view-like semantics for var-length types but literal takes the ownership.
- scalar supports nested types but literal does not.

src/iceberg/row/struct_like.h
```diff
@@ -0,0 +1,99 @@
...
+                            bool,            // for boolean
+                            int32_t,         // for int, date
+                            int64_t,  // for long, timestamp, timestamp_tz, and time
+                            float,    // for float
+                            double,   // for double
+                            std::string_view,  // for non-owned string, binary and fixed
+                            Decimal,           // for decimal
+                            std::shared_ptr<StructLike>,  // for struct
```
HuaHuaY, line 52: In which case will a class except `ArrowArrayStructLike` inherit `StructLike`?

wgtmac: Any object that might be evaluated against expression. I've provided implementation for `ManifestFile` and we may extend it for `ManifestList`, etc.

Fokko:
Thanks @wgtmac for working on this, and thanks @zhjwpku @HuaHuaY @HeartLinked and @mapleFU for reviewing this 🙌

PR #182, opened 2025-08-17
feat: add decimal value representation

zhjwpku:


zhjwpku:
This is ready for review. I've extracted the Decimal128 implementation and tests from Arrow. Please help to review. @wgtmac @lidavidm @mapleFU

LICENSE
```diff
@@ -253,3 +253,17 @@ https://github.com/apache/arrow/blob/main/cpp/src/arrow/visit_type_inline.h
 Copyright: 2016-2025 The Apache Software Foundation.
 Home page: https://arrow.apache.org/
 License: https://www.apache.org/licenses/LICENSE-2.0
+
+--------------------------------------------------------------------------------
+
+The file src/iceberg/expression/decimal.h contains code adapted from
```
wgtmac: Why not putting in under `src/iceberg/util/decimal.h`? I believe it is also useful out of the scope of expression.

src/iceberg/util/port.h
```diff
@@ -0,0 +1,45 @@
...
+
+/// \file iceberg/util/port.h
+/// \brief Portability macros and definitions for Iceberg C++ library
+
+#pragma once
+
+#if defined(_WIN32) /* Windows is always little endian */ \
+    || defined(__LITTLE_ENDIAN__) ||                      \
+    (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
+#  define ICEBERG_LITTLE_ENDIAN 1
```
wgtmac: Can we use `std::endian::little` and `std::endian::big` by including `<bit>`?

src/iceberg/util/port.h
```diff
@@ -0,0 +1,45 @@
...
+#elif defined(__BIG_ENDIAN__) || \
+    (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
+#  define ICEBERG_LITTLE_ENDIAN 0
+#else
+#  error "Unsupported or unknown endianness"
+#endif
+
+#if defined(_MSC_VER)
+#  include <__msvc_int128.hpp>
+using int128_t = std::_Signed128;
```
wgtmac: They are actually unused? Can we remove them in this case?

src/iceberg/expression/decimal.h
```diff
@@ -0,0 +1,274 @@
...
+
+/// \file iceberg/expression/decimal.h
+/// \brief 128-bit fixed-point decimal numbers.
+/// Adapted from Apache Arrow with only Decimal128 support.
+/// https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/decimal.h
+
+#include <array>
+#include <cstdint>
+#include <iosfwd>
+#include <string>
```
wgtmac: Please remove unused headers

src/iceberg/expression/decimal.cc
```diff
@@ -0,0 +1,1439 @@
...
+ */
+
+/// \file iceberg/expression/decimal.cc
+/// \brief 128-bit fixed-point decimal numbers.
+/// Adapted from Apache Arrow with only Decimal128 support.
+/// https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/decimal.cc
+
+#include "iceberg/expression/decimal.h"
+
+#include <algorithm>
```
wgtmac: Remove duplicate and unused ones.

src/iceberg/expression/decimal.h
```diff
@@ -0,0 +1,274 @@
...
+};
+
+ICEBERG_EXPORT std::ostream& operator<<(std::ostream& os, const Decimal& decimal);
+
+// Unary operators
+ICEBERG_EXPORT Decimal operator-(const Decimal& operand);
+ICEBERG_EXPORT Decimal operator~(const Decimal& operand);
+
+// Binary operators
+ICEBERG_EXPORT bool operator<=(const Decimal& lhs, const Decimal& rhs);
```
wgtmac: Can we provide `operator<=>` instead of these four?

zhjwpku:
> Could you explicitly mark which part is exactly ported from Arrow to make the current review and future sync process easier?

That's not easy, Arrow's Decimal type uses a complex inheritance hierarchy and supports bit lengths in any 64-bit multiple. Since we only need Decimal128, l omitted the inheritance layer. But the algorithms adapted from Arrow is quite straightforward and should be easy for future sync.

src/iceberg/util/port.h
zhjwpku: Some test cases rely on int128_t, so I'd prefer to keep it for now.

wgtmac: Then I'd rather rename it to `iceberg/util/int128.h` so both Literal::Value and Scalar::Value can directly use int128_t to represent a decimal value.

src/iceberg/expression/decimal.h
```diff
@@ -0,0 +1,274 @@
...
+  static constexpr int32_t kBitWidth = 128;
+  static constexpr int32_t kByteWidth = kBitWidth / 8;
+  static constexpr int32_t kMaxPrecision = 38;
+  static constexpr int32_t kMaxScale = 38;
+
+  /// \brief Default constructor initializes to zero.
+  constexpr Decimal() noexcept : data_{0, 0} {}
+
+  /// \brief Construct a Decimal from an array of 2 64-bit integers.
+  explicit Decimal(std::array<uint64_t, 2> data) noexcept : data_(std::move(data)) {}
```
wgtmac: We are missing a ctor of `Decimal(const uint8_t*, size_t)`. This is required to restore encoded decimal values from binary data.

mapleFU:
What about a `thirdparty/` vendor directory, and we can use it same as `<arrow/xxx>` but it's vendored by us? (Current solving is also ok to me)

wgtmac:
I guess we cannot port `arrow/util/decimal.h` without significant change due to dependency on arrow utilities.

My concern is that we use `std::array<uint64_t, 2>` as the internal representation but use `int128_t` for testing. I'm not sure if we can directly use `int128_t` as the physical representation for the `Decimal` class. I did exactly this when I worked for my previous employer and it should work well.

src/iceberg/expression/decimal.h
zhjwpku: I'm not sure I get this, there are some usages of std::string in the methods.

src/iceberg/expression/decimal.h
zhjwpku: There is a `FromBigEndian` that could serve the same purpose. I initially avoided the ctor because it may throw an exception on invalid input, maybe the `FromBigEndian` is good enough?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  }
+
+  // Bytes are coming in big-endian, so the first byte is the MSB and therefore holds the
+  // sign bit.
+  const bool is_negative = static_cast<int8_t>(bytes[0]) < 0;
+
+  // 1. Extract the high bytes
+  // Stop byte of the high bytes
+  const int32_t high_bits_offset = std::max(0, length - 8);
+  const auto high_bits = UInt64FromBigEndian(bytes, high_bits_offset);
```
zhjwpku: It seems we could directly memcpy the bytes into a uint128_t and handle sign extension, but gcc 13 does not support `byteswap` for uint128_t, see [0], which limits this approach.

[0] https://github.com/zhjwpku/iceberg-cpp/actions/runs/17339119789/job/49230499646#step:3:386

LICENSE
```diff
@@ -253,3 +253,17 @@ https://github.com/apache/arrow/blob/main/cpp/src/arrow/visit_type_inline.h
 Copyright: 2016-2025 The Apache Software Foundation.
 Home page: https://arrow.apache.org/
 License: https://www.apache.org/licenses/LICENSE-2.0
+
+--------------------------------------------------------------------------------
+
+The file src/iceberg/util/decimal.h contains code adapted from
```
wgtmac, line 253: Should we merge these with the above section? They are all from Arrow.

src/iceberg/util/decimal.h
```diff
@@ -0,0 +1,265 @@
...
+
+#include <array>
+#include <bit>
+#include <cstdint>
+#include <iosfwd>
+#include <string>
+#include <string_view>
+#include <type_traits>
+
+#include <iceberg/type.h>
```
wgtmac:
```suggestion
#include "iceberg/type.h"
```

Do we really need this?

test/CMakeLists.txt
```diff
@@ -75,7 +75,11 @@ add_iceberg_test(table_test
                  table_test.cc
                  schema_json_test.cc)
 
-add_iceberg_test(expression_test SOURCES expression_test.cc literal_test.cc)
+add_iceberg_test(expression_test
+                 SOURCES
+                 decimal_test.cc
```
wgtmac: Move it to `util_test`?

src/iceberg/util/decimal.h
```diff
@@ -0,0 +1,265 @@
...
+  /// \param scale The scale to use for the string representation.
+  /// \return The string representation of the Decimal value.
+  Result<std::string> ToString(int32_t scale = 0) const;
+
+  /// \brief Convert the Decimal value to an integer string.
+  std::string ToIntegerString() const;
+
+  /// \brief Convert the decimal string to a Decimal value, optionally including precision
+  /// and scale if they are provided not null.
+  static Result<Decimal> FromString(std::string_view str, int32_t* precision = nullptr,
```
wgtmac, line 159: Add `\param[out]` for precision and scale?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+    out = out * multiple + value;
+    pos += group_size;
+  }
+}
+
+static void AdjustIntegerStringWithScale(std::string* str, int32_t scale) {
+  if (scale == 0) {
+    return;
+  }
+  assert(str != nullptr);
```
wgtmac: What about define `ICEBERG_DCHECK` in the `iceberg/util/macros.h` like below?

```cpp
#include <cassert>

#define ICEBERG_DCHECK(exp, msg) assert((void(msg), exp))
```

src/iceberg/util/decimal.h
```diff
@@ -0,0 +1,265 @@
...
+  static constexpr int32_t kBitWidth = 128;
+  static constexpr int32_t kByteWidth = kBitWidth / 8;
+  static constexpr int32_t kMaxPrecision = 38;
+  static constexpr int32_t kMaxScale = 38;
+
+  /// \brief Default constructor initializes to zero.
+  constexpr Decimal() noexcept = default;
+
+  /// \brief Create a Decimal from a 128-bit integer.
+  constexpr Decimal(int128_t value) noexcept  // NOLINT
```
wgtmac: Why do we need to add `NOLINT`?

src/iceberg/util/decimal.h
```diff
@@ -0,0 +1,265 @@
...
+      : data_(value) {}
+
+  /// \brief Create a Decimal from any integer not wider than 64 bits.
+  template <typename T>
+    requires(std::is_integral_v<T> && (sizeof(T) <= sizeof(uint64_t)))
+  constexpr Decimal(T value) noexcept  // NOLINT
```
wgtmac: Same question as above

src/iceberg/util/decimal.h
```diff
@@ -0,0 +1,265 @@
...
+  friend Decimal operator-(const Decimal& lhs, const Decimal& rhs);
+  friend Decimal operator*(const Decimal& lhs, const Decimal& rhs);
+  friend Decimal operator/(const Decimal& lhs, const Decimal& rhs);
+  friend Decimal operator%(const Decimal& lhs, const Decimal& rhs);
+
+ private:
+  int128_t data_{0};
+};
+
+ICEBERG_EXPORT std::ostream& operator<<(std::ostream& os, const Decimal& decimal);
```
wgtmac, line 219: Should we inherit it from `util::Formattable` to enable it to be used in `std::format`?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  requires std::is_signed_v<SignedInt> && std::is_integral_v<Shift>
+constexpr SignedInt SafeLeftShift(SignedInt u, Shift shift) {
+  using UnsignedInt = std::make_unsigned_t<SignedInt>;
+  return static_cast<SignedInt>(static_cast<UnsignedInt>(u) << shift);
+}
+
+struct DecimalComponents {
+  std::string_view while_digits;
+  std::string_view fractional_digits;
+  int32_t exponent;
```
wgtmac:
```suggestion
  int32_t exponent{0};
```

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  char sign{0};
+  bool has_exponent{false};
+};
+
+inline bool IsSign(char c) { return c == '+' || c == '-'; }
```
wgtmac, line 55: Do we really need the inline keyword?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  pos = ParseDigitsRun(str, pos, &out->while_digits);
+  if (pos == str.size()) {
+    return !out->while_digits.empty();
+  }
+
+  // Optional dot
+  if (IsDot(str[pos])) {
+    ++pos;
+    // Second run of digits after the dot
+    pos = ParseDigitsRun(str, pos, &out->fractional_digits);
```
wgtmac:
```suggestion
    // Second run of digits after the dot
    pos = ParseDigitsRun(str, ++pos, &out->fractional_digits);
```

Is it safe to do this?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+    Decimal(542101086242LL, 13875954555633532928ULL),
+    Decimal(5421010862427LL, 9632337040368467968ULL),
+    Decimal(54210108624275LL, 4089650035136921600ULL),
+    Decimal(542101086242752LL, 4003012203950112768ULL),
+    Decimal(5421010862427522LL, 3136633892082024448ULL),
+    Decimal(54210108624275221LL, 12919594847110692864ULL),
+    Decimal(542101086242752217LL, 68739955140067328ULL),
+    Decimal(5421010862427522170LL, 687399551400673280ULL)};
+
+static inline void ShiftAndAdd(std::string_view input, uint128_t& out) {
```
wgtmac: Is `static inline` necessary?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  for (size_t pos = 0; pos < input.size();) {
+    const size_t group_size = std::min(kInt64DecimalDigits, input.size() - pos);
+    const uint64_t multiple = kUInt64PowersOfTen[group_size];
+    uint64_t value = 0;
+
+    std::from_chars(input.data() + pos, input.data() + pos + group_size, value);
```
wgtmac: We need to deal with parse error.

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+
+    out = out * multiple + value;
+    pos += group_size;
+  }
+}
+
+static void AdjustIntegerStringWithScale(std::string* str, int32_t scale) {
```
wgtmac:
```suggestion
void AdjustIntegerStringWithScale(int32_t scale, std::string* str) {
```

Let's keep the original signature from Arrow. Usually the output param is placed at last.

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  // After setting decimal point: *str = "-0.0123"
+  str->insert(is_negative_offset, scale - num_digits + 2, '0');
+  str->at(is_negative_offset + 1) = '.';
+}
+
+}  // namespace
+
+Decimal::Decimal(std::string_view str) {
+  auto result = Decimal::FromString(str);
+  if (!result) {
```
wgtmac, line 286: Can we add `Decimal::Make` to handle the error and do not throw in the ctor?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+    throw std::runtime_error(
+        std::format("Failed to parse Decimal from string: {}, error: {}", str,
+                    result.error().message));
+  }
+  *this = std::move(result.value());
+}
+
+Decimal& Decimal::Negate() {
+  uint128_t u = static_cast<uint128_t>(~data_) + 1;
```
wgtmac:
```suggestion
  uint128_t u = ~static_cast<uint128_t>(data_) + 1;
```

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  data_ |= other.data_;
+  return *this;
+}
+
+Decimal& Decimal::operator&=(const Decimal& other) {
+  data_ &= other.data_;
+  return *this;
+}
+
+Decimal& Decimal::operator<<=(uint32_t shift) {
```
wgtmac:
```suggestion
Decimal& Decimal::operator<<=(uint32_t bits) {
```

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+      data_ = static_cast<int128_t>(static_cast<uint128_t>(data_) << shift);
+    } else {
+      data_ = 0;
+    }
+  }
+
+  return *this;
+}
+
+Decimal& Decimal::operator>>=(uint32_t shift) {
```
wgtmac:
```suggestion
Decimal& Decimal::operator>>=(uint32_t bits) {
```

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  data_ &= other.data_;
+  return *this;
+}
+
+Decimal& Decimal::operator<<=(uint32_t shift) {
+  if (shift != 0) {
+    if (shift < 128) {
+      data_ = static_cast<int128_t>(static_cast<uint128_t>(data_) << shift);
+    } else {
+      data_ = 0;
```
wgtmac: Do we actually need the else branch?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+
+  return *this;
+}
+
+Decimal& Decimal::operator>>=(uint32_t shift) {
+  if (shift != 0) {
+    if (shift < 128) {
+      data_ >>= shift;
+    } else {
+      data_ = (data_ < 0) ? -1 : 0;
```
wgtmac: Same question here

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+      bits -= 64;
+    }
+    if (bits > 0) {
+      shifted = (result_lo << (64 - bits)) | (shifted > 0);
+      result_lo >>= bits;
+      result_lo |= static_cast<uint64_t>(result_hi) << (64 - bits);
+      result_hi >>= bits;
+    }
+    // We almost have our result, but now do the rounding.
+    constexpr uint64_t kHalf = 0x8000000000000000ULL;
```
wgtmac: We can define `constexpr uint128_t kHalf = ...` to simplify this just like what `Decimal64RealConversion::RoundedRightShift` does in Arrow.

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+        result_hi += (result_lo == 0);
+      }
+    } else {
+      // Strictly less than half => round down
+    }
+    return Decimal{result_hi, result_lo};
+  }
+
+  template <typename Real>
+  static Result<Decimal> FromPositiveApprox(Real real, int32_t precision, int32_t scale) {
```
wgtmac:
```suggestion
  static Result<Decimal> FromPositiveRealApprox(Real real, int32_t precision, int32_t scale) {
```

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+    // Approximate algorithm that operates in the FP domain (thus subject
+    // to precision loss).
+    const auto x = std::nearbyint(real * PowerOfTen<double>(scale));
+    const auto max_abs = PowerOfTen<double>(precision);
```
wgtmac:
```suggestion
    const auto x = std::nearbyint(real * PowerOfTen<Real>(scale));
    const auto max_abs = PowerOfTen<Real>(precision);
```

src/iceberg/util/decimal.h
```diff
@@ -0,0 +1,265 @@
...
+  std::string ToIntegerString() const;
+
+  /// \brief Convert the decimal string to a Decimal value, optionally including precision
+  /// and scale if they are provided not null.
+  static Result<Decimal> FromString(std::string_view str, int32_t* precision = nullptr,
+                                    int32_t* scale = nullptr);
+
+  /// \brief Convert the floating-point value to a Decimal value with the given
+  /// precision and scale.
+  static Result<Decimal> FromReal(double real, int32_t precision, int32_t scale);
```
wgtmac: TBH, I don't think we need `FromReal`, `ToReal`, `ToFloat`, `ToDouble` and even `ToInteger`. I know you've spent a lot of time in porting these functions. WDYT?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+  if (scale != nullptr) {
+    *scale = parsed_scale;
+  }
+
+  return result;
+}
+
+namespace {
+
+constexpr float kFloatInf = std::numeric_limits<float>::infinity();
```
wgtmac: As I have commented in the decimal.h, if we decide not to support floating-point conversion, we can avoid maintaining a large portion of code starting from here.

src/iceberg/util/decimal.cc
wgtmac: https://github.com/apache/iceberg-cpp/pull/196 is merged, is it possible to add `int128_t` and `uint128_t` to it as template specializations?

wgtmac: BTW, you don't have to call byteswap on uint128_t. You may reinterpret_cast it as uint64_t[2] and swap them individually.

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+    low = SafeLeftShift(low, low_bits_offset * CHAR_BIT);
+    // Preserve the upper bits by inplace OR-ing the int64_t
+    low |= low_bits;
+  }
+
+  return Decimal(high, static_cast<uint64_t>(low));
+}
+
+Result<Decimal> Decimal::Rescale(int32_t orig_scale, int32_t new_scale) const {
+  if (orig_scale < 0 || orig_scale > kMaxScale) {
```
wgtmac: Is this correct? Isn't negative scale supported?

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,1044 @@
...
+
+  assert(abs_delta_scale <= kMaxScale);
+
+  auto& multiplier = kDecimal128PowersOfTen[abs_delta_scale];
+
+  const bool rescale_would_cause_data_loss =
+      RescaleWouldCauseDataLoss(*this, delta_scale, multiplier, &out);
+
+  if (rescale_would_cause_data_loss) {
+    return Invalid("Rescale would cause data loss: {} -> {}", ToIntegerString(),
```
wgtmac: Isn't `out.ToIntegerString()` undefined in the error message?

src/iceberg/util/decimal.h
zhjwpku: no, that was added by IDE, sorry for the oversight.

LICENSE
zhjwpku: good point

src/iceberg/util/decimal.h
zhjwpku: Yeah, that's because it allows implicit constructor. I change that to `// NOLINT(google-explicit-constructor)` to make it more clear

src/iceberg/util/decimal.h
zhjwpku: will do, thanks

src/iceberg/util/decimal.cc
zhjwpku: `inline` suggests compiler to embed the code in the caller function for better performance, decision stilled made by compiler. Do you think that's not necessary?

src/iceberg/util/decimal.cc
zhjwpku: should be safe for that part.

src/iceberg/util/decimal.cc
zhjwpku: `static` is not necessary since it's in a anonymous namespace, but not sure for `inline` :(

src/iceberg/util/decimal.cc
zhjwpku: Make sense, I'm not aware of that before.

src/iceberg/util/decimal.cc
zhjwpku: We already have `Decimal::FromString` which seems to serve the same purpose as `Decimal::Make`. The ctor makes the tests easier, I'm also not satisfied with the throw though.

src/iceberg/util/decimal.cc
zhjwpku: removed, we should use the int128 shift operation directly.

src/iceberg/util/decimal.cc
zhjwpku: Good catch, I originally copy this from Decimal128 version of FromPositiveRealApprox, but after some dig, Real should be fine.

src/iceberg/util/decimal.h
zhjwpku: Ok, removing all these conversions can make the code cleaner. I will remove these and keep a branch in my repo in case someday we need them.

src/iceberg/util/decimal.cc
zhjwpku: Good catch, I checked arrow decimal again, this is wrong.

src/iceberg/util/decimal.cc
zhjwpku: Good catch, change to a safe error message.

src/iceberg/util/decimal.cc
zhjwpku: I choose the second ;)

src/iceberg/util/decimal.cc
zhjwpku: Added.

src/iceberg/util/decimal.cc
zhjwpku: done with ICEBERG_DCHECK

src/iceberg/util/decimal.cc
zhjwpku: No longer needed.

src/iceberg/util/decimal.cc
zhjwpku: done.

src/iceberg/util/decimal.h
```diff
@@ -0,0 +1,221 @@
...
+  constexpr Decimal(int128_t value) noexcept  // NOLINT(google-explicit-constructor)
+      : data_(value) {}
+
+  /// \brief Create a Decimal from any integer not wider than 64 bits.
+  template <typename T>
+    requires(std::is_integral_v<T> && (sizeof(T) <= sizeof(uint64_t)))
+  constexpr Decimal(T value) noexcept  // NOLINT(google-explicit-constructor)
+  {
+    data_ = static_cast<int128_t>(value);
+  }
```
wgtmac:
```suggestion
      : data_(static_cast<int128_t>(value)) {}
```

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,589 @@
...
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#include <format>
+#include <iomanip>
+#include <limits>
+#include <sstream>
+#include <string>
+#include <string_view>
+#include <utility>
```
wgtmac:
```suggestion
#include <bit>
#include <charconv>
#include <climits>
#include <cmath>
#include <cstring>
#include <format>
#include <iomanip>
#include <limits>
#include <sstream>
#include <utility>
```

src/iceberg/util/decimal.cc
wgtmac: Then please throw `IcebergError` and add a comment with `\throw` by recommending `Decimal::FromString` to be used in production.

src/iceberg/util/decimal.h
zhjwpku: I went with `// NOLINT implicit conversion`, since Arrow uses this convention and I think it's better not to include anything tied to `google`.

zhjwpku:
@mapleFU Can you take a look at this again, thanks.

wgtmac:
This is mainly ported from Arrow C++. Could you help take a quick pass? @lidavidm @zeroshade

src/iceberg/util/decimal.cc
```diff
@@ -0,0 +1,585 @@
...
+    auto adjusted_exponent = dec.exponent;
+    parsed_scale = static_cast<int32_t>(dec.fractional_digits.size()) - adjusted_exponent;
+  } else {
+    parsed_scale = static_cast<int32_t>(dec.fractional_digits.size());
+  }
+
+  uint128_t value = 0;
+  ShiftAndAdd(dec.while_digits, value);
+  ShiftAndAdd(dec.fractional_digits, value);
+  Decimal result(static_cast<int128_t>(value));
```
Fokko, line 445: Keep in mind that Iceberg might also store the Decimal as an int32 or int64:
image

Fokko (approved):
This looks good to me @zhjwpku Thanks for working on this, and thanks @wgtmac and @mapleFU for the review 🙌

src/iceberg/util/decimal.cc
zhjwpku: Thanks for the reminder, will keep in mind, thanks.

PR #185, opened 2025-08-20
feat: implement Literal class with serialization

HeartLinked:
This PR implements the Literal class for the Iceberg expression system, providing comprehensive support for literal values in query expressions.


src/iceberg/expression/literal.h
```diff
@@ -144,6 +144,19 @@ class ICEBERG_EXPORT Literal {
   Literal(Value value, std::shared_ptr<PrimitiveType> type);
 
   friend class LiteralCaster;
+  friend class LiteralSerializer;
+
+  /// \brief Format a date value as a string.
+  std::string FormatDate(int32_t days_since_epoch) const;
```
wgtmac: These `FormatXXX` functions are used only internally, right? We should either move them to `literal.cc` or add a separate utility class for formatting time values like `iceberg/util/time_utils.h`. I prefer the latter because we can add test cases for them.

test/manifest_list_reader_test.cc
```diff
@@ -76,43 +77,40 @@ class ManifestListReaderV1Test : public ManifestListReaderTestBase {
...
+         .added_files_count = 4,
+         .existing_files_count = 0,
+         .deleted_files_count = 0,
+         .added_rows_count = 6,
+         .existing_rows_count = 0,
+         .deleted_rows_count = 0,
+         .partitions = {{.contains_null = false,
+                         .contains_nan = false,
+                         .lower_bound =
+                             iceberg::Literal::String("2022-02-22").Serialize().value(),
```
wgtmac:
```suggestion
                             Literal::String("2022-02-22").Serialize().value(),
```

test/manifest_list_reader_test.cc
```diff
@@ -76,43 +77,40 @@ class ManifestListReaderV1Test : public ManifestListReaderTestBase {
...
+                         .upper_bound =
+                             iceberg::Literal::String("2022-2-23").Serialize().value()}}},
```
wgtmac:
```suggestion
                             Literal::String("2022-2-23").Serialize().value()}}},
```

test/manifest_list_reader_test.cc
```diff
@@ -76,43 +77,40 @@ class ManifestListReaderV1Test : public ManifestListReaderTestBase {
...
+         .added_files_count = 0,
+         .existing_files_count = 0,
+         .deleted_files_count = 2,
+         .added_rows_count = 0,
+         .existing_rows_count = 0,
+         .deleted_rows_count = 6,
+         .partitions = {
+             {.contains_null = false,
+              .contains_nan = false,
+              .lower_bound = iceberg::Literal::String("2022-2-22").Serialize().value(),
```
wgtmac:
```suggestion
              .lower_bound = Literal::String("2022-2-22").Serialize().value(),
```

test/manifest_list_reader_test.cc
```diff
@@ -76,43 +77,40 @@ class ManifestListReaderV1Test : public ManifestListReaderTestBase {
...
+              .upper_bound =
+                  iceberg::Literal::String("2022-2-23").Serialize().value()}}}};
```
wgtmac:
```suggestion
                  Literal::String("2022-2-23").Serialize().value()}}}};
```

test/manifest_reader_test.cc
```diff
@@ -59,24 +59,33 @@ class ManifestReaderV1Test : public TempFileTestBase {
...
-         {4, {0xc0, 0x19, '#', '=', 0xe2, 0x0f, 0x06, 0x00}}},
-        {{1, {'{', 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}},
-         {2, {0xc8, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}},
-         {3, {0x0e, '"'}},
-         {4, {0xc0, 0xd9, '7', 0x93, 0x1f, 0xf3, 0x05, 0x00}}},
-        {{1, {'{', 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}},
-         {2, {0xc8, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}},
-         {3, {0x0e, '!'}},
-         {4, {0xc0, 0x19, 0x10, '{', 0xc2, 0xb9, 0x05, 0x00}}},
+        {{1, iceberg::Literal::Long(1234).Serialize().value()},
```
wgtmac:
```suggestion
        {{1, Literal::Long(1234).Serialize().value()},
```

wgtmac: Same apply for below

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
 
 #include "iceberg/exception.h"
+#include "iceberg/util/macros.h"
 
 namespace iceberg {
 
+namespace {
+/// \brief Write a value in little-endian format to the buffer.
+template <typename T>
+void WriteLittleEndian(std::vector<uint8_t>& buffer, T value) {
```
wgtmac: I would recommend to add `iceberg/util/endian.h` like https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/endian.h

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+  static_assert(std::is_trivially_copyable_v<T>, "Type must be trivially copyable");
```
wgtmac: Can we use C++20 concepts instead of `static_assert` which is more type-safe? We should only support limited set of types.

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+  std::memcpy(&value, data.data(), sizeof(T));
+
+  if constexpr (std::endian::native != std::endian::little && sizeof(T) > 1) {
+    value = std::byteswap(value);
+  }
+  return value;
+}
+
+/// \brief Write a 16-byte value in big-endian format (for UUID and Decimal).
+void WriteBigEndian16(std::vector<uint8_t>& buffer,
```
wgtmac: Is it possible to reuse the same template function signature as `WriteBigEndian` with specialization for `std::array<uint8_t, 16>`?

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+                      const std::array<uint8_t, 16>& value) {
+  buffer.insert(buffer.end(), value.begin(), value.end());
+}
+
+/// \brief Read a 16-byte value in big-endian format (for UUID and Decimal).
+Result<std::array<uint8_t, 16>> ReadBigEndian16(std::span<const uint8_t> data) {
```
wgtmac: Same question as above.

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
                       target_type->ToString());
 }
 
+// LiteralSerializer implementation
+
+Result<std::vector<uint8_t>> LiteralSerializer::ToBytes(const Literal& literal) {
+  // Cannot serialize special values
+  if (literal.IsAboveMax()) {
+    return InvalidArgument("Cannot serialize AboveMax literal");
```
```suggestion
    return NotSupported("Cannot serialize AboveMax");
```

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+  }
+  if (literal.IsBelowMin()) {
+    return InvalidArgument("Cannot serialize BelowMin literal");
```
```suggestion
    return NotSupported("Cannot serialize BelowMin");
```

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+  }
+
+  std::vector<uint8_t> result;
+
+  // Null values serialize to empty buffer
+  if (literal.IsNull()) {
+    return result;
```
```suggestion
    return NotSupported("Cannot serialize null");
```

wgtmac: It shouldn't serialize null.

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+
+    case TypeId::kUuid: {
+      // 16-byte big-endian value
+      const auto& uuid_bytes = std::get<std::array<uint8_t, 16>>(value);
+      WriteBigEndian16(result, uuid_bytes);
+      return result;
+    }
+
+    default:
+      return NotImplemented("Serialization for type {} is not supported",
```
```suggestion
      return NotSupported("Serialization for type {} is not supported",
```

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+      if (data.size() == 4) {
+        // Type was promoted from int to long
+        ICEBERG_ASSIGN_OR_RAISE(auto int_value, ReadLittleEndian<int32_t>(data));
+        value = static_cast<int64_t>(int_value);
+      } else if (data.size() == 8) {
+        // Standard 8-byte long
+        ICEBERG_ASSIGN_OR_RAISE(auto long_value, ReadLittleEndian<int64_t>(data));
+        value = long_value;
+      } else {
+        const char* type_name = [type_id]() {
```
wgtmac: Add a `ICEBERG_EXPORT std::string_view ToString(TypeId id)` to `iceberg/util/type.h`? Please also use lowercase form like what we did in `Type::ToString()`.

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+      ICEBERG_ASSIGN_OR_RAISE(auto value, ReadLittleEndian<float>(data));
+      return Literal::Float(value);
+    }
+
+    case TypeId::kDouble: {
+      if (data.size() == 4) {
+        // Type was promoted from float to double
+        ICEBERG_ASSIGN_OR_RAISE(auto float_value, ReadLittleEndian<float>(data));
+        return Literal::Double(static_cast<double>(float_value));
+      } else if (data.size() == 8) {
```
wgtmac: Put `if (data.size() == 8)` before `if (data.size() == 4)`. The reason is that `8` should be more common than `4` in real cases.

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+      std::array<uint8_t, 16> decimal_bytes{};
+      // Copy data to the end of the array (big-endian format for decimals)
+      // This handles variable-length decimals by right-aligning them
+      std::ranges::copy(data, decimal_bytes.end() - data.size());
+      return Literal(Literal::Value{decimal_bytes}, type);
+    }
+
+    default:
+      return NotImplemented("Deserialization for type {} is not supported",
+                            type->ToString());
```
```suggestion
      return NotSupported("Cannot deserialize type: {}", type->ToString());
```

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+  }
+
+  // Should not reach here
+  return InvalidArgument("Unexpected error in deserialization");
```
```suggestion
  std::unreachable();
```

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+    default:
+      return InvalidArgument("Unsupported long type: {}", static_cast<int>(type_id));
+  }
+}
+
+}  // namespace
+
+/// \brief LiteralSerializer handles serialization/deserialization operations for Literal.
+/// This is an internal implementation class.
+class LiteralSerializer {
```
wgtmac: Checking the java code, it seems that single-value serialization is not only used by Literal. We need to create a separate utility file to support reuse. For example, we can create `iceberg/util/conversions.h` and provide following functions:

```cpp
class ICEBERG_EXPORT Conversions {
 public:
  static Result<std::vector<uint8_t>> ToBytes(const Primitive& type, const Literal::Value& value);

  static Result<std::vector<uint8_t>> ToBytes(const Literal& literal);

  static Result<Literal::Value> FromBytes(const Primitive& type, std::span<const uint8_t> data);

  static Result<Literal> FromBytes(std::shared_ptr<Primitive> type, std::span<const uint8_t> data);
};
```

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+              return "TimestampTz";
+            default:
+              return "Unknown";
+          }
+        }();
+        return InvalidArgument("{} requires 4 or 8 bytes, got {}", type_name,
+                               data.size());
+      }
+
+      return CreateLongTypeLiteral(value, type_id);
```
```suggestion
      return {value, std::move(type)};
```

Why not directly use `Literal(Value value, std::shared_ptr<PrimitiveType> type)` to create it? You don't need CreateLongTypeLiteral.

src/iceberg/expression/literal.cc
```diff
@@ -294,13 +388,21 @@ std::string Literal::ToString() const {
...
+      return FormatTimestampTz(std::get<int64_t>(value_));
+    }
     case TypeId::kDecimal:
     case TypeId::kUuid:
-    case TypeId::kFixed:
-    case TypeId::kDate:
-    case TypeId::kTime:
-    case TypeId::kTimestamp:
-    case TypeId::kTimestampTz: {
+    case TypeId::kFixed: {
```
wgtmac: BTW, please help me check the `ToString` should follow these rules:

1. String literals should be printed as is.
2. Binary and fixed literals should be printed as base16-encoded texts and quoted by `X''`.
3. Other values should be quoted in `""`.

I'd suggest to create a separate PR to fix this. WDYT?

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+  } else {
+    if constexpr (sizeof(T) > 1) {
+      T le_value = std::byteswap(value);
+      const auto* bytes = reinterpret_cast<const uint8_t*>(&le_value);
+      buffer.insert(buffer.end(), bytes, bytes + sizeof(T));
+    } else {
+      // For single byte types, no byteswap needed
+      buffer.push_back(static_cast<uint8_t>(value));
+    }
+  }
```
```suggestion
  } else if constexpr (sizeof(T) > 1) {
    T le_value = std::byteswap(value);
    const auto* bytes = reinterpret_cast<const uint8_t*>(&le_value);
    buffer.insert(buffer.end(), bytes, bytes + sizeof(T));
  } else {
    // For single byte, no byteswap needed
    buffer.push_back(static_cast<uint8_t>(value));
  }
```

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+}
+
+/// \brief Read a value in little-endian format from the data.
+template <typename T>
+Result<T> ReadLittleEndian(std::span<const uint8_t> data) {
+  static_assert(std::is_trivially_copyable_v<T>, "Type must be trivially copyable");
```
wgtmac: ditio, try to explore concepts.

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+
+  if (data.size() < sizeof(T)) {
```
```suggestion
  if (data.size() < sizeof(T)) [[unlikely]] {
```

src/iceberg/expression/literal.cc
mapleFU: In C, the abi can pass 16bit argument in registers under linux platform, so generally small value would be better when passing by value

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,107 @@
...
+    return InvalidArgument("Insufficient data to read {} bytes, got {}", sizeof(T),
+                           data.size());
+  }
+
+  T value;
+  std::memcpy(&value, data.data(), sizeof(T));
+
+  if constexpr (std::endian::native != std::endian::little && sizeof(T) > 1) {
+    value = std::byteswap(value);
+  }
```
mapleFU: Maybe we can add a `FromLittleEndian` and `ToLittleEndian` function here

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+    case TypeId::kFixed: {
+      // Fixed(L) - Binary value, could be stored in std::array<uint8_t, 16> or
+      // std::vector<uint8_t>
+      if (std::holds_alternative<std::array<uint8_t, 16>>(value)) {
+        const auto& fixed_bytes = std::get<std::array<uint8_t, 16>>(value);
+        result.insert(result.end(), fixed_bytes.begin(), fixed_bytes.end());
+      } else if (std::holds_alternative<std::vector<uint8_t>>(value)) {
+        result = std::get<std::vector<uint8_t>>(value);
+      } else {
+        return InvalidArgument("Invalid value type for Fixed literal");
```
mapleFU: Maybe can print the literal here

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+      // UTF-8 bytes (without length)
+      const auto& str = std::get<std::string>(value);
+      result.insert(result.end(), str.begin(), str.end());
+      return result;
+    }
+
+    case TypeId::kBinary: {
+      // Binary value (without length)
+      const auto& binary_data = std::get<std::vector<uint8_t>>(value);
+      result = binary_data;
```
mapleFU: The style here mixes assign and insert back, the code is right but I think keep a same style might be better ( .i.e using `result.insert` )

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+    }
+
+    default:
+      return NotImplemented("Serialization for type {} is not supported",
+                            literal.type()->ToString());
+  }
+}
+
+Result<Literal> LiteralSerializer::FromBytes(std::span<const uint8_t> data,
+                                             std::shared_ptr<PrimitiveType> type) {
```
mapleFU: why this type is not passed by reference?

src/iceberg/expression/literal.cc
```diff
@@ -355,4 +457,305 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+    return InvalidArgument("Type cannot be null");
+  }
+
+  // Empty data represents null value
+  if (data.empty()) {
+    return Literal::Null(type);
+  }
+
+  const auto type_id = type->type_id();
+
```
mapleFU: I found a weird thing here. This part of the code only checks the buffer size in few types, it assumes the buffer-size equals to the size upper layer passed in. Should we do some check here?

HeartLinked: fixed

src/iceberg/expression/literal.cc
HeartLinked: fixed

src/iceberg/expression/literal.cc
HeartLinked: fixed

src/iceberg/expression/literal.cc
HeartLinked: ok

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,93 @@
...
+#include <cstring>
+#include <span>
+#include <vector>
+
+#include "iceberg/result.h"
+
+/// \file iceberg/util/endian.h
+/// \brief Endianness conversion utilities
+
+namespace iceberg::util {
```
wgtmac:
```suggestion
namespace iceberg {
```

src/iceberg/expression/literal.h
```diff
@@ -144,8 +144,9 @@ class ICEBERG_EXPORT Literal {
   Literal(Value value, std::shared_ptr<PrimitiveType> type);
 
   friend class LiteralCaster;
+  friend class LiteralSerializer;
```
```suggestion
```

src/iceberg/expression/literal.cc
```diff
@@ -21,8 +21,12 @@
 
 #include <cmath>
 #include <concepts>
+#include <cstring>
+#include <utility>
```
```suggestion
```

src/iceberg/expression/literal.cc
```diff
@@ -151,11 +155,11 @@ Literal Literal::Binary(std::vector<uint8_t> value) {
 
 Result<Literal> Literal::Deserialize(std::span<const uint8_t> data,
                                      std::shared_ptr<PrimitiveType> type) {
-  return NotImplemented("Deserialization of Literal is not implemented yet");
+  return Conversions::FromBytes(type, data);
```
```suggestion
  return Conversions::FromBytes(std::move(type), data);
```

src/iceberg/util/conversions.h
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <vector>
+
+#include "iceberg/expression/literal.h"
+#include "iceberg/iceberg_export.h"
```
wgtmac:
```suggestion
```

src/iceberg/util/conversions.h
```diff
@@ -0,0 +1,43 @@
...
+#include "iceberg/result.h"
+#include "iceberg/type.h"
```
wgtmac:
```suggestion
#include "iceberg/type_fwd.h"
```

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,93 @@
...
+
+template <typename T>
+concept LittleEndianWritable =
+    std::is_same_v<T, int32_t> || std::is_same_v<T, int64_t> ||
+    std::is_same_v<T, float> || std::is_same_v<T, double> || std::is_same_v<T, uint8_t>;
```
wgtmac:
```suggestion
concept LittleEndianWritable = std::is_arithmetic_v<T>;
```

Is this enough?

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,93 @@
...
+  if constexpr (std::endian::native != std::endian::little && sizeof(T) > 1) {
+    return std::byteswap(value);
+  }
+  return value;
+}
+
+/// \brief Convert a value from little-endian format.
+template <LittleEndianWritable T>
+T FromLittleEndian(T value) {
+  return ToLittleEndian(value);
```
wgtmac: This looks weird. I'd rather copy the above function.

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/// \brief Write a value in little-endian format to the buffer.
+template <LittleEndianWritable T>
+void WriteLittleEndian(std::vector<uint8_t>& buffer, T value) {
+  T le_value = ToLittleEndian(value);
+  const auto* bytes = reinterpret_cast<const uint8_t*>(&le_value);
```
wgtmac:
```suggestion
  value = ToLittleEndian(value);
  const auto* bytes = reinterpret_cast<const uint8_t*>(&value);
```

src/iceberg/util/literal_format.cc
```diff
@@ -0,0 +1,76 @@
...
+
+#include "iceberg/util/literal_format.h"
+
+#include <chrono>
+#include <cstring>
+#include <iomanip>
+
+namespace iceberg {
+
+std::string FormatDate(int32_t days_since_epoch) {
```
wgtmac: Could you please leverage std::formatter https://en.cppreference.com/w/cpp/chrono/hh_mm_ss/formatter.html instead of manual conversions? There are a bunch of specialization for `std::formatter<std::chrono::xyz>`.

src/iceberg/util/literal_format.cc
```diff
@@ -0,0 +1,76 @@
...
+  auto micros = microseconds_since_midnight % 1000000LL;
+
+  std::ostringstream oss;
+  oss << std::setfill('0') << std::setw(2) << hours << ":" << std::setfill('0')
+      << std::setw(2) << minutes << ":" << std::setfill('0') << std::setw(2) << seconds
+      << "." << std::setfill('0') << std::setw(6) << micros;
+  return oss.str();
+}
+
+std::string FormatTimestamp(int64_t microseconds_since_epoch) {
```
wgtmac: Please note that `timestamp` and `timestamptz` are different. `timestamp` should print the text in local timezone while `timestamptz` should use UTC.

src/iceberg/util/literal_format.cc
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/util/literal_format.h"
```
wgtmac:
```suggestion
#include "iceberg/util/date_time_util.h"
```

I think we should eventually add all functions from https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/util/DateTimeUtil.java. So let's rename it first. We need also move all time transformation to this utility:

https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/transform_function.cc#L204-L405

wgtmac: It would be better if we can split it (i.e. adding `date_time_util.h/.cc` and do the suggested refactoring from transfrom_function.cc) from this PR and check in that first.

src/iceberg/util/endian.h
wgtmac: This isn't resolved, right?

wgtmac (changes requested):

src/iceberg/CMakeLists.txt
```diff
@@ -50,7 +50,8 @@ set(ICEBERG_SOURCES
     arrow_c_data_guard_internal.cc
     util/murmurhash3_internal.cc
     util/timepoint.cc
-    util/gzip_internal.cc)
+    util/gzip_internal.cc
+    util/conversions.cc)
```
wgtmac: Please sort them alphabetically.

src/iceberg/util/conversions.h
```diff
@@ -0,0 +1,43 @@
...
+#pragma once
+
+#include <span>
+#include <vector>
+
+#include "iceberg/expression/literal.h"
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
```
wgtmac, line 32:
```suggestion
namespace iceberg {

```

src/iceberg/util/conversions.h
```diff
@@ -0,0 +1,43 @@
...
+                                              const Literal::Value& value);
+
+  static Result<std::vector<uint8_t>> ToBytes(const Literal& literal);
+
+  static Result<Literal::Value> FromBytes(const PrimitiveType& type,
+                                          std::span<const uint8_t> data);
+
+  static Result<Literal> FromBytes(std::shared_ptr<PrimitiveType> type,
+                                   std::span<const uint8_t> data);
+};
```
wgtmac, line 63:
```suggestion
};

```

src/iceberg/util/conversions.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include <span>
+#include <vector>
+
+#include "iceberg/expression/literal.h"
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+class ICEBERG_EXPORT Conversions {
```
wgtmac, line 35: Please add comment to public classes and functions.

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,296 @@
...
+#include <string>
+
+#include "iceberg/util/endian.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+/// \brief Write a value in little-endian format to the buffer.
+template <EndianConvertible T>
+void WriteLittleEndian(std::vector<uint8_t>& buffer, T value) {
```
wgtmac:
```suggestion
void WriteLittleEndian(T value, std::vector<uint8_t>* buffer) {
```

Usually the output parameter is placed at last and use pointer instead of reference type.

wgtmac: BTW, why not `std::vector<uint8_t> WriteLittleEndian(T value)`?

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,296 @@
...
+}
+
+Result<std::vector<uint8_t>> Conversions::ToBytes(const PrimitiveType& type,
+                                                  const Literal::Value& value) {
+  std::vector<uint8_t> result;
+  const auto type_id = type.type_id();
+
+  switch (type_id) {
+    case TypeId::kBoolean: {
+      // 0x00 for false, 0x01 for true
```
wgtmac: Is it too verbose to add these comments?

wgtmac: Can we simply do the following?

```cpp

// In literal.h or a new literal_traits.h

template <TypeId type_id>
struct LiteralTraits {
  using ValueType = void;
};

template <>
struct LiteralTraits<TypeId::kBoolean> {
  using ValueType = bool;
};

template <>
struct LiteralTraits<TypeId::kInt> {
  using ValueType = int32_t;
};

template <>
struct LiteralTraits<TypeId::kLong> {
  using ValueType = int64_t;
};

// ...

// In this file

template <TypeId type_id>
Result<std::vector<uint8_t>> ToBytes(const Literal::Value& value) {
  return WriteLittleEndian(std::get<LiteralTraits<type_id>>(value));
}

#define DISPATCH_LITERAL_TO_BYTES(type_id) \
  case type_id: \
    return ToBytes<type_id>(value);

Result<std::vector<uint8_t>> ToBytes(TypeId type_id, const Literal::Value& value) {
  switch (type_id) {
    DISPATCH_LITERAL_TO_BYTES(TypeId::kBoolean)
    DISPATCH_LITERAL_TO_BYTES(TypeId::kInt)
    DISPATCH_LITERAL_TO_BYTES(TypeId::kLong)
    DISPATCH_LITERAL_TO_BYTES(TypeId::kFloat)
    DISPATCH_LITERAL_TO_BYTES(TypeId::kDouble)
    DISPATCH_LITERAL_TO_BYTES(TypeId::kString)
    DISPATCH_LITERAL_TO_BYTES(TypeId::kBinary)
    // ...
  }
}

#undef DISPATCH_LITERAL_TO_BYTES
```

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,296 @@
...
+  if (literal.IsNull()) {
+    return NotSupported("Cannot serialize null");
+  }
+
+  return ToBytes(*literal.type(), literal.value());
+}
+
+Result<Literal::Value> Conversions::FromBytes(const PrimitiveType& type,
+                                              std::span<const uint8_t> data) {
+  // Empty data represents null value
```
wgtmac: Are you sure about this? Should we return error instead?

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,296 @@
...
+    case TypeId::kLong:
+    case TypeId::kTime:
+    case TypeId::kTimestamp:
+    case TypeId::kTimestampTz: {
+      int64_t value;
+      if (data.size() == 8) {
+        // Standard 8-byte long
+        ICEBERG_ASSIGN_OR_RAISE(auto long_value, ReadLittleEndian<int64_t>(data));
+        value = long_value;
+      } else if (data.size() == 4) {
```
wgtmac: Isn't this branch applied to `case TypeId::kLong` only?

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,296 @@
...
+        // Type was promoted from int to long
+        ICEBERG_ASSIGN_OR_RAISE(auto int_value, ReadLittleEndian<int32_t>(data));
+        value = static_cast<int64_t>(int_value);
+      } else {
+        auto type_name_view = ToString(type_id);
+        std::string type_name{type_name_view};
+        if (!type_name.empty()) {
+          type_name[0] = static_cast<char>(std::toupper(type_name[0]));
+        }
+        return InvalidArgument("{} requires 4 or 8 bytes, got {}", type_name,
```
wgtmac: Shouldn't we use `ToString(type_id)` in the error message directly?

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto value, FromBytes(*type, data));
+
+  // If we got a null value (monostate), create a null Literal
+  if (std::holds_alternative<std::monostate>(value)) {
+    return Literal::Null(type);
+  }
+
+  return Literal(value, type);
```
wgtmac: Use `std::move`?

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,296 @@
...
+                                              std::span<const uint8_t> data) {
+  // Empty data represents null value
+  if (data.empty()) {
+    return Literal::Value{std::monostate{}};
+  }
+
+  const auto type_id = type.type_id();
+
+  switch (type_id) {
+    case TypeId::kBoolean: {
```
wgtmac, line 131: Can we use template FromBytes like below to make the switch case more compact?

```
template <TypeId type_id>
Result<Literal::Value> FromBytes(std::span<const uint8_t> data)
```

src/iceberg/expression/literal.h
```diff
@@ -144,8 +144,8 @@ class ICEBERG_EXPORT Literal {
   Literal(Value value, std::shared_ptr<PrimitiveType> type);
 
   friend class LiteralCaster;
+  friend class Conversions;
```
wgtmac, line 148: Sort them alphabetically?

src/iceberg/expression/literal.cc
```diff
@@ -151,11 +152,11 @@ Literal Literal::Binary(std::vector<uint8_t> value) {
 
 Result<Literal> Literal::Deserialize(std::span<const uint8_t> data,
                                      std::shared_ptr<PrimitiveType> type) {
-  return NotImplemented("Deserialization of Literal is not implemented yet");
+  return Conversions::FromBytes(std::move(type), data);
```
wgtmac, line 161: Do we actually need to add `std::move` on `std::span`?

test/literal_test.cc
```diff
@@ -383,4 +383,142 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+
+  // Deserialize again to verify
+  auto final_literal = Literal::Deserialize(*bytes_result, type);
+  ASSERT_TRUE(final_literal.has_value());
+  EXPECT_EQ(final_literal->type()->type_id(), expected_literal.type()->type_id());
+  EXPECT_EQ(final_literal->ToString(), expected_literal.ToString());
+}
+
+// binary serialization tests
+TEST(LiteralSerializationTest, BinaryBoolean) {
```
wgtmac: Can you please use parameterized test pattern to reuse a single test case for different types?

test/literal_test.cc
```diff
@@ -383,4 +383,142 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+  EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+  EXPECT_EQ(double_result->ToString(), "1.000000");
+
+  auto double_bytes = double_result->Serialize();
+  ASSERT_TRUE(double_bytes.has_value());
+  EXPECT_EQ(double_bytes->size(), 8);
+}
+
+// Edge case serialization tests
+TEST(LiteralSerializationTest, EdgeCases) {
```
wgtmac: This can also use parameterized test, right?

test/literal_test.cc
```diff
@@ -383,4 +383,142 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+}
+
+// Error case serialization tests
+TEST(LiteralSerializationTest, ErrorCases) {
+  // AboveMax/BelowMin values cannot be serialized
+  auto long_literal =
+      Literal::Long(static_cast<int64_t>(std::numeric_limits<int32_t>::max()) + 1);
+  auto above_max_result = long_literal.CastTo(int32());
+  ASSERT_TRUE(above_max_result.has_value());
+  EXPECT_TRUE(above_max_result->IsAboveMax());
```
wgtmac: Can we directly create AboveMax?

test/literal_test.cc
```diff
@@ -383,4 +383,142 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+
+  auto serialize_result = above_max_result->Serialize();
+  EXPECT_FALSE(serialize_result.has_value());
```
wgtmac: Use ASSERT_THAT to verify the error message. Same apply for below.

test/manifest_list_reader_test.cc
```diff
@@ -20,9 +20,10 @@
 #include <arrow/filesystem/localfs.h>
 #include <avro/GenericDatum.hh>
 #include <gtest/gtest.h>
+#include <iceberg/avro/avro_register.h>
```
wgtmac:
```suggestion
```

src/iceberg/util/conversions.cc
HeartLinked: > BTW, why not `std::vector<uint8_t> WriteLittleEndian(T value)`?

I think the current `void` signature treats the `std::vector<uint8_t>&` buffer as a write target (a sink), allowing data to be appended multiple times efficiently. In contrast, returning a vector would create a new std::vector object on the heap with each call, making the existing approach superior?

src/iceberg/util/conversions.cc
HeartLinked: no. All these 4 types should be int64_t.   `using Value = std::variant<std::monostate, bool,int32_t, int64_t,         // for long, timestamp, timestamp_tz, time`

src/iceberg/util/endian.h
```diff
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include <algorithm>
+#include <array>
```
wgtmac:
```suggestion
```

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+  const auto* bytes = reinterpret_cast<const uint8_t*>(&value);
+  std::vector<uint8_t> result;
+  result.insert(result.end(), bytes, bytes + sizeof(T));
+  return result;
+}
+
+/// \brief Read a value in little-endian format from the data.
+template <EndianConvertible T>
+Result<T> ReadLittleEndian(std::span<const uint8_t> data) {
+  if (data.size() < sizeof(T)) [[unlikely]] {
```
wgtmac:
```suggestion
  if (data.size() == sizeof(T)) [[unlikely]] {
```

Shouldn't they be strictly equal?

test/literal_test.cc
```diff
@@ -383,4 +383,118 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+                              Literal::Float(std::numeric_limits<float>::quiet_NaN()),
+                              float32()}
+        // TODO(Li Feiyang): Add tests for Date, Time, Timestamp, TimestampTz
+        ),
+
+    [](const testing::TestParamInfo<LiteralSerializationParamTest::ParamType>& info) {
+      return info.param.test_name;
+    });
+
+TEST(LiteralSerializationEdgeCaseTest, EmptyStringSerialization) {
```
wgtmac:
```suggestion
TEST(LiteralSerializationTest, EmptyString) {
```

test/literal_test.cc
```diff
@@ -383,4 +383,118 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
 }
 
+// Type promotion tests
+TEST(LiteralSerializationTest, TypePromotion) {
+  // 4-byte int data can be deserialized as long
+  std::vector<uint8_t> int_data = {32, 0, 0, 0};
+  auto long_result = Literal::Deserialize(int_data, int64());
+  ASSERT_TRUE(long_result.has_value());
+  EXPECT_EQ(long_result->type()->type_id(), TypeId::kLong);
+  EXPECT_EQ(long_result->ToString(), "32");
```
wgtmac: I think it is not a good practice to have an indirection like this.

```suggestion
  EXPECT_EQ(std::get<int64_t>(long_result->value()), 32L);
```

test/literal_test.cc
```diff
@@ -383,4 +383,118 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+  const auto& param = GetParam();
+
+  // Deserialize from bytes
+  Result<Literal> literal_result = Literal::Deserialize(param.input_bytes, param.type);
+  ASSERT_TRUE(literal_result.has_value())
+      << "Deserialization failed: " << literal_result.error().message;
+
+  // Check type and value
+  EXPECT_EQ(literal_result->type()->type_id(), param.expected_literal.type()->type_id());
+  EXPECT_EQ(literal_result->ToString(), param.expected_literal.ToString());
```
wgtmac:
```suggestion
  EXPECT_EQ(*literal_result, param.expected_literal);
```

test/literal_test.cc
```diff
@@ -383,4 +383,118 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+  ASSERT_TRUE(bytes_result.has_value())
+      << "Serialization failed: " << bytes_result.error().message;
+  EXPECT_EQ(*bytes_result, param.input_bytes);
+
+  // Deserialize again to verify idempotency
+  Result<Literal> final_literal = Literal::Deserialize(*bytes_result, param.type);
+  ASSERT_TRUE(final_literal.has_value())
+      << "Final deserialization failed: " << final_literal.error().message;
+  EXPECT_EQ(final_literal->type()->type_id(), param.expected_literal.type()->type_id());
+  EXPECT_EQ(final_literal->ToString(), param.expected_literal.ToString());
```
wgtmac:
```suggestion
  EXPECT_EQ(*final_literal, param.expected_literal);
```

test/literal_test.cc
```diff
@@ -383,4 +383,118 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    BinarySerializationTests, LiteralSerializationParamTest,
```
wgtmac: Can we eliminate `Tests` and `Test` from their names to be shorter?

test/literal_test.cc
```diff
@@ -383,4 +383,118 @@ TEST(LiteralTest, DoubleZeroComparison) {
   EXPECT_EQ(neg_zero <=> pos_zero, std::partial_ordering::less);
 }
 
+// Type promotion tests
+TEST(LiteralSerializationTest, TypePromotion) {
```
wgtmac: This case is better to be placed after the roundtrip test.

src/iceberg/util/conversions.h
```diff
@@ -0,0 +1,51 @@
...
+namespace iceberg {
+
+/// \brief Conversion utilities for primitive types
+class ICEBERG_EXPORT Conversions {
+ public:
+  /// \brief Convert a literal value to bytes
+  static Result<std::vector<uint8_t>> ToBytes(const PrimitiveType& type,
+                                              const Literal::Value& value);
+
+  static Result<std::vector<uint8_t>> ToBytes(const Literal& literal);
```
wgtmac, line 47: Please add docstring.

src/iceberg/util/conversions.h
```diff
@@ -0,0 +1,51 @@
...
+
+  /// \brief Convert bytes to a literal value
+  static Result<Literal::Value> FromBytes(const PrimitiveType& type,
+                                          std::span<const uint8_t> data);
+
+  static Result<Literal> FromBytes(std::shared_ptr<PrimitiveType> type,
```
wgtmac, line 61: ditto

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kInt)
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kDate)
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kLong)
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kTime)
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kTimestamp)
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kTimestampTz)
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kFloat)
+    DISPATCH_LITERAL_TO_BYTES(TypeId::kDouble)
+    case TypeId::kBoolean: {
+      return std::vector<uint8_t>{std::get<bool>(value) ? static_cast<uint8_t>(0x01)
```
wgtmac: Can we add a template specialization of ToBytesImpl for TypeId::kBoolean? Then you can simply write `DISPATCH_LITERAL_TO_BYTES(TypeId::kBoolean)`

wgtmac: I believe we can also specialize all types below.

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+      const auto& str = std::get<std::string>(value);
+      return std::vector<uint8_t>(str.begin(), str.end());
+    }
+
+    case TypeId::kBinary: {
+      return std::get<std::vector<uint8_t>>(value);
+    }
+
+    case TypeId::kFixed: {
+      if (std::holds_alternative<std::array<uint8_t, 16>>(value)) {
```
wgtmac: I think we can blindly use `std::vector<uint8_t>` for fixed type.

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+    return NotSupported("Cannot serialize null");
+  }
+
+  return ToBytes(*literal.type(), literal.value());
+}
+
+Result<Literal::Value> Conversions::FromBytes(const PrimitiveType& type,
+                                              std::span<const uint8_t> data) {
+  if (data.empty()) {
+    return InvalidArgument("Data cannot be empty");
```
wgtmac:
```suggestion
    return InvalidArgument("Cannot deserialize empty value");
```

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+Result<Literal> Conversions::FromBytes(std::shared_ptr<PrimitiveType> type,
+                                       std::span<const uint8_t> data) {
+  if (!type) {
+    return InvalidArgument("Type cannot be null");
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto value, FromBytes(*type, data));
+
+  // If we got a null value (monostate), create a null Literal
+  if (std::holds_alternative<std::monostate>(value)) {
```
wgtmac: Will it ever happen?

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+    return InvalidArgument("Data cannot be empty");
+  }
+
+  const auto type_id = type.type_id();
+
+  switch (type_id) {
+    case TypeId::kBoolean: {
+      if (data.size() != 1) {
+        return InvalidArgument("Boolean requires 1 byte, got {}", data.size());
+      }
```
wgtmac:
```suggestion
```

We have checked it in ReadLittleEndian, so no need to check it again here.

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+      ICEBERG_ASSIGN_OR_RAISE(auto value, ReadLittleEndian<uint8_t>(data));
+      return Literal::Value{static_cast<bool>(value != 0x00)};
+    }
+
+    case TypeId::kInt: {
+      if (data.size() != sizeof(int32_t)) {
```
wgtmac: ditto

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+      } else {
+        return InvalidArgument("{} requires 4 or 8 bytes, got {}", ToString(type_id),
+                               data.size());
+      }
+
+      return Literal::Value{value};
+    }
+
+    case TypeId::kFloat: {
+      if (data.size() != sizeof(float)) {
```
wgtmac: ditto

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+      if (data.size() == 8) {
+        ICEBERG_ASSIGN_OR_RAISE(auto double_value, ReadLittleEndian<double>(data));
+        return Literal::Value{double_value};
+      } else if (data.size() == 4) {
+        // Type was promoted from float to double
+        ICEBERG_ASSIGN_OR_RAISE(auto float_value, ReadLittleEndian<float>(data));
+        return Literal::Value{static_cast<double>(float_value)};
+      } else {
+        return InvalidArgument("Double requires 4 or 8 bytes, got {}", data.size());
+      }
```
wgtmac:
```suggestion
      if (data.size() < 8) {
        // Type was promoted from float to double
        ICEBERG_ASSIGN_OR_RAISE(auto float_value, ReadLittleEndian<float>(data));
        return Literal::Value{static_cast<double>(float_value)};
      } else {
        ICEBERG_ASSIGN_OR_RAISE(auto double_value, ReadLittleEndian<double>(data));
        return Literal::Value{double_value};
      }
```

Let's simplify this

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+      } else if (data.size() == 4) {
+        // Type was promoted from int to long
+        ICEBERG_ASSIGN_OR_RAISE(auto int_value, ReadLittleEndian<int32_t>(data));
+        value = static_cast<int64_t>(int_value);
+      } else {
+        return InvalidArgument("{} requires 4 or 8 bytes, got {}", ToString(type_id),
+                               data.size());
+      }
+
+      return Literal::Value{value};
```
wgtmac, line 156:
```suggestion
      int64_t value;
      if (data.size() < 8) {
        // Type was promoted from int to long
        ICEBERG_ASSIGN_OR_RAISE(auto int_value, ReadLittleEndian<int32_t>(data));
        value = static_cast<int64_t>(int_value);
      } else {
        ICEBERG_ASSIGN_OR_RAISE(auto long_value, ReadLittleEndian<int64_t>(data));
        value = long_value;
      }
      return Literal::Value{value};
```

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+          std::string(reinterpret_cast<const char*>(data.data()), data.size())};
+    }
+
+    case TypeId::kBinary: {
+      return Literal::Value{std::vector<uint8_t>(data.begin(), data.end())};
+    }
+
+    case TypeId::kFixed: {
+      if (data.size() == 16) {
+        std::array<uint8_t, 16> fixed_bytes;
```
wgtmac: Please do not use std::array<uint8_t, 16> for fixed type, this makes cases complex.

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,248 @@
...
+        std::ranges::copy(data, fixed_bytes.begin());
+        return Literal::Value{fixed_bytes};
+      } else {
+        return Literal::Value{std::vector<uint8_t>(data.begin(), data.end())};
```
wgtmac: Please check data.size() against type.

src/iceberg/expression/literal.cc
zhjwpku: Why wasn't the suggested code adopted? I share the same idea ;)

zhjwpku (approved):
LGTM

src/iceberg/expression/literal.cc
HeartLinked: > Why wasn't the suggested code adopted? I share the same idea ;)

You're right. There have been quite a few comments in the review, so I missed this one. Thanks.

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,226 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/util/conversions.h"
+
+#include <array>
```
HuaHuaY:
```suggestion
```

src/iceberg/expression/literal.cc
```diff
@@ -149,13 +150,17 @@ Literal Literal::Binary(std::vector<uint8_t> value) {
   return {Value{std::move(value)}, binary()};
 }
 
+Literal Literal::Fixed(std::vector<uint8_t> value, int32_t length) {
+  return {Value{std::move(value)}, fixed(length)};
```
wgtmac, line 156: nit: add `ICEBERG_DCHECK(value.size() == length)`

src/iceberg/expression/literal.cc
```diff
@@ -249,6 +255,20 @@ std::partial_ordering Literal::operator<=>(const Literal& other) const {
...
     }
 
+    case TypeId::kFixed: {
+      // Fixed types can only be compared if they have the same length
+      auto& this_fixed_type = static_cast<const FixedType&>(*type_);
+      auto& other_fixed_type = static_cast<const FixedType&>(*other.type_);
+
+      if (this_fixed_type.length() != other_fixed_type.length()) {
+        return std::partial_ordering::unordered;
+      }
```
wgtmac: I think we can remove line 259 to line 256 if we modify line 197 to 

```
  // If types are different, comparison is unordered
  if (*type_ != *other.type_) {
    return std::partial_ordering::unordered;
  }
```

src/iceberg/expression/literal.h
```diff
@@ -71,6 +71,7 @@ class ICEBERG_EXPORT Literal {
   static Literal Double(double value);
   static Literal String(std::string value);
   static Literal Binary(std::vector<uint8_t> value);
+  static Literal Fixed(std::vector<uint8_t> value, int32_t length);
```
```suggestion
  static Literal Fixed(std::vector<uint8_t> value);
```

I think a single vector is enough and the length can be deduced from value.size(). Otherwise, it is annoying when `length != value.size()`, should we throw then?

src/iceberg/expression/literal.cc
```diff
@@ -249,6 +255,20 @@ std::partial_ordering Literal::operator<=>(const Literal& other) const {
...
+
```
```suggestion
```

I'd suggest to remove these lines and change the line 196 above as below:

```cpp
  // If types are different, comparison is unordered
  if (*type_ != *other.type_) {
    return std::partial_ordering::unordered;
  }
```

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,226 @@
...
+template <>
+Result<std::vector<uint8_t>> ToBytesImpl<TypeId::kFixed>(const Literal::Value& value) {
+  if (std::holds_alternative<std::vector<uint8_t>>(value)) {
+    return std::get<std::vector<uint8_t>>(value);
+  } else {
+    std::string actual_type =
+        std::visit([](auto&& arg) -> std::string { return typeid(arg).name(); }, value);
+    return InvalidArgument("Invalid value type for Fixed literal, got type: {}",
+                           actual_type);
+  }
```
wgtmac:
```suggestion
  return std::get<std::vector<uint8_t>>(value);
```

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,226 @@
...
+
+  const auto type_id = type.type_id();
+
+  switch (type_id) {
+    case TypeId::kBoolean: {
+      ICEBERG_ASSIGN_OR_RAISE(auto value, ReadLittleEndian<uint8_t>(data));
+      return Literal::Value{static_cast<bool>(value != 0x00)};
+    }
+
```
wgtmac: Generally I don't think it is a good practice to add blank lines between cases which makes the switch block lengthy.

src/iceberg/util/conversions.cc
```diff
@@ -0,0 +1,226 @@
...
+      } else {
+        ICEBERG_ASSIGN_OR_RAISE(auto double_value, ReadLittleEndian<double>(data));
+        return Literal::Value{double_value};
+      }
+    }
+
+    case TypeId::kString: {
+      return Literal::Value{
+          std::string(reinterpret_cast<const char*>(data.data()), data.size())};
+    }
```
wgtmac:
```suggestion
    case TypeId::kString:
      return Literal::Value{
          std::string(reinterpret_cast<const char*>(data.data()), data.size())};
```

Usually we don't add curly braces to a single-line case.

test/literal_test.cc
```diff
@@ -136,8 +136,8 @@ TEST(LiteralTest, LongCastTo) {
   EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
 }
 
+// Test overflow cases
```
wgtmac:
```suggestion
```

The comment is redundant.

test/literal_test.cc
```diff
@@ -383,4 +383,211 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+    });
+
+TEST(LiteralSerializationTest, EmptyString) {
+  auto empty_string = Literal::String("");
+  auto empty_bytes = empty_string.Serialize();
+  ASSERT_TRUE(empty_bytes.has_value());
+  EXPECT_TRUE(empty_bytes->empty());
+
+  auto deserialize_result = Literal::Deserialize(*empty_bytes, string());
+  EXPECT_THAT(deserialize_result, IsError(ErrorKind::kInvalidArgument));
```
wgtmac: Could you double check the spec and java impl on how to deal with empty values? I think this is a valid case which shouldn't return an error.

test/literal_test.cc
```diff
@@ -383,4 +383,211 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+  // 4-byte int data can be deserialized as long
+  std::vector<uint8_t> int_data = {32, 0, 0, 0};
+  auto long_result = Literal::Deserialize(int_data, int64());
+  ASSERT_TRUE(long_result.has_value());
+  EXPECT_EQ(long_result->type()->type_id(), TypeId::kLong);
+  EXPECT_EQ(std::get<int64_t>(long_result->value()), 32L);
+
+  auto long_bytes = long_result->Serialize();
+  ASSERT_TRUE(long_bytes.has_value());
+  EXPECT_EQ(long_bytes->size(), 8);
```
wgtmac:
```suggestion
```

These lines are meaningless to this case.

test/literal_test.cc
```diff
@@ -383,4 +383,211 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+  // 4-byte float data can be deserialized as double
+  std::vector<uint8_t> float_data = {0, 0, 128, 63};
+  auto double_result = Literal::Deserialize(float_data, float64());
+  ASSERT_TRUE(double_result.has_value());
+  EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+  EXPECT_EQ(std::get<double>(double_result->value()), 1.0);
+
+  auto double_bytes = double_result->Serialize();
+  ASSERT_TRUE(double_bytes.has_value());
+  EXPECT_EQ(double_bytes->size(), 8);
```
wgtmac:
```suggestion
```

ditto

test/literal_test.cc
```diff
@@ -383,4 +383,211 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+  auto long_bytes = long_result->Serialize();
+  ASSERT_TRUE(long_bytes.has_value());
+  EXPECT_EQ(long_bytes->size(), 8);
+
+  // 4-byte float data can be deserialized as double
+  std::vector<uint8_t> float_data = {0, 0, 128, 63};
+  auto double_result = Literal::Deserialize(float_data, float64());
+  ASSERT_TRUE(double_result.has_value());
+  EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+  EXPECT_EQ(std::get<double>(double_result->value()), 1.0);
```
wgtmac:
```suggestion
  EXPECT_DOUBLE_EQ(std::get<double>(double_result->value()), 1.0);
```

src/iceberg/expression/literal.h
HeartLinked: This comment seems contradictions with the above？So should we use 
```
Literal Literal::Fixed(std::vector<uint8_t> value) {
  auto length = static_cast<int32_t>(value.size());
  return {Value{std::move(value)}, fixed(length)};
}
``` 
to implement it, not use ICEBERG_DCHECK?

HeartLinked:
@wgtmac I've handled these comments.

src/iceberg/expression/literal.h
wgtmac: Exactly

src/iceberg/expression/literal.h
```diff
@@ -143,11 +144,76 @@ class ICEBERG_EXPORT Literal {
  private:
   Literal(Value value, std::shared_ptr<PrimitiveType> type);
 
+  friend class Conversions;
   friend class LiteralCaster;
 
- private:
```
wgtmac, line 149: Why this gets removed?

HeartLinked: There is another one above. I guess this is repeated.

test/literal_test.cc
```diff
@@ -383,4 +382,214 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
 }
 
+struct LiteralRoundTripParam {
+  std::string test_name;
+  std::vector<uint8_t> input_bytes;
+  Literal expected_literal;
+  std::shared_ptr<PrimitiveType> type;
+};
+
+class LiteralSerializationParam : public ::testing::TestWithParam<LiteralRoundTripParam> {
```
wgtmac:
```suggestion
class LiteralSerDeTest : public ::testing::TestWithParam<LiteralParam> {
```

test/literal_test.cc
```diff
@@ -383,4 +382,214 @@ TEST(LiteralTest, DoubleZeroComparison) {
   EXPECT_EQ(neg_zero <=> pos_zero, std::partial_ordering::less);
 }
 
+struct LiteralRoundTripParam {
+  std::string test_name;
+  std::vector<uint8_t> input_bytes;
+  Literal expected_literal;
+  std::shared_ptr<PrimitiveType> type;
```
wgtmac:
```suggestion
struct LiteralParam {
  std::string test_name;
  std::vector<uint8_t> serialized;
  Literal value;
  std::shared_ptr<PrimitiveType> type;
```

test/literal_test.cc
```diff
@@ -383,4 +382,214 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+        LiteralRoundTripParam{"TimestampTzOneHour",
+                              {0, 164, 147, 214, 0, 0, 0, 0},
+                              Literal::TimestampTz(3600000000),
+                              timestamp_tz()}),
+
+    [](const testing::TestParamInfo<LiteralSerializationParam::ParamType>& info) {
+      return info.param.test_name;
+    });
+
+TEST(LiteralSerializationTest, EmptyString) {
```
wgtmac: Shouldn't this be a case of the above test? For example:

```cpp
        LiteralRoundTripParam{"EmptyString",
                              {},
                              Literal::String({""}),
                              string()},
```

test/literal_test.cc
```diff
@@ -383,4 +382,214 @@ TEST(LiteralTest, DoubleZeroComparison) {
...
+  auto empty_bytes = empty_string.Serialize();
+  ASSERT_TRUE(empty_bytes.has_value());
+  EXPECT_TRUE(empty_bytes->empty());
+
+  auto deserialize_result = Literal::Deserialize(*empty_bytes, string());
+  ASSERT_THAT(deserialize_result, IsOk());
+  EXPECT_TRUE(std::get<std::string>(deserialize_result->value()).empty());
+}
+
+TEST(LiteralSerializationTest, EmptyBinary) {
```
wgtmac: ditto

PR #186, opened 2025-08-20
chore: use more from_chars

zhjwpku:


src/iceberg/schema_internal.cc
```diff
@@ -192,7 +193,11 @@ int32_t GetFieldId(const ArrowSchema& schema) {
     return kUnknownFieldId;
   }
 
-  return std::stoi(std::string(field_id_value.data, field_id_value.size_bytes));
+  int32_t field_id = kUnknownFieldId;
+  std::from_chars(field_id_value.data, field_id_value.data + field_id_value.size_bytes,
```
zeroshade, line 197: should this check the ec return also?

zhjwpku: If ec is std::errc(), the field_id is still kUnknownFieldId, this check is unnecessary.

Issue #188, opened 2025-08-21
feat: Do you plan to make the library available in e.g. Conan Center for easier access?


ormandi:
Context:
- We rely on large scale monorepo with Bazel as a build system.
- We have a way to bring in Conan Center third party libs easily.
- Without this we need to build the library from source in the context of Bazel using something like [`rules_foreign_cc`](https://github.com/bazel-contrib/rules_foreign_cc) which is usually doable but requires significantly more effort compare to a Conan import.

Why:
- Would help accelerate adoption of the library.

Curious to see your thoughts. Thank you!

ormandi:
Awesome, thank you!

PR #193, opened 2025-08-26
chore: add release script and github workflow

HeartLinked:
This contribution introduces release automation tooling including:
- `/dev/release/README.md`: Release process documentation
- `/dev/release/*.sh`: Release script utilities
- `/dev/release/check_rat_report.py`: Check Apache lisence script for release
- `/dev/release/rat_exclude_files.txt`: List of files excluded from RAT license checks
- `.github/workflows/rc.yml`: GitHub Actions automation for release

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -74,21 +74,10 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-latest
-          - ubuntu-latest
+          - macos-15
+          - ubuntu-24.04
           # - windows-latest
     steps:
-      - name: Install C++ Build Tools (Ubuntu)
-        if: runner.os == 'Linux'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y build-essential cmake ninja-build libssl-dev
-
-      - name: Install C++ Build Tools (macOS)
-        if: runner.os == 'macOS'
-        run: |
-          brew reinstall cmake ninja openssl
-
       - name: Checkout
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -31,7 +31,7 @@ Run `dev/release/release_rc.sh` on a working copy of
 `git@github.com:apache/iceberg-cpp` not from your fork:
 
 ```console
-$ git clone git@github.com:apache/iceberg-cpp.git
+$ git clone git@github.com:apache/iceberg-cpp.git && cd iceberg-cpp
 $ dev/release/release_rc.sh ${VERSION} ${RC}
 (Send a vote email to dev@iceberg.apache.org.
  You can use a draft shown by release_rc.sh for the email.)


.github/workflows/rc.yml
```diff
@@ -0,0 +1,147 @@
...
+  contents: read
+
+jobs:
+  archive:
+    name: Archive
+    runs-on: ubuntu-latest
+    timeout-minutes: 5 # Can be reduced as we are not checking out submodules
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
zhjwpku:
```suggestion
        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```

same for the below.

.github/workflows/rc.yml
```diff
@@ -0,0 +1,147 @@
...
+  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
+  cancel-in-progress: true
+permissions:
+  contents: read
+
+jobs:
+  archive:
+    name: Archive
+    runs-on: ubuntu-latest
+    timeout-minutes: 5 # Can be reduced as we are not checking out submodules
```
zhjwpku: We don't checkout submodules, but we need to download and build some thirdparties, is 5 minutes enough?

zhjwpku: Ah, seems no need to run the build, sorry for the noise.

.github/workflows/rc.yml
HeartLinked: why not simply use v5?

wgtmac: Because versions are not immutable. It is recommended by ASF to use commit hash: https://github.com/apache/infrastructure-actions/blob/main/actions.yml

dev/release/README.md
```diff
@@ -0,0 +1,105 @@
...
+`git@github.com:apache/iceberg-cpp` not from your fork:
+
+```console
+$ git clone git@github.com:apache/iceberg-cpp.git
+$ dev/release/release_rc.sh ${VERSION} ${RC}
+(Send a vote email to dev@iceberg.apache.org.
+ You can use a draft shown by release_rc.sh for the email.)
+```
+
+Here is an example to release RC1:
```
wgtmac:
```suggestion
Here is an example to release RC0 of version 0.1.0:
```

dev/release/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+```console
+$ GH_TOKEN=${YOUR_GITHUB_TOKEN} dev/release/release_rc.sh 0.1.0 1
```
wgtmac:
```suggestion
$ GH_TOKEN=${YOUR_GITHUB_TOKEN} dev/release/release_rc.sh 0.1.0 0
```

.github/workflows/rc.yml
```diff
@@ -0,0 +1,147 @@
...
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+          echo "VERSION=${version}"
+          echo "RC=${rc}"
+
+      - name: Archive
+        run: |
+          id="apache-iceberg-cpp-${VERSION}-rc${RC}"
+          tar_gz="${id}.tar.gz"
+          echo "TAR_GZ=${tar_gz}" >> ${GITHUB_ENV}
+          # Reverted to the git archive command from the Go implementation
```
raulcd: is this comment correct?

dev/release/README.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+The arguments of `release_rc.sh` are the version and the RC number. If RC1 has a problem, we'll increment the RC number such as RC2, RC3 and so on.
```
wgtmac:
```suggestion
The arguments of `release_rc.sh` are the version and the RC number. If RC0 has a problem, we'll increment the RC number such as RC1, RC2 and so on.
```

.github/workflows/rc.yml
```diff
@@ -0,0 +1,147 @@
...
+      - archive
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          # Consider adding windows-latest if your project supports it
+          # - windows-latest
```
raulcd: Is the plan to add windows verification in the future? Maybe on future releases? We probably should update the comment and potentially open an issue to add Windows verification

dev/release/README.md
```diff
@@ -0,0 +1,105 @@
...
+$ cd iceberg
+$ editor KEYS
+$ svn ci KEYS
+```
+
+### Publish
+
+We need to do the following to publish a new release:
+
+    * Publish to apache.org
```
wgtmac:
```suggestion
We need to publish to apache.org for the new release:
```

.github/workflows/rc.yml
```diff
@@ -0,0 +1,147 @@
...
+          version=${version%%-rc*}
+          if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
+            rc=${GITHUB_REF_NAME##*-rc}
+          else
+            rc=100
+          fi
+          echo "VERSION=${version}"
+          echo "RC=${rc}"
+          # You MUST create this verify_rc.sh script for C++
+          # It should untar, cmake, build, and run tests (ctest)
```
raulcd:
```suggestion
          # The verify_rc.sh script will untar and 
          # run cmake, build, tests (ctest) and install
```

dev/release/README.md
raulcd: minor nit, is the plan to start with RC1 or 0? Potentially add the example for the first one (0 on case of Arrow)

dev/release/rat_exclude_files.txt
```diff
@@ -0,0 +1,32 @@
...
+.gitignore
+LICENSE
+NOTICE
+CONTRIBUTING.md
+CHANGELOG.md
+build/**
+dist/**
+.git/**
+cmake_modules/**
+.cmake/**
```
wgtmac:
```suggestion
```

dev/release/verify_rc.sh
```diff
@@ -0,0 +1,194 @@
...
+  fi
+  download_rc_file "${ARCHIVE_BASE_NAME}.tar.gz.sha512"
+  ${sha512_verify} "${ARCHIVE_BASE_NAME}.tar.gz.sha512"
+}
+
+ensure_source_directory() {
+  tar xf "${ARCHIVE_BASE_NAME}".tar.gz
+}
+
+check_compiler() {
```
wgtmac: We don't need this. CMake script already enforce this.

.github/workflows/rc.yml
```diff
@@ -0,0 +1,147 @@
...
+          dev/release/verify_rc.sh "${version}" "${rc}"
```
raulcd: I think this will fail as is.
Signatures are created via the `release_rc.sh` script once this workflow is finished but we are verifying signatures here, we probably should use `VERIFY_SIGN=0 dev/release/verify_rc.sh "${version}" "${rc}"` as the signatures are not yet created when this is executed.

Go suspiciously seems to just do `VERIFY_DEFAULT=0 dev/release/verify_rc.sh "${version}" "${rc}"` which means it won't even test. Probably we should validate the verification is being exercised on the Go repository or open an issue there :)

.github/workflows/rc.yml
HeartLinked: fixed.

dev/release/README.md
HeartLinked: fixed.

.github/workflows/rc.yml
HeartLinked: fixed.

dev/release/rat_exclude_files.txt
```diff
@@ -0,0 +1,32 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+.gitignore
+LICENSE
+NOTICE
+CONTRIBUTING.md
```
Fokko: I don't see this file, same for` CHANGELOG.md`. I don't think there is any issue with adding the ASv2 headers to the Markdown files, example: https://github.com/apache/iceberg-cpp/blob/main/README.md?plain=1

dev/release/rat_exclude_files.txt
```diff
@@ -0,0 +1,32 @@
...
+CHANGELOG.md
+build/**
+dist/**
+.git/**
+cmake_modules/**
+.cmake/**
+test/resources/**
+*.avro
+*.json
+*.parquet
```
Fokko, line 27: We want to avoid binary files in the release distribution.

.github/workflows/rc.yml
HeartLinked: Thanks for your detailed review and for catching that critical issue!

You're absolutely right, the verification would fail as the signature doesn't exist yet. I'll fix this by explicitly adding `VERIFY_SIGN=0` and `VERIFY_DOWNLOAD=0` to the command.

Regarding the Go script, I took a closer look. It seems that while `VERIFY_DEFAULT=0` is equal to `VERIFY_SIGN=0 & VERIFY_DOWNLOAD=0`, and the tests are still being executed. Thanks again for the great feedback! I'm pushing the fix now.

raulcd: Oh! you are correct about the go tests, in Arrow `VERIFY_DEFAULT` also drives `TEST_SOURCE`, `TEST_BINARY` etcetera but those do not exist on iceberg-go, tests are exercised unconditionally on that case. I talked from memory from the other cases.

dev/release/rat_exclude_files.txt
HeartLinked: Thanks for your detailed review! I'm pushing the fix now.

dev/release/rat_exclude_files.txt
HeartLinked: Thank you for your review! I still believe we need to include these binary files, especially as they are part of our tests and are used for unit testing.

I found that the release version of iceberg-rust also contains similar files (e.g., `iceberg-rust/crates/iceberg/testdata/manifests_lists/manifest-list-v2-1.avro`), so I think it is necessary to keep them here for now. This seems to be related to the current lack of support for relative paths in Apache Iceberg (see: [Google Docs link](https://docs.google.com/document/d/1a6tXvbWVbvOxiRexCiaIsIA6NOOqcbxPKjeoJGqgYtg/edit?pli=1&tab=t.p92mo7cvg08q)).In the future, we could potentially store these test files in a separate repository, similar to how apache/parquet-testing is managed.

.github/workflows/rc.yml
wgtmac: This comment is not resolved yet.

.github/workflows/rc.yml
```diff
@@ -0,0 +1,144 @@
...
+  contents: read
+
+jobs:
+  archive:
+    name: Archive
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
wgtmac:
```suggestion
        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0

File: .github/workflows/rc.yml [hunk 16]
```diff
@@ -0,0 +1,144 @@
...
+          sudo apt-get update
+          sudo apt-get install -y build-essential cmake ninja-build libssl-dev
+
+      - name: Install C++ Build Tools (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          brew install cmake ninja openssl
+
+      - name: Checkout
+        uses: actions/checkout@v4
```
wgtmac:
```suggestion
        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```

dev/release/rat_exclude_files.txt
wgtmac: @Fokko Do you still have concern about this? I think we need at least `.avro` and `.parquet` files before we have full writer implementation.

wgtmac:
nit: the PR title can be `chore: add release script and github workflow`

.github/workflows/rc.yml
HeartLinked: fixed

wgtmac:
@Fokko @raulcd @zeroshade Do you have more comment on this PR?

wgtmac:
@raulcd I'm not sure why my announcement email for parquet release always gets rejected by it :/

wgtmac (approved):

HeartLinked:
@raulcd I've added a draft email for announce@apache.org at `release.sh`. Thanks for your advice. In the future, maybe after we upload the release to platforms like Conan Center(#188), we can modify the script and add relevant links? @wgtmac CC

.github/workflows/rc.yml
```diff
@@ -0,0 +1,144 @@
...
+          echo "VERSION=${version}"
+          echo "RC=${rc}"
+
+      - name: Archive
+        run: |
+          id="apache-iceberg-cpp-${VERSION}-rc${RC}"
+          tar_gz="${id}.tar.gz"
+          echo "TAR_GZ=${tar_gz}" >> ${GITHUB_ENV}
+          git archive HEAD --prefix "${id}/" --output "${tar_gz}"
+          sha512sum "${tar_gz}" > "${tar_gz}.sha512"
```
Fokko: This works, but personally, I would change this into:

```
./dev/release/archive.sh "${RC}" "${ARCHIVE_FILE_NAME}"
```

.github/workflows/rc.yml
```diff
@@ -0,0 +1,144 @@
...
+          version=${version%-rc*}
+          rc=${GITHUB_REF_NAME##*-rc}
+          echo "VERSION=${version}" >> ${GITHUB_ENV}
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+          echo "VERSION=${version}"
+          echo "RC=${rc}"
+
+      - name: Archive
+        run: |
+          id="apache-iceberg-cpp-${VERSION}-rc${RC}"
```
Fokko: Maybe something more descriptive?
```suggestion
          ARCHIVE_FILE_NAME="apache-iceberg-cpp-${VERSION}-rc${RC}"
```

dev/release/README.md
```diff
@@ -0,0 +1,103 @@
...
+Your PGP key must be registered to the following:
+
+    * https://downloads.apache.org/iceberg/KEYS
+
+See the header comment of them for how to add a PGP key.
+
+Apache Iceberg committers can update them by Subversion client with their ASF account.
+e.g.:
+
+```console
```
Fokko, line 66: I always try to avoid editing the KEYS file manually, how about something like: https://iceberg.apache.org/how-to-release/#setup

dev/release/rat_exclude_files.txt
Fokko: I thought this was not allowed per ASF, but I can't find the source. Since rust also contains binaries, let's continue with this for now.

Fokko (approved):
Some nits. I haven't tried the script myself, but they look good 👍

HeartLinked:
@Fokko Thanks! All nits addressed.

dev/release/release.sh
```diff
@@ -0,0 +1,126 @@
...
+echo "  https://dist.apache.org/repos/dist/release/iceberg/${release_id}"
+echo
+echo "Add this release to ASF's report database:"
+echo "  https://reporter.apache.org/addrelease.html?iceberg"
+
+echo "Draft email for announce@apache.org mailing list"
+echo ""
+echo "---------------------------------------------------------"
+cat <<MAIL
+To: announce@apache.org
```
wgtmac: This should be sent to `dev@iceberg.apache.org` and cc `announce@apache.org`

dev/release/release.sh
```diff
@@ -0,0 +1,126 @@
...
+    "https://dist.apache.org/repos/dist/release/iceberg/${old_release_version}"
+done
+
+echo "Success! The release is available here:"
+echo "  https://dist.apache.org/repos/dist/release/iceberg/${release_id}"
+echo
+echo "Add this release to ASF's report database:"
+echo "  https://reporter.apache.org/addrelease.html?iceberg"
+
+echo "Draft email for announce@apache.org mailing list"
```
wgtmac:
```suggestion
echo "Draft email for announcement"
```

.github/workflows/rc.yml
```diff
@@ -0,0 +1,144 @@
...
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential cmake ninja-build libssl-dev
+
+      - name: Install C++ Build Tools (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          brew reinstall cmake ninja openssl
+
```
wgtmac:
```suggestion
```

Do we really need these lines? I just checked the CI workflow and they don't have anything similar:

https://github.com/apache/iceberg-cpp/blob/main/.github/workflows/test.yml#L40-L73

The only difference is that our CI uses `ubuntu-24.04` and `macos-15` but here are all `latest`. If we use the same versions as in ci, then we can safely remove the install steps as suggested above.

dev/release/README.md
```diff
@@ -0,0 +1,110 @@
...
+    2. Prepare RC and vote (detailed later)
+    3. Publish (detailed later)
+
+### Prepare RC and vote
+
+Run `dev/release/release_rc.sh` on a working copy of
+`git@github.com:apache/iceberg-cpp` not from your fork:
+
+```console
+$ git clone git@github.com:apache/iceberg-cpp.git
```
wgtmac:
```suggestion
$ git clone git@github.com:apache/iceberg-cpp.git && cd iceberg-cpp
```

wgtmac (approved):

HeartLinked:
@wgtmac I've fixed these comments.Thanks!

wgtmac (approved):

PR #180, opened 2025-08-15
feat: add find field (by id and name) support to schema

nullccxsy:
1. add insensitive way to find schemafield(list, struct, map)
2. change the complexity of find name to O(1)
3. test insensitive way to  find schemafield(list, struct, map)

PR #194, opened 2025-08-26
feat: add find field (by name) support to NestedType

nullccxsy:
… MapType

    - Implemented case-insensitive GetFieldByName in NestedType subclasses.
    - Added lazy initialization for maps in StructType
    - Handled duplicate names/IDs with Status returns instead of throws.

### DIFF
--- a/src/iceberg/type.h
+++ b/src/iceberg/type.h
@@ -70,15 +70,13 @@ class ICEBERG_EXPORT PrimitiveType : public Type {
 
 /// \brief A data type that has child fields.
 class ICEBERG_EXPORT NestedType : public Type {
- protected:
-  using SchemaFieldConstRef = std::reference_wrapper<const SchemaField>;
-
  public:
   bool is_primitive() const override { return false; }
   bool is_nested() const override { return true; }
 
   /// \brief Get a view of the child fields.
   [[nodiscard]] virtual std::span<const SchemaField> fields() const = 0;
+  using SchemaFieldConstRef = std::reference_wrapper<const SchemaField>;
   /// \brief Get a field by field ID.
   ///
   /// \note This is O(1) complexity.
@@ -132,7 +130,6 @@ class ICEBERG_EXPORT StructType : public NestedType {
   Status InitFieldByName() const;
   Status InitFieldByLowerCaseName() const;
 
- protected:
   std::vector<SchemaField> fields_;
   mutable std::unordered_map<int32_t, SchemaFieldConstRef> field_by_id_;
   mutable std::unordered_map<std::string_view, SchemaFieldConstRef> field_by_name_;


test/schema_test.cc
```diff
@@ -51,14 +51,6 @@ TEST(SchemaTest, Basics) {
     ASSERT_EQ(std::nullopt, schema.GetFieldByIndex(-1));
     ASSERT_EQ(std::nullopt, schema.GetFieldByName("element"));
   }
-  ASSERT_THAT(
-      []() {
-        iceberg::SchemaField field1(5, "foo", iceberg::int32(), true);
-        iceberg::SchemaField field2(5, "bar", iceberg::string(), true);
-        iceberg::Schema schema({field1, field2}, 100);
-      },
-      ::testing::ThrowsMessage<std::runtime_error>(
```
wgtmac, line 60: Why this gets removed?

src/iceberg/type.h
```diff
@@ -78,20 +79,23 @@ class ICEBERG_EXPORT NestedType : public Type {
...
+  [[nodiscard]] virtual Result<std::optional<std::reference_wrapper<const SchemaField>>>
   GetFieldById(int32_t field_id) const = 0;
   /// \brief Get a field by index.
   ///
   /// \note This is O(1) complexity.
-  [[nodiscard]] virtual std::optional<std::reference_wrapper<const SchemaField>>
+  [[nodiscard]] virtual Result<std::optional<std::reference_wrapper<const SchemaField>>>
   GetFieldByIndex(int32_t index) const = 0;
-  /// \brief Get a field by name (case-sensitive).  Behavior is undefined if
+  /// \brief Get a field by name.  Behavior is not allowed if
```
HeartLinked:
```suggestion
  /// \brief Get a field by name.  Returns an error Status if
```

src/iceberg/type.cc
```diff
@@ -105,23 +149,29 @@ std::string ListType::ToString() const {
...
-  if (name == element_.name()) {
+Result<std::optional<std::reference_wrapper<const SchemaField>>> ListType::GetFieldByName(
+    std::string_view name, bool case_sensitive) const {
+  if (case_sensitive) {
+    if (name == element_.name()) {
+      return std::cref(element_);
+    }
+    return std::nullopt;
+  }
+  if (StringUtils::ToLower(name) == StringUtils::ToLower(element_.name())) {
```
HeartLinked: maybe we can simply use `if (StringUtils::ToLower(name) == kLowerCaseElementName) {` ?

HeartLinked: I have one suggestion regarding its scope. I believe the case-insensitive logic should apply exclusively to StructType and not to ListType or MapType.

The reason is that the internal field names for ListType ("element") and MapType ("key", "value") are defined by the Iceberg specification as structural components, rather than user-defined columns. As such, they should probably be treated as case-sensitive constants.

Look for iceberg Java implementation `TypeUtil.java` and `Type.java`, you'll find the`indexByLowerCaseName` function is only for StructType.

src/iceberg/type.h
```diff
@@ -78,20 +79,23 @@ class ICEBERG_EXPORT NestedType : public Type {
   /// \brief Get a field by field ID.
   ///
   /// \note This is O(1) complexity.
-  [[nodiscard]] virtual std::optional<std::reference_wrapper<const SchemaField>>
+  [[nodiscard]] virtual Result<std::optional<std::reference_wrapper<const SchemaField>>>
```
wgtmac: nit: `using SchemaFieldConstRef = std::reference_wrapper<const SchemaField>;` to make it shorter and replace all return types below?

src/iceberg/type.cc
```diff
@@ -25,23 +25,18 @@
...
-      throw IcebergError(
-          std::format("StructType: duplicate field ID {} (field indices {} and {})",
-                      field.field_id(), it->second, index));
-    }
-
-    ++index;
-  }
+Result<std::optional<std::reference_wrapper<const SchemaField>>>
+NestedType::GetFieldByName(std::string_view name) const {
+  return GetFieldByName(name, true);
```
wgtmac:
```suggestion
  return GetFieldByName(name, /*case_sensitive=*/true);
```

src/iceberg/type.cc
```diff
@@ -53,27 +48,34 @@ std::string StructType::ToString() const {
...
+  auto it = field_by_id_.find(field_id);
+  if (it == field_by_id_.end()) return std::nullopt;
+  return it->second;
 }
-std::optional<std::reference_wrapper<const SchemaField>> StructType::GetFieldByIndex(
-    int32_t index) const {
-  if (index < 0 || index >= static_cast<int32_t>(fields_.size())) {
+Result<std::optional<std::reference_wrapper<const SchemaField>>>
+StructType::GetFieldByIndex(int32_t index) const {
+  if (index < 0 || static_cast<size_t>(index) >= fields_.size()) {
```
wgtmac, line 60: Perhaps we should return a `InvalidArgument` error now?

src/iceberg/type.cc
```diff
@@ -84,6 +86,48 @@ bool StructType::Equals(const Type& other) const {
...
   return fields_ == struct_.fields_;
 }
+Status StructType::InitFieldById() const {
+  if (!field_by_id_.empty()) {
+    return {};
+  }
+  for (const auto& field : fields_) {
+    auto it = field_by_id_.try_emplace(field.field_id(), field);
+    if (!it.second) {
+      return NotAllowed("Duplicate field id found: {} (prev name: {}, curr name: {})",
```
wgtmac:
```suggestion
      return InvalidSchema("Duplicate field id found: {} (prev name: {}, curr name: {})",
```

src/iceberg/type.cc
```diff
@@ -84,6 +86,48 @@ bool StructType::Equals(const Type& other) const {
...
+  return {};
+}
+Status StructType::InitFieldByName() const {
+  if (!field_by_name_.empty()) {
+    return {};
+  }
+  for (const auto& field : fields_) {
+    auto it = field_by_name_.try_emplace(field.name(), field);
+    if (!it.second) {
+      return NotAllowed("Duplicate field name found: {} (prev id: {}, curr id: {})",
```
wgtmac:
```suggestion
      return InvalidSchema("Duplicate field name found: {} (prev id: {}, curr id: {})",
```

src/iceberg/type.cc
```diff
@@ -84,6 +86,48 @@ bool StructType::Equals(const Type& other) const {
...
+}
+Status StructType::InitFieldByLowerCaseName() const {
+  if (!field_by_lowercase_name_.empty()) {
+    return {};
+  }
+  for (const auto& field : fields_) {
+    auto it =
+        field_by_lowercase_name_.try_emplace(StringUtils::ToLower(field.name()), field);
+    if (!it.second) {
+      return NotAllowed("Duplicate field name found: {} (prev id: {}, curr id: {})",
```
wgtmac:
```suggestion
      return InvalidSchema("Duplicate field name found: {} (prev id: {}, curr id: {})",
```

src/iceberg/type.cc
```diff
@@ -105,23 +149,29 @@ std::string ListType::ToString() const {
...
   return std::nullopt;
 }
-std::optional<std::reference_wrapper<const SchemaField>> ListType::GetFieldByIndex(
-    int index) const {
+Result<std::optional<std::reference_wrapper<const SchemaField>>>
+ListType::GetFieldByIndex(int index) const {
   if (index == 0) {
     return std::cref(element_);
   }
   return std::nullopt;
```
wgtmac: Return `InvalidArgument`?

src/iceberg/type.cc
```diff
@@ -177,11 +227,19 @@ std::optional<std::reference_wrapper<const SchemaField>> MapType::GetFieldByInde
...
+    std::string_view name, bool case_sensitive) const {
+  if (case_sensitive) {
+    if (name == kKeyName) {
+      return key();
+    } else if (name == kValueName) {
+      return value();
+    }
+    return std::nullopt;
+  }
+  if (StringUtils::ToLower(name) == StringUtils::ToLower(kKeyName)) {
```
wgtmac:
```suggestion
  if (StringUtils::ToLower(name) == kKeyName) {
```

wgtmac (changes requested):

src/iceberg/type.cc
nullccxsy: You can take a look at this code in `schema.java` — it converts all names to lowercase before querying. Therefore, even if the name is "KEY", it should still be searchable. Here, type.h reuses the functionality from schema to support select function.
```
public NestedField caseInsensitiveFindField(String name) {
    Preconditions.checkArgument(!name.isEmpty(), "Invalid column name: (empty)");
    Integer id = lazyLowerCaseNameToId().get(name.toLowerCase(Locale.ROOT));
    if (id != null) {
      return lazyIdToField().get(id);
    }
    return null;
 }
```

test/schema_test.cc
nullccxsy: The logic for detecting duplicate IDs has been moved to GetFieldById. Didn’t we agree earlier that no exceptions would be thrown?

src/iceberg/type.cc
HeartLinked: @wgtmac see see

test/schema_test.cc
wgtmac: You're right.

src/iceberg/type.cc
wgtmac: I think it's fine to support case-sensitive search for map and list type. This is consistent with the `Schema` implementation.

test/schema_test.cc
nullccxsy: For list's element and map's key, value, do we need to do the same processing? However, I think it's a bit strange to handle it in `GetFieldByName()`, they are easy,  so I didn't change its exception-throwing form.

src/iceberg/type.cc
```diff
@@ -105,23 +149,28 @@ std::string ListType::ToString() const {
...
   return std::nullopt;
 }
-std::optional<std::reference_wrapper<const SchemaField>> ListType::GetFieldByIndex(
-    int index) const {
+Result<std::optional<SchemaFieldConstRef>> ListType::GetFieldByIndex(int index) const {
   if (index == 0) {
     return std::cref(element_);
   }
-  return std::nullopt;
+  return InvalidArgument("index {} is out of range[0, {})", index, 1);
```
wgtmac:
```suggestion
  return InvalidArgument("Invalid index {} to get field from list", index);
```

src/iceberg/type.cc
```diff
@@ -159,29 +208,35 @@ std::string MapType::ToString() const {
...
-std::optional<std::reference_wrapper<const SchemaField>> MapType::GetFieldByIndex(
-    int32_t index) const {
+Result<std::optional<SchemaFieldConstRef>> MapType::GetFieldByIndex(int32_t index) const {
   if (index == 0) {
     return key();
   } else if (index == 1) {
     return value();
   }
-  return std::nullopt;
+  return InvalidArgument("index {} is out of range[0, {})", index, 2);
```
wgtmac:
```suggestion
  return InvalidArgument("Invalid index {} to get field from map", index);
```

src/iceberg/type.cc
```diff
@@ -159,29 +208,35 @@ std::string MapType::ToString() const {
...
+    std::string_view name, bool case_sensitive) const {
+  if (case_sensitive) {
+    if (name == kKeyName) {
+      return key();
+    } else if (name == kValueName) {
+      return value();
+    }
+    return std::nullopt;
+  }
+  if (StringUtils::ToLower(name) == kKeyName) {
```
wgtmac:
```suggestion
  const auto lower_case_name = StringUtils::ToLower(name);
  if (lower_case_name == kKeyName) {
```

src/iceberg/type.cc
```diff
@@ -159,29 +208,35 @@ std::string MapType::ToString() const {
...
     return key();
-  } else if (name == kValueName) {
+  } else if (StringUtils::ToLower(name) == kValueName) {
```
wgtmac:
```suggestion
  } else if (lower_case_name == kValueName) {
```

src/iceberg/type.cc
wgtmac: Let's avoid converting it twice in the worst case.

src/iceberg/type.h
```diff
@@ -67,6 +68,7 @@ class ICEBERG_EXPORT PrimitiveType : public Type {
   bool is_nested() const override { return false; }
 };
 
+using SchemaFieldConstRef = std::reference_wrapper<const SchemaField>;
```
wgtmac: Can we put it inside the definition of `NestedType`?

src/iceberg/type.h
```diff
@@ -109,18 +114,26 @@ class ICEBERG_EXPORT StructType : public NestedType {
...
  protected:
   bool Equals(const Type& other) const override;
+  Status InitFieldById() const;
+  Status InitFieldByName() const;
+  Status InitFieldByLowerCaseName() const;
 
+ protected:
   std::vector<SchemaField> fields_;
-  std::unordered_map<int32_t, size_t> field_id_to_index_;
+  mutable std::unordered_map<int32_t, SchemaFieldConstRef> field_by_id_;
```
wgtmac, line 134: Add a TODO comment to make it thread-safe as well.

wgtmac (commented):
Generally looks good. Left some nits.

src/iceberg/type.h
nullccxsy: if do this, in type.cc, we should use `NestedType::ScheamFieldConstRef`

wgtmac: Yes, this is by purpose.

HeartLinked (approved):

src/iceberg/type.h
```diff
@@ -72,26 +73,30 @@ class ICEBERG_EXPORT NestedType : public Type {
  public:
   bool is_primitive() const override { return false; }
   bool is_nested() const override { return true; }
+  using SchemaFieldConstRef = std::reference_wrapper<const SchemaField>;
```
wgtmac:
```
class ICEBERG_EXPORT NestedType : public Type {
  using SchemaFieldConstRef = std::reference_wrapper<const SchemaField>;

 public:
  bool is_primitive() const override { return false; }
```

It would be better to put the type alias in the non-public scope

src/iceberg/type.h
```diff
@@ -101,6 +106,7 @@ class ICEBERG_EXPORT NestedType : public Type {
 /// \brief A data type representing a struct with nested fields.
 class ICEBERG_EXPORT StructType : public NestedType {
  public:
+  using NestedType::GetFieldByName;
```
wgtmac:
```suggestion
```

Move it closer to its friends.

src/iceberg/type.h
```diff
@@ -109,23 +115,32 @@ class ICEBERG_EXPORT StructType : public NestedType {
...
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldById(
+  Result<std::optional<SchemaFieldConstRef>> GetFieldById(
       int32_t field_id) const override;
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldByIndex(
+  Result<std::optional<SchemaFieldConstRef>> GetFieldByIndex(
       int32_t index) const override;
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldByName(
-      std::string_view name) const override;
+  Result<std::optional<SchemaFieldConstRef>> GetFieldByName(
+      std::string_view name, bool case_sensitive) const override;
```
wgtmac, line 122:
```suggestion
  Result<std::optional<SchemaFieldConstRef>> GetFieldByName(
      std::string_view name, bool case_sensitive) const override;
  using NestedType::GetFieldByName;
```

src/iceberg/type.h
```diff
@@ -109,23 +115,32 @@ class ICEBERG_EXPORT StructType : public NestedType {
...
-  std::unordered_map<int32_t, size_t> field_id_to_index_;
+  mutable std::unordered_map<int32_t, SchemaFieldConstRef> field_by_id_;
+  mutable std::unordered_map<std::string_view, SchemaFieldConstRef> field_by_name_;
+  mutable std::unordered_map<std::string, SchemaFieldConstRef> field_by_lowercase_name_;
 };
 
 /// \brief A data type representing a list of values.
 class ICEBERG_EXPORT ListType : public NestedType {
  public:
+  using NestedType::GetFieldByName;
```
wgtmac:
```suggestion
```

src/iceberg/type.h
```diff
@@ -140,12 +155,12 @@ class ICEBERG_EXPORT ListType : public NestedType {
...
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldById(
+  Result<std::optional<SchemaFieldConstRef>> GetFieldById(
       int32_t field_id) const override;
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldByIndex(
+  Result<std::optional<SchemaFieldConstRef>> GetFieldByIndex(
       int32_t index) const override;
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldByName(
-      std::string_view name) const override;
+  Result<std::optional<SchemaFieldConstRef>> GetFieldByName(
+      std::string_view name, bool case_sensitive) const override;
```
wgtmac, line 161:
```suggestion
  Result<std::optional<SchemaFieldConstRef>> GetFieldByName(
      std::string_view name, bool case_sensitive) const override;
  using NestedType::GetFieldByName;
```

src/iceberg/type.h
```diff
@@ -156,6 +171,7 @@ class ICEBERG_EXPORT ListType : public NestedType {
 /// \brief A data type representing a dictionary of values.
 class ICEBERG_EXPORT MapType : public NestedType {
  public:
+  using NestedType::GetFieldByName;
```
wgtmac:
```suggestion
```

src/iceberg/type.h
```diff
@@ -172,12 +188,12 @@ class ICEBERG_EXPORT MapType : public NestedType {
...
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldById(
+  Result<std::optional<SchemaFieldConstRef>> GetFieldById(
       int32_t field_id) const override;
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldByIndex(
+  Result<std::optional<SchemaFieldConstRef>> GetFieldByIndex(
       int32_t index) const override;
-  std::optional<std::reference_wrapper<const SchemaField>> GetFieldByName(
-      std::string_view name) const override;
+  Result<std::optional<SchemaFieldConstRef>> GetFieldByName(
+      std::string_view name, bool case_sensitive) const override;
```
wgtmac, line 194:
```suggestion
  Result<std::optional<SchemaFieldConstRef>> GetFieldByName(
      std::string_view name, bool case_sensitive) const override;
  using NestedType::GetFieldByName;
```

src/iceberg/type.h
```diff
@@ -109,23 +115,32 @@ class ICEBERG_EXPORT StructType : public NestedType {
...
+  Status InitFieldById() const;
+  Status InitFieldByName() const;
+  Status InitFieldByLowerCaseName() const;
 
+ protected:
   std::vector<SchemaField> fields_;
-  std::unordered_map<int32_t, size_t> field_id_to_index_;
+  mutable std::unordered_map<int32_t, SchemaFieldConstRef> field_by_id_;
+  mutable std::unordered_map<std::string_view, SchemaFieldConstRef> field_by_name_;
+  mutable std::unordered_map<std::string, SchemaFieldConstRef> field_by_lowercase_name_;
```
wgtmac, line 136: We need to wait for https://github.com/apache/iceberg-cpp/pull/180 and rebase on it to use `StringHash` here.

src/iceberg/type.h
```diff
@@ -72,26 +73,30 @@ class ICEBERG_EXPORT NestedType : public Type {
...
+  [[nodiscard]] virtual Result<std::optional<SchemaFieldConstRef>> GetFieldByIndex(
+      int32_t index) const = 0;
+  /// \brief Get a field by name.  Return an error Status if
   ///   the field name is not unique; prefer GetFieldById or GetFieldByIndex
   ///   when possible.
   ///
-  /// \note This is currently O(n) complexity.
-  [[nodiscard]] virtual std::optional<std::reference_wrapper<const SchemaField>>
-  GetFieldByName(std::string_view name) const = 0;
+  /// \note This is currently O(1) complexity.
```
wgtmac:
```suggestion
  /// \note This is O(1) complexity.
```

src/iceberg/type.h
```diff
@@ -72,26 +73,30 @@ class ICEBERG_EXPORT NestedType : public Type {
...
+  [[nodiscard]] virtual Result<std::optional<SchemaFieldConstRef>> GetFieldByName(
+      std::string_view name, bool case_sensitive) const = 0;
+  /// \brief Get a field by name(case-sensitive).
```
wgtmac:
```suggestion
  /// \brief Get a field by name (case-sensitive).
```

src/iceberg/type.cc
```diff
@@ -84,6 +86,48 @@ bool StructType::Equals(const Type& other) const {
...
+}
+Status StructType::InitFieldByLowerCaseName() const {
+  if (!field_by_lowercase_name_.empty()) {
+    return {};
+  }
+  for (const auto& field : fields_) {
+    auto it =
+        field_by_lowercase_name_.try_emplace(StringUtils::ToLower(field.name()), field);
+    if (!it.second) {
+      return InvalidSchema("Duplicate field name found: {} (prev id: {}, curr id: {})",
```
wgtmac:
```suggestion
      return InvalidSchema("Duplicate lowercase field name found: {} (prev id: {}, curr id: {})",
```

src/iceberg/util/macros.h
```diff
@@ -19,10 +19,13 @@
...
 
-#define ICEBERG_RETURN_UNEXPECTED(result)          \
-  if (!result) [[unlikely]] {                      \
-    return std::unexpected<Error>(result.error()); \
-  }
+#define ICEBERG_RETURN_UNEXPECTED(result)                         \
+  do {                                                            \
+    auto&& iceberg_temp_result = (result);                        \
+    if (!iceberg_temp_result) [[unlikely]] {                      \
+      return std::unexpected<Error>(iceberg_temp_result.error()); \
```
wgtmac:
```suggestion
    auto&& result_name = (result);                        \
    if (!result_name) [[unlikely]] {                      \
      return std::unexpected<Error>(result_name.error()); \
```

test/schema_test.cc
```diff
@@ -47,18 +47,16 @@ TEST(SchemaTest, Basics) {
     ASSERT_THAT(schema.GetFieldByName("bar"), ::testing::Optional(field2));
 
     ASSERT_EQ(std::nullopt, schema.GetFieldById(0));
-    ASSERT_EQ(std::nullopt, schema.GetFieldByIndex(2));
-    ASSERT_EQ(std::nullopt, schema.GetFieldByIndex(-1));
+    auto result = schema.GetFieldByIndex(2);
+    ASSERT_EQ(result.error().kind, iceberg::ErrorKind::kInvalidArgument);
+    ASSERT_THAT(result.error().message,
+                ::testing::HasSubstr("Invalid index 2 to get field from struct"));
```
wgtmac:
```suggestion
    ASSERT_THAT(result, IsError(ErrorKind::kInvalidArgument));
    ASSERT_THAT(result, HasErrorMessage("Invalid index 2 to get field from struct"));
```

wgtmac: Same for other similar change in the test files.

wgtmac:
BTW, the PR title is broken...

src/iceberg/util/macros.h
```diff
@@ -19,10 +19,13 @@
 
 #pragma once
 
-#define ICEBERG_RETURN_UNEXPECTED(result)          \
-  if (!result) [[unlikely]] {                      \
-    return std::unexpected<Error>(result.error()); \
-  }
+#define ICEBERG_RETURN_UNEXPECTED(result)                 \
+  do {                                                    \
+    auto&& result_name = (result);                        \
```
HuaHuaY: What is this change doing?

nullccxsy: `result` might be a function. In this case, both `result` and `result.error()` would call this function once. This would cause issues when we're checking for errors like duplicate names.

HuaHuaY: Got it. Maybe we can remove `do ... while(false)`.

src/iceberg/util/macros.h
```diff
@@ -19,10 +19,13 @@
...
+    if (!result_name) [[unlikely]] {                      \
+      return std::unexpected<Error>(result_name.error()); \
+    }                                                     \
+  } while (false);
```
HuaHuaY:
```suggestion
#define ICEBERG_RETURN_UNEXPECTED(result)                         \
  if (auto&& result_name = result; !result_name) [[unlikely]] { \
    return std::unexpected<Error>(result_name.error());           \
  }
```

nullccxsy: Thank you, it will be fixed.

src/iceberg/type.h
```diff
@@ -69,6 +70,9 @@ class ICEBERG_EXPORT PrimitiveType : public Type {
 
 /// \brief A data type that has child fields.
 class ICEBERG_EXPORT NestedType : public Type {
+ protected:
```
wgtmac:
```suggestion
```

nit: we can just use the default visibility

src/iceberg/type.h
```diff
@@ -109,18 +116,27 @@ class ICEBERG_EXPORT StructType : public NestedType {
...
 
  protected:
   bool Equals(const Type& other) const override;
+  // TODO(nullccxsy): Lazy initialization has concurrency issues, need to add proper
+  // synchronization mechanism
+  Status InitFieldById() const;
+  Status InitFieldByName() const;
+  Status InitFieldByLowerCaseName() const;
 
+ protected:
```
wgtmac:
```suggestion
```

wgtmac (approved):
LGTM. Left some nits. Thanks for working on this!

wgtmac (approved):

Fokko (approved):

PR #196, opened 2025-08-27
feat: implement endian conversion utilities

HeartLinked:
Implements cross-platform endianness conversion utilities in `src/iceberg/util/endian.h` with full support for integer and floating-point types, including comprehensive test coverage.


src/iceberg/util/endian.h
```diff
@@ -0,0 +1,124 @@
...
+        uint32_t int_representation = std::bit_cast<uint32_t>(value);
+        int_representation = std::byteswap(int_representation);
+        return std::bit_cast<T>(int_representation);
+      } else if constexpr (sizeof(T) == sizeof(uint64_t)) {
+        uint64_t int_representation = std::bit_cast<uint64_t>(value);
+        int_representation = std::byteswap(int_representation);
+        return std::bit_cast<T>(int_representation);
+      }
+    }
+  }
```
mapleFU: A un-matched should compile-error rather than do-nothing?

HeartLinked: I don't understand your concern. This is a template method that only accepts types where std::is_arithmetic_v && !std::same_as<T, bool> evaluates to true. Given these constraints, I believe failure is impossible for such types.

mapleFU: Since we have C++23, so, maybe: https://en.cppreference.com/w/cpp/types/floating-point.html 
:-)

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,124 @@
...
+#include <concepts>
+
+/// \file iceberg/util/endian.h
+/// \brief Endianness conversion utilities
+
+namespace iceberg {
+
+/// \brief Concept for values that can be converted to/from another endian format.
+template <typename T>
+concept EndianConvertible = std::is_arithmetic_v<T> && !std::same_as<T, bool>;
```
wgtmac: Why excluding `bool`?

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,125 @@
...
+
+/// \brief Convert a value to little-endian format.
+template <EndianConvertible T>
+constexpr T ToLittleEndian(T value) {
+  if constexpr (std::endian::native == std::endian::little || sizeof(T) <= 1) {
+    return value;
+  } else {
+    if constexpr (std::is_integral_v<T>) {
```
wgtmac:
```suggestion
  } else if constexpr (std::is_integral_v<T>) {
```

Can we unnest the else branch?

test/CMakeLists.txt
```diff
@@ -89,7 +89,8 @@ add_iceberg_test(util_test
                  formatter_test.cc
                  config_test.cc
                  visit_type_test.cc
-                 string_utils_test.cc)
+                 string_utils_test.cc
+                 endian_test.cc)
```
wgtmac: Please sort them alphabetically.

src/iceberg/util/endian.h
HeartLinked: You are right, I've added static_assert to provide clearer compile-error information.

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,135 @@
...
+      return std::byteswap(value);
+    } else if constexpr (std::is_floating_point_v<T>) {
+      // For floats, use the bit_cast -> byteswap -> bit_cast pattern.
+      // We only support 32-bit and 64-bit floats.
+      if constexpr (sizeof(T) == sizeof(uint32_t)) {
+        uint32_t int_representation = std::bit_cast<uint32_t>(value);
+        int_representation = std::byteswap(int_representation);
+        return std::bit_cast<T>(int_representation);
```
wgtmac:
```suggestion
        return std::bit_cast<T>(std::byteswap(std::bit_cast<uint32_t>(value)));
```

We can use one-liner here.

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,135 @@
...
+
+namespace iceberg {
+
+/// \brief Concept for values that can be converted to/from another endian format.
+template <typename T>
+concept EndianConvertible = std::is_arithmetic_v<T> && !std::same_as<T, bool>;
+
+/// \brief Convert a value to little-endian format.
+template <EndianConvertible T>
+constexpr T ToLittleEndian(T value) {
```
wgtmac, line 58: There four functions are essentially same. Could we write our own `ByteSwap` function and reuse it by those functions?

src/iceberg/util/endian.h
HeartLinked: Sizeof(bool) is 1 byte, and endianness is irrelevant for single-byte data types; a byte-swap is a no-op. 

Also, according to the Iceberg specification, a boolean is represented as 0x00 for false and any non-zero byte for true. This definition is inherently independent of endianness, which is different from multi-byte types like int or double, where the specification explicitly defines the value as being stored in a particular endian format .

zhjwpku: What about char? There is a `sizeof(T) <= 1` branch, so maybe treat bool the same.

HeartLinked: > What about char? There is a `sizeof(T) <= 1` branch, so maybe treat bool the same.

I see your point, but I think performing an endian conversion on a bool is semantically less reasonable. The main benefit of the current design is that it explicitly catches this kind of logical error at compile time.

wgtmac: If we don't support bool here, users may need to special case bool to avoid passing it to `ByteSwap`. For simplicity, should we support bool as well?

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,94 @@
...
+/// \brief Byte-swap a value. For floating-point types, only support 32-bit and 64-bit
+/// floats.
+template <EndianConvertible T>
+constexpr T ByteSwap(T value) {
+  if constexpr (sizeof(T) <= 1) {
+    return value;
+  } else if constexpr (std::is_integral_v<T>) {
+    return std::byteswap(value);
+  } else if constexpr (std::is_floating_point_v<T>) {
+    if constexpr (sizeof(T) == sizeof(uint32_t)) {
```
wgtmac: Should we add `if constexpr (sizeof(T) == sizeof(uint16_t))` to be future-proof?

There are std::float16_t and std::bfloat16_t.

test/endian_test.cc
```diff
@@ -0,0 +1,121 @@
...
+
+#include <array>
+#include <cmath>
+#include <limits>
+
+#include <gtest/gtest.h>
+
+namespace iceberg {
+
+// test round trip preserves value
```
wgtmac:
```suggestion
```

The comment is redundant.

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,94 @@
...
+
+/// \brief Concept for values that can be converted to/from another endian format.
+template <typename T>
+concept EndianConvertible = std::is_arithmetic_v<T> && !std::same_as<T, bool>;
+
+/// \brief Byte-swap a value. For floating-point types, only support 32-bit and 64-bit
+/// floats.
+template <EndianConvertible T>
+constexpr T ByteSwap(T value) {
+  if constexpr (sizeof(T) <= 1) {
```
wgtmac, line 39: It seems that you have already supported bool type here.

src/iceberg/util/endian.h
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/// \brief Convert a value to little-endian format.
+template <EndianConvertible T>
+constexpr T ToLittleEndian(T value) {
+  if constexpr (std::endian::native == std::endian::little || sizeof(T) <= 1) {
+    return value;
+  } else {
+    return ByteSwap(value);
+  }
```
wgtmac:
```suggestion
  if constexpr (std::endian::native == std::endian::little) {
    return value;
  } else {
    return ByteSwap(value);
  }
```

Doesn't it look better?

test/endian_test.cc
```diff
@@ -0,0 +1,121 @@
...
+  EXPECT_EQ(FromLittleEndian(ToLittleEndian(-std::numeric_limits<double>::infinity())),
+            -std::numeric_limits<double>::infinity());
+  EXPECT_TRUE(
+      std::isnan(FromBigEndian(ToBigEndian(std::numeric_limits<double>::quiet_NaN()))));
+  EXPECT_EQ(FromLittleEndian(ToLittleEndian(0.0)), 0.0);
+  EXPECT_EQ(FromBigEndian(ToBigEndian(-0.0)), -0.0);
+}
+
+// test constexpr evaluation
+TEST(EndianTest, ConstexprEvaluation) {
```
wgtmac: Is there any difference with the above case?

test/endian_test.cc
```diff
@@ -0,0 +1,121 @@
...
+
+  static_assert(ToBigEndian<uint8_t>(0xFF) == 0xFF);
+  static_assert(FromLittleEndian<int8_t>(-1) == -1);
+
+  static_assert(FromLittleEndian(ToLittleEndian(3.14f)) == 3.14f);
+  static_assert(FromBigEndian(ToBigEndian(2.71)) == 2.71);
+}
+
+// test platform dependent behavior
+TEST(EndianTest, PlatformDependentBehavior) {
```
wgtmac: I don't think we need this case.

test/endian_test.cc
```diff
@@ -0,0 +1,121 @@
...
+#include <cmath>
+#include <limits>
+
+#include <gtest/gtest.h>
+
+namespace iceberg {
+
+// test round trip preserves value
+TEST(EndianTest, RoundTripPreservesValue) {
+  EXPECT_EQ(FromLittleEndian(ToLittleEndian<uint16_t>(0x1234)), 0x1234);
```
wgtmac: There are a lot of duplicate lines here. Can we use a simple macro for this?

```
#define EXPECT_ROUNDTRIP(T value)                            \
  EXPECT_EQ(FromLittleEndian(ToLittleEndian(value)), value); \
  EXPECT_EQ(FromBigEndian(ToBigEndian(value)), value);
```

test/endian_test.cc
```diff
@@ -0,0 +1,121 @@
...
+    EXPECT_EQ(FromBigEndian(test_value), test_value);
+    EXPECT_NE(ToLittleEndian(test_value), test_value);
+  }
+
+  EXPECT_EQ(ToLittleEndian<uint8_t>(0xAB), 0xAB);
+  EXPECT_EQ(ToBigEndian<uint8_t>(0xAB), 0xAB);
+}
+
+// test specific byte pattern validation
+TEST(EndianTest, SpecificBytePatternValidation) {
```
wgtmac:
```suggestion
TEST(EndianTest, ByteWiseValidation) {
```

test/endian_test.cc
```diff
@@ -0,0 +1,121 @@
...
+  uint32_t original_int = 0x12345678;
+  uint32_t little_endian_int = ToLittleEndian(original_int);
+  uint32_t big_endian_int = ToBigEndian(original_int);
+
+  auto little_int_bytes = std::bit_cast<std::array<uint8_t, 4>>(little_endian_int);
+  auto big_int_bytes = std::bit_cast<std::array<uint8_t, 4>>(big_endian_int);
+
+  EXPECT_EQ(little_int_bytes, (std::array<uint8_t, 4>{0x78, 0x56, 0x34, 0x12}));
+  EXPECT_EQ(big_int_bytes, (std::array<uint8_t, 4>{0x12, 0x34, 0x56, 0x78}));
```
wgtmac, line 74:
```suggestion
  EXPECT_EQ(little_int_bytes, std::array<uint8_t, 4>{0x78, 0x56, 0x34, 0x12});
  EXPECT_EQ(big_int_bytes, std::array<uint8_t, 4>{0x12, 0x34, 0x56, 0x78});
```

test/endian_test.cc
```diff
@@ -0,0 +1,121 @@
...
+
+  float original_float = 3.14f;
+  float little_endian_float = ToLittleEndian(original_float);
+  float big_endian_float = ToBigEndian(original_float);
+
+  auto little_float_bytes = std::bit_cast<std::array<uint8_t, 4>>(little_endian_float);
+  auto big_float_bytes = std::bit_cast<std::array<uint8_t, 4>>(big_endian_float);
+
+  EXPECT_EQ(little_float_bytes, (std::array<uint8_t, 4>{0xC3, 0xF5, 0x48, 0x40}));
+  EXPECT_EQ(big_float_bytes, (std::array<uint8_t, 4>{0x40, 0x48, 0xF5, 0xC3}));
```
wgtmac, line 84:
```suggestion
  EXPECT_EQ(little_float_bytes, std::array<uint8_t, 4>{0xC3, 0xF5, 0x48, 0x40});
  EXPECT_EQ(big_float_bytes, std::array<uint8_t, 4>{0x40, 0x48, 0xF5, 0xC3});
```

test/endian_test.cc
wgtmac: Or we can use template function to leverage parameterized test to improve readability.

test/endian_test.cc
HeartLinked: A macro argument containing commas must be wrapped in extra parentheses to prevent the preprocessor from misinterpreting it as multiple separate arguments.

test/endian_test.cc
HeartLinked: I'll delete this case.

src/iceberg/util/endian.h
HeartLinked: fixed

HeartLinked:
@wgtmac I've updated the code to address the feedback from the review. Thank you.

wgtmac (approved):
Thanks @HeartLinked!

PR #198, opened 2025-08-28
feat: implement basic parquet writer and add roundtrip tests

HuaHuaY:
Add parquet writer factory and basic parquet writer without metrics.

### DIFF
--- a/src/iceberg/parquet/parquet_reader.cc
+++ b/src/iceberg/parquet/parquet_reader.cc
@@ -125,9 +125,9 @@ class ParquetReader::Impl {
     arrow_reader_properties.set_arrow_extensions_enabled(true);
 
     // Open the Parquet file reader
-    ICEBERG_ASSIGN_OR_RAISE(auto input_stream, OpenInputStream(options));
+    ICEBERG_ASSIGN_OR_RAISE(input_stream_, OpenInputStream(options));
     auto file_reader =
-        ::parquet::ParquetFileReader::Open(std::move(input_stream), reader_properties);
+        ::parquet::ParquetFileReader::Open(input_stream_, reader_properties);
     ICEBERG_ARROW_RETURN_NOT_OK(::parquet::arrow::FileReader::Make(
         pool_, std::move(file_reader), arrow_reader_properties, &reader_));
 
@@ -169,6 +169,7 @@ class ParquetReader::Impl {
     }
 
     reader_.reset();
+    ICEBERG_ARROW_RETURN_NOT_OK(input_stream_->Close());
     return {};
   }
 
@@ -240,6 +241,8 @@ class ParquetReader::Impl {
   std::unique_ptr<::parquet::arrow::FileReader> reader_;
   // The context to keep track of the reading progress.
   std::unique_ptr<ReadContext> context_;
+  // The input stream to read Parquet file.
+  std::shared_ptr<::arrow::io::RandomAccessFile> input_stream_;
 };
 
 ParquetReader::~ParquetReader() = default;
--- a/src/iceberg/parquet/parquet_reader.cc
+++ b/src/iceberg/parquet/parquet_reader.cc
@@ -237,12 +237,12 @@ class ParquetReader::Impl {
   std::shared_ptr<::iceberg::Schema> read_schema_;
   // The projection result to apply to the read schema.
   SchemaProjection projection_;
+  // The input stream to read Parquet file.
+  std::shared_ptr<::arrow::io::RandomAccessFile> input_stream_;
   // Parquet file reader to create RecordBatchReader.
   std::unique_ptr<::parquet::arrow::FileReader> reader_;
   // The context to keep track of the reading progress.
   std::unique_ptr<ReadContext> context_;
-  // The input stream to read Parquet file.
-  std::shared_ptr<::arrow::io::RandomAccessFile> input_stream_;
 };
 
 ParquetReader::~ParquetReader() = default;
--- a/src/iceberg/parquet/parquet_writer.cc
+++ b/src/iceberg/parquet/parquet_writer.cc
@@ -66,12 +66,9 @@ class ParquetWriter::Impl {
     auto schema_node = std::static_pointer_cast<::parquet::schema::GroupNode>(
         schema_descriptor->schema_root());
 
-    std::shared_ptr<::arrow::KeyValueMetadata> metadata =
-        ::arrow::key_value_metadata(options.properties);
-
     ICEBERG_ASSIGN_OR_RAISE(output_stream_, OpenOutputStream(options));
     auto file_writer = ::parquet::ParquetFileWriter::Open(output_stream_, schema_node,
-                                                          writer_properties, metadata);
+                                                          writer_properties);
     ICEBERG_ARROW_RETURN_NOT_OK(::parquet::arrow::FileWriter::Make(
         pool_, std::move(file_writer), arrow_schema_, arrow_writer_properties, &writer_));
 
@@ -106,10 +103,10 @@ class ParquetWriter::Impl {
   ::arrow::MemoryPool* pool_ = ::arrow::default_memory_pool();
   // Schema to write from the Parquet file.
   std::shared_ptr<::arrow::Schema> arrow_schema_;
-  // Parquet file writer to write ArrowArray.
-  std::unique_ptr<::parquet::arrow::FileWriter> writer_;
   // The output stream to write Parquet file.
   std::shared_ptr<::arrow::io::OutputStream> output_stream_;
+  // Parquet file writer to write ArrowArray.
+  std::unique_ptr<::parquet::arrow::FileWriter> writer_;
 };
 
 ParquetWriter::~ParquetWriter() = default;
--- a/src/iceberg/parquet/parquet_writer.cc
+++ b/src/iceberg/parquet/parquet_writer.cc
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/parquet/parquet_writer.h"
+
+#include <memory>
+
+#include <arrow/c/bridge.h>
+#include <arrow/record_batch.h>
+#include <arrow/util/key_value_metadata.h>
+#include <parquet/arrow/schema.h>
+#include <parquet/arrow/writer.h>
+#include <parquet/file_writer.h>
+#include <parquet/properties.h>
+
+#include "iceberg/arrow/arrow_error_transform_internal.h"
+#include "iceberg/arrow/arrow_fs_file_io_internal.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg::parquet {
+
+namespace {
+
+Result<std::shared_ptr<::arrow::io::OutputStream>> OpenOutputStream(
+    const WriterOptions& options) {
+  auto io = internal::checked_pointer_cast<arrow::ArrowFileSystemFileIO>(options.io);
+  ICEBERG_ARROW_ASSIGN_OR_RETURN(auto output, io->fs()->OpenOutputStream(options.path));
+  return output;
+}
+
+}  // namespace
+
+class ParquetWriter::Impl {
+ public:
+  Status Open(const WriterOptions& options) {
+    auto writer_properties =
+        ::parquet::WriterProperties::Builder().memory_pool(pool_)->build();
+    auto arrow_writer_properties = ::parquet::default_arrow_writer_properties();
+
+    ArrowSchema c_schema;
+    ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*options.schema, &c_schema));
+    ICEBERG_ARROW_ASSIGN_OR_RETURN(arrow_schema_, ::arrow::ImportSchema(&c_schema));
+
+    std::shared_ptr<::parquet::SchemaDescriptor> schema_descriptor;
+    ICEBERG_ARROW_RETURN_NOT_OK(
+        ::parquet::arrow::ToParquetSchema(arrow_schema_.get(), *writer_properties,
+                                          *arrow_writer_properties, &schema_descriptor));
+    auto schema_node = std::static_pointer_cast<::parquet::schema::GroupNode>(
+        schema_descriptor->schema_root());
+
+    std::shared_ptr<::arrow::KeyValueMetadata> metadata =
+        ::arrow::key_value_metadata(options.properties);
+
+    ICEBERG_ASSIGN_OR_RAISE(auto output_stream, OpenOutputStream(options));
+    auto file_writer = ::parquet::ParquetFileWriter::Open(
+        std::move(output_stream), schema_node, writer_properties, metadata);
+    ICEBERG_ARROW_RETURN_NOT_OK(::parquet::arrow::FileWriter::Make(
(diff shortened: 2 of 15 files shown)


src/iceberg/parquet/parquet_reader.cc
```diff
@@ -240,6 +241,8 @@ class ParquetReader::Impl {
   std::unique_ptr<::parquet::arrow::FileReader> reader_;
   // The context to keep track of the reading progress.
   std::unique_ptr<ReadContext> context_;
+  // The input stream to read Parquet file.
+  std::shared_ptr<::arrow::io::RandomAccessFile> input_stream_;
```
wgtmac: nit: move it before `reader_`?

src/iceberg/schema_internal.h
```diff
@@ -33,7 +34,7 @@ namespace iceberg {
 // Apache Arrow C++ uses "PARQUET:field_id" to store field IDs for Parquet.
 // Here we follow a similar convention for Iceberg but we might also add
 // "PARQUET:field_id" in the future once we implement a Parquet writer.
-constexpr std::string_view kFieldIdKey = "ICEBERG:field_id";
+constexpr std::string_view kFieldIdKey = kParquetFieldIdKey;
```
wgtmac: What about removing this line and directly use `kParquetFieldIdKey` in the source file?

test/parquet_test.cc
```diff
@@ -17,38 +17,113 @@
...
 #include <arrow/json/from_string.h>
 #include <arrow/record_batch.h>
 #include <arrow/table.h>
+#include <arrow/type.h>
 #include <arrow/util/key_value_metadata.h>
+#include <iceberg/file_reader.h>
+#include <iceberg/file_writer.h>
+#include <iceberg/result.h>
+#include <iceberg/schema_field.h>
+#include <iceberg/schema_internal.h>
```
wgtmac:
```suggestion
```

They are redundant, right? Or if really required, use `""` to include them.

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,154 @@
...
+
+    std::shared_ptr<::parquet::SchemaDescriptor> schema_descriptor;
+    ICEBERG_ARROW_RETURN_NOT_OK(
+        ::parquet::arrow::ToParquetSchema(arrow_schema_.get(), *writer_properties,
+                                          *arrow_writer_properties, &schema_descriptor));
+    auto schema_node = std::static_pointer_cast<::parquet::schema::GroupNode>(
+        schema_descriptor->schema_root());
+
+    std::shared_ptr<::arrow::KeyValueMetadata> metadata =
+        ::arrow::key_value_metadata(options.properties);
```
wgtmac: Let's remove this first. These properties are not supposed for metadata. They are writer configs instead.

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,154 @@
...
+
+ private:
+  // TODO(gangwu): make memory pool configurable
+  ::arrow::MemoryPool* pool_ = ::arrow::default_memory_pool();
+  // Schema to write from the Parquet file.
+  std::shared_ptr<::arrow::Schema> arrow_schema_;
+  // Parquet file writer to write ArrowArray.
+  std::unique_ptr<::parquet::arrow::FileWriter> writer_;
+  // The output stream to write Parquet file.
+  std::shared_ptr<::arrow::io::OutputStream> output_stream_;
```
wgtmac, line 119: nit: define them in the creation order?

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,154 @@
...
+Status ParquetWriter::Close() { return impl_->Close(); }
+
+std::optional<Metrics> ParquetWriter::metrics() {
+  if (!impl_->Closed()) {
+    return std::nullopt;
+  }
+  return {};
+}
+
+std::optional<int64_t> ParquetWriter::length() {
```
wgtmac, line 146: Why not implement this?

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,154 @@
...
+    return std::nullopt;
+  }
+  return {};
+}
+
+std::vector<int64_t> ParquetWriter::split_offsets() {
+  if (!impl_->Closed()) {
+    return {};
+  }
+  return {};
```
wgtmac: ditto

test/parquet_test.cc
```diff
@@ -17,38 +17,113 @@
...
+                  const WriterOptions& writer_options) {
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto writer, WriterFactoryRegistry::Open(FileFormatType::kParquet, writer_options));
+  ArrowArray arr;
+  ICEBERG_ARROW_RETURN_NOT_OK(::arrow::ExportArray(*data, &arr));
+  ICEBERG_RETURN_UNEXPECTED(writer->Write(arr));
+  return writer->Close();
+}
+
+Status ReadTable(std::optional<std::shared_ptr<::arrow::Array>>* out,
```
wgtmac:
```suggestion
Status ReadTable(std::shared_ptr<::arrow::Array>* out,
```

It seems we can get rid of optional.

test/parquet_test.cc
```diff
@@ -204,4 +280,85 @@ TEST_F(ParquetReaderTest, ReadSplit) {
...
+  });
+
+  ArrowSchema arrow_c_schema;
+  ASSERT_THAT(ToArrowSchema(*schema, &arrow_c_schema), IsOk());
+  auto arrow_schema = ::arrow::ImportType(&arrow_c_schema).ValueOrDie();
+
+  auto array = ::arrow::json::ArrayFromJSONString(
+                   ::arrow::struct_(arrow_schema->fields()),
+                   R"([[true, 1, 2, 1.1, 1.2, "abc", 44614000, 1756696503000000],
+                            [false, 0, 0, 0, 0, "", 0, 0],
```
wgtmac: The alignment is a little bit strange...

test/parquet_test.cc
```diff
@@ -204,4 +280,85 @@ TEST_F(ParquetReaderTest, ReadSplit) {
...
+
+  auto array = ::arrow::json::ArrayFromJSONString(
+                   ::arrow::struct_(arrow_schema->fields()), R"([null, {}])")
+                   .ValueOrDie();
+
+  std::shared_ptr<FileIO> file_io = arrow::ArrowFileSystemFileIO::MakeMockFileIO();
+  const std::string basePath = "base.parquet";
+
+  ASSERT_THAT(WriteTable(array, {.path = basePath, .schema = schema, .io = file_io}),
+              IsError(ErrorKind::kUnknownError));
```
wgtmac: Why this is a kUnknownError?

test/parquet_test.cc
HuaHuaY: If reader doesn't have value to be read, store `nullptr` in `shared_ptr`?

wgtmac: Yes

test/parquet_test.cc
HuaHuaY: `ToErrorKind(const ::arrow::Status& status)` in `arrow_error_transform_internal.h` only maps arrow's IOError to iceberg's ErrorKind. I add a NotImplemented mapping first and will open another PR to organize errorcode mappings.

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,166 @@
...
+    std::shared_ptr<::parquet::SchemaDescriptor> schema_descriptor;
+    ICEBERG_ARROW_RETURN_NOT_OK(
+        ::parquet::arrow::ToParquetSchema(arrow_schema_.get(), *writer_properties,
+                                          *arrow_writer_properties, &schema_descriptor));
+    auto schema_node = std::static_pointer_cast<::parquet::schema::GroupNode>(
+        schema_descriptor->schema_root());
+
+    ICEBERG_ASSIGN_OR_RAISE(output_stream_, OpenOutputStream(options));
+    auto file_writer = ::parquet::ParquetFileWriter::Open(output_stream_, schema_node,
+                                                          writer_properties);
```
wgtmac:
```suggestion
    auto file_writer = ::parquet::ParquetFileWriter::Open(output_stream_, std::move(schema_node),
                                                          std::move(writer_properties));
```

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,166 @@
...
+    ICEBERG_ARROW_RETURN_NOT_OK(::parquet::arrow::FileWriter::Make(
+        pool_, std::move(file_writer), arrow_schema_, arrow_writer_properties, &writer_));
```
wgtmac:
```suggestion
        pool_, std::move(file_writer), arrow_schema_, std::move(arrow_writer_properties), &writer_));
```

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,166 @@
...
+    ICEBERG_ARROW_RETURN_NOT_OK(writer_->Close());
+    auto& metadata = writer_->metadata();
+    split_offsets_.reserve(metadata->num_row_groups());
+    for (int i = 0; i < metadata->num_row_groups(); ++i) {
+      split_offsets_.push_back(metadata->RowGroup(i)->file_offset());
+    }
+    writer_.reset();
+
+    ICEBERG_ARROW_ASSIGN_OR_RETURN(total_bytes_, output_stream_->Tell());
+    ICEBERG_ARROW_RETURN_NOT_OK(output_stream_->Close());
```
wgtmac, line 103:
```suggestion
    ICEBERG_ARROW_RETURN_NOT_OK(output_stream_->Close());
    ICEBERG_ARROW_ASSIGN_OR_RETURN(total_bytes_, output_stream_->Tell());
```

Shouldn't we close it before getting the length? It may contain an internal buffer that is not flushed yet.

test/parquet_test.cc
```diff
@@ -17,38 +17,121 @@
...
+#include "iceberg/util/macros.h"
 #include "matchers.h"
-#include "temp_file_test_base.h"
 
 namespace iceberg::parquet {
 
-class ParquetReaderTest : public TempFileTestBase {
+namespace {
+
+Status WriteTableInner(Writer& writer, std::shared_ptr<::arrow::Array> data) {
```
wgtmac:
```suggestion
Status WriteTable(Writer& writer, std::shared_ptr<::arrow::Array> data) {
```

Adding a `Inner` suffix looks weird.

wgtmac: This is not a Table actually. In the other test cases, I use Table because I've created a `::arrow::Table` instance to write to Parquet. To be precise, we should use `WriteArray` instead.

test/parquet_test.cc
```diff
@@ -17,38 +17,121 @@
...
+  ArrowArray arr;
+  ICEBERG_ARROW_RETURN_NOT_OK(::arrow::ExportArray(*data, &arr));
+  ICEBERG_RETURN_UNEXPECTED(writer.Write(arr));
+  return writer.Close();
+}
+
+Status WriteTable(std::shared_ptr<::arrow::Array> data,
```
wgtmac: ditto

test/parquet_test.cc
```diff
@@ -17,38 +17,121 @@
...
+                  const WriterOptions& writer_options) {
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto writer, WriterFactoryRegistry::Open(FileFormatType::kParquet, writer_options));
+  return WriteTableInner(*writer, data);
+}
+
+Status ReadTable(std::shared_ptr<::arrow::Array>& out,
```
wgtmac: ditto

test/parquet_test.cc
```diff
@@ -17,38 +17,121 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto read_data, reader->Next());
+
+  if (!read_data.has_value()) {
+    out = nullptr;
+    return {};
+  }
+  auto arrow_c_array = read_data.value();
+
+  ArrowSchema arrow_schema;
+  ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*reader_options.projection, &arrow_schema));
```
wgtmac: nit: you may use reader->Schema() to get it to save some typing.

test/parquet_test.cc
```diff
@@ -17,38 +17,121 @@
...
+  auto writer = std::move(writer_data.value());
+  ASSERT_THAT(WriteTableInner(*writer, data), IsOk());
+
+  ASSERT_THAT(ReadTable(out, {.path = basePath,
+                              .length = writer->length(),
+                              .io = file_io,
+                              .projection = schema}),
+              IsOk());
+
+  ASSERT_TRUE(out != nullptr) << "Reader.Next() returned no data";
```
wgtmac, line 101: Why not comparing `data` and `out` here?

test/parquet_test.cc
```diff
@@ -17,38 +17,121 @@
...
   }
 
   void CreateSimpleParquetFile() {
+    auto schema = std::make_shared<Schema>(
+        std::vector<SchemaField>{SchemaField::MakeRequired(1, "id", int32()),
+                                 SchemaField::MakeOptional(2, "name", string())});
+
+    ArrowSchema arrow_c_schema;
+    ASSERT_THAT(ToArrowSchema(*schema, &arrow_c_schema), IsOk());
+    auto arrow_schema = ::arrow::ImportType(&arrow_c_schema).ValueOrDie();
```
wgtmac, line 122: nit: use arrow::ImportSchema

test/parquet_test.cc
```diff
@@ -204,4 +288,81 @@ TEST_F(ParquetReaderTest, ReadSplit) {
...
+  });
+
+  ArrowSchema arrow_c_schema;
+  ASSERT_THAT(ToArrowSchema(*schema, &arrow_c_schema), IsOk());
+  auto arrow_schema = ::arrow::ImportType(&arrow_c_schema).ValueOrDie();
+
+  auto array =
+      ::arrow::json::ArrayFromJSONString(
+          ::arrow::struct_(arrow_schema->fields()),
+          R"([[true, 1, 2, 1.1, 1.2, "abc", 44614000, 1756696503000000], [false, 0, 0, 0, 0, "", 0, 0], [null, null, null, null, null, null, null, null]])")
```
wgtmac:
```suggestion
          R"([[true, 1, 2, 1.1, 1.2, "abc", 44614000, 1756696503000000],
              [false, 0, 0, 0, 0, "", 0, 0],
              [null, null, null, null, null, null, null, null]])")
```

test/parquet_test.cc
wgtmac: Same for below

src/iceberg/parquet/parquet_writer.cc
HuaHuaY: According to `Tell` method implementation of `MockFSOutputStream`, `Tell` should be called before `Close`.

test/parquet_test.cc
HuaHuaY: I follow arrow's `DoRoundtrip`. `out` is a reference parameter as a return value. `data` and `out` will be compared in caller.

test/parquet_test.cc
HuaHuaY: `::arrow::ImportSchema` converts a `ArrowSchema` to `std::shared_ptr<Schema>` and it doesn't seems to be helpful here.

wgtmac: My bad. I have mixed Schema with StructType.

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -0,0 +1,167 @@
...
+  // TODO(gangwu): make memory pool configurable
+  ::arrow::MemoryPool* pool_ = ::arrow::default_memory_pool();
+  // Schema to write from the Parquet file.
+  std::shared_ptr<::arrow::Schema> arrow_schema_;
+  // The output stream to write Parquet file.
+  std::shared_ptr<::arrow::io::OutputStream> output_stream_;
+  // Parquet file writer to write ArrowArray.
+  std::unique_ptr<::parquet::arrow::FileWriter> writer_;
+  // Total length of the written Parquet file.
+  int64_t total_bytes_;
```
wgtmac: `total_bytes_` should be initialized to 0 or -1.

test/parquet_test.cc
```diff
@@ -17,38 +17,120 @@
...
+#include "iceberg/util/macros.h"
 #include "matchers.h"
-#include "temp_file_test_base.h"
 
 namespace iceberg::parquet {
 
-class ParquetReaderTest : public TempFileTestBase {
+namespace {
+
+Status WriteArrayInner(Writer& writer, std::shared_ptr<::arrow::Array> data) {
```
wgtmac: Can we rename it to `WriteArray`?

test/parquet_test.cc
wgtmac: Fair enough.

PR #200, opened 2025-08-29
feat: Introduce ArrowArrayReader factory on FileScanTask

HeartLinked:
- Introduced `ArrowArrayReader` Interface: Added a new abstract interface, `ArrowArrayReader`, to represent a stream of Arrow data. This provides a unified interface for all file format readers.
- Added `FileScanTask::ToArrowArrayReader()` Factory Method: The FileScanTask now has a factory method that creates a file-format-specific reader (e.g., Parquet) and returns it as a `unique_ptr<ArrowArrayReader>`. This encapsulates the reader instantiation logic.

### DIFF
--- a/src/iceberg/arrow_array_reader.h
+++ b/src/iceberg/arrow_array_reader.h
@@ -19,29 +19,28 @@
 
 #pragma once
 
-#include "iceberg/file_reader.h"
-#include "iceberg/iceberg_bundle_export.h"
+#include <optional>
 
-namespace iceberg::parquet {
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
 
-/// \brief A reader that reads ArrowArray from Parquet files.
-class ICEBERG_BUNDLE_EXPORT ParquetReader : public Reader {
+namespace iceberg {
+
+class ICEBERG_EXPORT ArrowArrayReader {
  public:
-  ParquetReader() = default;
+  /// \brief Read next batch of data.
+  ///
+  /// \return std::nullopt if the reader has no more data, otherwise `ArrowArray`.
+  virtual Result<std::optional<ArrowArray>> Next() = 0;
 
-  ~ParquetReader() override;
+  /// \brief Get schema of data returned by `Next`.
+  virtual Result<ArrowSchema> Schema() const = 0;
 
-  Status Open(const ReaderOptions& options) final;
+  /// \brief Close this reader and release all resources.
+  virtual Status Close() = 0;
 
-  Status Close() final;
-
-  Result<std::optional<ArrowArray>> Next() final;
-
-  Result<ArrowSchema> Schema() const final;
-
- private:
-  class Impl;
-  std::unique_ptr<Impl> impl_;
+  virtual ~ArrowArrayReader() = default;
 };
 
-}  // namespace iceberg::parquet
+}  // namespace iceberg
--- a/src/iceberg/arrow_array_reader.h
+++ b/src/iceberg/arrow_array_reader.h
@@ -27,6 +27,7 @@
 
 namespace iceberg {
 
+/// \brief A reader interface that returns ArrowArray in a streaming fashion.
 class ICEBERG_EXPORT ArrowArrayReader {
  public:
   /// \brief Read next batch of data.
--- a/src/iceberg/file_reader.h
+++ b/src/iceberg/file_reader.h
@@ -26,6 +26,7 @@
 #include <memory>
 #include <optional>
 
+#include "iceberg/arrow_array_reader.h"
 #include "iceberg/arrow_c_data.h"
 #include "iceberg/file_format.h"
 #include "iceberg/result.h"
@@ -34,7 +35,7 @@
 namespace iceberg {
 
 /// \brief Base reader class to read data from different file formats.
-class ICEBERG_EXPORT Reader {
+class ICEBERG_EXPORT Reader : public ArrowArrayReader {
  public:
   virtual ~Reader() = default;
   Reader() = default;
@@ -45,15 +46,15 @@ class ICEBERG_EXPORT Reader {
   virtual Status Open(const struct ReaderOptions& options) = 0;
 
   /// \brief Close the reader.
-  virtual Status Close() = 0;
+  Status Close() override = 0;
 
   /// \brief Read next data from file.
   ///
   /// \return std::nullopt if the reader has no more data, otherwise `ArrowArray`.
-  virtual Result<std::optional<ArrowArray>> Next() = 0;
+  Result<std::optional<ArrowArray>> Next() override = 0;
 
   /// \brief Get the schema of the data.
-  virtual Result<ArrowSchema> Schema() = 0;
+  Result<ArrowSchema> Schema() const override = 0;
 };
 
 /// \brief A split of the file to read.
--- a/src/iceberg/file_reader.h
+++ b/src/iceberg/file_reader.h
@@ -37,24 +37,12 @@ namespace iceberg {
 /// \brief Base reader class to read data from different file formats.
 class ICEBERG_EXPORT Reader : public ArrowArrayReader {
  public:
-  virtual ~Reader() = default;
   Reader() = default;
   Reader(const Reader&) = delete;
   Reader& operator=(const Reader&) = delete;
 
   /// \brief Open the reader.
   virtual Status Open(const struct ReaderOptions& options) = 0;
-
-  /// \brief Close the reader.
-  Status Close() override = 0;
-
-  /// \brief Read next data from file.
-  ///
-  /// \return std::nullopt if the reader has no more data, otherwise `ArrowArray`.
-  Result<std::optional<ArrowArray>> Next() override = 0;
-
-  /// \brief Get the schema of the data.
-  Result<ArrowSchema> Schema() const override = 0;
 };
 
 /// \brief A split of the file to read.
--- a/src/iceberg/table_scan.cc
+++ b/src/iceberg/table_scan.cc
@@ -21,7 +21,11 @@
 
 #include <algorithm>
 #include <ranges>
+#include <utility>
 
+#include <iceberg/file_format.h>
+
+#include "iceberg/file_reader.h"
 #include "iceberg/manifest_entry.h"
 #include "iceberg/manifest_list.h"
 #include "iceberg/manifest_reader.h"
@@ -32,18 +36,6 @@
 #include "iceberg/util/macros.h"
 
 namespace iceberg {
-
-// implement FileScanTask
-FileScanTask::FileScanTask(std::shared_ptr<DataFile> data_file)
-    : data_file_(std::move(data_file)) {}
-
-const std::shared_ptr<DataFile>& FileScanTask::data_file() const { return data_file_; }
-
-int64_t FileScanTask::size_bytes() const { return data_file_->file_size_in_bytes; }
-
-int32_t FileScanTask::files_count() const { return 1; }
-
-int64_t FileScanTask::estimated_row_count() const { return data_file_->record_count; }
 
 TableScanBuilder::TableScanBuilder(std::shared_ptr<TableMetadata> table_metadata,
                                    std::shared_ptr<FileIO> file_io)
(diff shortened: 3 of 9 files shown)


src/iceberg/arrow_array_reader.h
```diff
@@ -0,0 +1,46 @@
...
+
+class ICEBERG_EXPORT ArrowArrayReader {
+ public:
+  /// \brief Read next batch of data.
+  ///
+  /// \return std::nullopt if the reader has no more data, otherwise `ArrowArray`.
+  virtual Result<std::optional<ArrowArray>> Next() = 0;
+
+  /// \brief Get schema of data returned by `Next`.
+  virtual Result<ArrowSchema> Schema() = 0;
```
mapleFU: should `Schema()` be const?

HeartLinked: I'll fix this.

src/iceberg/file_reader.h
```diff
@@ -34,7 +35,7 @@
 namespace iceberg {
 
 /// \brief Base reader class to read data from different file formats.
-class ICEBERG_EXPORT Reader {
+class ICEBERG_EXPORT Reader : public ArrowArrayReader {
  public:
   virtual ~Reader() = default;
```
wgtmac: We can remove this dtor now.

src/iceberg/file_reader.h
```diff
@@ -45,15 +46,15 @@ class ICEBERG_EXPORT Reader {
   virtual Status Open(const struct ReaderOptions& options) = 0;
 
   /// \brief Close the reader.
-  virtual Status Close() = 0;
```
wgtmac, line 48: We don't need to redefine this and below.

test/CMakeLists.txt
```diff
@@ -122,5 +122,6 @@ if(ICEBERG_BUILD_BUNDLE)
                    SOURCES
                    parquet_data_test.cc
                    parquet_schema_test.cc
-                   parquet_test.cc)
+                   parquet_test.cc
+                   file_scan_task_test.cc)
```
wgtmac: Add a new test for table scan?

```
  add_iceberg_test(scan_test
                   USE_BUNDLE
                   SOURCES
                   file_scan_task_test.cc)
```

test/file_scan_task_test.cc
```diff
@@ -0,0 +1,168 @@
...
+#include "iceberg/arrow/arrow_fs_file_io_internal.h"
+#include "iceberg/arrow_array_reader.h"
+#include "iceberg/file_format.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/parquet/parquet_register.h"
+#include "iceberg/schema.h"
+#include "iceberg/table_scan.h"
+#include "iceberg/type.h"
+#include "iceberg/util/checked_cast.h"
+#include "matchers.h"
```
wgtmac, line 39:
```suggestion

#include "matchers.h"
```

src/iceberg/table_scan.cc
```diff
@@ -21,7 +21,11 @@
 
 #include <algorithm>
 #include <ranges>
+#include <utility>
 
+#include <iceberg/file_format.h>
+
```
```suggestion
```

src/iceberg/table_scan.h
```diff
@@ -185,4 +169,40 @@ class ICEBERG_EXPORT DataTableScan : public TableScan {
...
+  int64_t size_bytes() const override;
+
+  /// \brief The number of files that should be read by this scan task.
+  int32_t files_count() const override;
+
+  /// \brief The number of rows that should be read by this scan task.
+  int64_t estimated_row_count() const override;
+
+  /**
+   * \brief Creates and returns an ArrowArrayReader to read the data for this task.
```
wgtmac:
```suggestion
   * \brief Returns an ArrowArrayReader to read the data for this task.
```

src/iceberg/table_scan.h
```diff
@@ -185,4 +169,40 @@ class ICEBERG_EXPORT DataTableScan : public TableScan {
...
+   *
+   * This acts as a factory to instantiate a file-format-specific reader (e.g., Parquet)
+   * based on the metadata in this task and the provided context.
+   *
+   * \param context The table scan context, used to configure the reader (e.g., with the
+   * projected schema).
+   * \param io The FileIO instance for accessing the file data.
+   * \return A Result containing a unique pointer to the reader, or an error on failure.
+   */
+  Result<std::unique_ptr<ArrowArrayReader>> ToArrowArrayReader(
```
wgtmac: We should not use `TableScanContext` here since it is a transient object used inside TableScan. Downstreams no longer have access to it. So we need to pass required parameters one by one to this function.

In this case, can you please move it back to the original location to minimize lines of change?

src/iceberg/table_scan.cc
```diff
@@ -178,4 +170,29 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
   return tasks;
 }
 
+FileScanTask::FileScanTask(std::shared_ptr<DataFile> data_file)
```
wgtmac: ditto, move them back to the original location.

HeartLinked:
@wgtmac I've updated the code to address the feedback from the review. Thanks!

test/parquet_test.cc
```diff
@@ -28,7 +28,7 @@
 #include <parquet/metadata.h>
 
 #include "iceberg/arrow/arrow_fs_file_io_internal.h"
-#include "iceberg/parquet/parquet_reader.h"
+#include "iceberg/file_reader.h"
```
wgtmac, line 31: Perhaps revert this line to minimize number of files changed in this PR?

src/iceberg/arrow_array_reader.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <optional>
+
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+class ICEBERG_EXPORT ArrowArrayReader {
```
wgtmac, line 31:
```suggestion
/// \brief A reader interface that returns ArrowArray in a streaming fashion.
class ICEBERG_EXPORT ArrowArrayReader {
```

src/iceberg/table_scan.cc
```diff
@@ -45,6 +47,21 @@ int32_t FileScanTask::files_count() const { return 1; }
...
+  const ReaderOptions options{.path = data_file_->file_path,
+                              .length = data_file_->file_size_in_bytes,
+                              .io = io,
+                              .projection = projected_schema,
+                              .filter = filter};
+
+  ICEBERG_ASSIGN_OR_RAISE(auto reader,
+                          ReaderFactoryRegistry::Open(data_file_->file_format, options));
+
+  return std::move(reader);
```
```suggestion
  return reader;
```

Is the move required?

src/iceberg/table_scan.h
```diff
@@ -50,10 +51,30 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
+
+  /// \brief The number of rows that should be read by this scan task.
   int64_t estimated_row_count() const override;
 
+  /**
+   * \brief Returns an ArrowArrayReader to read the data for this task.
+   *
+   * This acts as a factory to instantiate a file-format-specific reader (e.g., Parquet)
+   * based on the metadata in this task and the provided parameters.
+   *
```
wgtmac:
```suggestion
```

This is only the current impl. Will be changed later.

src/iceberg/table_scan.h
```diff
@@ -50,10 +51,30 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
   /// \brief The data file that should be read by this scan task.
   const std::shared_ptr<DataFile>& data_file() const;
 
+  /// \brief The total size in bytes of the file split to be read.
```
wgtmac:
```suggestion
```

We don't need to add inherited comment unless it is different.

src/iceberg/table_scan.h
```diff
@@ -50,10 +51,30 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
   int64_t size_bytes() const override;
+
+  /// \brief The number of files that should be read by this scan task.
```
wgtmac:
```suggestion
```

src/iceberg/table_scan.h
```diff
@@ -50,10 +51,30 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
   int32_t files_count() const override;
+
+  /// \brief The number of rows that should be read by this scan task.
```
wgtmac:
```suggestion
```

src/iceberg/table_scan.cc
```diff
@@ -21,7 +21,9 @@
 
 #include <algorithm>
 #include <ranges>
+#include <utility>
```
wgtmac: Why do we need this?

test/file_scan_task_test.cc
```diff
@@ -0,0 +1,166 @@
...
+};
+
+TEST_F(FileScanTaskTest, ReadFullSchema) {
+  auto data_file = std::make_shared<DataFile>();
+  data_file->file_path = temp_parquet_file_;
+  data_file->file_format = FileFormatType::kParquet;
+
+  auto io_internal = internal::checked_cast<arrow::ArrowFileSystemFileIO&>(*file_io_);
+  data_file->file_size_in_bytes =
+      io_internal.fs()->GetFileInfo(temp_parquet_file_).ValueOrDie().size();
```
wgtmac:
```suggestion
  auto io = internal::checked_cast<arrow::ArrowFileSystemFileIO&>(*file_io_);
  data_file->file_size_in_bytes =
      io.fs()->GetFileInfo(temp_parquet_file_).ValueOrDie().size();
```

test/file_scan_task_test.cc
```diff
@@ -0,0 +1,166 @@
...
+}
+
+TEST_F(FileScanTaskTest, ReadProjectedAndReorderedSchema) {
+  auto data_file = std::make_shared<DataFile>();
+  data_file->file_path = temp_parquet_file_;
+  data_file->file_format = FileFormatType::kParquet;
+
+  auto io_internal = internal::checked_cast<arrow::ArrowFileSystemFileIO&>(*file_io_);
+  data_file->file_size_in_bytes =
+      io_internal.fs()->GetFileInfo(temp_parquet_file_).ValueOrDie().size();
```
wgtmac:
```suggestion
  auto io = internal::checked_cast<arrow::ArrowFileSystemFileIO&>(*file_io_);
  data_file->file_size_in_bytes =
      io.fs()->GetFileInfo(temp_parquet_file_).ValueOrDie().size();
```

wgtmac (approved):

src/iceberg/arrow_array_reader.h
```diff
@@ -0,0 +1,47 @@
...
+#include <optional>
+
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+/// \brief A reader interface that returns ArrowArray in a streaming fashion.
+class ICEBERG_EXPORT ArrowArrayReader {
```
zeroshade, line 31: Why ArrowArray instead of RecordBatches?

src/iceberg/table_scan.cc
```diff
@@ -45,6 +43,21 @@ int32_t FileScanTask::files_count() const { return 1; }
 
 int64_t FileScanTask::estimated_row_count() const { return data_file_->record_count; }
 
+Result<std::unique_ptr<ArrowArrayReader>> FileScanTask::ToArrowArrayReader(
```
zeroshade, line 46: Shouldn't a `FileScanTask` return a stream of `RecordBatch`? Not `ArrowArray`?

zeroshade (changes requested):
Just one question really

src/iceberg/table_scan.cc
wgtmac: We have two libraries, one is `libiceberg` and the other is `libiceberg-bundle`. For `libiceberg`, it only depends on Arrow C ABI so it cannot have access to `::arrow::RecordBatch`. An alternative is to use `ArrowArrayStream` but it requires more work so I'm not sure if it worths the effort at this moment. Since the API is not stable before 1.0.0, we can change our mind at any time for any better idea.

src/iceberg/arrow_array_reader.h
wgtmac: @zeroshade Could you help review https://github.com/apache/iceberg-cpp/pull/214 which returns `ArrowArrayStream`?

PR #204, opened 2025-09-01
feat: implement AvroWriter Write method

nullccxsy:
- Complete AvroWriter::Write implementation using ExtractDatumFromArray

### DIFF
--- a/src/iceberg/avro/avro_stream_internal.h
+++ b/src/iceberg/avro/avro_stream_internal.h
@@ -87,7 +87,7 @@ class AvroOutputStream : public ::avro::OutputStream {
   /// store, if any.
   void flush() override;
 
-  std::shared_ptr<::arrow::io::OutputStream> get_output_stream() const;
+  const std::shared_ptr<::arrow::io::OutputStream>& arrow_output_stream() const;
 
  private:
   std::shared_ptr<::arrow::io::OutputStream> output_stream_;
--- a/src/iceberg/avro/avro_writer.cc
+++ b/src/iceberg/avro/avro_writer.cc
@@ -53,9 +53,6 @@ Result<std::unique_ptr<AvroOutputStream>> CreateOutputStream(const WriterOptions
 
 }  // namespace
 
-// A stateful context to keep track of the writing progress.
-struct WriteContext {};
-
 class AvroWriter::Impl {
  public:
   Status Open(const WriterOptions& options) {
@@ -70,7 +67,7 @@ class AvroWriter::Impl {
     constexpr int64_t kDefaultBufferSize = 1024 * 1024;
     ICEBERG_ASSIGN_OR_RAISE(auto output_stream,
                             CreateOutputStream(options, kDefaultBufferSize));
-    arrow_output_stream_ = output_stream->get_output_stream();
+    arrow_output_stream_ = output_stream->arrow_output_stream();
     writer_ = std::make_unique<::avro::DataFileWriter<::avro::GenericDatum>>(
         std::move(output_stream), *avro_schema_);
     datum_ = std::make_unique<::avro::GenericDatum>(*avro_schema_);
@@ -105,19 +102,19 @@ class AvroWriter::Impl {
   int64_t length() { return total_bytes_; }
 
  private:
-  int64_t total_bytes_ = 0;
   // The schema to write.
   std::shared_ptr<::iceberg::Schema> write_schema_;
   // The avro schema to write.
   std::shared_ptr<::avro::ValidSchema> avro_schema_;
+  // Arrow output stream of the Avro file to write
+  std::shared_ptr<::arrow::io::OutputStream> arrow_output_stream_;
   // The avro writer to write the data into a datum.
   std::unique_ptr<::avro::DataFileWriter<::avro::GenericDatum>> writer_;
-  // Arrow schema for data conversion (C API format)
-  ArrowSchema arrow_schema_;
-  // Reusable Avro datum for writing individual records
+  // Reusable Avro datum for writing individual records.
   std::unique_ptr<::avro::GenericDatum> datum_;
-  // Arrow output stream for writing data to filesystem
-  std::shared_ptr<::arrow::io::OutputStream> arrow_output_stream_;
+  // Arrow schema to write data.
+  ArrowSchema arrow_schema_;
+  int64_t total_bytes_ = 0;
 };
 
 AvroWriter::~AvroWriter() = default;
--- a/test/avro_test.cc
+++ b/test/avro_test.cc
@@ -135,9 +135,11 @@ class AvroReaderTest : public TempFileTestBase {
     ASSERT_TRUE(file_info_result.ok());
     ASSERT_EQ(file_info_result->size(), writer->length().value());
 
-    auto reader_result = ReaderFactoryRegistry::Open(
-        FileFormatType::kAvro,
-        {.path = temp_avro_file_, .io = file_io_, .projection = schema});
+    auto reader_result = ReaderFactoryRegistry::Open(FileFormatType::kAvro,
+                                                     {.path = temp_avro_file_,
+                                                      .length = file_info_result->size(),
+                                                      .io = file_io_,
+                                                      .projection = schema});
     ASSERT_THAT(reader_result, IsOk());
     auto reader = std::move(reader_result.value());
     ASSERT_NO_FATAL_FAILURE(VerifyNextBatch(*reader, expected_string));
--- a/src/iceberg/avro/avro_stream_internal.cc
+++ b/src/iceberg/avro/avro_stream_internal.cc
@@ -145,7 +145,8 @@ void AvroOutputStream::flush() {
   }
 }
 
-std::shared_ptr<::arrow::io::OutputStream> AvroOutputStream::get_output_stream() const {
+const std::shared_ptr<::arrow::io::OutputStream>& AvroOutputStream::arrow_output_stream()
+    const {
   return output_stream_;
 }
 


test/avro_test.cc
```diff
@@ -28,11 +27,14 @@
 #include <avro/Generic.hh>
 #include <avro/GenericDatum.hh>
 #include <gtest/gtest.h>
+#include <nanoarrow/nanoarrow.hpp>
```
wgtmac: See my comment below. We don't need nanoarrow api to create arrays.

test/avro_test.cc
```diff
@@ -105,6 +107,30 @@ class AvroReaderTest : public TempFileTestBase {
     ASSERT_FALSE(data.value().has_value());
   }
 
+  void WriteAndVerify(std::shared_ptr<Schema> schema, const std::string& expected_string,
+                      ArrowArray array) {
+    iceberg::WriterOptions options;
```
wgtmac: nit: use aggregate initialization like the way you created the reader.

test/avro_test.cc
```diff
@@ -105,6 +107,30 @@ class AvroReaderTest : public TempFileTestBase {
...
+    options.schema = schema;
+    options.path = temp_avro_file_;
+    options.io = file_io_;
+
+    auto writer_result =
+        iceberg::WriterFactoryRegistry::Open(iceberg::FileFormatType::kAvro, options);
+    ASSERT_TRUE(writer_result.has_value());
+    auto writer = std::move(writer_result).value();
```
wgtmac:
```suggestion
    auto writer = std::move(writer_result.value());
```

test/avro_test.cc
```diff
@@ -105,6 +107,30 @@ class AvroReaderTest : public TempFileTestBase {
...
 
+  void WriteAndVerify(std::shared_ptr<Schema> schema, const std::string& expected_string,
+                      ArrowArray array) {
+    iceberg::WriterOptions options;
+    options.schema = schema;
+    options.path = temp_avro_file_;
+    options.io = file_io_;
+
+    auto writer_result =
+        iceberg::WriterFactoryRegistry::Open(iceberg::FileFormatType::kAvro, options);
```
wgtmac:
```suggestion
       WriterFactoryRegistry::Open(FileFormatType::kAvro, options);
```

src/iceberg/avro/avro_writer.cc
```diff
@@ -73,9 +76,22 @@ class AvroWriter::Impl {
     return {};
   }
 
-  Status Write(ArrowArray /*data*/) {
-    // TODO(xiao.dong) convert data and write to avro
-    // total_bytes_+= written_bytes;
+  Status Write(ArrowArray data) {
+    ArrowSchema arrow_schema;
+    ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*write_schema_, &arrow_schema));
```
wgtmac: Consider caching `ArrowSchema` in the `WriterContext` to avoid recreating it every time. This is an expensive operation.

src/iceberg/avro/avro_writer.cc
```diff
@@ -73,9 +76,22 @@ class AvroWriter::Impl {
...
+    auto import_result = ::arrow::ImportArray(&data, &arrow_schema);
+    if (!import_result.ok()) {
```
wgtmac: Use `ICEBERG_ARROW_ASSIGN_OR_RETURN`

src/iceberg/avro/avro_writer.cc
```diff
@@ -73,9 +76,22 @@ class AvroWriter::Impl {
...
+      std::cout << import_result.status() << std::endl;
+      return InvalidArgument("Failed to import ArrowArray: {}",
+                             import_result.status().ToString());
+    }
+
+    auto arrow_array = import_result.ValueOrDie();
+    ::avro::GenericDatum datum(*avro_schema_);
```
wgtmac: `GenericDatum` should be cached as well.

src/iceberg/avro/avro_writer.cc
```diff
@@ -73,9 +76,22 @@ class AvroWriter::Impl {
...
 
-  Status Write(ArrowArray /*data*/) {
-    // TODO(xiao.dong) convert data and write to avro
-    // total_bytes_+= written_bytes;
+  Status Write(ArrowArray data) {
+    ArrowSchema arrow_schema;
+    ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*write_schema_, &arrow_schema));
+    auto import_result = ::arrow::ImportArray(&data, &arrow_schema);
+    if (!import_result.ok()) {
+      std::cout << import_result.status() << std::endl;
```
zhjwpku: I think this is for debug purpose, so should be deleted.

test/avro_test.cc
```diff
@@ -161,4 +187,71 @@ TEST_F(AvroReaderTest, ReadWithBatchSize) {
...
+
+  ArrowSchema struct_schema;
+  ASSERT_THAT(ToArrowSchema(*schema, &struct_schema), IsOk());
+
+  ArrowArray array;
+  NANOARROW_THROW_NOT_OK(ArrowArrayInitFromSchema(&array, &struct_schema, nullptr));
+  NANOARROW_THROW_NOT_OK(ArrowArrayStartAppending(&array));
+
+  std::vector<std::string> str_values{"Hello", "世界", "nanoarrow"};
+  std::string expected_string = R"([["Hello"], ["世界"], ["nanoarrow"]])";
```
wgtmac, line 209: You may use `::arrow::json::ArrayFromJsonString` to directly create an `::arrow::Array` and then use `::arrow::ExportArray` to transform it to `ArrowArray`. See related code from test files.

src/iceberg/avro/avro_writer.cc
```diff
@@ -73,9 +76,22 @@ class AvroWriter::Impl {
     return {};
   }
 
-  Status Write(ArrowArray /*data*/) {
-    // TODO(xiao.dong) convert data and write to avro
-    // total_bytes_+= written_bytes;
```
wgtmac, line 78: Please make sure `total_bytes_` is maintained properly and tested.

wgtmac (changes requested):

test/avro_test.cc
wgtmac: See https://github.com/apache/iceberg-cpp/pull/198/files#diff-47d6cca8c5afb86a1ac5a0d7306af8521a5cb2e9eaddd469b0b278287d6bc89a

src/iceberg/avro/avro_writer.cc
```diff
@@ -73,9 +76,22 @@ class AvroWriter::Impl {
     return {};
```
HuaHuaY, line 75: A needed change which is not caused by this PR: we need to save `output_stream` temporarily and close it in `Close` method, due to it is created in `Open`.

wgtmac: Good catch! Let's fix it in this PR too.

src/iceberg/avro/avro_writer.cc
nullccxsy: how can we total_bytes?

src/iceberg/avro/avro_writer.cc
nullccxsy: Here, the output_stream resource has been moved to writer_, so we shouldn't need to close it manually.

src/iceberg/avro/avro_writer.cc
wgtmac: You can call `AvroOutputStream::byteCount()` after closing the writer.

src/iceberg/avro/avro_writer.cc
wgtmac: The avro writer does not close the stream. We need to handle it by ourselves.

```
DataFileWriterBase::~DataFileWriterBase() {
    if (stream_) {
        try {
            close();
        } catch (...) {}
    }
}

void DataFileWriterBase::close() {
    flush();
    stream_.reset();
}
```

HuaHuaY: My mistakes in the explanation here. We need to save `output_stream_` in `AvroOutputStream` but not `AvroOutputStream` itself. `AvroOutputStream` is moved to writer_, but the `output_stream_` in it is a `shared_ptr` and it's not closed.

src/iceberg/avro/avro_stream_internal.h
```diff
@@ -87,6 +87,8 @@ class AvroOutputStream : public ::avro::OutputStream {
   /// store, if any.
   void flush() override;
 
+  std::shared_ptr<::arrow::io::OutputStream> get_output_stream() const;
```
wgtmac:
```suggestion
  const std::shared_ptr<::arrow::io::OutputStream>& arrow_output_stream() const;
```

Usually we don't use `get` as the function name. It would be good to be clear about what output stream it is from the name.

src/iceberg/avro/avro_writer.cc
```diff
@@ -67,22 +70,32 @@ class AvroWriter::Impl {
     constexpr int64_t kDefaultBufferSize = 1024 * 1024;
     ICEBERG_ASSIGN_OR_RAISE(auto output_stream,
                             CreateOutputStream(options, kDefaultBufferSize));
-
+    arrow_output_stream_ = output_stream->get_output_stream();
```
wgtmac: Could you help delete `WriterContext` and its comment at line 57 above?

src/iceberg/avro/avro_writer.cc
```diff
@@ -99,6 +112,12 @@ class AvroWriter::Impl {
   std::shared_ptr<::avro::ValidSchema> avro_schema_;
   // The avro writer to write the data into a datum.
   std::unique_ptr<::avro::DataFileWriter<::avro::GenericDatum>> writer_;
+  // Arrow schema for data conversion (C API format)
```
wgtmac:
```suggestion
  // Arrow schema to write data
```

wgtmac: BTW, could you help reorder these fields based on their creation order?

src/iceberg/avro/avro_writer.cc
```diff
@@ -99,6 +112,12 @@ class AvroWriter::Impl {
...
+  ArrowSchema arrow_schema_;
+  // Reusable Avro datum for writing individual records
+  std::unique_ptr<::avro::GenericDatum> datum_;
+  // Arrow output stream for writing data to filesystem
```
wgtmac:
```suggestion
  // Arrow output stream of the Avro file to write
```

test/avro_test.cc
```diff
@@ -105,6 +106,44 @@ class AvroReaderTest : public TempFileTestBase {
...
+    ASSERT_THAT(writer->Write(arrow_array), IsOk());
+    ASSERT_THAT(writer->Close(), IsOk());
+
+    auto file_info_result = local_fs_->GetFileInfo(temp_avro_file_);
+    ASSERT_TRUE(file_info_result.ok());
+    ASSERT_EQ(file_info_result->size(), writer->length().value());
+
+    auto reader_result = ReaderFactoryRegistry::Open(
+        FileFormatType::kAvro,
+        {.path = temp_avro_file_, .io = file_io_, .projection = schema});
```
wgtmac: Please pass the file size to the reader option.

wgtmac (commented):
Thanks, it generally looks good!

PR #206, opened 2025-09-03
feat: Implement Type Casting and toString for Literals

HeartLinked:
- Implements the complete type casting logic for `iceberg::Literal` in the `LiteralCaster` class to align with the Java reference implementation. This is critical for expression evaluation and predicate pushdown.
- Add basic implementation for fixed type.
- Updated `ToString()` to match Java's output format for better consistency (e.g., `X'...'` for binary).
- Added comprehensive unit tests to validate all new casting logic and `ToString()` formatting.

### DIFF
--- a/src/iceberg/expression/literal.cc
+++ b/src/iceberg/expression/literal.cc
@@ -446,7 +446,8 @@ std::string Literal::ToString() const {
     case TypeId::kUuid: {
       return std::get<Uuid>(value_).ToString();
     }
-    case TypeId::kBinary: {
+    case TypeId::kBinary:
+    case TypeId::kFixed: {
       const auto& binary_data = std::get<std::vector<uint8_t>>(value_);
       std::string result = "X'";
       result.reserve(/*prefix*/ 2 + /*suffix*/ 1 + /*data*/ binary_data.size() * 2);


src/iceberg/expression/literal.cc
```diff
@@ -26,6 +26,16 @@
 
 namespace iceberg {
 
+namespace {
+
+constexpr int64_t kMicrosPerDay = 86400000000LL;  // 24 * 60 * 60 * 1000 * 1000
+
+int32_t MicrosToDays(int64_t micros) {
+  return static_cast<int32_t>(std::floor(static_cast<double>(micros) / kMicrosPerDay));
```
nullccxsy: we can use `static_cast<int32_t>(micros / kMicrosPerDay))`

src/iceberg/expression/literal.cc
```diff
@@ -120,6 +169,131 @@ Result<Literal> LiteralCaster::CastFromFloat(
...
+Result<Literal> LiteralCaster::CastFromDouble(
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  auto double_val = std::get<double>(literal.value_);
+
+  switch (target_type->type_id()) {
+    case TypeId::kFloat: {
+      if (double_val > static_cast<double>(std::numeric_limits<float>::max())) {
+        return AboveMaxLiteral(target_type);
+      }
+      if (double_val < -static_cast<double>(std::numeric_limits<float>::max())) {
```
nullccxsy: `-static_cast<double>(std::numeric_limits<float>::max())` can change to 
`static_cast<double>(std::numeric_limits<float>::lowest())`

src/iceberg/expression/literal.cc
```diff
@@ -285,22 +466,28 @@ std::string Literal::ToString() const {
...
     case TypeId::kTimestamp:
     case TypeId::kTimestampTz: {
+      return std::to_string(std::get<int64_t>(value_));
+    }
+    case TypeId::kDate: {
+      return std::to_string(std::get<int32_t>(value_));
+    }
+    case TypeId::kDecimal:
+    case TypeId::kUuid: {
       throw IcebergError("Not implemented: ToString for " + type_->ToString());
```
nullccxsy: There we should `return NotImplement(....)`

src/iceberg/expression/literal.cc
nullccxsy: if (micros) < 0, we can use `static_cast<int32_t>((micros + kMicrosPerDay - 1) / kMicrosPerDay))`

HeartLinked: Good catch on the floating-point conversion. I investigated using simple integer division, but it truncates rather than floors, which fails for pre-1970 dates (e.g., -1us becomes day 0, not -1).

I've switched to using std::chrono::floor, as it's the correct, safe, and idiomatic C++ solution for this conversion.

src/iceberg/expression/literal.cc
HeartLinked: you are right. fixed.

src/iceberg/expression/literal.cc
HeartLinked: fixed.

src/iceberg/expression/literal.cc
```diff
@@ -74,6 +109,8 @@ Result<Literal> LiteralCaster::CastFromInt(
       return Literal::Float(static_cast<float>(int_val));
     case TypeId::kDouble:
       return Literal::Double(static_cast<double>(int_val));
+    case TypeId::kDate:
+      return Literal::Date(int_val);
```
wgtmac, line 106:
```suggestion
      return Literal::Date(int_val);
    // TODO(foo): support decimal
```

src/iceberg/expression/literal.cc
```diff
@@ -100,6 +136,21 @@ Result<Literal> LiteralCaster::CastFromLong(
       return Literal::Float(static_cast<float>(long_val));
     case TypeId::kDouble:
       return Literal::Double(static_cast<double>(long_val));
+    case TypeId::kDate: {
+      if (long_val > static_cast<int64_t>(std::numeric_limits<int32_t>::max())) {
+        return AboveMaxLiteral(target_type);
+      }
+      if (long_val < static_cast<int64_t>(std::numeric_limits<int32_t>::min())) {
+        return BelowMinLiteral(target_type);
+      }
```
```suggestion
      if (long_val > std::numeric_limits<int32_t>::max()) {
        return AboveMaxLiteral(target_type);
      }
      if (long_val < std::numeric_limits<int32_t>::min()) {
        return BelowMinLiteral(target_type);
      }
```

src/iceberg/expression/literal.cc
```diff
@@ -100,6 +136,21 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+      return Literal::Date(static_cast<int32_t>(long_val));
+    }
+    case TypeId::kTime:
+      return Literal::Time(long_val);
+    case TypeId::kTimestamp:
+      return Literal::Timestamp(long_val);
+    case TypeId::kTimestampTz:
+      return Literal::TimestampTz(long_val);
```
wgtmac, line 146: ditto, add TODO for decimal and perhaps TimestampNs / TimestampTzNs

src/iceberg/expression/literal.cc
```diff
@@ -109,9 +160,8 @@ Result<Literal> LiteralCaster::CastFromLong(
 Result<Literal> LiteralCaster::CastFromFloat(
     const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
   auto float_val = std::get<float>(literal.value_);
-  auto target_type_id = target_type->type_id();
 
-  switch (target_type_id) {
+  switch (target_type->type_id()) {
```
wgtmac, line 157: add TODO for decimal

src/iceberg/expression/literal.cc
```diff
@@ -120,6 +170,131 @@ Result<Literal> LiteralCaster::CastFromFloat(
...
+      return Literal::Binary(fixed_val);
+    }
+    case TypeId::kFixed: {
+      auto target_fixed_type = std::dynamic_pointer_cast<FixedType>(target_type);
+      if (fixed_val.size() == target_fixed_type->length()) {
+        return literal;
+      }
+      return NotSupported("Cannot cast Fixed({}) to Fixed({}) due to mismatched lengths",
+                          fixed_val.size(), target_fixed_type->length());
+    }
```
```suggestion
```

We don't need this. Identity cast has already been handled in the `CastTo`.

src/iceberg/expression/literal.cc
```diff
@@ -120,6 +170,131 @@ Result<Literal> LiteralCaster::CastFromFloat(
...
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  auto micros = std::get<int64_t>(literal.value_);
+
+  switch (target_type->type_id()) {
+    case TypeId::kDate: {
+      return Literal::Date(MicrosToDays(micros));
+    }
+    case TypeId::kTimestamp: {
+      return Literal::Timestamp(micros);
+    }
```
```suggestion
    case TypeId::kDate:
      return Literal::Date(MicrosToDays(micros));
    case TypeId::kTimestamp:
      return Literal::Timestamp(micros);
```

For consistency as above.

src/iceberg/expression/literal.cc
```diff
@@ -120,6 +170,131 @@ Result<Literal> LiteralCaster::CastFromFloat(
...
+                          target_type->ToString());
+  }
+}
+
+Result<Literal> LiteralCaster::CastFromString(
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  switch (target_type->type_id()) {
+    case TypeId::kDate: {
+      // TODO(Li Feiyang): Implement parsing for "YYYY-MM-DD" using std::chrono::parse
+      // once it becomes available in the target libc++.
```
```suggestion
      // TODO(Li Feiyang): Implement parsing "YYYY-MM-DD"
```

`libc++` is not a blocker so please remove that explanation. For example you can leverage https://en.cppreference.com/w/cpp/io/c/fscanf to parse the string. Same for below.

src/iceberg/expression/literal.h
```diff
@@ -71,6 +71,7 @@ class ICEBERG_EXPORT Literal {
   static Literal Double(double value);
   static Literal String(std::string value);
   static Literal Binary(std::vector<uint8_t> value);
+  static Literal Fixed(std::vector<uint8_t> value);
```
wgtmac, line 78: Finally I think `String`, `Binary` and `Fixed` should also accept std::string_view, std::span<const uint8_t> and other acceptable value types.

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +172,230 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+  std::istringstream in{str_val};
+  std::tm tm = {};
+
+  switch (target_type->type_id()) {
+    case TypeId::kDate: {
+      // Parse "YYYY-MM-DD" into days since 1970-01-01 epoch.
+      in >> std::get_time(&tm, "%Y-%m-%d");
+
+      if (in.fail() || tm.tm_mday == 0 || in.peek() != EOF) {
+        return NotSupported("Failed to parse '{}' as a valid Date (expected YYYY-MM-DD)",
```
wgtmac: `NotSupported` means that we cannot do an operation. Here we just fail to do an operation so it is better to use `InvalidArgument`.

Same for below.

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +172,230 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+
+Result<Literal> LiteralCaster::CastFromString(
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  const auto& str_val = std::get<std::string>(literal.value_);
+  std::istringstream in{str_val};
+  std::tm tm = {};
+
+  switch (target_type->type_id()) {
+    case TypeId::kDate: {
+      // Parse "YYYY-MM-DD" into days since 1970-01-01 epoch.
```
wgtmac: As discussed before, we need to move these utility functions to a new `iceberg/util/date_time_util.h` and add exhaustive test cases for them.

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +172,230 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+      in >> std::get_time(&tm, "%Y-%m-%d");
+
+      if (in.fail() || tm.tm_mday == 0 || in.peek() != EOF) {
```
```suggestion
      if (in.fail() || tm.tm_mday == 0 || !in.eof()) {
```

wgtmac: Why only checking tm.tm_mday but not other fields? Will `in.fail()` report this kind of errors?

test/literal_test.cc
```diff
@@ -134,9 +140,33 @@ TEST(LiteralTest, LongCastTo) {
   auto double_result = long_literal.CastTo(iceberg::float64());
   ASSERT_THAT(double_result, IsOk());
   EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+
+  // Cast to Date
+  auto date_result = long_literal.CastTo(iceberg::date());
+  ASSERT_THAT(date_result, IsOk());
+  EXPECT_EQ(date_result->type()->type_id(), TypeId::kDate);
+  EXPECT_EQ(date_result->ToString(), "42");
```
wgtmac: Shouldn't we directly test the value instead of converting to string?

src/iceberg/expression/literal.cc
```diff
@@ -19,13 +19,34 @@
...
 
+namespace {
+
+int32_t MicrosToDays(int64_t micros_since_epoch) {
+  std::chrono::microseconds micros(micros_since_epoch);
+  auto days_duration = std::chrono::floor<std::chrono::days>(micros);
+  return static_cast<int32_t>(days_duration.count());
+}
+
+time_t timegm_custom(std::tm* tm) {
```
wgtmac: Do you mean https://en.cppreference.com/w/cpp/chrono/c/mktime.html?

test/literal_test.cc
```diff
@@ -92,6 +92,12 @@ TEST(LiteralTest, IntCastTo) {
   auto double_result = int_literal.CastTo(iceberg::float64());
   ASSERT_THAT(double_result, IsOk());
   EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+
+  // Cast to Date
+  auto date_result = int_literal.CastTo(iceberg::date());
```
wgtmac: These four lines are repeated for a lot of times. Can we add a function or macro?

test/literal_test.cc
```diff
@@ -227,6 +290,120 @@ TEST(LiteralTest, StringComparison) {
...
+  EXPECT_EQ(result1->ToString(), "1684152000000000");
+
+  // Test with full fractional part
+  auto literal2 = Literal::String("2023-05-15T12:34:56.123456");
+  auto result2 = literal2.CastTo(iceberg::timestamp());
+  ASSERT_THAT(result2, IsOk());
+  EXPECT_EQ(result2->ToString(), "1684154096123456");
+
+  // Invalid Formats
+  auto invalid1 = Literal::String("2023-05-15 12:00:00");
```
wgtmac: Why this is an invalid format?

src/iceberg/test/literal_test.cc
```diff
@@ -254,6 +431,214 @@ TEST(LiteralTest, BinaryComparison) {
...
+
+// Date type tests
+TEST(LiteralTest, DateBasics) {
+  auto date_literal = Literal::Date(19489);  // May 15, 2023
+  auto negative_date = Literal::Date(-1);    // December 31, 1969
+
+  EXPECT_EQ(date_literal.type()->type_id(), TypeId::kDate);
+  EXPECT_EQ(negative_date.type()->type_id(), TypeId::kDate);
+
+  EXPECT_EQ(date_literal.ToString(), "19489");
```
wgtmac: Shouldn't we print it as a human readable date time string? What does the Java impl do?

test/literal_test.cc
```diff
@@ -254,6 +431,214 @@ TEST(LiteralTest, BinaryComparison) {
...
+  ASSERT_THAT(same_fixed_result, IsOk());
+  EXPECT_EQ(same_fixed_result->type()->type_id(), TypeId::kFixed);
+
+  // Cast to Fixed with different length should fail
+  auto different_fixed_result = fixed_literal.CastTo(iceberg::fixed(5));
+  EXPECT_THAT(different_fixed_result, IsError(ErrorKind::kNotSupported));
+}
+
+// Microseconds to Days conversion tests
+TEST(LiteralTest, MicrosToDaysConversion) {
```
wgtmac: Can we try to refactor these cases to use parameterized test pattern?

src/iceberg/expression/literal.cc
HeartLinked: Checking tm_mday ensures the entire format was parsed because it is the last required field, a necessary check since in.fail() often won't report an error for incomplete input, only for an explicit format mismatch.

src/iceberg/expression/literal.cc
HeartLinked: The key difference here is that std::mktime interprets the input std::tm as being in the system's local timezone, while timegm (and its Windows equivalent _mkgmtime) interprets it as UTC.
According to the Iceberg specification, timestamp strings without an explicit timezone (like "2023-05-15T12:00:00") must be interpreted as UTC. Since our parser creates a std::tm that represents a UTC time, we must use a function like timegm that performs a direct conversion without applying any local timezone offsets.

test/literal_test.cc
HeartLinked: ISO 8601 stipulates that when a date and time are combined in an expression, the capital letter T should be used as a separator. For example: 2023-05-15T12:00:00

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +172,230 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+}
+
+Result<Literal> LiteralCaster::CastFromBinary(
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  auto binary_val = std::get<std::vector<uint8_t>>(literal.value_);
+  switch (target_type->type_id()) {
+    case TypeId::kFixed: {
+      auto target_fixed_type = std::dynamic_pointer_cast<FixedType>(target_type);
+      if (binary_val.size() == target_fixed_type->length()) {
+        return Literal::Fixed(binary_val);
```
nullccxsy: There we can use `std::move(binary_val)` or use `const auto& binary_val` in line259

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +172,230 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+  }
+}
+
+Result<Literal> LiteralCaster::CastFromFixed(
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  const auto& fixed_val = std::get<std::vector<uint8_t>>(literal.value_);
+
+  switch (target_type->type_id()) {
+    case TypeId::kBinary:
+      return Literal::Binary(fixed_val);
```
nullccxsy: There we have know target_type is FixedType, so may we can use `static_cast`

src/iceberg/test/literal_test.cc
HeartLinked: I've verified the Java reference implementation, and its DateLiteral(19489).toString() also returns the string "19489". So my current implementation is consistent with the reference behavior.
```
// In org.apache.iceberg.expressions.Literal
@Override
public String toString() {
  return String.valueOf(value);
}
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +172,230 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+
+      if (in.fail()) {
+        return NotSupported(
+            "Failed to parse '{}' as a valid Time (expected HH:MM:SS.ffffff)", str_val);
+      }
+
+      int64_t total_micros =
+          (tm.tm_hour * 3600LL + tm.tm_min * 60LL + tm.tm_sec) * 1000000LL;
+
+      if (in.peek() == '.') {
```
nullccxsy: These cods is same with the line286-309, I think we can add a function to reduce duplicate code.

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +172,230 @@ Result<Literal> LiteralCaster::CastFromLong(
 Result<Literal> LiteralCaster::CastFromFloat(
     const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
   auto float_val = std::get<float>(literal.value_);
-  auto target_type_id = target_type->type_id();
 
-  switch (target_type_id) {
+  switch (target_type->type_id()) {
```
nullccxsy, line 157: Why there don't cast to `int`?

src/iceberg/expression/literal.cc
```diff
@@ -343,10 +632,18 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
-    case TypeId::kBoolean:
+      return CastFromDouble(literal, target_type);
     case TypeId::kString:
+      return CastFromString(literal, target_type);
     case TypeId::kBinary:
-      break;
+      return CastFromBinary(literal, target_type);
+    case TypeId::kFixed:
+      return CastFromFixed(literal, target_type);
+    case TypeId::kTimestamp:
```
nullccxsy, line 518: Why there don't have `CastFromTime()`?

src/iceberg/expression/literal.cc
```diff
@@ -285,23 +568,29 @@ std::string Literal::ToString() const {
...
     }
-    case TypeId::kBinary: {
+    case TypeId::kBinary:
+    case TypeId::kFixed: {
       const auto& binary_data = std::get<std::vector<uint8_t>>(value_);
-      std::string result;
-      result.reserve(binary_data.size() * 2);  // 2 chars per byte
+      std::string result = "X'";
+      result.reserve(2 + binary_data.size() * 2 +
+                     1);  // 2 chars per byte and 2 + 1 for prefix and suffix
```
```suggestion
      result.reserve(/*prefix*/2 + binary_data.size() * 2 + /*suffix*/1);
```

src/iceberg/expression/literal.cc
HeartLinked: > Why there don't cast to `int`?

In java implementation, float cast to double/ decimal, while decimal not implemented in our project.
```
  static class FloatLiteral extends ComparableLiteral<Float> {
    FloatLiteral(Float value) {
      super(value);
    }

    @Override
    @SuppressWarnings("unchecked")
    public <T> Literal<T> to(Type type) {
      switch (type.typeId()) {
        case FLOAT:
          return (Literal<T>) this;
        case DOUBLE:
          return (Literal<T>) new DoubleLiteral(value().doubleValue());
        case DECIMAL:
          int scale = ((Types.DecimalType) type).scale();
          return (Literal<T>)
              new DecimalLiteral(BigDecimal.valueOf(value()).setScale(scale, RoundingMode.HALF_UP));
        default:
          return null;
      }
    }
```

src/iceberg/expression/literal.cc
HeartLinked: Please check the java implementation, there isn't any cast from TimeLiteral to other Literal type in java version.

src/iceberg/expression/literal.h
```diff
@@ -71,6 +74,11 @@ class ICEBERG_EXPORT Literal {
...
+
+  /// \brief Factory methods for decimal type
+  static Literal Decimal(const Decimal& value, int32_t precision, int32_t scale);
+  static Literal Decimal(int128_t value, int32_t precision, int32_t scale);
```
```suggestion
  static Literal Decimal(int128_t value, int32_t precision, int32_t scale);
```

- Decimal type is also a primitive type so let's put them together without blank line.
- I think `int128_t` is enough and it duplicates with `Decimal`

src/iceberg/expression/literal.cc
```diff
@@ -74,6 +101,13 @@ Result<Literal> LiteralCaster::CastFromInt(
       return Literal::Float(static_cast<float>(int_val));
     case TypeId::kDouble:
       return Literal::Double(static_cast<double>(int_val));
+    case TypeId::kDate:
+      return Literal::Date(int_val);
+    case TypeId::kDecimal: {
+      auto decimal_type = std::static_pointer_cast<DecimalType>(target_type);
```
```suggestion
      auto decimal_type = internal::checked_pointer_cast<DecimalType>(target_type);
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +162,151 @@ Result<Literal> LiteralCaster::CastFromLong(
...
     const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
   auto float_val = std::get<float>(literal.value_);
-  auto target_type_id = target_type->type_id();
 
-  switch (target_type_id) {
+  switch (target_type->type_id()) {
     case TypeId::kDouble:
       return Literal::Double(static_cast<double>(float_val));
+    case TypeId::kDecimal: {
+      auto decimal_type = std::static_pointer_cast<DecimalType>(target_type);
```
wgtmac: ditto

src/iceberg/expression/literal.cc
```diff
@@ -100,6 +133,26 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+      return Literal::Date(static_cast<int32_t>(long_val));
+    }
+    case TypeId::kTime:
+      return Literal::Time(long_val);
+    case TypeId::kTimestamp:
+      return Literal::Timestamp(long_val);
+    case TypeId::kTimestampTz:
+      return Literal::TimestampTz(long_val);
+    case TypeId::kDecimal: {
+      auto decimal_type = std::static_pointer_cast<DecimalType>(target_type);
```
```suggestion
      auto decimal_type = internal::checked_pointer_cast<DecimalType>(target_type);
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +162,151 @@ Result<Literal> LiteralCaster::CastFromLong(
...
     case TypeId::kDouble:
       return Literal::Double(static_cast<double>(float_val));
+    case TypeId::kDecimal: {
+      auto decimal_type = std::static_pointer_cast<DecimalType>(target_type);
+      std::string float_str = std::to_string(float_val);
+
+      iceberg::Decimal parsed_val;
+      int32_t parsed_scale = 0;
+      ICEBERG_ASSIGN_OR_RAISE(
+          parsed_val, iceberg::Decimal::FromString(float_str, nullptr, &parsed_scale));
```
```suggestion
          parsed_val, iceberg::Decimal::FromString(float_str, /*precision=*/nullptr, &parsed_scale));
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +162,151 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+      if (double_val > static_cast<double>(std::numeric_limits<float>::max())) {
+        return AboveMaxLiteral(target_type);
+      }
+      if (double_val < static_cast<double>(std::numeric_limits<float>::lowest())) {
+        return BelowMinLiteral(target_type);
+      }
+      return Literal::Float(static_cast<float>(double_val));
+    }
+    case TypeId::kDecimal: {
+      auto decimal_type = std::static_pointer_cast<DecimalType>(target_type);
```
```suggestion
      auto decimal_type = internal::checked_pointer_cast<DecimalType>(target_type);
```

src/iceberg/expression/literal.cc
```suggestion
      int32_t parsed_scale = 0;
      ICEBERG_ASSIGN_OR_RAISE(
          auto parsed_val, iceberg::Decimal::FromString(float_str, nullptr, &parsed_scale));
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +162,151 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+      std::string float_str = std::to_string(float_val);
+
+      iceberg::Decimal parsed_val;
+      int32_t parsed_scale = 0;
+      ICEBERG_ASSIGN_OR_RAISE(
+          parsed_val, iceberg::Decimal::FromString(float_str, nullptr, &parsed_scale));
+      iceberg::Decimal rescaled_val;
+      ICEBERG_ASSIGN_OR_RAISE(rescaled_val,
+                              parsed_val.Rescale(parsed_scale, decimal_type->scale()));
```
```suggestion
      ICEBERG_ASSIGN_OR_RAISE(auto rescaled_val,
                              parsed_val.Rescale(parsed_scale, decimal_type->scale()));
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +162,151 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+      auto decimal_type = std::static_pointer_cast<DecimalType>(target_type);
+      std::string double_str = std::to_string(double_val);
+
+      iceberg::Decimal parsed_val;
+      int32_t parsed_scale = 0;
+      ICEBERG_ASSIGN_OR_RAISE(
+          parsed_val, iceberg::Decimal::FromString(double_str, nullptr, &parsed_scale));
+      iceberg::Decimal rescaled_val;
+      ICEBERG_ASSIGN_OR_RAISE(rescaled_val,
+                              parsed_val.Rescale(parsed_scale, decimal_type->scale()));
```
```suggestion
      int32_t parsed_scale = 0;
      ICEBERG_ASSIGN_OR_RAISE(
          auto parsed_val, iceberg::Decimal::FromString(double_str, /*precision=*/nullptr, &parsed_scale));
      iceberg::Decimal rescaled_val;
      ICEBERG_ASSIGN_OR_RAISE(rescaled_val,
                              parsed_val.Rescale(parsed_scale, decimal_type->scale()));
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +162,151 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+    case TypeId::kTime:
+    case TypeId::kTimestamp:
+    case TypeId::kTimestampTz:
+    case TypeId::kUuid:
+      return NotImplemented("Cast from String to {} is not implemented yet",
+                            target_type->ToString());
+    case TypeId::kDecimal: {
+      auto decimal_type = std::static_pointer_cast<DecimalType>(target_type);
+      iceberg::Decimal dec_val;
+      ICEBERG_ASSIGN_OR_RAISE(dec_val, iceberg::Decimal::FromString(str_val));
```
```suggestion
      ICEBERG_ASSIGN_OR_RAISE(auto dec_val, iceberg::Decimal::FromString(str_val));
```

src/iceberg/expression/literal.cc
```diff
@@ -109,17 +162,151 @@ Result<Literal> LiteralCaster::CastFromLong(
...
+                          target_type->ToString());
+  }
+}
+
+Result<Literal> LiteralCaster::CastFromBinary(
+    const Literal& literal, const std::shared_ptr<PrimitiveType>& target_type) {
+  auto binary_val = std::get<std::vector<uint8_t>>(literal.value_);
+  switch (target_type->type_id()) {
+    case TypeId::kFixed: {
+      auto target_fixed_type = std::static_pointer_cast<FixedType>(target_type);
```
```suggestion
      auto target_fixed_type = internal::checked_pointer_cast<FixedType>(target_type);
```

src/iceberg/expression/literal.cc
```diff
@@ -243,19 +445,33 @@ std::partial_ordering Literal::operator<=>(const Literal& other) const {
...
   }
 }
 
 std::string Literal::ToString() const {
+  auto unsupported_error = [this]() {
+    return std::format("ToString not supported for type: {}", type_->ToString());
+  };
+  auto invalid_argument = [this]() {
+    return std::format("Invalid argument for type: {}", type_->ToString());
+  };
```
```suggestion
  auto invalid_literal = [this]() {
    return std::format("Invalid literal of type: {}", type_->ToString());
  };
```

This function will be called in printing an expression so it is better to print a consistent short message.

src/iceberg/expression/literal.cc
```diff
@@ -283,28 +499,42 @@ std::string Literal::ToString() const {
...
+      const auto unscaled_value = std::get<int128_t>(value_);
+      auto decimal_type = std::static_pointer_cast<DecimalType>(type_);
+      int32_t scale = decimal_type->scale();
+
+      iceberg::Decimal decimal_val(unscaled_value);
+      Result<std::string> str_res = decimal_val.ToString(scale);
+      if (str_res.has_value()) {
+        return str_res.value();
+      }
+      return invalid_argument();
```
wgtmac: We can use std::expected.value_or to simplify this.

src/iceberg/expression/literal.h
```diff
@@ -71,6 +74,11 @@ class ICEBERG_EXPORT Literal {
   static Literal Double(double value);
   static Literal String(std::string value);
   static Literal Binary(std::vector<uint8_t> value);
+  static Literal Fixed(std::vector<uint8_t> value);
+
+  /// \brief Factory methods for decimal type
+  static Literal Decimal(const Decimal& value, int32_t precision, int32_t scale);
```
HuaHuaY: Will it be better to use `Decimal` in `Value` instead of `int128_t` and remove the second overload function of `Decimal`? We already have the wrapper class `Decimal` and it has clearer semantics.

And I think it's better to use this signature `Decimal(Decimal value, int32_t precision, int32_t scale)` instead.

test/literal_test.cc
```diff
@@ -30,6 +30,22 @@
...
+
+// Helper function to assert that a CastTo operation succeeds and checks
+// the resulting type and value.
+template <typename T>
+void AssertCastSucceeds(const Result<Literal>& result, TypeId expected_type_id,
+                        const T& expected_value) {
+  ASSERT_THAT(result, IsOk());
+  EXPECT_EQ(result->type()->type_id(), expected_type_id);
+  ASSERT_NO_THROW(EXPECT_EQ(std::get<T>(result->value()), expected_value))
+      << "Value type mismatch in std::get. Expected type for TypeId "
```
HuaHuaY: value mismatch here?

src/iceberg/expression/literal.h
HeartLinked: I don't understand what you're saying. Both of these two methods were just added by me. The reason I added them is that I noticed our wrapper class, Decimal, has clearer semantics. These two methods not only support the complete Decimal type but also are intended to conveniently support the original int128_t type. So, is your comment suggesting that the constructor using int128_t is unnecessary, we only need one new constructor using Decimal?

test/literal_test.cc
```diff
@@ -120,23 +137,35 @@ TEST(LiteralTest, LongCastTo) {
...
-  ASSERT_THAT(double_result, IsOk());
-  EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+  AssertCastSucceeds(long_literal.CastTo(iceberg::float64()), TypeId::kDouble, 42.0);
+
+  // Cast to Date
+  AssertCastSucceeds(long_literal.CastTo(iceberg::date()), TypeId::kDate, 42);
+
+  // Cast to Time
+  AssertCastSucceeds(long_literal.CastTo(iceberg::time()), TypeId::kTime,
+                     static_cast<int64_t>(42L));
```
HuaHuaY: `static_cast<int64_t>(42)` or `42LL`?

src/iceberg/expression/literal.h
HuaHuaY: Yes. Only reserve `Decimal(Decimal value, int32_t precision, int32_t scale)` and define `Value` like this:
```cpp
  using Value = std::variant<..., 
                             Decimal,                 // for decimal
                             ...>;
```

HeartLinked: I've removed decimal related logic. This will be handled in the following PRs like #238 .

src/iceberg/test/literal_test.cc
```diff
@@ -263,22 +457,18 @@ TEST(LiteralTest, CrossTypeComparison) {
   EXPECT_EQ(int_literal <=> string_literal, std::partial_ordering::unordered);
 }
 
-// Special value tests
-TEST(LiteralTest, SpecialValues) {
+// Same type cast tests
+TEST(LiteralTest, SameTypeCast) {
```
wgtmac: As discussed offline, please refactor this file to use parameterized test to use common code for testing different data types.

src/iceberg/expression/literal.cc
```diff
@@ -283,28 +433,29 @@ std::string Literal::ToString() const {
...
     }
-    case TypeId::kDecimal:
-    case TypeId::kUuid:
-    case TypeId::kFixed:
-    case TypeId::kDate:
     case TypeId::kTime:
     case TypeId::kTimestamp:
     case TypeId::kTimestampTz: {
-      throw IcebergError("Not implemented: ToString for " + type_->ToString());
+      return std::to_string(std::get<int64_t>(value_));
```
wgtmac, line 463: I think these switch cases can be easily rewritten by `return std::to_string(std::get<typename LiteralTraits<type_id>::ValueType>(value_));` once https://github.com/apache/iceberg-cpp/pull/185 is merged.

src/iceberg/expression/literal.cc
```diff
@@ -256,6 +402,10 @@ std::partial_ordering Literal::operator<=>(const Literal& other) const {
 }
 
 std::string Literal::ToString() const {
+  auto unsupported_error = [this]() {
+    return std::format("ToString not supported for type: {}", type_->ToString());
+  };
+
```
```suggestion
```

src/iceberg/expression/literal.cc
```diff
@@ -283,28 +433,29 @@ std::string Literal::ToString() const {
...
     case TypeId::kTimestampTz: {
-      throw IcebergError("Not implemented: ToString for " + type_->ToString());
+      return std::to_string(std::get<int64_t>(value_));
+    }
+    case TypeId::kDate: {
+      return std::to_string(std::get<int32_t>(value_));
     }
     default: {
-      throw IcebergError("Unknown type: " + type_->ToString());
+      return unsupported_error();
```
```suggestion
      return std::format("invalid literal of type {}", type_->ToString());
```

src/iceberg/expression/literal.h
```diff
@@ -27,6 +27,8 @@
 
 #include "iceberg/result.h"
 #include "iceberg/type.h"
+#include "iceberg/util/decimal.h"
+#include "iceberg/util/int128.h"
```
wgtmac: Revert these related lines?

src/iceberg/expression/literal.h
```diff
@@ -56,7 +58,7 @@ class ICEBERG_EXPORT Literal {
                              double,          // for double
                              std::string,     // for string
                              std::vector<uint8_t>,     // for binary, fixed
-                             std::array<uint8_t, 16>,  // for uuid and decimal
+                             std::array<uint8_t, 16>,  // for uuid
```
wgtmac: ditto

src/iceberg/transform_function.cc
```diff
@@ -27,6 +27,7 @@
 
 #include "iceberg/expression/literal.h"
 #include "iceberg/type.h"
+#include "iceberg/util/int128.h"
```
wgtmac: ditto

src/iceberg/test/literal_test.cc
```diff
@@ -30,6 +30,22 @@
...
 
+namespace {
+
+// Helper function to assert that a CastTo operation succeeds and checks
+// the resulting type and value.
+template <typename T>
+void AssertCastSucceeds(const Result<Literal>& result, TypeId expected_type_id,
+                        const T& expected_value) {
+  ASSERT_THAT(result, IsOk());
+  EXPECT_EQ(result->type()->type_id(), expected_type_id);
```
wgtmac:
```suggestion
  ASSERT_EQ(result->type()->type_id(), expected_type_id);
```

src/iceberg/test/literal_test.cc
```diff
@@ -30,6 +30,22 @@
...
+  ASSERT_NO_THROW(EXPECT_EQ(std::get<T>(result->value()), expected_value))
```
wgtmac:
```suggestion
  EXPECT_EQ(std::get<T>(result->value()), expected_value)
```

Why adding `ASSERT_NO_THROW` here if we already have `EXPECT_EQ`?

src/iceberg/test/literal_test.cc
```diff
@@ -120,23 +133,31 @@ TEST(LiteralTest, LongCastTo) {
...
-  ASSERT_THAT(double_result, IsOk());
-  EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+  AssertCastSucceeds(long_literal.CastTo(iceberg::float64()), TypeId::kDouble, 42.0);
+
+  // Cast to Date
+  AssertCastSucceeds(long_literal.CastTo(iceberg::date()), TypeId::kDate, 42);
+
+  // Cast to Time
+  AssertCastSucceeds(long_literal.CastTo(iceberg::time()), TypeId::kTime,
+                     static_cast<int64_t>(42));
```
wgtmac:
```suggestion
                     int64_t{42});
```

Same for below

src/iceberg/test/literal_test.cc
```diff
@@ -150,6 +171,14 @@ TEST(LiteralTest, LongCastToIntOverflow) {
   auto min_result = min_long.CastTo(iceberg::int32());
   ASSERT_THAT(min_result, IsOk());
   EXPECT_TRUE(min_result->IsBelowMin());
+
+  max_result = max_long.CastTo(iceberg::date());
```
wgtmac:
```suggestion
  max_result = max_long.CastTo(date());
```

src/iceberg/test/literal_test.cc
```diff
@@ -175,12 +204,11 @@ TEST(LiteralTest, FloatComparison) {
...
 TEST(LiteralTest, FloatCastTo) {
-  auto float_literal = Literal::Float(3.14f);
+  auto float_literal = Literal::Float(2.0f);
 
   // Cast to Double
-  auto double_result = float_literal.CastTo(iceberg::float64());
-  ASSERT_THAT(double_result, IsOk());
-  EXPECT_EQ(double_result->type()->type_id(), TypeId::kDouble);
+  AssertCastSucceeds(float_literal.CastTo(iceberg::float64()), TypeId::kDouble,
+                     static_cast<double>(2.0f));
```
wgtmac:
```suggestion
                     2.0);
```

src/iceberg/test/literal_test.cc
```diff
@@ -205,6 +233,29 @@ TEST(LiteralTest, DoubleComparison) {
...
+}
+
+TEST(LiteralTest, DoubleCastToOverflow) {
+  // Test overflow cases for Double to Float
+  auto max_double =
+      Literal::Double(static_cast<double>(std::numeric_limits<float>::max()) * 2);
+  auto min_double =
+      Literal::Double(-static_cast<double>(std::numeric_limits<float>::max()) * 2);
+
+  auto max_result = max_double.CastTo(iceberg::float32());
```
wgtmac:
```suggestion
  auto max_result = max_double.CastTo(float32());
```

src/iceberg/test/literal_test.cc
```diff
@@ -254,6 +305,149 @@ TEST(LiteralTest, BinaryComparison) {
   EXPECT_EQ(binary2 <=> binary1, std::partial_ordering::greater);
 }
 
+TEST(LiteralTest, BinaryCastTo) {
+  std::vector<uint8_t> data4 = {0x01, 0x02, 0x03, 0x04};
+  auto binary_literal = Literal::Binary(data4);
+
+  // Cast to Fixed with matching length
+  AssertCastSucceeds(binary_literal.CastTo(iceberg::fixed(4)), TypeId::kFixed, data4);
```
wgtmac:
```suggestion
  AssertCastSucceeds(binary_literal.CastTo(fixed(4)), TypeId::kFixed, data4);
```

src/iceberg/test/literal_test.cc
```diff
@@ -254,6 +305,149 @@ TEST(LiteralTest, BinaryComparison) {
...
+
+  // Cast to Fixed with different length should fail
+  EXPECT_THAT(binary_literal.CastTo(iceberg::fixed(5)),
```
wgtmac:
```suggestion
  EXPECT_THAT(binary_literal.CastTo(fixed(5)),
```

src/iceberg/test/literal_test.cc
```diff
@@ -30,257 +30,147 @@
...
-  EXPECT_EQ(false_literal.type()->type_id(), TypeId::kBoolean);
-
-  EXPECT_EQ(true_literal.ToString(), "true");
-  EXPECT_EQ(false_literal.ToString(), "false");
+namespace {
+
+// Helper function to assert that a CastTo operation succeeds and checks
+// the resulting type and value.
+template <typename T>
+void AssertCastSucceeds(const Result<Literal>& result, TypeId expected_type_id,
```
wgtmac: Let's remove it since not used anywhere.

PR #207, opened 2025-09-03
feat: implement schema selection and projection methods

nullccxsy:
- Added select and project methods to the Schema class for creating projection schemas based on specified field names or IDs.
- Introduced PruneColumnVisitor to handle the logic for selecting and projecting fields, including support for nested structures.

### DIFF
--- a/src/iceberg/schema.cc
+++ b/src/iceberg/schema.cc
@@ -257,4 +257,259 @@ void NameToIdVisitor::Finish() {
   }
 }
 
+class PruneColumnVisitor {
+ public:
+  explicit PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                              bool select_full_types = false);
+  Status Visit(const ListType& type);
+  Status Visit(const MapType& type);
+  Status Visit(const StructType& type);
+  Status Visit(const PrimitiveType& type);
+  std::shared_ptr<const Type> GetResult() const;
+  void SetResult(std::shared_ptr<const Type> result);
+  Status ProjectList(const SchemaField& element,
+                     std::shared_ptr<const Type> element_result);
+  Status ProjectMap(const SchemaField& key_field, const SchemaField& value_field,
+                    std::shared_ptr<const Type> value_result);
+
+ private:
+  const std::unordered_set<int32_t>& selected_ids_;
+  bool select_full_types_;
+  std::shared_ptr<const Type> result_;
+};
+
+Result<std::shared_ptr<const Schema>> Schema::select(
+    const std::vector<std::string>& names, bool case_sensitive) const {
+  return internalSelect(names, case_sensitive);
+}
+
+Result<std::shared_ptr<const Schema>> Schema::select(
+    const std::initializer_list<std::string>& names, bool case_sensitive) const {
+  return internalSelect(std::vector<std::string>(names), case_sensitive);
+}
+
+Result<std::shared_ptr<const Schema>> Schema::internalSelect(
+    const std::vector<std::string>& names, bool case_sensitive) const {
+  const std::string ALL_COLUMNS = "*";
+  if (std::ranges::find(names, ALL_COLUMNS) != names.end()) {
+    return shared_from_this();
+  }
+
+  std::unordered_set<int32_t> selected_ids;
+  for (const auto& name : names) {
+    ICEBERG_ASSIGN_OR_RAISE(auto result, FindFieldByName(name, case_sensitive));
+    if (result.has_value()) {
+      selected_ids.insert(result.value().get().field_id());
+    }
+  }
+
+  PruneColumnVisitor visitor(selected_ids, /*select_full_types=*/true);
+  ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*this, &visitor));
+
+  auto projected_type = visitor.GetResult();
+  if (!projected_type) {
+    return std::make_shared<Schema>(std::vector<SchemaField>{}, schema_id_);
+  }
+
+  if (projected_type->type_id() != TypeId::kStruct) {
+    return InvalidSchema("Projected type must be a struct type");
+  }
+
+  const auto& projected_struct =
+      internal::checked_cast<const StructType&>(*projected_type);
+
+  std::vector<SchemaField> fields_vec(projected_struct.fields().begin(),
+                                      projected_struct.fields().end());
+  return std::make_shared<Schema>(std::move(fields_vec), schema_id_);
+}
+
+Result<std::shared_ptr<const Schema>> Schema::project(
+    std::unordered_set<int32_t>& selected_ids) const {
+  PruneColumnVisitor visitor(selected_ids, /*select_full_types=*/false);
+  ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*this, &visitor));
+
+  auto projected_type = visitor.GetResult();
+  if (!projected_type) {
+    return std::make_shared<Schema>(std::vector<SchemaField>{}, schema_id_);
+  }
+
+  if (projected_type->type_id() != TypeId::kStruct) {
+    return InvalidSchema("Projected type must be a struct type");
+  }
+
+  const auto& projected_struct =
+      internal::checked_cast<const StructType&>(*projected_type);
+  std::vector<SchemaField> fields_vec(projected_struct.fields().begin(),
+                                      projected_struct.fields().end());
+  return std::make_shared<Schema>(std::move(fields_vec), schema_id_);
+}
+
+PruneColumnVisitor::PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                                       bool select_full_types)
+    : selected_ids_(selected_ids), select_full_types_(select_full_types) {}
+
+std::shared_ptr<const Type> PruneColumnVisitor::GetResult() const { return result_; }
+
+void PruneColumnVisitor::SetResult(std::shared_ptr<const Type> result) {
+  result_ = std::move(result);
+}
+
+Status PruneColumnVisitor::Visit(const StructType& type) {
+  std::vector<std::shared_ptr<const Type>> selected_types;
+  const auto& fields = type.fields();
+  for (const auto& field : fields) {
+    PruneColumnVisitor field_visitor(selected_ids_, select_full_types_);
+    ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*field.type(), &field_visitor));
+    auto result = field_visitor.GetResult();
+    if (selected_ids_.contains(field.field_id())) {
+      // select
+      if (select_full_types_) {
+        selected_types.emplace_back(field.type());
+      } else if (field.type()->type_id() == TypeId::kStruct) {
+        // project(kstruct)
+        if (!result) {
+          result = std::make_shared<StructType>(std::vector<SchemaField>{});
+        }
+        selected_types.emplace_back(std::move(result));
+      } else {
+        // project(list, map, primitive)
+        if (!field.type()->is_primitive()) {
+          return InvalidArgument(
+              "Cannot explicitly project List or Map types, {}:{} of type {} was "
+              "selected",
+              field.field_id(), field.name(), field.type()->ToString());
+        }
+        selected_types.emplace_back(field.type());
+      }
+    } else if (result) {
+      // project, select
+      selected_types.emplace_back(std::move(result));
+    } else {
+      // project, select
+      selected_types.emplace_back(nullptr);
+    }
+  }
+
+  bool same_types = true;
+  std::vector<SchemaField> selected_fields;
+  for (size_t i = 0; i < fields.size(); i++) {
+    if (fields[i].type() == selected_types[i]) {
+      selected_fields.emplace_back(fields[i]);
+    } else if (selected_types[i]) {
+      same_types = false;
+      selected_fields.emplace_back(fields[i].field_id(), std::string(fields[i].name()),
+                                   std::const_pointer_cast<Type>(selected_types[i]),
+                                   fields[i].optional(), std::string(fields[i].doc()));
+    }
+  }
+
+  if (!selected_fields.empty()) {
+    if (selected_fields.size() == fields.size() && same_types) {
+      result_ = std::make_shared<StructType>(type);
+    } else {
+      result_ = std::make_shared<StructType>(std::move(selected_fields));
+    }
+  }
+
+  return {};
+}
+
+Status PruneColumnVisitor::Visit(const ListType& type) {
+  const auto& element_field = type.fields()[0];
+
+  PruneColumnVisitor element_visitor(selected_ids_, select_full_types_);
+  ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*element_field.type(), &element_visitor));
+
+  auto element_result = element_visitor.GetResult();
+
+  if (selected_ids_.contains(element_field.field_id())) {
+    if (select_full_types_) {
+      result_ = std::make_shared<ListType>(element_field);
+    } else if (element_field.type()->type_id() == TypeId::kStruct) {
+      ICEBERG_RETURN_UNEXPECTED(ProjectList(element_field, element_result));
+    } else {
+      if (!element_field.type()->is_primitive()) {
+        return InvalidArgument(
+            "Cannot explicitly project List or Map types, List element {} of type {} was "
+            "selected",
+            element_field.field_id(), element_field.name());
+      }
+      result_ = std::make_shared<ListType>(element_field);
+    }
+  } else if (element_result) {
+    ICEBERG_RETURN_UNEXPECTED(ProjectList(element_field, element_result));
+  }
+
+  return {};
+}
+
+Status PruneColumnVisitor::Visit(const MapType& type) {
+  const auto& key_field = type.fields()[0];
+  const auto& value_field = type.fields()[1];
+
+  PruneColumnVisitor key_visitor(selected_ids_, select_full_types_);
+  ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*key_field.type(), &key_visitor));
+  auto key_result = key_visitor.GetResult();
+
+  PruneColumnVisitor value_visitor(selected_ids_, select_full_types_);
+  ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*value_field.type(), &value_visitor));
+  auto value_result = value_visitor.GetResult();
+
+  if (selected_ids_.contains(value_field.field_id())) {
+    if (select_full_types_) {
+      result_ = std::make_shared<MapType>(type);
+    } else if (value_field.type()->type_id() == TypeId::kStruct) {
+      ICEBERG_RETURN_UNEXPECTED(ProjectMap(key_field, value_field, value_result));
+    } else {
+      if (!value_field.type()->is_primitive()) {
+        return InvalidArgument(
+            "Cannot explicitly project List or Map types, Map value {} of type {} was "
+            "selected",
+            value_field.field_id(), type.ToString());
+      }
+      result_ = std::make_shared<MapType>(type);
+    }
+  } else if (value_result) {
+    ICEBERG_RETURN_UNEXPECTED(ProjectMap(key_field, value_field, value_result));
+  } else if (selected_ids_.contains(key_field.field_id())) {
+    result_ = std::make_shared<MapType>(type);
+  }
+
+  return {};
+}
+
+Status PruneColumnVisitor::Visit(const PrimitiveType& type) { return {}; }
+
+Status PruneColumnVisitor::ProjectList(const SchemaField& element_field,
+                                       std::shared_ptr<const Type> element_result) {
+  if (!element_result) {
+    return InvalidArgument("Cannot project a list when the element result is null");
+  }
+  if (element_field.type() == element_result) {
+    result_ = std::make_shared<ListType>(element_field);
+  } else {
+    result_ = std::make_shared<ListType>(element_field.field_id(),
+                                         std::const_pointer_cast<Type>(element_result),
+                                         element_field.optional());
+  }
+  return {};
+}
+
+Status PruneColumnVisitor::ProjectMap(const SchemaField& key_field,
+                                      const SchemaField& value_field,
+                                      std::shared_ptr<const Type> value_result) {
+  if (!value_result) {
+    return InvalidArgument("Attempted to project a map without a defined map value type");
+  }
+  if (value_field.type() == value_result) {
+    result_ = std::make_shared<MapType>(key_field, value_field);
+  } else {
+    result_ = std::make_shared<MapType>(
+        key_field,
+        SchemaField(value_field.field_id(), std::string(value_field.name()),
+                    std::const_pointer_cast<Type>(value_result), value_field.optional()));
+  }
+  return {};
+}
+
 }  // namespace iceberg
--- a/src/iceberg/schema.h
+++ b/src/iceberg/schema.h
@@ -26,6 +26,7 @@
 #include <cstdint>
 #include <optional>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 #include "iceberg/iceberg_export.h"
@@ -41,7 +42,8 @@ namespace iceberg {
 /// A schema is a list of typed columns, along with a unique integer ID.  A
 /// Table may have different schemas over its lifetime due to schema
 /// evolution.
-class ICEBERG_EXPORT Schema : public StructType {
+class ICEBERG_EXPORT Schema : public StructType,
+                              public std::enable_shared_from_this<Schema> {
  public:
   static constexpr int32_t kInitialSchemaId = 0;
 
@@ -52,9 +54,9 @@ class ICEBERG_EXPORT Schema : public StructType {
   ///
   /// A schema is identified by a unique ID for the purposes of schema
   /// evolution.
-  [[nodiscard]] std::optional<int32_t> schema_id() const;
+  std::optional<int32_t> schema_id() const;
 
-  [[nodiscard]] std::string ToString() const override;
+  std::string ToString() const override;
 
   /// \brief Find the SchemaField by field name.
   ///
@@ -65,18 +67,40 @@ class ICEBERG_EXPORT Schema : public StructType {
   /// canonical name 'm.value.x'
   /// FIXME: Currently only handles ASCII lowercase conversion; extend to support
   /// non-ASCII characters (e.g., using std::towlower or ICU)
-  [[nodiscard]] Result<std::optional<std::reference_wrapper<const SchemaField>>>
-  FindFieldByName(std::string_view name, bool case_sensitive = true) const;
+  Result<std::optional<std::reference_wrapper<const SchemaField>>> FindFieldByName(
+      std::string_view name, bool case_sensitive = true) const;
 
   /// \brief Find the SchemaField by field id.
-  [[nodiscard]] Result<std::optional<std::reference_wrapper<const SchemaField>>>
-  FindFieldById(int32_t field_id) const;
+  Result<std::optional<std::reference_wrapper<const SchemaField>>> FindFieldById(
+      int32_t field_id) const;
+
+  /// \brief Creates a projection schema for a subset of columns, selected by name.
+  Result<std::shared_ptr<const Schema>> select(const std::vector<std::string>& names,
+                                               bool case_sensitive = true) const;
+
+  /// \brief Creates a projection schema for a subset of columns, selected by name.
+  Result<std::shared_ptr<const Schema>> select(
+      const std::initializer_list<std::string>& names, bool case_sensitive = true) const;
+
+  /// \brief Creates a projection schema for a subset of columns, selected by name.
+  template <typename... Args>
+  Result<std::shared_ptr<const Schema>> select(Args&&... names,
+                                               bool case_sensitive = true) const {
+    static_assert((std::is_convertible_v<Args, std::string> && ...),
+                  "All arguments must be convertible to std::string");
+    return select({std::string(names)...}, case_sensitive);
+  }
+
+  Result<std::shared_ptr<const Schema>> project(std::unordered_set<int32_t>& ids) const;
 
   friend bool operator==(const Schema& lhs, const Schema& rhs) { return lhs.Equals(rhs); }
 
  private:
   /// \brief Compare two schemas for equality.
   [[nodiscard]] bool Equals(const Schema& other) const;
+
+  Result<std::shared_ptr<const Schema>> internalSelect(
+      const std::vector<std::string>& names, bool case_sensitive) const;
 
   // TODO(nullccxsy): Address potential concurrency issues in lazy initialization (e.g.,
   // use std::call_once)
--- a/test/schema_test.cc
+++ b/test/schema_test.cc
@@ -25,8 +25,10 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "iceberg/result.h"
 #include "iceberg/schema_field.h"
 #include "iceberg/util/formatter.h"  // IWYU pragma: keep
+#include "matchers.h"
 
 TEST(SchemaTest, Basics) {
   {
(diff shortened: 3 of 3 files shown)


src/iceberg/schema.h
```diff
@@ -41,7 +42,8 @@ namespace iceberg {
 /// A schema is a list of typed columns, along with a unique integer ID.  A
 /// Table may have different schemas over its lifetime due to schema
 /// evolution.
-class ICEBERG_EXPORT Schema : public StructType {
+class ICEBERG_EXPORT Schema : public StructType,
```
wgtmac: We shouldn't inherit enable_shared_from_this, otherwise it greatly limits its usage.

src/iceberg/schema.h
```diff
@@ -65,19 +67,41 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+      std::string_view name, bool case_sensitive = true) const;
 
   /// \brief Find the SchemaField by field id.
-  [[nodiscard]] Result<std::optional<std::reference_wrapper<const SchemaField>>>
-  FindFieldById(int32_t field_id) const;
+  Result<std::optional<std::reference_wrapper<const SchemaField>>> FindFieldById(
+      int32_t field_id) const;
+
+  /// \brief Creates a projection schema for a subset of columns, selected by name.
+  Result<std::shared_ptr<const Schema>> select(const std::vector<std::string>& names,
```
wgtmac: Function names should be capitalized unless it is a trivial getter. So please rename `select` to `Select`.

BTW, why the return value is a `shared_ptr` not `unique_ptr`? And why it is `const`? I'd suggest to return `Result<std::unique_ptr<Schema>>` so users can accept it as both  unique_ptr and shared_ptr.

src/iceberg/schema.h
```diff
@@ -65,19 +67,41 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+                                               bool case_sensitive = true) const;
+
+  /// \brief Creates a projection schema for a subset of columns, selected by name.
+  Result<std::shared_ptr<const Schema>> select(
+      const std::initializer_list<std::string>& names, bool case_sensitive = true) const;
```
wgtmac: Usually `initializer_list` is used to create a container type and more likely be used in the testing. But I'm fine to keep it.

src/iceberg/schema.h
wgtmac: Same suggestion applies to below functions.

src/iceberg/schema.h
```diff
@@ -65,19 +67,41 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+
+  /// \brief Creates a projection schema for a subset of columns, selected by name.
+  template <typename... Args>
+  Result<std::shared_ptr<const Schema>> select(Args&&... names,
+                                               bool case_sensitive = true) const {
+    static_assert((std::is_convertible_v<Args, std::string> && ...),
```
wgtmac: Shouldn't we accept `std::string_view` as well?

src/iceberg/schema.h
```diff
@@ -65,19 +67,41 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+
+  Result<std::shared_ptr<const Schema>> project(std::unordered_set<int32_t>& ids) const;
 
   friend bool operator==(const Schema& lhs, const Schema& rhs) { return lhs.Equals(rhs); }
 
  private:
   /// \brief Compare two schemas for equality.
   [[nodiscard]] bool Equals(const Schema& other) const;
 
+  Result<std::shared_ptr<const Schema>> internalSelect(
```
wgtmac:
```suggestion
  Result<std::shared_ptr<const Schema>> SelectInternal(
```

src/iceberg/schema.h
wgtmac: Should we use `std::span<const std::string>` instead of `const std::vector<std::string>&` to accept more std container types? It can support both vector and initializer_list.

src/iceberg/schema.cc
```diff
@@ -257,4 +257,259 @@ void NameToIdVisitor::Finish() {
   }
 }
 
+class PruneColumnVisitor {
+ public:
+  explicit PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                              bool select_full_types = false);
```
wgtmac: Can we avoid default parameter as much as possible?

src/iceberg/schema.h
```diff
@@ -65,19 +67,41 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+  Result<std::optional<std::reference_wrapper<const SchemaField>>> FindFieldByName(
+      std::string_view name, bool case_sensitive = true) const;
 
   /// \brief Find the SchemaField by field id.
-  [[nodiscard]] Result<std::optional<std::reference_wrapper<const SchemaField>>>
-  FindFieldById(int32_t field_id) const;
+  Result<std::optional<std::reference_wrapper<const SchemaField>>> FindFieldById(
+      int32_t field_id) const;
+
+  /// \brief Creates a projection schema for a subset of columns, selected by name.
```
wgtmac:
```suggestion
  /// \brief Creates a projected schema from selected field names.
```

BTW, it would be clear about the distinction between `Select` and `Project` functions and what are the requirements for the `names` parameter. For example, does it support selecting fields in a nested field?

src/iceberg/schema.h
```diff
@@ -65,19 +67,41 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+  template <typename... Args>
+  Result<std::shared_ptr<const Schema>> select(Args&&... names,
+                                               bool case_sensitive = true) const {
+    static_assert((std::is_convertible_v<Args, std::string> && ...),
+                  "All arguments must be convertible to std::string");
+    return select({std::string(names)...}, case_sensitive);
+  }
+
+  Result<std::shared_ptr<const Schema>> project(std::unordered_set<int32_t>& ids) const;
```
wgtmac: Please add comment for it and rename `ids` to `field_ids` to be clear. Also the comment should explain what to expect if a non-leaf id has been specified.

src/iceberg/schema.cc
```diff
@@ -257,4 +257,259 @@ void NameToIdVisitor::Finish() {
   }
 }
 
+class PruneColumnVisitor {
```
wgtmac, line 272: Why not directly put the function definitions in the class? We don't have to split class declaration and definition if this class is not declared in a header file. If you really want to split them, at least those function definitions should be right after this class.

src/iceberg/schema.cc
```diff
@@ -257,4 +257,259 @@ void NameToIdVisitor::Finish() {
...
+  std::shared_ptr<const Type> GetResult() const;
+  void SetResult(std::shared_ptr<const Type> result);
+  Status ProjectList(const SchemaField& element,
+                     std::shared_ptr<const Type> element_result);
+  Status ProjectMap(const SchemaField& key_field, const SchemaField& value_field,
+                    std::shared_ptr<const Type> value_result);
+
+ private:
+  const std::unordered_set<int32_t>& selected_ids_;
+  bool select_full_types_;
```
wgtmac: Can you add comment to explain what is this?

src/iceberg/schema.cc
```diff
@@ -257,4 +257,260 @@ void NameToIdVisitor::Finish() {
   }
 }
 
+class PruneColumnVisitor {
+ public:
+  explicit PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                              bool select_full_types = false);
+  Status Visit(const ListType& type, std::shared_ptr<const Type>& result);
```
wgtmac:
```suggestion
  Status Visit(const ListType& type, std::shared_ptr<Type>* result);
```

For output parameter, use pointer instead of reference for better readability per the google coding style: https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

src/iceberg/schema.cc
```diff
@@ -257,4 +257,260 @@ void NameToIdVisitor::Finish() {
...
+}
+
+Result<std::shared_ptr<const Schema>> Schema::select(
+    const std::initializer_list<std::string>& names, bool case_sensitive) const {
+  return internalSelect(std::vector<std::string>(names), case_sensitive);
+}
+
+Result<std::shared_ptr<const Schema>> Schema::internalSelect(
+    const std::vector<std::string>& names, bool case_sensitive) const {
+  const std::string ALL_COLUMNS = "*";
```
wgtmac:
```suggestion
  const std::string kAllColumns = "*";
```

src/iceberg/schema.cc
```diff
@@ -257,4 +257,260 @@ void NameToIdVisitor::Finish() {
...
+  if (std::ranges::find(names, ALL_COLUMNS) != names.end()) {
+    return shared_from_this();
```
wgtmac: It is not worthy to use `shared_from_this` only for this case.  In this case we can just return a full copy of the current schema.

src/iceberg/schema.cc
wgtmac: BTW, I prefer to use `out` to replace `result` because it may confuse readers with the `Result` type.

src/iceberg/schema.cc
```diff
@@ -257,4 +257,260 @@ void NameToIdVisitor::Finish() {
...
+
+  if (!result) {
+    return std::make_shared<Schema>(std::vector<SchemaField>{}, schema_id_);
+  }
+
+  if (result->type_id() != TypeId::kStruct) {
+    return InvalidSchema("Projected type must be a struct type");
+  }
+
+  const auto& projected_struct = internal::checked_cast<const StructType&>(*result);
```
wgtmac: nit: use `FromStructType` function in the schema_internal.h to save some lines.

src/iceberg/schema.cc
```diff
@@ -257,4 +257,260 @@ void NameToIdVisitor::Finish() {
...
+
+PruneColumnVisitor::PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                                       bool select_full_types)
+    : selected_ids_(selected_ids), select_full_types_(select_full_types) {}
+
+Status PruneColumnVisitor::Visit(const StructType& type,
+                                 std::shared_ptr<const Type>& result) {
+  std::vector<std::shared_ptr<const Type>> selected_types;
+  const auto& fields = type.fields();
+  for (const auto& field : fields) {
```
wgtmac:
```suggestion
  for (const auto& field : type.fields()) {
```

Let's make it shorter.

src/iceberg/schema.cc
```diff
@@ -257,4 +257,260 @@ void NameToIdVisitor::Finish() {
...
+    std::shared_ptr<const Type> child_result;
+    ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*field.type(), this, child_result));
+    if (selected_ids_.contains(field.field_id())) {
+      // select
+      if (select_full_types_) {
+        selected_types.emplace_back(field.type());
```
wgtmac: In this case, line 355 is wasted.

src/iceberg/schema.cc
```diff
@@ -257,4 +257,260 @@ void NameToIdVisitor::Finish() {
...
 
+class PruneColumnVisitor {
+ public:
+  explicit PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                              bool select_full_types = false);
+  Status Visit(const ListType& type, std::shared_ptr<const Type>& result);
+  Status Visit(const MapType& type, std::shared_ptr<const Type>& result);
+  Status Visit(const StructType& type, std::shared_ptr<const Type>& result);
+  Status Visit(const PrimitiveType& type, std::shared_ptr<const Type>& result);
+  Status ProjectList(const SchemaField& element,
```
wgtmac: Why `ProjectSelectedStruct` from Java impl is missing here?

src/iceberg/schema.cc
wgtmac: Can we change the class definition to one of the two forms below?

```cpp
class PruneColumnVisitor {
 public:
  PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
                     bool select_full_types);

  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<ListType>& type) const;
  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<MapType>& type) const;
  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<StructType>& type) const;
  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<PrimitiveType>& type) const;
  Result<std::unique_ptr<Schema>> Visit(const Schema& schema) const;

 private:
  const std::unordered_set<int32_t>& selected_ids_;
  bool select_full_types_;
};
```

or 

```cpp
class PruneColumnVisitor {
 public:
  PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
                     bool select_full_types);

  Status Visit(const std::shared_ptr<ListType>& type, std::shared_ptr<Type>* out) const;
  Status Visit(const std::shared_ptr<MapType>& type, std::shared_ptr<Type>* out) const;
  Status Visit(const std::shared_ptr<StructType>& type, std::shared_ptr<Type>* out) const;
  Status Visit(const std::shared_ptr<PrimitiveType>& type,
               std::shared_ptr<Type>* out) const;
  Status Visit(const Schema& schema, std::unique_ptr<Schema>* out) const;

 private:
  const std::unordered_set<int32_t>& selected_ids_;
  bool select_full_types_;
};
```

src/iceberg/schema.cc
nullccxsy: Yes, `ProjectSelectedStruct ` is need here to make `child_result` not null

src/iceberg/schema.cc
```diff
@@ -257,4 +258,268 @@ void NameToIdVisitor::Finish() {
...
+/// This visitor traverses a schema and creates a projected version containing only
+/// the specified fields. It handles different projection modes:
+/// - select_full_types=true: Include entire fields when their ID is selected
+/// - select_full_types=false: Recursively project nested fields within selected structs
+///
+/// \warning Error conditions that will cause projection to fail:
+/// - Attempting to explicitly project List or Map types (returns InvalidArgument)
+/// - Projecting a List when element result is null (returns InvalidArgument)
+/// - Projecting a Map without a defined map value type (returns InvalidArgument)
+/// - Projecting a struct when result is not StructType (returns InvalidArgument)
```
wgtmac: I think it is easy and valid to support projections in the nested map and list types and don't know why the Java impl does not support this. The code will be much simpler (shorter) if we support them.

@Fokko Do you have any context on the Java impl?

src/iceberg/schema.cc
```diff
@@ -257,4 +258,268 @@ void NameToIdVisitor::Finish() {
...
+    const auto& key_field = type.fields()[0];
+    const auto& value_field = type.fields()[1];
+
+    if (select_full_types_ and selected_ids_.contains(value_field.field_id())) {
+      *out = std::make_unique<MapType>(type);
+      return {};
+    }
+
+    std::unique_ptr<Type> key_result;
+    ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*key_field.type(), this, &key_result));
```
wgtmac: Are you sure about this? In the Java impl, key type should be fully projected.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,268 @@ void NameToIdVisitor::Finish() {
...
+/// - Projecting a struct when result is not StructType (returns InvalidArgument)
+class PruneColumnVisitor {
+ public:
+  PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                     bool select_full_types)
+      : selected_ids_(selected_ids), select_full_types_(select_full_types) {}
+
+  Status Visit(const StructType& type, std::unique_ptr<Type>* out) const {
+    std::vector<std::shared_ptr<const Type>> selected_types;
+    for (const auto& field : type.fields()) {
```
wgtmac: There are a lot of duplicate (or at least similar code) in this visitor class and can be greatly simplified by adding a Visit function for `const SchemaField&`. Please check that Struct, List, and Map are all nested types with a list of `SchemaField`s.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,268 @@ void NameToIdVisitor::Finish() {
...
+/// - Projecting a List when element result is null (returns InvalidArgument)
+/// - Projecting a Map without a defined map value type (returns InvalidArgument)
+/// - Projecting a struct when result is not StructType (returns InvalidArgument)
+class PruneColumnVisitor {
+ public:
+  PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
+                     bool select_full_types)
+      : selected_ids_(selected_ids), select_full_types_(select_full_types) {}
+
+  Status Visit(const StructType& type, std::unique_ptr<Type>* out) const {
```
wgtmac:
```suggestion
  Status Visit(const std::shared_ptr<StructType>& type, std::shared_ptr<Type>* out) const {
```

I still think we need to change the signature to this to enable returning original type (which is a common case for select). Or even better, we can simplify it as `Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<StructType>& type) const` and add our own dispatching in the `Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<Type>& type) const`.

src/iceberg/schema.h
```diff
@@ -65,18 +66,110 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+  ///
+  /// Selects fields by their names using dot notation for nested fields.
+  /// Supports both canonical names (e.g., "user.address.street") and short names
+  /// (e.g., "user.street" for map values, "list.element" for list elements).
+  ///
+  /// \param names Field names to select (supports nested field paths)
+  /// \param case_sensitive Whether name matching is case-sensitive (default: true)
+  /// \return Projected schema containing only the specified fields
+  ///
+  /// \example
```
wgtmac: The comments here and below are a little bit over-complicated. I think we can remove the concrete examples if the parameter docstrings are clear enough.

src/iceberg/schema.cc
nullccxsy: Here, we only visited the key; when creating the map, we still use `key_field`. In my view, we still need to perform the visit operation. When visiting the key, there may be some errors that need to be handled internally. For example, when projecting the elements of a list.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,242 @@ void NameToIdVisitor::Finish() {
...
 
+/// \brief Visitor class for pruning schema columns based on selected field IDs.
+///
+/// This visitor traverses a schema and creates a projected version containing only
+/// the specified fields. It handles different projection modes:
+/// - select_full_types=true: Include entire fields when their ID is selected
+/// - select_full_types=false: Recursively project nested fields within selected structs
+///
+/// \warning Error conditions that will cause projection to fail:
+/// - Project or Select a Map with just key or value (returns InvalidArgument)
```
wgtmac:
```suggestion
/// \brief Visitor for pruning schema columns based on selected field IDs.
///
/// This visitor traverses a schema and creates a projected version containing only
/// the requested fields. When `select_full_types` is true, a field with all its
/// sub-fields are selected if its field-id has been selected; otherwise, only leaf
/// fields of selected field-ids are selected.
///
/// \note It returns an error when projection is not successful.
```

src/iceberg/schema.cc
```diff
@@ -257,4 +258,242 @@ void NameToIdVisitor::Finish() {
...
+class PruneColumnVisitor {
+ public:
+  PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
```
wgtmac, line 274:
```cpp
class PruneColumnVisitor {
 public:
  PruneColumnVisitor(const std::unordered_set<int32_t>& selected_ids,
                     bool select_full_types)
      : selected_ids_(selected_ids), select_full_types_(select_full_types) {}

  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<Type>& type) const {
    switch (type->type_id()) {
      case TypeId::kStruct:
        return Visit(internal::checked_pointer_cast<StructType>(type));
      case TypeId::kList:
        return Visit(internal::checked_pointer_cast<ListType>(type));
      case TypeId::kMap:
        return Visit(internal::checked_pointer_cast<MapType>(type));
      default:
        return nullptr;
    }
  }

  Result<std::shared_ptr<Type>> Visit(const SchemaField& field) const {
    if (selected_ids_.contains(field.field_id())) {
      return (select_full_types_ || field.type()->is_primitive()) ? field.type()
                                                                  : Visit(field.type());
    }
    return Visit(field.type());
  }

  static SchemaField MakeField(const SchemaField& field, std::shared_ptr<Type> type) {
    return {field.field_id(), std::string(field.name()), std::move(type),
            field.optional(), std::string(field.doc())};
  }

  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<StructType>& type) const {
    std::vector<SchemaField> selected_fields;
    bool same_types = true;
    for (const auto& field : type->fields()) {
      ICEBERG_ASSIGN_OR_RAISE(auto child_type, Visit(field.type()));
      if (child_type != nullptr) {
        same_types = same_types && (child_type == field.type());
        selected_fields.push_back(MakeField(field, std::move(child_type)));
      }
    }
    if (selected_fields.empty()) {
      return nullptr;
    } else if (same_types && selected_fields.size() == type->fields().size()) {
      return type;
    }
    return std::make_shared<StructType>(std::move(selected_fields));
  }

  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<ListType>& type) const {
    const auto& elem_field = type->fields()[0];
    ICEBERG_ASSIGN_OR_RAISE(auto elem_type, Visit(elem_field));
    if (elem_type == nullptr) {
      return nullptr;
    } else if (elem_type == elem_field.type()) {
      return type;
    }
    return std::make_shared<ListType>(MakeField(elem_field, std::move(elem_type)));
  }

  Result<std::shared_ptr<Type>> Visit(const std::shared_ptr<MapType>& type) const {
    const auto& key_field = type->fields()[0];
    const auto& value_field = type->fields()[1];
    ICEBERG_ASSIGN_OR_RAISE(auto key_type, Visit(key_field.type()));
    ICEBERG_ASSIGN_OR_RAISE(auto value_type, Visit(value_field.type()));

    if (key_type == nullptr && value_type == nullptr) {
      return nullptr;
    } else if (value_type == value_field.type() &&
               (key_type == key_field.type() || key_type == nullptr)) {
      return type;
    } else if (value_type == nullptr) {
      return InvalidArgument("Cannot project Map without value field");
    }
    return std::make_shared<MapType>(
        (key_type == nullptr ? key_field : MakeField(key_field, std::move(key_type))),
        MakeField(value_field, std::move(value_type)));
  }

 private:
  const std::unordered_set<int32_t>& selected_ids_;
  const bool select_full_types_;
};
```

I think we need `Result<std::shared_ptr<Type>> Visit(const SchemaField& field) const` to make it look simpler by eliminating duplicate codes.

wgtmac (changes requested):

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+  /// \brief Creates a projected schema from selected field names.
+  ///
+  /// Selects fields by their names using dot notation for nested fields.
+  /// Supports both canonical names (e.g., "user.address.street") and short names
+  /// (e.g., "user.street" for map values, "list.element" for list elements).
+  ///
+  /// \param names Field names to select (supports nested field paths)
+  /// \param case_sensitive Whether name matching is case-sensitive (default: true)
+  /// \return Projected schema containing only the specified fields
+  Result<std::unique_ptr<const Schema>> Select(std::span<const std::string> names,
```
wgtmac:
```suggestion
  Result<std::unique_ptr<Schema>> Select(std::span<const std::string> names,
```

It is an overkill to return const here

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+                                               bool case_sensitive = true) const;
+
+  /// \brief Creates a projected schema from selected field names.
+  Result<std::unique_ptr<const Schema>> Select(
+      const std::initializer_list<std::string>& names, bool case_sensitive = true) const;
```
wgtmac:
```suggestion
```

We have `std::span<const std::string>` as the input now, so this overload is unnecessary.

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+
+  /// \brief Creates a projected schema from selected field names.
+  template <typename... Args>
+  Result<std::unique_ptr<const Schema>> Select(Args&&... names,
```
wgtmac:
```suggestion
  Result<std::unique_ptr<Schema>> Select(Args&&... names,
```

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+                                               bool case_sensitive = true) const {
+    static_assert(((std::is_convertible_v<Args, std::string> ||
+                    std::is_convertible_v<Args, std::string>) &&
```
wgtmac:
```suggestion
                    std::is_convertible_v<Args, std::string_view>) &&
```

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+                   ...),
+                  "All arguments must be convertible to std::string");
```
wgtmac:
```suggestion
                  "All arguments must be convertible to std::string or std::string_view");
```

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+    return select({(names)...}, case_sensitive);
```
wgtmac: We don't have test for this function, otherwise it must fail

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+
+  /// \brief Creates a projected schema from selected field names.
+  ///
+  /// Selects fields by their names using dot notation for nested fields.
+  /// Supports both canonical names (e.g., "user.address.street") and short names
+  /// (e.g., "user.street" for map values, "list.element" for list elements).
+  ///
+  /// \param names Field names to select (supports nested field paths)
+  /// \param case_sensitive Whether name matching is case-sensitive (default: true)
+  /// \return Projected schema containing only the specified fields
```
wgtmac:
```suggestion
  /// \brief Creates a projected schema from selected field names.
  ///
  /// \param names Selected field names and nested names are dot-concatenated.
  /// \param case_sensitive Whether name matching is case-sensitive (default: true).
  /// \return Projected schema containing only selected fields.
  /// \note If the field name of a nested type has been selected, all of its
  /// sub-fields will be selected.
```

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+  /// Selects fields by their numeric IDs. More efficient than Select() when you
+  /// already know the field IDs. Handles recursive projection of nested structs.
+  ///
+  /// \param field_ids Set of field IDs to select
+  /// \return Projected schema containing only the specified fields
+  ///
+  /// \note When a struct field ID is specified:
+  ///       - If nested field IDs are also in field_ids, they are recursively projected
+  ///       - If no nested field IDs are in field_ids, an empty struct is included
+  ///       - List/Map types cannot be explicitly projected (returns error)
```
wgtmac:
```suggestion
  /// \brief Creates a projected schema from selected field IDs.
  ///
  /// \param field_ids Set of field IDs to project.
  /// \return Projected schema containing only selected field IDs.
  /// \note Field ID of a nested field may not be projected unless at least
  /// one of its sub-fields has been projected.
```

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
+  Result<std::unique_ptr<const Schema>> Project(
+      std::unordered_set<int32_t>& field_ids) const;
```
wgtmac:
```suggestion
  Result<std::unique_ptr<Schema>> Project(
      const std::unordered_set<int32_t>& field_ids) const;
```

Return value should not have `const` but the input parameter should.

src/iceberg/schema.h
```diff
@@ -65,18 +66,63 @@ class ICEBERG_EXPORT Schema : public StructType {
...
 
   friend bool operator==(const Schema& lhs, const Schema& rhs) { return lhs.Equals(rhs); }
 
  private:
   /// \brief Compare two schemas for equality.
-  [[nodiscard]] bool Equals(const Schema& other) const;
+  bool Equals(const Schema& other) const;
+
+  Result<std::unique_ptr<const Schema>> SelectInternal(std::span<const std::string> names,
```
wgtmac:
```suggestion
  Result<std::unique_ptr<Schema>> SelectInternal(std::span<const std::string> names,
```

src/iceberg/schema.cc
```diff
@@ -257,4 +258,164 @@ void NameToIdVisitor::Finish() {
...
+      }
+      case TypeId::kList: {
+        return Visit(internal::checked_pointer_cast<ListType>(type));
+      }
+      case TypeId::kMap: {
+        return Visit(internal::checked_pointer_cast<MapType>(type));
+      }
+      default: {
+        return nullptr;
+      }
```
wgtmac:
```suggestion
      case TypeId::kStruct:
        return Visit(internal::checked_pointer_cast<StructType>(type));
      case TypeId::kList:
        return Visit(internal::checked_pointer_cast<ListType>(type));
      case TypeId::kMap:
        return Visit(internal::checked_pointer_cast<MapType>(type));
      default:
        return nullptr;
```

Less is more :)

src/iceberg/schema.cc
```diff
@@ -257,4 +258,164 @@ void NameToIdVisitor::Finish() {
...
+      return InvalidArgument("Cannot project Map without value field");
+    }
+    return std::make_shared<MapType>(
+        (key_type == nullptr ? key_field : MakeField(key_field, std::move(key_type))),
+        MakeField(value_field, std::move(value_type)));
+  }
+
+ private:
+  const std::unordered_set<int32_t>& selected_ids_;
+  bool select_full_types_;
```
wgtmac:
```suggestion
  const bool select_full_types_;
```

src/iceberg/schema.cc
```diff
@@ -257,4 +258,164 @@ void NameToIdVisitor::Finish() {
...
+                                                     bool case_sensitive) const {
+  return SelectInternal(names, case_sensitive);
+}
+
+Result<std::unique_ptr<const Schema>> Schema::Select(
+    const std::initializer_list<std::string>& names, bool case_sensitive) const {
+  return SelectInternal(names, case_sensitive);
+}
+
+Result<std::unique_ptr<const Schema>> Schema::SelectInternal(
```
wgtmac: Can we remove `SelectInternal` now? We can remove the `Select` function overload with `std::initializer_list`.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,164 @@ void NameToIdVisitor::Finish() {
...
+    ICEBERG_ASSIGN_OR_RAISE(auto result, FindFieldByName(name, case_sensitive));
+    if (result.has_value()) {
+      selected_ids.insert(result.value().get().field_id());
+    }
+  }
+
+  PruneColumnVisitor visitor(selected_ids, /*select_full_types=*/true);
+  auto self = std::shared_ptr<const StructType>(this, [](const StructType*) {});
+  ICEBERG_ASSIGN_OR_RAISE(auto result,
+                          visitor.Visit(std::const_pointer_cast<StructType>(self)));
```
wgtmac: Please do not hack like this. You may add a function `std::unique_ptr<StructType> ToStructType(const Schema& schema);` to `iceberg/schema_internal.h` by copying all fields to the struct type and return.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,164 @@ void NameToIdVisitor::Finish() {
...
+
+  auto& projected_struct = internal::checked_cast<const StructType&>(*result);
+
+  return FromStructType(std::move(const_cast<StructType&>(projected_struct)), schema_id_);
+}
+
+Result<std::unique_ptr<const Schema>> Schema::Project(
+    std::unordered_set<int32_t>& field_ids) const {
+  PruneColumnVisitor visitor(field_ids, /*select_full_types=*/false);
+  auto self = std::shared_ptr<const StructType>(this, [](const StructType*) {});
```
wgtmac: Same comment as above.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,164 @@ void NameToIdVisitor::Finish() {
...
+  if (!result) {
+    return std::make_unique<Schema>(std::vector<SchemaField>{}, schema_id_);
+  }
+
+  if (result->type_id() != TypeId::kStruct) {
+    return InvalidSchema("Projected type must be a struct type");
+  }
+
+  auto& projected_struct = internal::checked_cast<const StructType&>(*result);
+
```
wgtmac: Why there is a blank line but line 417 does not have one? I think we can have a one-liner for 397 and 399.

test/schema_test.cc
```diff
@@ -491,3 +493,921 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+  if (param.should_succeed) {
+    ASSERT_TRUE(result.has_value());
+    auto actual_schema = std::move(result.value());
+    auto expected_schema = param.expected_schema();
+    ASSERT_EQ(*actual_schema, *expected_schema)
+        << "Schema mismatch for: " << param.test_name;
+  } else {
+    ASSERT_FALSE(result.has_value());
+    ASSERT_THAT(result, iceberg::IsError(iceberg::ErrorKind::kInvalidArgument));
+
```
wgtmac:
```suggestion
```

test/schema_test.cc
```diff
@@ -491,3 +493,921 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+    ASSERT_THAT(result, iceberg::HasErrorMessage(param.expected_error_message));
+  }
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    SelectTestCases, SelectParamTest,
+    ::testing::Values(
+        SelectTestParam{.test_name = "SelectAllColumns",
+                        .create_schema =
```
wgtmac: If the same schema has been used for multiple time, we can add a function above to reuse. The point is to make these lines as short as possible.

test/schema_test.cc
```diff
@@ -491,3 +493,921 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+                            []() {
+                              auto field1 = std::make_unique<iceberg::SchemaField>(
```
wgtmac: We shouldn't make a unique_ptr and then only copy its data below. Instead, we can:

```
SchemaField field1(1, "id", iceberg::int32(), false);
...
return std::make_shared<iceberg::Schema>(
                                  std::vector<iceberg::SchemaField>{std::move(field1), ...});
```

wgtmac: If multiple fields are used, perhaps we can also do this to reuse them in all test cases:

```
static SchemaField MakeId() {
  return {1, "id", iceberg::int32(), false};
}

static SchemaField MakeXXX() {
  return ...
}
```

src/iceberg/schema.cc
```diff
@@ -257,4 +258,149 @@ void NameToIdVisitor::Finish() {
...
+      ICEBERG_ASSIGN_OR_RAISE(auto child_type, Visit(field));
+      if (child_type) {
+        same_types = same_types && (child_type == field.type());
+        selected_fields.emplace_back(MakeField(field, std::move(child_type)));
+      }
+    }
+
+    if (selected_fields.empty()) {
+      return nullptr;
+    } else if (same_types and selected_fields.size() == type->fields().size()) {
```
wgtmac:
```suggestion
    } else if (same_types && selected_fields.size() == type->fields().size()) {
```

We don't use `and` across the repo so please be consistent.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,149 @@ void NameToIdVisitor::Finish() {
...
+      selected_ids.insert(result.value().get().field_id());
+    }
+  }
+
+  PruneColumnVisitor visitor(selected_ids, /*select_full_types=*/true);
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto result, visitor.Visit(std::shared_ptr<StructType>(ToStructType(*this))));
+
+  if (!result) {
+    return std::make_unique<Schema>(std::vector<SchemaField>{}, schema_id_);
```
wgtmac:
```suggestion
    return std::make_unique<Schema>(std::vector<SchemaField>{}, std::nullopt);
```

This is no longer the original schema so it cannot share the schema id.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,149 @@ void NameToIdVisitor::Finish() {
...
+  }
+
+  if (result->type_id() != TypeId::kStruct) {
+    return InvalidSchema("Projected type must be a struct type");
+  }
+
+  return FromStructType(std::move(const_cast<StructType&>(
```
wgtmac: ditto

src/iceberg/schema.cc
```diff
@@ -257,4 +258,149 @@ void NameToIdVisitor::Finish() {
...
+}
+
+Result<std::unique_ptr<Schema>> Schema::Project(
+    const std::unordered_set<int32_t>& field_ids) const {
+  PruneColumnVisitor visitor(field_ids, /*select_full_types=*/false);
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto result, visitor.Visit(std::shared_ptr<StructType>(ToStructType(*this))));
+
+  if (!result) {
+    return std::make_unique<Schema>(std::vector<SchemaField>{}, schema_id_);
```
wgtmac: ditto

src/iceberg/schema.cc
```diff
@@ -257,4 +258,149 @@ void NameToIdVisitor::Finish() {
   }
 }
 
+/// \brief Visitor class for pruning schema columns based on selected field IDs.
```
wgtmac:
```suggestion
/// \brief Visitor for pruning columns based on selected field IDs.
```

src/iceberg/schema.cc
```diff
@@ -257,4 +258,149 @@ void NameToIdVisitor::Finish() {
...
+  for (const auto& name : names) {
+    ICEBERG_ASSIGN_OR_RAISE(auto result, FindFieldByName(name, case_sensitive));
+    if (result.has_value()) {
+      selected_ids.insert(result.value().get().field_id());
+    }
+  }
+
+  PruneColumnVisitor visitor(selected_ids, /*select_full_types=*/true);
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto result, visitor.Visit(std::shared_ptr<StructType>(ToStructType(*this))));
```
wgtmac:
```suggestion
      auto pruned_type, visitor.Visit(std::shared_ptr<StructType>(ToStructType(*this))));
```

Let's avoid vague variable names.

src/iceberg/schema.cc
```diff
@@ -257,4 +258,149 @@ void NameToIdVisitor::Finish() {
...
+    return std::make_unique<Schema>(std::vector<SchemaField>{}, schema_id_);
+  }
+
+  if (result->type_id() != TypeId::kStruct) {
+    return InvalidSchema("Projected type must be a struct type");
+  }
+
+  return FromStructType(std::move(const_cast<StructType&>(
+                            internal::checked_cast<const StructType&>(*result))),
+                        schema_id_);
```
wgtmac:
```suggestion
  return FromStructType(std::move(internal::checked_cast<StructType&>(*pruned_type)),
                        std::nullopt);
```

test/schema_test.cc
```diff
@@ -491,3 +493,688 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
   EXPECT_THAT(result.error().message,
               ::testing::HasSubstr("Duplicate field id found: 1"));
 }
+
+namespace {
+
+struct TestFieldFactory {
+  static std::unique_ptr<iceberg::SchemaField> Id(int32_t field_id = 1,
```
wgtmac: The main point of adding these convenience functions is to make lines shorter. However, these function now are significantly longer than simply writing `SchemaField field{1, "id", iceberg::int32(), true};`

wgtmac: I have seen a common misuse like this for multiple times. Please check that `NestedType` accepts `SchemaField` but not `std::shared_ptr<SchemaField>` or `std::unique_ptr<SchemaField>`, so it is pointless to create a smart point of it then make a copy by dereferencing it. To be short, you should never create a smart pointer of `SchemaField`.

wgtmac: For readability, functions to create `Id`, `Name` and the like, we should not accept any input parameters because their definition should be consistent across these test cases. Otherwise, simply using `SchemaField field{1, "id", iceberg::int32(), true}` looks better.

wgtmac: What's in my mind is something like below:

```cpp
namespace {

iceberg::SchemaField MakeIdField() { return {1, "id", iceberg::int32(), true}; }

iceberg::SchemaField MakeNameField() { return {2, "name", iceberg::string(), false}; }

template <typename... Args>
std::shared_ptr<iceberg::StructType> MakeStructType(Args... args) {
  return std::make_shared<iceberg::StructType>(
      std::vector<iceberg::SchemaField>{args...});
}

template <typename... Args>
std::shared_ptr<iceberg::Schema> MakeSchema(Args... args) {
  return std::make_shared<iceberg::Schema>(std::vector<iceberg::SchemaField>{args...},
                                           std::nullopt);
}

}  // namespace

TEST(SchemaTest, Dummy) {
  auto schema = MakeSchema(MakeIdField(), MakeNameField());
  auto struct_type = MakeStructType(MakeIdField(), MakeNameField());
}
```

test/schema_test.cc
```diff
@@ -490,3 +503,474 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
   EXPECT_THAT(result.error().message,
               ::testing::HasSubstr("Duplicate field id found: 1"));
 }
+
+namespace {
+
+namespace TestFields {
```
wgtmac:
```suggestion
```

We should not define nested namespace in an anonymous namespace

test/schema_test.cc
```diff
@@ -490,3 +503,474 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+iceberg::SchemaField City() { return {12, "city", iceberg::string(), true}; }
+iceberg::SchemaField Zip() { return {13, "zip", iceberg::int32(), true}; }
+iceberg::SchemaField Theme() { return {24, "theme", iceberg::string(), true}; }
+iceberg::SchemaField Key() { return {31, "key", iceberg::int32(), false}; }
+iceberg::SchemaField Value() { return {32, "value", iceberg::string(), false}; }
+iceberg::SchemaField Element() { return {41, "element", iceberg::string(), false}; }
+
+}  // namespace TestFields
+
+struct TestSchemaFactory {
```
wgtmac: Why not simply putting them in the anonymous namespace as well?

test/schema_test.cc
```diff
@@ -490,3 +503,474 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+  static std::shared_ptr<iceberg::Schema> BasicSchema(int32_t schema_id = 100) {
```
wgtmac:
```suggestion
  static std::shared_ptr<iceberg::Schema> BasicSchema() {
```

`schema_id` is not used and unnecessary. Same for below.

test/schema_test.cc
```diff
@@ -490,3 +503,474 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+    return MakeSchema(TestFields::Id(), TestFields::Name(), TestFields::Age(),
+                      TestFields::Email());
+  }
+
+  static std::shared_ptr<iceberg::Schema> AddressSchema(int32_t schema_id = 200) {
+    auto address_type =
+        MakeStructType(TestFields::Street(), TestFields::City(), TestFields::Zip());
+    auto address_field = iceberg::SchemaField{14, "address", address_type, true};
```
wgtmac:
```suggestion
    auto address_field = iceberg::SchemaField{14, "address", std::move(address_type), true};
```

Avoid unnecessary copy. Same for below.

test/schema_test.cc
```diff
@@ -490,3 +503,474 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+    return MakeSchema(TestFields::Id(), TestFields::Name(), address_field);
```
wgtmac:
```suggestion
    return MakeSchema(TestFields::Id(), TestFields::Name(), std::move(address_field));
```

test/schema_test.cc
```diff
@@ -490,3 +503,474 @@ TEST(SchemaTest, NestedDuplicateFieldIdError) {
...
+}
+
+TEST_P(SelectParamTest, SelectFields) {
+  const auto& param = GetParam();
+  auto input_schema = param.create_schema();
+  auto result = input_schema->Select(param.select_fields, param.case_sensitive);
+
+  if (param.should_succeed) {
+    ASSERT_TRUE(result.has_value());
+    CompareSchema(std::move(result.value()), param.expected_schema());
```
wgtmac: We don't need the `CompareSchema`. Schema has implemented `operator==`.
```suggestion
    ASSERT_EQ(*result.value(), *param.expected_schema());
```

PR #209, opened 2025-09-04
chore: update CI workflow and enhance dependency management

nullccxsy:
- Updated the dependencies installation command to include `nlohmann-json` and `nanoarrow`.
- Refactored the `IcebergThirdpartyToolchain.cmake` to improve handling of `nanoarrow` and `nlohmann_json` dependencies, allowing for better integration and installation based on their availability.


.github/workflows/test.yml
```diff
@@ -80,10 +80,10 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           fetch-depth: 0
-      - name: Install ZLIB
+      - name: Install dependencies
         shell: cmd
         run: |
-          vcpkg install zlib:x64-windows
+          vcpkg install zlib:x64-windows nlohmann-json:x64-windows nanoarrow:x64-windows
```
wgtmac, line 86: nit: add `arrow` here to reduce ci time. windows ci is so slow.

.github/workflows/test.yml
```diff
@@ -80,10 +80,10 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           fetch-depth: 0
-      - name: Install ZLIB
+      - name: Install dependencies
         shell: cmd
         run: |
-          vcpkg install zlib:x64-windows
+          vcpkg install zlib:x64-windows nlohmann-json:x64-windows arrow:x64-windows
```
wgtmac: Why nanoarrow is missing?

nullccxsy: This is just a test to check if there are any problems with the other parts after removing nanoarrow.

.github/workflows/test.yml
wgtmac, line 86: Can you try to add `spdlog` as well?

src/iceberg/CMakeLists.txt
```diff
@@ -72,13 +72,13 @@ list(APPEND
      ZLIB::ZLIB)
 list(APPEND
      ICEBERG_STATIC_INSTALL_INTERFACE_LIBS
-     "Iceberg::nanoarrow_static"
-     "Iceberg::nlohmann_json"
+     "$<IF:$<BOOL:${NANOARROW_VENDORED}>,Iceberg::nanoarrow_static,$<IF:$<TARGET_EXISTS:nanoarrow::nanoarrow_static>,nanoarrow::nanoarrow_static,nanoarrow::nanoarrow_shared>>"
+     "$<IF:$<BOOL:${NLOHMANN_JSON_VENDORED}>,Iceberg::nlohmann_json,$<IF:$<TARGET_EXISTS:nlohmann_json::nlohmann_json>,nlohmann_json::nlohmann_json,nlohmann_json::nlohmann_json>>"
      "$<IF:$<BOOL:${SPDLOG_VENDORED}>,Iceberg::spdlog,spdlog::spdlog>")
```
wgtmac, line 77: I guess we don't need to change this for spdlog

PR #211, opened 2025-09-04
refactor: remove -isystem on local include header files

HuaHuaY:


src/iceberg/table_metadata.h
```diff
@@ -87,11 +87,11 @@ struct ICEBERG_EXPORT TableMetadata {
   /// The highest assigned column ID for the table
   int32_t last_column_id;
   /// A list of schemas
-  std::vector<std::shared_ptr<Schema>> schemas;
```
zhjwpku, line 90: I don't think this is necessary since we are inside a iceberg namespace here, same for the below.

src/iceberg/transform.h
```diff
@@ -56,8 +57,40 @@ enum class TransformType {
   kVoid,
 };
 
+constexpr std::string_view kUnknownName = "unknown";
```
zhjwpku: Why move these constants to the header file? I don't see the point :(

HuaHuaY: The constexpr function below needs these constants. Constexpr function is implicitly inline so we have to move the definition to header file to avoid warning. Or we can remove constexpr. Either solution is fine for me.

src/iceberg/table_metadata.h
HuaHuaY: If we don't add `iceberg::`, gcc will report a compile error due to `-Werror` and `-Wchanges-meaning` (included in `-Wall`).

HuaHuaY:
I use clang in macos for coding but ci uses gcc. The two environments have some different compile flags. I can compile the code in my environment but ci compiles fail. Let me convert this PR to draft. I will fix compile errors soon.

src/iceberg/table_metadata.h
HuaHuaY: https://github.com/apache/iceberg-cpp/actions/runs/17463347891/job/49593031029?pr=211
```sh
In file included from /home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/catalog/in_memory_catalog.cc:27:
/home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/table_metadata.h:125:35: error: declaration of ‘iceberg::Result<std::shared_ptr<iceberg::Schema> > iceberg::TableMetadata::Schema() const’ changes meaning of ‘Schema’ [-Wchanges-meaning]
  125 |   Result<std::shared_ptr<Schema>> Schema() const;
      |                                   ^~~~~~
```

src/iceberg/table_metadata.h
```diff
@@ -145,7 +145,7 @@ ICEBERG_EXPORT std::string ToString(const SnapshotLogEntry& entry);
 ICEBERG_EXPORT std::string ToString(const MetadataLogEntry& entry);
 
 /// \brief The codec type of the table metadata file.
-ICEBERG_EXPORT enum class MetadataFileCodecType {
+enum class ICEBERG_EXPORT MetadataFileCodecType {
```
HuaHuaY, line 148: https://github.com/apache/iceberg-cpp/actions/runs/17463347891/job/49593031039

```sh
In file included from /home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/catalog/in_memory_catalog.cc:27:
/home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/table_metadata.h:148:27: error: attribute ignored in declaration of ‘enum class iceberg::MetadataFileCodecType’ [-Werror=attributes]
  148 | ICEBERG_EXPORT enum class MetadataFileCodecType {
      |                           ^~~~~~~~~~~~~~~~~~~~~
/home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/table_metadata.h:148:27: note: attribute for ‘enum class iceberg::MetadataFileCodecType’ must follow the ‘enum class’ keyword
```

src/iceberg/transform.h
zhjwpku: So these warnings are reported from CI, can you post a link?

HuaHuaY: > So these warnings are reported from CI, can you post a link?

Other warnings are reported from CI, this constexpr function is reported in my environment. Without this fix, I can't compile in my MacOS.

```sh
[build] In file included from /Users/huahua/github/iceberg-cpp/test/transform_test.cc:20:
[build] /Users/huahua/github/iceberg-cpp/src/iceberg/transform.h:60:43: error: inline function 'iceberg::TransformTypeToString' is not defined [-Werror,-Wundefined-inline]
[build]    60 | ICEBERG_EXPORT constexpr std::string_view TransformTypeToString(TransformType type);
[build]       |                                           ^
[build] /Users/huahua/github/iceberg-cpp/test/transform_test.cc:91:39: note: used here
[build]    91 |                 std::format("{}[{}]", TransformTypeToString(c.type), *c.param));
[build]       |                                       ^
[build] 1 error generated.

```

cmake_modules/IcebergBuildUtils.cmake
```diff
@@ -131,7 +131,7 @@ function(add_iceberg_lib LIB_NAME)
     endif()
 
     if(LIB_INCLUDES)
-      target_include_directories(${LIB_NAME}_shared SYSTEM PUBLIC ${ARG_EXTRA_INCLUDES})
```
wgtmac, line 134: @raulcd I think I just copied the same thing from Apache Arrow. Do you have any idea why it was added there?

HuaHuaY: I read arrow's cmake files. It doesn't use the parameter `EXTRA_INCLUDES` of `add_arrow_lib`. So maybe we misunderstand the meaning of `EXTRA_INCLUDES` and we should modify the caller of `add_iceberg_lib` instead of here.

wgtmac: Adding these header files as build time dependencies makes it easier for `libiceberg-bundle` and unit test executables to find them. But I also think we don't need to add `SYSTEM` to them in this case.

raulcd: This was added ~7 years ago here:
https://github.com/apache/arrow/commit/5c97cd6bfd447dcc3d8d2b8e6debe6b92b8c1b42

I also think `SYSTEM` is not required, we could try removing it on ARROW too and validate CI, Python bindings builds, etcetera.

PR #212, opened 2025-09-04
feat: define table properties with default values

wgtmac:
Just copied everything from the TableProperties.java as of today

### DIFF
--- a/src/iceberg/table_properties.h
+++ b/src/iceberg/table_properties.h
@@ -242,9 +242,6 @@ class ICEBERG_EXPORT TableProperties : public ConfigBase<TableProperties> {
   inline static Entry<int64_t> kDeleteTargetFileSizeBytes{
       "write.delete.target-file-size-bytes", int64_t{64} * 1024 * 1024};  // 64 MB
 
-  inline static Entry<bool> kSnapshotIdInheritanceEnabled{
-      "compatibility.snapshot-id-inheritance.enabled", false};
-
   // Garbage collection properties
 
   inline static Entry<bool> kGcEnabled{"gc.enabled", true};


src/iceberg/table_properties.h
```diff
@@ -0,0 +1,303 @@
...
+                                                    "gzip"};
+  inline static Entry<std::string> kDeleteAvroCompression{
+      "write.delete.avro.compression-codec", "gzip"};
+  inline static Entry<std::string> kAvroCompressionLevel{"write.avro.compression-level",
+                                                         ""};
+  inline static Entry<std::string> kDeleteAvroCompressionLevel{
+      "write.delete.avro.compression-level", ""};
+
+  // ORC properties
+  inline static Entry<int64_t> kOrcStripeSizeBytes{"write.orc.stripe-size-bytes",
```
Fokko, line 163: Why are we going with int64 here, and with int32 in Parquet? Int32 seems to be sufficient?

src/iceberg/table_properties.h
```diff
@@ -0,0 +1,303 @@
...
+  inline static std::string_view kDefaultNameMapping{"schema.name-mapping.default"};
+
+  inline static Entry<bool> kWriteAuditPublishEnabled{"write.wap.enabled", false};
+  inline static Entry<int64_t> kWriteTargetFileSizeBytes{
+      "write.target-file-size-bytes", int64_t{512} * 1024 * 1024};  // 512 MB
+  inline static Entry<int64_t> kDeleteTargetFileSizeBytes{
+      "write.delete.target-file-size-bytes", int64_t{64} * 1024 * 1024};  // 64 MB
+
+  inline static Entry<bool> kSnapshotIdInheritanceEnabled{
+      "compatibility.snapshot-id-inheritance.enabled", false};
```
Fokko: I would not backport this part. It was used before V2 was released to test out the snapshot inheritance, but I don't think anyone is really using it since upgrading to V2 is easier.

src/iceberg/table_properties.h
wgtmac: These align with the Java definitions.

src/iceberg/table_properties.h
```diff
@@ -0,0 +1,303 @@
...
+
```
wgtmac:
```suggestion
```

src/iceberg/table_properties.h
wgtmac: Just removed it. Thanks for the information!

PR #215, opened 2025-09-07
chore: use ninja for ci build

zhjwpku:
Our CI runs quite slow, especially on windows, I observed task being terminated when it hit 30 minutes, not sure if that's the limit of gh actions.

By using Ninja build, CI goes a lot faster, in general, it can save us more that 10 minutes in general. For specific jobs, below are some numbers(unit minutes):

- ASAN 18 -> 7
- Cpp linter 17 -> 7
- Test Mac 13 -> 5
- Test Ubuntu 17 -> 7
- Test Windows 29 -> 17


wgtmac (approved):
I think this makes sense. Thanks a lot @zhjwpku!

cc @HuaHuaY as you have observed the same issue last week.

Fokko (approved):
Nice, this is great! This is especially helpful for first time contributors where we have to trigger the CI. Thanks @zhjwpku for working on this, and thanks @wgtmac for the review 🙌

PR #216, opened 2025-09-09
feat: manifest writer and adapter impl part2

dongxiao1198:
1 parse v1v2v3 manifest schema in adapter
2 convert ManifestFile&ManifestEntry into ArrowArray 
3 add e2e case

### DIFF
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -48,6 +48,10 @@ set(ICEBERG_SOURCES
     manifest_reader.cc
     manifest_reader_internal.cc
     manifest_writer.cc
+    manifest_adapter.cc
+    v1_metadata.cc
+    v2_metadata.cc
+    v3_metadata.cc
     arrow_c_data_guard_internal.cc
     util/murmurhash3_internal.cc
     util/timepoint.cc
--- a/src/iceberg/manifest_adapter.cc
+++ b/src/iceberg/manifest_adapter.cc
@@ -1,0 +1,702 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/manifest_adapter.h"
+
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/util/macros.h"
+#include "nanoarrow/nanoarrow.h"
+
+namespace iceberg {
+
+Status ManifestAdapter::StartAppending() {
+  if (size_ > 0) {
+    return InvalidArgument("Adapter buffer not empty, cannot start appending.");
+  }
+  if (is_initialized_) {
+    // reset buffer
+    ArrowArrayRelease(&array_);
+  }
+  array_ = {};
+  size_ = 0;
+  ArrowError error;
+  auto status = ArrowArrayInitFromSchema(&array_, &schema_, &error);
+  NANOARROW_RETURN_IF_NOT_OK(status, error);
+  status = ArrowArrayStartAppending(&array_);
+  NANOARROW_RETURN_IF_NOT_OK(status, error);
+  is_initialized_ = true;
+  return {};
+}
+
+Result<ArrowArray> ManifestAdapter::FinishAppending() {
+  ArrowError error;
+  auto status = ArrowArrayFinishBuildingDefault(&array_, &error);
+  NANOARROW_RETURN_IF_NOT_OK(status, error);
+  return array_;
+}
+
+Status ManifestAdapter::AppendField(ArrowArray* arrowArray, int64_t value) {
+  auto status = ArrowArrayAppendInt(arrowArray, value);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestAdapter::AppendField(ArrowArray* arrowArray, uint64_t value) {
+  auto status = ArrowArrayAppendUInt(arrowArray, value);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestAdapter::AppendField(ArrowArray* arrowArray, double value) {
+  auto status = ArrowArrayAppendDouble(arrowArray, value);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestAdapter::AppendField(ArrowArray* arrowArray, std::string_view value) {
+  ArrowStringView view(value.data(), value.size());
+  auto status = ArrowArrayAppendString(arrowArray, view);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestAdapter::AppendField(ArrowArray* arrowArray,
+                                    const std::vector<uint8_t>& value) {
+  ArrowBufferViewData data;
+  data.as_char = reinterpret_cast<const char*>(value.data());
+  ArrowBufferView view(data, value.size());
+  auto status = ArrowArrayAppendBytes(arrowArray, view);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+ManifestEntryAdapter::~ManifestEntryAdapter() {
+  // arrow::ImportedArrayData::Release() bridge.cc:1478 will release the
+  // internal array, so we have no need to release it here.
+  //  if (is_initialized_) {
+  //   ArrowArrayRelease(&array_);
+  //  }
+}
+
+std::shared_ptr<StructType> ManifestEntryAdapter::GetManifestEntryStructType() {
+  return ManifestEntry::TypeFromPartitionType(std::move(partition_schema_));
+}
+
+Status ManifestEntryAdapter::AppendPartitions(
+    ArrowArray* arrow_array, const std::shared_ptr<StructType>& partition_type,
+    const std::vector<Literal>& partitions) {
+  if (arrow_array->n_children != partition_type->fields().size()) {
+    return InvalidManifest("Partition arrow not match partition type.");
+  }
+  auto fields = partition_type->fields();
+
+  for (const auto& partition : partitions) {
+    for (int32_t i = 0; i < fields.size(); i++) {
+      const auto& field = fields[i];
+      auto array = arrow_array->children[i];
+      if (partition.IsNull()) {
+        auto status = ArrowArrayAppendNull(array, 1);
+        NANOARROW_RETURN_IF_FAILED(status);
+        continue;
+      }
+      switch (field.type()->type_id()) {
+        case TypeId::kBoolean:
+          ICEBERG_RETURN_UNEXPECTED(AppendField(
+              array, static_cast<uint64_t>(
+                         std::get<bool>(partition.value()) == true ? 1L : 0L)));
+          break;
+        case TypeId::kInt:
+          ICEBERG_RETURN_UNEXPECTED(AppendField(
+              array, static_cast<int64_t>(std::get<int32_t>(partition.value()))));
+          break;
+        case TypeId::kLong:
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, std::get<int64_t>(partition.value())));
+          break;
+        case TypeId::kFloat:
+          ICEBERG_RETURN_UNEXPECTED(AppendField(
+              array, static_cast<double>(std::get<float>(partition.value()))));
+          break;
+        case TypeId::kDouble:
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, std::get<double>(partition.value())));
+          break;
+        case TypeId::kString:
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, std::get<std::string>(partition.value())));
+          break;
+        case TypeId::kBinary:
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, std::get<std::vector<uint8_t>>(partition.value())));
+          break;
+        case TypeId::kDate:
+          ICEBERG_RETURN_UNEXPECTED(AppendField(
+              array, static_cast<int64_t>(std::get<int32_t>(partition.value()))));
+          break;
+        case TypeId::kTime:
+        case TypeId::kTimestamp:
+        case TypeId::kTimestampTz:
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, std::get<int64_t>(partition.value())));
+          break;
+        case TypeId::kDecimal:
+        case TypeId::kUuid:
+        case TypeId::kFixed:
+          // TODO(xiao.dong) currently literal does not support those types
+        default:
+          return InvalidManifest("Unsupported partition type: {}", field.ToString());
+      }
+    }
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestEntryAdapter::AppendList(ArrowArray* arrow_array,
+                                        const std::vector<int32_t>& list_value) {
+  auto list_array = arrow_array->children[0];
+  for (const auto& value : list_value) {
+    auto status = ArrowArrayAppendInt(list_array, static_cast<int64_t>(value));
+    NANOARROW_RETURN_IF_FAILED(status);
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestEntryAdapter::AppendList(ArrowArray* arrow_array,
+                                        const std::vector<int64_t>& list_value) {
+  auto list_array = arrow_array->children[0];
+  for (const auto& value : list_value) {
+    auto status = ArrowArrayAppendInt(list_array, value);
+    NANOARROW_RETURN_IF_FAILED(status);
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestEntryAdapter::AppendMap(ArrowArray* arrow_array,
+                                       const std::map<int32_t, int64_t>& map_value) {
+  auto map_array = arrow_array->children[0];
+  if (map_array->n_children != 2) {
+    return InvalidManifest("Invalid map array.");
+  }
+  for (const auto& [key, value] : map_value) {
+    auto key_array = map_array->children[0];
+    auto value_array = map_array->children[1];
+    ICEBERG_RETURN_UNEXPECTED(AppendField(key_array, static_cast<int64_t>(key)));
+    ICEBERG_RETURN_UNEXPECTED(AppendField(value_array, value));
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestEntryAdapter::AppendMap(
+    ArrowArray* arrow_array, const std::map<int32_t, std::vector<uint8_t>>& map_value) {
+  auto map_array = arrow_array->children[0];
+  if (map_array->n_children != 2) {
+    return InvalidManifest("Invalid map array.");
+  }
+  for (const auto& [key, value] : map_value) {
+    auto key_array = map_array->children[0];
+    auto value_array = map_array->children[1];
+    ICEBERG_RETURN_UNEXPECTED(AppendField(key_array, static_cast<int64_t>(key)));
+    ICEBERG_RETURN_UNEXPECTED(AppendField(value_array, value));
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Status ManifestEntryAdapter::AppendDataFile(
+    ArrowArray* arrow_array, const std::shared_ptr<StructType>& data_file_type,
+    const std::shared_ptr<DataFile>& file) {
+  auto fields = data_file_type->fields();
+  for (int32_t i = 0; i < fields.size(); i++) {
+    const auto& field = fields[i];
+    auto array = arrow_array->children[i];
+
+    switch (field.field_id()) {
+      case 134:  // content (optional int32)
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendField(array, static_cast<int64_t>(file->content)));
+        break;
+      case 100:  // file_path (required string)
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, file->file_path));
+        break;
+      case 101:  // file_format (required string)
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, ToString(file->file_format)));
+        break;
+      case 102:  // partition (required struct)
+      {
+        auto partition_type = std::dynamic_pointer_cast<StructType>(field.type());
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendPartitions(array, partition_type, file->partition));
+      } break;
+      case 103:  // record_count (required int64)
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, file->record_count));
+        break;
+      case 104:  // file_size_in_bytes (required int64)
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, file->file_size_in_bytes));
+        break;
+      case 105:  // block_size_in_bytes (compatible in v1)
+        // always 64MB for v1
+        static const int64_t kBlockSizeInBytes = 64 * 1024 * 1024L;
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, kBlockSizeInBytes));
+        break;
+      case 108:  // column_sizes (optional map)
+        ICEBERG_RETURN_UNEXPECTED(AppendMap(array, file->column_sizes));
+        break;
+      case 109:  // value_counts (optional map)
+        ICEBERG_RETURN_UNEXPECTED(AppendMap(array, file->value_counts));
+        break;
+      case 110:  // null_value_counts (optional map)
+        ICEBERG_RETURN_UNEXPECTED(AppendMap(array, file->null_value_counts));
+        break;
+      case 137:  // nan_value_counts (optional map)
+        ICEBERG_RETURN_UNEXPECTED(AppendMap(array, file->nan_value_counts));
+        break;
+      case 125:  // lower_bounds (optional map)
+        ICEBERG_RETURN_UNEXPECTED(AppendMap(array, file->lower_bounds));
+        break;
+      case 128:  // upper_bounds (optional map)
+        ICEBERG_RETURN_UNEXPECTED(AppendMap(array, file->upper_bounds));
+        break;
+      case 131:  // key_metadata (optional binary)
+        if (!file->key_metadata.empty()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(array, file->key_metadata));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 132:  // split_offsets (optional list)
+        ICEBERG_RETURN_UNEXPECTED(AppendList(array, file->split_offsets));
+        break;
+      case 135:  // equality_ids (optional list)
+        ICEBERG_RETURN_UNEXPECTED(AppendList(array, file->equality_ids));
+        break;
+      case 140:  // sort_order_id (optional int32)
+        if (file->sort_order_id.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, static_cast<int64_t>(file->sort_order_id.value())));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 142:  // first_row_id (optional int64)
+        if (file->first_row_id.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(array, file->first_row_id.value()));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 143:  // referenced_data_file (optional string)
+      {
+        ICEBERG_ASSIGN_OR_RAISE(auto referenced_data_file,
+                                GetWrappedReferenceDataFile(file));
+        if (referenced_data_file.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(array, referenced_data_file.value()));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      }
+      case 144:  // content_offset (optional int64)
+        if (file->content_offset.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(array, file->content_offset.value()));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 145:  // content_size_in_bytes (optional int64)
+        if (file->content_size_in_bytes.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, file->content_size_in_bytes.value()));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      default:
+        return InvalidManifest("Unknown data file field id: {} ", field.field_id());
+    }
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Result<std::optional<int64_t>> ManifestEntryAdapter::GetWrappedSequenceNumber(
+    const iceberg::ManifestEntry& entry) {
+  return entry.sequence_number;
+}
+
+Result<std::optional<std::string>> ManifestEntryAdapter::GetWrappedReferenceDataFile(
+    const std::shared_ptr<DataFile>& file) {
+  return file->referenced_data_file;
+}
+
+Result<std::optional<int64_t>> ManifestEntryAdapter::GetWrappedFirstRowId(
+    const std::shared_ptr<DataFile>& file) {
+  return file->first_row_id;
+}
+
+Result<std::optional<int64_t>> ManifestEntryAdapter::GetWrappedContentOffset(
+    const std::shared_ptr<DataFile>& file) {
+  return file->content_offset;
+}
+
+Result<std::optional<int64_t>> ManifestEntryAdapter::GetWrappedContentSizeInBytes(
+    const std::shared_ptr<DataFile>& file) {
+  return file->content_size_in_bytes;
+}
+
+Status ManifestEntryAdapter::AppendInternal(const iceberg::ManifestEntry& entry) {
+  const auto& fields = manifest_schema_->fields();
+  for (int32_t i = 0; i < fields.size(); i++) {
+    const auto& field = fields[i];
+    auto array = array_.children[i];
+
+    switch (field.field_id()) {
+      case 0:  // status (required int32)
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendField(array, static_cast<int64_t>(static_cast<int32_t>(entry.status))));
+        break;
+      case 1:  // snapshot_id (optional int64)
+        if (entry.snapshot_id.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(array, entry.snapshot_id.value()));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 2:  // data_file (required struct)
+        if (entry.data_file) {
+          // Get the data file type from the field
+          auto data_file_type = std::dynamic_pointer_cast<StructType>(field.type());
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendDataFile(array, data_file_type, entry.data_file));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 3:  // sequence_number (optional int64)
+      {
+        ICEBERG_ASSIGN_OR_RAISE(auto sequence_num, GetWrappedSequenceNumber(entry));
+        if (sequence_num.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(array, sequence_num.value()));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      }
+      case 4:  // file_sequence_number (optional int64)
+        if (entry.file_sequence_number.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, entry.file_sequence_number.value()));
+        } else {
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      default:
+        return InvalidManifest("Unknown manifest entry field id: {}", field.field_id());
+    }
+  }
+
+  auto status = ArrowArrayFinishElement(&array_);
+  NANOARROW_RETURN_IF_FAILED(status);
+  size_++;
+  return {};
+}
+
+Status ManifestEntryAdapter::InitSchema(const std::unordered_set<int32_t>& fields_ids) {
+  auto manifest_entry_schema = GetManifestEntryStructType();
+  auto fields_span = manifest_entry_schema->fields();
+  std::vector<SchemaField> fields;
+  // TODO(xiao.dong) make this a common function to recursive handle
+  // all nested fields in schema
+  for (const auto& field : fields_span) {
+    if (field.field_id() == 2) {
+      // handle data_file field
+      auto data_file_struct = std::dynamic_pointer_cast<StructType>(field.type());
+      std::vector<SchemaField> data_file_fields;
+      for (const auto& data_file_field : data_file_struct->fields()) {
+        if (fields_ids.contains(data_file_field.field_id())) {
+          data_file_fields.emplace_back(data_file_field);
+        }
+      }
+      auto type = std::make_shared<StructType>(data_file_fields);
+      auto data_file_field = SchemaField::MakeRequired(
+          field.field_id(), std::string(field.name()), std::move(type));
+      fields.emplace_back(std::move(data_file_field));
+    } else {
+      if (fields_ids.contains(field.field_id())) {
+        fields.emplace_back(field);
+      }
+    }
+  }
+  manifest_schema_ = std::make_shared<Schema>(fields);
+  ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*manifest_schema_, &schema_));
+  return {};
+}
+
+ManifestFileAdapter::~ManifestFileAdapter() {
+  // arrow::ImportedArrayData::Release() bridge.cc:1478 will release the
+  // internal array, so we have no need to release it here.
+  //  if (is_initialized_) {
+  //    ArrowArrayRelease(&array_);
+  //  }
+}
+
+Status ManifestFileAdapter::AppendPartitions(
+    ArrowArray* arrow_array, const std::shared_ptr<ListType>& partition_type,
+    const std::vector<PartitionFieldSummary>& partitions) {
+  auto& array = arrow_array->children[0];
+  if (array->n_children != 4) {
+    return InvalidManifestList("Invalid partition array.");
+  }
+  auto partition_struct =
+      std::dynamic_pointer_cast<StructType>(partition_type->fields()[0].type());
+  auto fields = partition_struct->fields();
+  for (const auto& partition : partitions) {
+    for (const auto& field : fields) {
+      switch (field.field_id()) {
+        case 509:  // contains_null (required bool)
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array->children[0],
+                          static_cast<uint64_t>(partition.contains_null ? 1 : 0)));
+          break;
+        case 518:  // contains_nan (optional bool)
+        {
+          auto field_array = array->children[1];
+          if (partition.contains_nan.has_value()) {
+            ICEBERG_RETURN_UNEXPECTED(AppendField(
+                field_array,
+                static_cast<uint64_t>(partition.contains_nan.value() ? 1 : 0)));
+          } else {
+            auto status = ArrowArrayAppendNull(field_array, 1);
+            NANOARROW_RETURN_IF_FAILED(status);
+          }
+          break;
+        }
+        case 510:  // lower_bound (optional binary)
+        {
+          auto field_array = array->children[2];
+          if (partition.lower_bound.has_value() && !partition.lower_bound->empty()) {
+            ICEBERG_RETURN_UNEXPECTED(
+                AppendField(field_array, partition.lower_bound.value()));
+          } else {
+            auto status = ArrowArrayAppendNull(field_array, 1);
+            NANOARROW_RETURN_IF_FAILED(status);
+          }
+          break;
+        }
+        case 511:  // upper_bound (optional binary)
+        {
+          auto field_array = array->children[3];
+          if (partition.upper_bound.has_value() && !partition.upper_bound->empty()) {
+            ICEBERG_RETURN_UNEXPECTED(
+                AppendField(field_array, partition.upper_bound.value()));
+          } else {
+            auto status = ArrowArrayAppendNull(field_array, 1);
+            NANOARROW_RETURN_IF_FAILED(status);
+          }
+          break;
+        }
+        default:
+          return InvalidManifestList("Unknown field id: {}", field.field_id());
+      }
+    }
+    auto status = ArrowArrayFinishElement(array);
+    NANOARROW_RETURN_IF_FAILED(status);
+  }
+
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Result<int64_t> ManifestFileAdapter::GetWrappedSequenceNumber(
+    const iceberg::ManifestFile& file) {
+  return file.sequence_number;
+}
+
+Result<int64_t> ManifestFileAdapter::GetWrappedMinSequenceNumber(
+    const iceberg::ManifestFile& file) {
+  return file.min_sequence_number;
+}
+
+Result<std::optional<int64_t>> ManifestFileAdapter::GetWrappedFirstRowId(
+    const iceberg::ManifestFile& file) {
+  return file.first_row_id;
+}
+
+Status ManifestFileAdapter::AppendInternal(const iceberg::ManifestFile& file) {
+  const auto& fields = manifest_list_schema_->fields();
+  for (int32_t i = 0; i < fields.size(); i++) {
+    const auto& field = fields[i];
+    auto array = array_.children[i];
+    switch (field.field_id()) {
+      case 500:  // manifest_path
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, file.manifest_path));
+        break;
+      case 501:  // manifest_length
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, file.manifest_length));
+        break;
+      case 502:  // partition_spec_id
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendField(array, static_cast<int64_t>(file.partition_spec_id)));
+        break;
+      case 517:  // content
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendField(array, static_cast<int64_t>(static_cast<int32_t>(file.content))));
+        break;
+      case 515:  // sequence_number
+      {
+        ICEBERG_ASSIGN_OR_RAISE(auto sequence_num, GetWrappedSequenceNumber(file));
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, sequence_num));
+        break;
+      }
+      case 516:  // min_sequence_number
+      {
+        ICEBERG_ASSIGN_OR_RAISE(auto min_sequence_num, GetWrappedMinSequenceNumber(file));
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, min_sequence_num));
+        break;
+      }
+      case 503:  // added_snapshot_id
+        ICEBERG_RETURN_UNEXPECTED(AppendField(array, file.added_snapshot_id));
+        break;
+      case 504:  // added_files_count
+        if (file.added_files_count.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, static_cast<int64_t>(file.added_files_count.value())));
+        } else {
+          // Append null for optional field
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 505:  // existing_files_count
+        if (file.existing_files_count.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(
+              array, static_cast<int64_t>(file.existing_files_count.value())));
+        } else {
+          // Append null for optional field
+          auto status = ArrowArrayAppendNull(array, 1);
+          NANOARROW_RETURN_IF_FAILED(status);
+        }
+        break;
+      case 506:  // deleted_files_count
+        if (file.deleted_files_count.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(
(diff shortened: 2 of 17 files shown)


src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,100 @@ class ICEBERG_EXPORT ManifestAdapter {
...
-  virtual Status StartAppending() = 0;
-  virtual Result<ArrowArray> FinishAppending() = 0;
+  Status StartAppending();
+  Result<ArrowArray> FinishAppending();
   int64_t size() const { return size_; }
+  virtual std::shared_ptr<Schema> schema() const = 0;
+
+ protected:
+  Status AppendField(ArrowArray* arrowArray, int64_t value);
+  Status AppendField(ArrowArray* arrowArray, uint64_t value);
```
HeartLinked: Why these function are independent of the class's fields and looks like a static method, but it is not defined as static？

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,704 @@
...
+          break;
+        case TypeId::kTime:
+        case TypeId::kTimestamp:
+        case TypeId::kTimestampTz:
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, std::get<int64_t>(partition.value())));
+          break;
+        case TypeId::kDecimal:
+        case TypeId::kUuid:
+        case TypeId::kFixed:
```
HeartLinked: Only Decimal and Uuid are not supported yet? Fixed should be as same as Binary(using vector<uint8_t>)?

src/iceberg/v1_metadata.cc
```diff
@@ -0,0 +1,91 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/v1_metadata.h"
+
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/partition_spec.h"
```
HeartLinked:
```suggestion
```

src/iceberg/v1_metadata.cc
```diff
@@ -0,0 +1,91 @@
...
+  metadata_["format-version"] = "1";
+  return InitSchema(compatible_fields);
+}
+
+Status ManifestEntryAdapterV1::Append(const iceberg::ManifestEntry& entry) {
+  return AppendInternal(entry);
+}
+
+std::shared_ptr<StructType> ManifestEntryAdapterV1::GetManifestEntryStructType() {
+  // V1 compatible fields. Official suggestions:
```
HeartLinked:
```suggestion
  // 'block_size_in_bytes' (ID 105) is a deprecated field that is REQUIRED in the v1 data_file schema for backward compatibility.
```

test/manifest_reader_writer_test.cc
```diff
@@ -28,7 +26,12 @@
 #include "iceberg/avro/avro_register.h"
 #include "iceberg/manifest_entry.h"
 #include "iceberg/manifest_list.h"
+#include "iceberg/manifest_reader.h"
+#include "iceberg/manifest_writer.h"
+#include "iceberg/partition_spec.h"
```
HeartLinked:
```suggestion
```

src/iceberg/manifest_adapter.h
dongxiao1198: Yes those func should be defined in static, already fixed.

src/iceberg/manifest_adapter.cc
dongxiao1198: Nested type was missed，they should be rejected. 
Actually decimal and fixed are both allowed for partition literal I will fix this.
But for uuid, here stores the partition values after transform,  so it should not  be supported.

src/iceberg/util/macros.h
```diff
@@ -36,3 +36,13 @@
 #define ICEBERG_ASSIGN_OR_RAISE(lhs, rexpr)                                             \
   ICEBERG_ASSIGN_OR_RAISE_IMPL(ICEBERG_ASSIGN_OR_RAISE_NAME(result_, __COUNTER__), lhs, \
                                rexpr)
+
+#define NANOARROW_RETURN_IF_FAILED(status)                       \
```
wgtmac: All public macros need to be prefixed with `ICEBERG_`. BTW, we need to hide any symbol of nanoarrow so they must not appear in this file. Can we do something similar to `iceberg/arrow/arrow_error_transform_internal.h` and add `src/iceberg/nanoarrow_error_transform_internal.h`?

src/iceberg/CMakeLists.txt
```diff
@@ -48,6 +48,10 @@ set(ICEBERG_SOURCES
     manifest_reader.cc
     manifest_reader_internal.cc
     manifest_writer.cc
+    manifest_adapter.cc
+    v1_metadata.cc
+    v2_metadata.cc
+    v3_metadata.cc
```
wgtmac: Could you help sort these files alphabetically?

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
   ManifestAdapter() = default;
   virtual ~ManifestAdapter() = default;
+  virtual Status Init() = 0;
 
-  virtual Status StartAppending() = 0;
-  virtual Result<ArrowArray> FinishAppending() = 0;
+  Status StartAppending();
+  Result<ArrowArray> FinishAppending();
   int64_t size() const { return size_; }
+  virtual std::shared_ptr<Schema> schema() const = 0;
```
wgtmac:
```suggestion
  virtual const std::shared_ptr<Schema>& schema() const = 0;
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+
+ protected:
+  static Status AppendField(ArrowArray* arrowArray, int64_t value);
+  static Status AppendField(ArrowArray* arrowArray, uint64_t value);
+  static Status AppendField(ArrowArray* arrowArray, double value);
+  static Status AppendField(ArrowArray* arrowArray, std::string_view value);
+  static Status AppendField(ArrowArray* arrowArray, const std::vector<uint8_t>& value);
+  static Status AppendField(ArrowArray* arrowArray, const std::array<uint8_t, 16>& value);
```
wgtmac:
```suggestion
  static Status AppendField(ArrowArray* arrowArray, std::span<const uint8_t> value);
```

It seems that we can simplify these two into one?

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
 };
 
 // \brief Implemented by different versions with different schemas to
 // append a list of `ManifestEntry`s to an `ArrowArray`.
 class ICEBERG_EXPORT ManifestEntryAdapter : public ManifestAdapter {
  public:
-  ManifestEntryAdapter() = default;
-  ~ManifestEntryAdapter() override = default;
+  explicit ManifestEntryAdapter(std::shared_ptr<Schema> partition_schema,
+                                std::shared_ptr<PartitionSpec> partition_spec)
```
wgtmac: Is `std::shared_ptr<PartitionSpec> partition_spec` enough? I think we miss the Java method `public StructType PartitionSpec.partitionType()` to return the partition schema directly from it.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+      : partition_spec_(std::move(partition_spec)),
```
wgtmac: It is strange that the orders of input parameters and member variables mismatch.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+        partition_schema_(std::move(partition_schema)) {}
+  ~ManifestEntryAdapter() override;
 
   virtual Status Append(const ManifestEntry& entry) = 0;
+
+  std::shared_ptr<Schema> schema() const override { return manifest_schema_; }
```
wgtmac:
```suggestion
  const std::shared_ptr<Schema>& schema() const override { return manifest_schema_; }
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+
+ protected:
+  virtual std::shared_ptr<StructType> GetManifestEntryStructType();
```
wgtmac:
```suggestion
  virtual const std::shared_ptr<StructType>& GetManifestEntryStructType();
```

Is it possible to return const reference?

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+  Status InitSchema(const std::unordered_set<int32_t>& fields_ids);
+  Status AppendInternal(const ManifestEntry& entry);
+  Status AppendDataFile(ArrowArray* arrow_array,
+                        const std::shared_ptr<StructType>& data_file_type,
+                        const std::shared_ptr<DataFile>& file);
+  static Status AppendPartitions(ArrowArray* arrow_array,
```
wgtmac:
```suggestion
  static Status AppendPartition(ArrowArray* arrow_array,
```

This is a single partition value so we need to avoid using plural form.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+  static Status AppendList(ArrowArray* arrow_array,
+                           const std::vector<int32_t>& list_value);
+  static Status AppendList(ArrowArray* arrow_array,
+                           const std::vector<int64_t>& list_value);
+  static Status AppendMap(ArrowArray* arrow_array,
+                          const std::map<int32_t, int64_t>& map_value);
+  static Status AppendMap(ArrowArray* arrow_array,
+                          const std::map<int32_t, std::vector<uint8_t>>& map_value);
+
+  virtual Result<std::optional<int64_t>> GetWrappedSequenceNumber(
```
wgtmac: Why calling them wrapped? Can we simply name it `GetSequenceNumber`? Same question for below.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +40,104 @@ class ICEBERG_EXPORT ManifestAdapter {
...
 // append a list of `ManifestFile`s to an `ArrowArray`.
 class ICEBERG_EXPORT ManifestFileAdapter : public ManifestAdapter {
  public:
   ManifestFileAdapter() = default;
-  ~ManifestFileAdapter() override = default;
+  ~ManifestFileAdapter() override;
 
   virtual Status Append(const ManifestFile& file) = 0;
+
+  std::shared_ptr<Schema> schema() const override { return manifest_list_schema_; }
```
wgtmac:
```suggestion
  const std::shared_ptr<Schema>& schema() const override { return manifest_list_schema_; }
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/util/macros.h"
+#include "nanoarrow/nanoarrow.h"
+
+namespace {
+static constexpr int64_t kBlockSizeInBytes = 64 * 1024 * 1024L;
```
wgtmac:
```suggestion
static constexpr int64_t kBlockSizeInBytesV1 = 64 * 1024 * 1024L;
```

Some functions use the same name. So let's just rename it for v1 explicitly.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+  if (size_ > 0) {
+    return InvalidArgument("Adapter buffer not empty, cannot start appending.");
+  }
+  array_ = {};
+  size_ = 0;
+  ArrowError error;
+  auto status = ArrowArrayInitFromSchema(&array_, &schema_, &error);
+  NANOARROW_RETURN_IF_NOT_OK(status, error);
+  status = ArrowArrayStartAppending(&array_);
+  NANOARROW_RETURN_IF_NOT_OK(status, error);
```
wgtmac:
```suggestion
  NANOARROW_RETURN_IF_NOT_OK(ArrowArrayInitFromSchema(&array_, &schema_, &error), error);
  NANOARROW_RETURN_IF_NOT_OK(ArrowArrayStartAppending(&array_), error);
```

We can make it shorter?

wgtmac: Same question applies to below.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+ManifestEntryAdapter::~ManifestEntryAdapter() {
+  if (is_initialized_) {
+    // arrow::ImportedArrayData::Release() bridge.cc:1478 will release the
+    // internal array, so we have no need to release it here.
+    // ArrowArrayRelease(&array_);
+    ArrowSchemaRelease(&schema_);
```
wgtmac, line 147: We have `ArrowSchemaGuard` added by you. Can we reuse that?

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+  data.as_char = reinterpret_cast<const char*>(value.data());
+  ArrowBufferView view(data, value.size());
+  auto status = ArrowArrayAppendBytes(arrowArray, view);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+ManifestEntryAdapter::~ManifestEntryAdapter() {
+  if (is_initialized_) {
+    // arrow::ImportedArrayData::Release() bridge.cc:1478 will release the
```
wgtmac: If any error occurs, we may not have called `arrow::ImportedArrayData` so we still need to guard it?

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+    // internal array, so we have no need to release it here.
+    // ArrowArrayRelease(&array_);
+    ArrowSchemaRelease(&schema_);
+  }
+}
+
+std::shared_ptr<StructType> ManifestEntryAdapter::GetManifestEntryStructType() {
```
wgtmac: OK, so it cannot return a const reference. Please ignore my previous comment about this.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+  return ManifestEntry::TypeFromPartitionType(std::move(partition_schema_));
+}
+
+Status ManifestEntryAdapter::AppendPartitions(
+    ArrowArray* arrow_array, const std::shared_ptr<StructType>& partition_type,
+    const std::vector<Literal>& partitions) {
+  if (arrow_array->n_children != partition_type->fields().size()) {
+    return InvalidManifest("Partition arrow not match partition type.");
```
wgtmac:
```suggestion
    return InvalidArrowData("Arrow array of partition does not match partition type.");
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+}
+
+std::shared_ptr<StructType> ManifestEntryAdapter::GetManifestEntryStructType() {
+  return ManifestEntry::TypeFromPartitionType(std::move(partition_schema_));
+}
+
+Status ManifestEntryAdapter::AppendPartitions(
+    ArrowArray* arrow_array, const std::shared_ptr<StructType>& partition_type,
+    const std::vector<Literal>& partitions) {
+  if (arrow_array->n_children != partition_type->fields().size()) {
```
wgtmac:
```suggestion
  if (arrow_array->n_children != partition_type->fields().size()) [[unlikely]] {
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+          break;
+        case TypeId::kDecimal:
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendField(array, std::get<std::array<uint8_t, 16>>(partition.value())));
+          break;
+        case TypeId::kUuid:
+        case TypeId::kStruct:
+        case TypeId::kList:
+        case TypeId::kMap:
+          // TODO(xiao.dong) currently literal does not support those types
```
wgtmac: FYI: I may use `StructLike` from https://github.com/apache/iceberg-cpp/pull/181 to replace the partition literals.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+
+Status ManifestEntryAdapter::AppendPartitions(
+    ArrowArray* arrow_array, const std::shared_ptr<StructType>& partition_type,
+    const std::vector<Literal>& partitions) {
+  if (arrow_array->n_children != partition_type->fields().size()) {
+    return InvalidManifest("Partition arrow not match partition type.");
+  }
+  auto fields = partition_type->fields();
+
+  for (const auto& partition : partitions) {
```
wgtmac: I'm confused here. Isn't `partitions` is for a single row of partition values? If true, `partitions.size()` should match `fields.size()`, then `partition[i]` should be appended to `arrow_array->children[i]` as type of `fields[i]`.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+        case TypeId::kList:
+        case TypeId::kMap:
+          // TODO(xiao.dong) currently literal does not support those types
+        default:
+          return InvalidManifest("Unsupported partition type: {}", field.ToString());
+      }
+    }
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
```
wgtmac:
```suggestion
  NANOARROW_RETURN_IF_FAILED(ArrowArrayFinishElement(arrow_array));
```

wgtmac: Same for below

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+        return InvalidManifest("Unknown data file field id: {} ", field.field_id());
+    }
+  }
+  auto status = ArrowArrayFinishElement(arrow_array);
+  NANOARROW_RETURN_IF_FAILED(status);
+  return {};
+}
+
+Result<std::optional<int64_t>> ManifestEntryAdapter::GetWrappedSequenceNumber(
+    const iceberg::ManifestEntry& entry) {
```
wgtmac:
```suggestion
    const ManifestEntry& entry) {
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+Status ManifestEntryAdapter::InitSchema(const std::unordered_set<int32_t>& fields_ids) {
+  auto manifest_entry_schema = GetManifestEntryStructType();
+  auto fields_span = manifest_entry_schema->fields();
+  std::vector<SchemaField> fields;
+  // TODO(xiao.dong) make this a common function to recursive handle
+  // all nested fields in schema
+  for (const auto& field : fields_span) {
+    if (field.field_id() == 2) {
+      // handle data_file field
+      auto data_file_struct = std::dynamic_pointer_cast<StructType>(field.type());
```
wgtmac:
```suggestion
      auto data_file_struct = internal::checked_pointer_cast<StructType>(field.type());
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+  }
+  manifest_schema_ = std::make_shared<Schema>(fields);
+  ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*manifest_schema_, &schema_));
+  is_initialized_ = true;
+  return {};
+}
+
+ManifestFileAdapter::~ManifestFileAdapter() {
+  if (is_initialized_) {
+    // arrow::ImportedArrayData::Release() bridge.cc:1478 will release the
```
wgtmac: Same question as above.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,720 @@
...
+    }
+  }
+
+  auto status = ArrowArrayFinishElement(&array_);
+  NANOARROW_RETURN_IF_FAILED(status);
+  size_++;
+  return {};
+}
+
+Status ManifestEntryAdapter::InitSchema(const std::unordered_set<int32_t>& fields_ids) {
```
wgtmac, line 436: Perhaps add a minor comment for `fields_ids`. It took me a while to understand for it is for.

src/iceberg/v1_metadata.cc
```diff
@@ -0,0 +1,91 @@
...
+      DataFile::kKeyMetadata, DataFile::kSplitOffsets, DataFile::kSortOrderId});
+
+  return std::make_shared<StructType>(
+      std::vector<SchemaField>{ManifestEntry::kStatus, ManifestEntry::kSnapshotId,
+                               SchemaField::MakeRequired(2, ManifestEntry::kDataFileField,
+                                                         std::move(datafile_type))});
+}
+
+Status ManifestFileAdapterV1::Init() {
+  static std::unordered_set<int32_t> compatible_fields{
```
wgtmac:
```suggestion
  static const std::unordered_set<int32_t> kManifestFileFieldIds{
```

src/iceberg/v1_metadata.cc
```diff
@@ -0,0 +1,91 @@
...
+Status ManifestEntryAdapterV1::Append(const iceberg::ManifestEntry& entry) {
+  return AppendInternal(entry);
+}
+
+std::shared_ptr<StructType> ManifestEntryAdapterV1::GetManifestEntryStructType() {
+  // 'block_size_in_bytes' (ID 105) is a deprecated field that is REQUIRED
+  // in the v1 data_file schema for backward compatibility.
+  // Deprecated. Always write a default in v1. Do not write in v2 or v3.
+  static const SchemaField kBlockSizeInBytes = SchemaField::MakeRequired(
+      105, "block_size_in_bytes", iceberg::int64(), "Block size in bytes");
```
wgtmac:
```suggestion
      105, "block_size_in_bytes", int64(), "Block size in bytes");
```

src/iceberg/v1_metadata.cc
```diff
@@ -0,0 +1,91 @@
...
+#include "iceberg/v1_metadata.h"
+
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+
+namespace iceberg {
+
+Status ManifestEntryAdapterV1::Init() {
+  static std::unordered_set<int32_t> compatible_fields{
```
wgtmac:
```suggestion
  static const std::unordered_set<int32_t> kManifestEntryFieldIds{
```

src/iceberg/v1_metadata.cc
```diff
@@ -0,0 +1,91 @@
...
+      std::vector<SchemaField>{ManifestEntry::kStatus, ManifestEntry::kSnapshotId,
+                               SchemaField::MakeRequired(2, ManifestEntry::kDataFileField,
+                                                         std::move(datafile_type))});
+}
+
+Status ManifestFileAdapterV1::Init() {
+  static std::unordered_set<int32_t> compatible_fields{
+      500, 501, 502, 503, 504, 505, 506, 512, 513, 514, 507, 519,
+  };
+  metadata_["snapshot-id"] = std::to_string(snapshot_id_);
```
wgtmac, line 105: Should we define these in `iceberg/constant.h` instead of magic strings?

src/iceberg/util/macros.h
dongxiao1198: OK I will handle this

src/iceberg/manifest_adapter.h
dongxiao1198: Not enough now，since PartitionSpec currently don't have a individual partition schema. PartitionSpec.Schema is the full schema of table, we can mark a todo here.

dongxiao1198: Oh， I see the impl of java, it's ok for us. I will add a transformer to get this schema from spec directly

src/iceberg/manifest_adapter.h
dongxiao1198: partition_schema was removed

src/iceberg/manifest_adapter.h
dongxiao1198: this has been refactored into Result<xxx>

src/iceberg/manifest_adapter.h
dongxiao1198: v1 directly return seq from manifest
v2 v3 has some special check for it in ManifestFileAdapterV2::GetWrappedSequenceNumber

src/iceberg/manifest_adapter.cc
dongxiao1198: sure

src/iceberg/manifest_adapter.cc
dongxiao1198: Writer::Write(ArrowArray array) will internal release the object, but since the `array` is a copyed obj not a ref, the soure object of array which hold by the caller is not changed(the release func will set array->release =NULL to mar this is already released).
I need to refactor the writer function param to Writer::Write(ArrowArray& array) to fix this

src/iceberg/manifest_adapter.cc
dongxiao1198: But I think this is already a dtor, there is no need to import other guard

src/iceberg/manifest_adapter.cc
dongxiao1198: is_initialized_ was removed since after refactor of those code we can do it like:
  if (array_ != nullptr && array_->release != nullptr) {
    ArrowArrayRelease(array_.get());
  }
  if (schema_.release != nullptr) {
    ArrowSchemaRelease(&schema_);
  }

src/iceberg/v1_metadata.cc
dongxiao1198: may be like this: ManifestFile::kManifestLength.field_id()

src/iceberg/manifest_adapter.cc
dongxiao1198: Yes. This is a bug, and the ut has only one partition field so can not find this bug.
I will fix this and more ut need to be added.

dongxiao1198: we need to check
arrow_array->n_children  ==  partition_type->fields().size() == partitions.size()

src/iceberg/arrow/nanoarrow_error_transform_internal.h
```diff
@@ -0,0 +1,25 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#define ICEBERG_NANOARROW_RETURN_IF_NOT_OK(status, error)              \
+  if (status != NANOARROW_OK) [[unlikely]] {                           \
+    return InvalidArrowData("Nanoarrow error msg: {}", error.message); \
```
wgtmac:
```suggestion
    return InvalidArrowData("nanoarrow error: {}", error.message); \
```

src/iceberg/file_writer.h
```diff
@@ -65,7 +65,7 @@ class ICEBERG_EXPORT Writer {
   /// \brief Write arrow data to the file.
   ///
   /// \return Status of write results.
-  virtual Status Write(ArrowArray data) = 0;
+  virtual Status Write(ArrowArray& data) = 0;
```
wgtmac: `ArrowArray` is a simple C struct and will be moved away (be invalid) after the call. Therefore it looks a little bit strange to use a reference here. Perhaps changing it to `ArrowArray*` and document the behavior?

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,682 @@
...
+#include "iceberg/manifest_adapter.h"
+
+#include "iceberg/arrow/nanoarrow_error_transform_internal.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/macros.h"
+#include "nanoarrow/nanoarrow.h"
```
wgtmac:
```suggestion
#include <nanoarrow/nanoarrow.h>
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
-  ArrowArray array_;
+  static Status AppendField(ArrowArray* arrowArray, int64_t value);
+  static Status AppendField(ArrowArray* arrowArray, uint64_t value);
+  static Status AppendField(ArrowArray* arrowArray, double value);
+  static Status AppendField(ArrowArray* arrowArray, std::string_view value);
+  static Status AppendField(ArrowArray* arrowArray,
+                            const std::span<const uint8_t>& value);
+
+ protected:
+  std::shared_ptr<ArrowArray> array_;
```
wgtmac: It is not recommended to use smart pointers on a C struct, especially when we don't add a custom deleter. Can we revert this line?

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,682 @@
...
+
+#define NANOARROW_RETURN_IF_FAILED(status)                       \
+  if (status != NANOARROW_OK) [[unlikely]] {                     \
+    return InvalidArrowData("Nanoarrow error code: {}", status); \
+  }
```
wgtmac:
```suggestion
```

We should use `ICEBERG_NANOARROW_RETURN_IF_NOT_OK` consistently.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,682 @@
...
+Status ManifestAdapter::StartAppending() {
+  if (size_ > 0) {
+    return InvalidArgument("Adapter buffer not empty, cannot start appending.");
+  }
+  array_ = std::make_shared<ArrowArray>();
+  size_ = 0;
+  ArrowError error;
+  ICEBERG_NANOARROW_RETURN_IF_NOT_OK(
+      ArrowArrayInitFromSchema(array_.get(), &schema_, &error), error);
+  NANOARROW_RETURN_IF_FAILED(ArrowArrayStartAppending(array_.get()));
```
wgtmac:
```suggestion
  ICEBERG_NANOARROW_RETURN_IF_NOT_OK(ArrowArrayStartAppending(array_.get()));
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,682 @@
...
+        break;
+      case TypeId::kTime:
+      case TypeId::kTimestamp:
+      case TypeId::kTimestampTz:
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendField(array, std::get<int64_t>(partition.value())));
+        break;
+      case TypeId::kDecimal:
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendField(array, std::get<std::array<uint8_t, 16>>(partition.value())));
```
wgtmac:
```suggestion
            AppendField(array, std::get<int128_t>(partition.value())));
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
  public:
   ManifestAdapter() = default;
   virtual ~ManifestAdapter() = default;
+  virtual Status Init() = 0;
 
-  virtual Status StartAppending() = 0;
-  virtual Result<ArrowArray> FinishAppending() = 0;
+  Status StartAppending();
+  Result<std::shared_ptr<ArrowArray>> FinishAppending();
```
wgtmac: As commented below, I don't think `std::shared_ptr<ArrowArray>` is a good idea.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
   int64_t size() const { return size_; }
 
  protected:
-  ArrowArray array_;
+  static Status AppendField(ArrowArray* arrowArray, int64_t value);
+  static Status AppendField(ArrowArray* arrowArray, uint64_t value);
+  static Status AppendField(ArrowArray* arrowArray, double value);
+  static Status AppendField(ArrowArray* arrowArray, std::string_view value);
+  static Status AppendField(ArrowArray* arrowArray,
+                            const std::span<const uint8_t>& value);
```
wgtmac:
```suggestion
                            std::span<const uint8_t> value);
```

std::span is a view so we don't need pass by const ref.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+  explicit ManifestEntryAdapter(std::shared_ptr<PartitionSpec> partition_spec)
+      : partition_spec_(std::move(partition_spec)) {}
+  ~ManifestEntryAdapter() override;
 
   virtual Status Append(const ManifestEntry& entry) = 0;
+
+  const std::shared_ptr<Schema>& schema() const { return manifest_schema_; }
+
+ protected:
+  virtual Result<std::shared_ptr<StructType>> GetManifestEntryStructType();
```
wgtmac:
```suggestion
  virtual Result<std::shared_ptr<StructType>> GetManifestEntryType();
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+
+  /// \brief Init version-specific schema for each version.
+  ///
+  /// \param fields_ids each version of manifest schema has schema, we will init this
+  /// schema based on the fields_ids.
+  Status InitSchema(const std::unordered_set<int32_t>& fields_ids);
+  Status AppendInternal(const ManifestEntry& entry);
+  Status AppendDataFile(ArrowArray* arrow_array,
+                        const std::shared_ptr<StructType>& data_file_type,
+                        const std::shared_ptr<DataFile>& file);
```
wgtmac:
```suggestion
                        const StructType& data_file_type,
                        const DataFile& file);
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+
+  virtual Result<std::optional<int64_t>> GetSequenceNumber(const ManifestEntry& entry);
+  virtual Result<std::optional<std::string>> GetWrappedReferenceDataFile(
+      const std::shared_ptr<DataFile>& file);
+  virtual Result<std::optional<int64_t>> GetWrappedFirstRowId(
+      const std::shared_ptr<DataFile>& file);
+  virtual Result<std::optional<int64_t>> GetWrappedContentOffset(
+      const std::shared_ptr<DataFile>& file);
+  virtual Result<std::optional<int64_t>> GetWrappedContentSizeInBytes(
+      const std::shared_ptr<DataFile>& file);
```
wgtmac:
```suggestion
  virtual Result<std::optional<std::string>> GetWrappedReferenceDataFile(
      const DataFile& file);
  virtual Result<std::optional<int64_t>> GetWrappedFirstRowId(
      const DataFile& file);
  virtual Result<std::optional<int64_t>> GetWrappedContentOffset(
      const DataFile& file);
  virtual Result<std::optional<int64_t>> GetWrappedContentSizeInBytes(
      const DataFile& file);
```

Should we also remove `Wrapped` from these functions?

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+  /// \param fields_ids each version of manifest schema has schema, we will init this
+  /// schema based on the fields_ids.
+  Status InitSchema(const std::unordered_set<int32_t>& fields_ids);
+  Status AppendInternal(const ManifestEntry& entry);
+  Status AppendDataFile(ArrowArray* arrow_array,
+                        const std::shared_ptr<StructType>& data_file_type,
+                        const std::shared_ptr<DataFile>& file);
+  static Status AppendPartition(ArrowArray* arrow_array,
+                                const std::shared_ptr<StructType>& partition_type,
+                                const std::vector<Literal>& partitions);
```
wgtmac:
```suggestion
  static Status AppendPartition(ArrowArray* arrow_array,
                                const StructType& partition_type,
                                const std::vector<Literal>& partition_value);
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+  /// \param fields_ids each version of manifest schema has schema, we will init this
+  /// schema based on the fields_ids.
+  Status InitSchema(const std::unordered_set<int32_t>& fields_ids);
+  Status AppendInternal(const ManifestFile& file);
+  static Status AppendPartitions(ArrowArray* arrow_array,
+                                 const std::shared_ptr<ListType>& partition_type,
+                                 const std::vector<PartitionFieldSummary>& partitions);
+
+  virtual Result<int64_t> GetSequenceNumber(const ManifestFile& file);
+  virtual Result<int64_t> GetWrappedMinSequenceNumber(const ManifestFile& file);
```
wgtmac: Same question for removing `Wrapped` from the name

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+ protected:
+  /// \brief Init version-specific schema for each version.
+  ///
+  /// \param fields_ids each version of manifest schema has schema, we will init this
+  /// schema based on the fields_ids.
+  Status InitSchema(const std::unordered_set<int32_t>& fields_ids);
+  Status AppendInternal(const ManifestFile& file);
+  static Status AppendPartitions(ArrowArray* arrow_array,
+                                 const std::shared_ptr<ListType>& partition_type,
+                                 const std::vector<PartitionFieldSummary>& partitions);
```
wgtmac:
```suggestion
  static Status AppendPartitionSummary(ArrowArray* arrow_array,
                                       const ListType& partition_summary_type,
                                       const std::vector<PartitionFieldSummary>& summaries);
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,682 @@
...
+    auto array = arrow_array->children[i];
+    if (partition.IsNull()) {
+      NANOARROW_RETURN_IF_FAILED(ArrowArrayAppendNull(array, 1));
+      continue;
+    }
+    switch (field.type()->type_id()) {
+      case TypeId::kBoolean:
+        ICEBERG_RETURN_UNEXPECTED(AppendField(
+            array,
+            static_cast<uint64_t>(std::get<bool>(partition.value()) == true ? 1L : 0L)));
```
wgtmac:
```suggestion
            static_cast<uint64_t>(std::get<bool>(partition.value()) ? 1L : 0L)));
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,682 @@
...
+        }
+        break;
+      case 2:  // data_file (required struct)
+        if (entry.data_file) {
+          // Get the data file type from the field
+          auto data_file_type = internal::checked_pointer_cast<StructType>(field.type());
+          ICEBERG_RETURN_UNEXPECTED(
+              AppendDataFile(array, data_file_type, entry.data_file));
+        } else {
+          NANOARROW_RETURN_IF_FAILED(ArrowArrayAppendNull(array, 1));
```
wgtmac: We should return error since data_file is required.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,682 @@
...
+  NANOARROW_RETURN_IF_FAILED(ArrowArrayFinishElement(array_.get()));
+  size_++;
+  return {};
+}
+
+Status ManifestEntryAdapter::InitSchema(const std::unordered_set<int32_t>& fields_ids) {
+  ICEBERG_ASSIGN_OR_RAISE(auto manifest_entry_schema, GetManifestEntryStructType())
+  auto fields_span = manifest_entry_schema->fields();
+  std::vector<SchemaField> fields;
+  // TODO(xiao.dong) make this a common function to recursive handle
```
wgtmac:
```suggestion
  // TODO(xiao.dong) make this a common function to recursively handle
```

src/iceberg/partition_spec.cc
```diff
@@ -57,6 +60,49 @@ int32_t PartitionSpec::spec_id() const { return spec_id_; }
 
 std::span<const PartitionField> PartitionSpec::fields() const { return fields_; }
 
+Result<std::shared_ptr<Schema>> PartitionSpec::partition_schema() {
+  if (fields_.empty()) {
+    return nullptr;
+  }
+  {
+    std::scoped_lock<std::mutex> lock(mutex_);
```
wgtmac, line 68: Can we use `std::once_flag` and `std::call_once` instead?

src/iceberg/partition_spec.h
```diff
@@ -67,6 +69,8 @@ class ICEBERG_EXPORT PartitionSpec : public util::Formattable {
   /// \brief Get a view of the partition fields.
   std::span<const PartitionField> fields() const;
 
+  Result<std::shared_ptr<Schema>> partition_schema();
```
```suggestion
  Result<std::shared_ptr<Schema>> GetPartitionSchema();
```

This is not a trivial getter so we cannot use the snake case form.

src/iceberg/partition_spec.cc
```diff
@@ -57,6 +60,49 @@ int32_t PartitionSpec::spec_id() const { return spec_id_; }
...
+                            partition_field.transform()->Bind(source_field_type));
+
+    auto result_type = transform_function->ResultType();
+
+    // Create the partition field with the transform result type
+    // Partition fields are always optional (can be null)
+    partition_fields.emplace_back(partition_field.field_id(),
+                                  std::string(partition_field.name()),
+                                  std::move(result_type),
+                                  true  // optional
```
```suggestion
                                  /*optional=*/true
```

src/iceberg/partition_spec.cc
```diff
@@ -57,6 +60,49 @@ int32_t PartitionSpec::spec_id() const { return spec_id_; }
...
+    }
+  }
+
+  std::vector<SchemaField> partition_fields;
+  for (const auto& partition_field : fields_) {
+    // Get the source field from the original schema by source_id
+    ICEBERG_ASSIGN_OR_RAISE(auto source_field,
+                            schema_->FindFieldById(partition_field.source_id()));
+    if (!source_field.has_value()) {
+      return InvalidSchema("Cannot find source field for partition field:{}",
```
wgtmac, line 82: Add a TODO comment to use unknown type when source field is missing.

src/iceberg/partition_spec.h
wgtmac: BTW, can we add a test case for this?

src/iceberg/test/manifest_reader_writer_test.cc
```diff
@@ -144,6 +166,20 @@ TEST_F(ManifestReaderV1Test, PartitionedTest) {
...
+  auto partition_schema =
+      std::make_shared<Schema>(std::vector<SchemaField>({partition_field}));
+  auto identity_transform = Transform::Identity();
+  std::vector<PartitionField> fields{
+      PartitionField(1000, 1000, "order_ts_hour", identity_transform)};
+  auto partition_spec = std::make_shared<PartitionSpec>(partition_schema, 1, fields);
+
+  auto expected_entries = PreparePartitionedTestData();
+  auto write_manifest_path = CreateNewTempFilePath();
+  TestWriteManifest(write_manifest_path, partition_spec, expected_entries);
```
wgtmac, line 191: `TestManifestReadingByPath` is missing? Same for below.

src/iceberg/v2_metadata.cc
```diff
@@ -0,0 +1,157 @@
...
+
+namespace iceberg {
+
+Status ManifestEntryAdapterV2::Init() {
+  static std::unordered_set<int32_t> kManifestEntryFieldIds{
+      ManifestEntry::kStatus.field_id(),
+      ManifestEntry::kSnapshotId.field_id(),
+      ManifestEntry::kDataFileFieldId,
+      ManifestEntry::kSequenceNumber.field_id(),
+      ManifestEntry::kFileSequenceNumber.field_id(),
```
wgtmac:
```suggestion
      ManifestEntry::kSequenceNumber.field_id(),
      ManifestEntry::kFileSequenceNumber.field_id(),
      ManifestEntry::kDataFileFieldId,
```

Reorder them to match the actual schema.

src/iceberg/test/manifest_reader_writer_test.cc
```diff
@@ -133,6 +141,20 @@ class ManifestReaderV1Test : public ManifestReaderTestBase {
     }
     return manifest_entries;
   }
+
+  void TestWriteManifest(const std::string& manifest_list_path,
+                         std::shared_ptr<PartitionSpec> partition_spec,
+                         const std::vector<ManifestEntry>& manifest_entries) {
+    std::cout << "Writing manifest list to " << manifest_list_path << std::endl;
```
wgtmac: remove this?

src/iceberg/test/manifest_reader_writer_test.cc
```diff
@@ -209,6 +246,20 @@ class ManifestReaderV2Test : public ManifestReaderTestBase {
   std::vector<ManifestEntry> PrepareMetadataInheritanceTestData() {
     return CreateV2TestData(/*sequence_number=*/15, /*partition_spec_id*/ 12);
   }
+
+  void TestWriteManifest(int64_t snapshot_id, const std::string& manifest_list_path,
+                         std::shared_ptr<PartitionSpec> partition_spec,
+                         const std::vector<ManifestEntry>& manifest_entries) {
+    std::cout << "Writing manifest list to " << manifest_list_path << std::endl;
```
wgtmac: remove it?

src/iceberg/test/manifest_reader_writer_test.cc
```diff
@@ -144,6 +166,21 @@ TEST_F(ManifestReaderV1Test, PartitionedTest) {
                       partition_schema);
 }
 
+TEST_F(ManifestReaderV1Test, WritePartitionedTest) {
+  iceberg::SchemaField partition_field(1000, "order_ts_hour", iceberg::int32(), true);
+  auto partition_schema =
+      std::make_shared<Schema>(std::vector<SchemaField>({partition_field}));
+  auto identity_transform = Transform::Identity();
+  std::vector<PartitionField> fields{
+      PartitionField(1000, 1000, "order_ts_hour", identity_transform)};
```
wgtmac: Usually `source_id` and `field_id` cannot be the same.

src/iceberg/test/manifest_reader_writer_test.cc
```diff
@@ -230,4 +281,21 @@ TEST_F(ManifestReaderV2Test, MetadataInheritanceTest) {
...
+  TestWriteManifest(679879563479918846LL, write_manifest_path, nullptr, expected_entries);
+  TestManifestReadingByPath(write_manifest_path, expected_entries);
+}
+
+TEST_F(ManifestReaderV2Test, WriteInheritancePartitionedTest) {
+  auto expected_entries = PrepareMetadataInheritanceTestData();
+  auto write_manifest_path = CreateNewTempFilePath();
+  TestWriteManifest(679879563479918846LL, write_manifest_path, nullptr, expected_entries);
+  for (auto& entry : expected_entries) {
+    entry.data_file->partition_spec_id = PartitionSpec::kInitialSpecId;
```
wgtmac: Why do we need this? Isn't it the default value?

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+  Result<ArrowArray*> FinishAppending();
   int64_t size() const { return size_; }
 
+ protected:
+  static Status AppendField(ArrowArray* arrowArray, int64_t value);
+  static Status AppendField(ArrowArray* arrowArray, uint64_t value);
+  static Status AppendField(ArrowArray* arrowArray, double value);
+  static Status AppendField(ArrowArray* arrowArray, std::string_view value);
+  static Status AppendField(ArrowArray* arrowArray,
+                            const std::span<const uint8_t>& value);
```
wgtmac:
```suggestion
  static Status AppendField(ArrowArray* arrowArray,
                            std::span<const uint8_t> value);
```

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+                                 const std::shared_ptr<ListType>& partition_type,
+                                 const std::vector<PartitionFieldSummary>& partitions);
+
+  virtual Result<int64_t> GetSequenceNumber(const ManifestFile& file);
+  virtual Result<int64_t> GetWrappedMinSequenceNumber(const ManifestFile& file);
+  virtual Result<std::optional<int64_t>> GetWrappedFirstRowId(const ManifestFile& file);
+
+ protected:
+  std::shared_ptr<Schema> manifest_list_schema_;
+  std::unordered_map<std::string, std::string> metadata_;
```
wgtmac: nit: these two variables can be moved to the base class.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+
+  virtual Result<std::optional<int64_t>> GetSequenceNumber(const ManifestEntry& entry);
+  virtual Result<std::optional<std::string>> GetWrappedReferenceDataFile(
+      const std::shared_ptr<DataFile>& file);
+  virtual Result<std::optional<int64_t>> GetWrappedFirstRowId(
+      const std::shared_ptr<DataFile>& file);
+  virtual Result<std::optional<int64_t>> GetWrappedContentOffset(
+      const std::shared_ptr<DataFile>& file);
+  virtual Result<std::optional<int64_t>> GetWrappedContentSizeInBytes(
+      const std::shared_ptr<DataFile>& file);
```
wgtmac:
```suggestion
  virtual Result<std::optional<std::string>> GetReferenceDataFile(
      const DataFile& file);
  virtual Result<std::optional<int64_t>> GetFirstRowId(
      const DataFile& file);
  virtual Result<std::optional<int64_t>> GetContentOffset(
      const DataFile& file);
  virtual Result<std::optional<int64_t>> GetContentSizeInBytes(
      const DataFile& file);
```

Let's be consistent with `GetSequenceNumber` to use const reference as input and remove `Wrapped`.

src/iceberg/manifest_adapter.h
```diff
@@ -33,34 +41,107 @@ class ICEBERG_EXPORT ManifestAdapter {
...
+  /// schema based on the fields_ids.
+  Status InitSchema(const std::unordered_set<int32_t>& fields_ids);
+  Status AppendInternal(const ManifestFile& file);
+  static Status AppendPartitions(ArrowArray* arrow_array,
+                                 const std::shared_ptr<ListType>& partition_type,
+                                 const std::vector<PartitionFieldSummary>& partitions);
+
+  virtual Result<int64_t> GetSequenceNumber(const ManifestFile& file);
+  virtual Result<int64_t> GetWrappedMinSequenceNumber(const ManifestFile& file);
+  virtual Result<std::optional<int64_t>> GetWrappedFirstRowId(const ManifestFile& file);
```
wgtmac:
```suggestion
  virtual Result<int64_t> GetMinSequenceNumber(const ManifestFile& file);
  virtual Result<std::optional<int64_t>> GetFirstRowId(const ManifestFile& file);
```

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,685 @@
...
+#include "iceberg/manifest_list.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/macros.h"
+
+#define NANOARROW_RETURN_IF_FAILED(status)                  \
+  if (status != NANOARROW_OK) [[unlikely]] {                \
+    return InvalidArrowData("nanoarrow error: {}", status); \
+  }
```
wgtmac:
```suggestion
```

Remove it by using `ICEBERG_NANOARROW_RETURN_IF_NOT_OK`

src/iceberg/arrow/nanoarrow_error_transform_internal.h
```diff
@@ -0,0 +1,25 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#define ICEBERG_NANOARROW_RETURN_IF_NOT_OK(status, error)          \
+  if (status != NANOARROW_OK) [[unlikely]] {                       \
+    return InvalidArrowData("nanoarrow error: {}", error.message); \
```
wgtmac:
```suggestion
    return iceberg::InvalidArrowData("nanoarrow error: {}", error.message); \
```

wgtmac: Do we want to include headers that are used here?

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,685 @@
...
+  if (array_.release != nullptr) {
+    ArrowArrayRelease(&array_);
+  }
+  if (schema_.release != nullptr) {
+    ArrowSchemaRelease(&schema_);
+  }
+}
+
+Result<std::shared_ptr<StructType>> ManifestEntryAdapter::GetManifestEntryStructType() {
+  if (partition_spec_ == nullptr) {
```
wgtmac: This shouldn't happen. Even for an unpartitioned table, it has a special partition spec.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,685 @@
...
+
+  for (int32_t i = 0; i < fields.size(); i++) {
+    const auto& partition = partitions[i];
+    const auto& field = fields[i];
+    auto array = arrow_array->children[i];
+    if (partition.IsNull()) {
+      NANOARROW_RETURN_IF_FAILED(ArrowArrayAppendNull(array, 1));
+      continue;
+    }
+    switch (field.type()->type_id()) {
```
wgtmac: We might need to add more test cases for different partition types. Could you please add a TODO comment in the test file?

src/iceberg/manifest_adapter.cc
wgtmac: This is not resolved yet.

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,685 @@
...
+    ArrowArrayRelease(&array_);
+  }
+  if (schema_.release != nullptr) {
+    ArrowSchemaRelease(&schema_);
+  }
+}
+
+Status ManifestFileAdapter::AppendPartitions(
+    ArrowArray* arrow_array, const std::shared_ptr<ListType>& partition_type,
+    const std::vector<PartitionFieldSummary>& partitions) {
```
wgtmac:
```suggestion
Status ManifestFileAdapter::AppendPartitionSummary(
    ArrowArray* arrow_array, const std::shared_ptr<ListType>& partition_type,
    const std::vector<PartitionFieldSummary>& summaries) {
```

It is misleading to call it `partitions`

src/iceberg/manifest_adapter.cc
```diff
@@ -0,0 +1,685 @@
...
+  auto& array = arrow_array->children[0];
+  if (array->n_children != 4) {
+    return InvalidManifestList("Invalid partition array.");
+  }
+  auto partition_struct =
+      internal::checked_pointer_cast<StructType>(partition_type->fields()[0].type());
+  auto fields = partition_struct->fields();
+  for (const auto& partition : partitions) {
+    for (const auto& field : fields) {
```
wgtmac: This looks wrong. There is one summary per partition field.

src/iceberg/manifest_adapter.h
wgtmac: I think `StructType` is redundant.

src/iceberg/partition_spec.cc
wgtmac: This is not resolved yet.

src/iceberg/partition_spec.h
```diff
@@ -67,6 +69,8 @@ class ICEBERG_EXPORT PartitionSpec : public util::Formattable {
   /// \brief Get a view of the partition fields.
   std::span<const PartitionField> fields() const;
 
+  Result<std::shared_ptr<Schema>> GetPartitionSchema();
```
```suggestion
  Result<std::shared_ptr<StructType>> PartitionType();
```

It is an overkill to use Schema here because it does not need the schema_id.

src/iceberg/partition_spec.h
```diff
@@ -83,6 +87,8 @@ class ICEBERG_EXPORT PartitionSpec : public util::Formattable {
   const int32_t spec_id_;
   std::vector<PartitionField> fields_;
   int32_t last_assigned_field_id_;
+  std::mutex mutex_;
+  std::shared_ptr<Schema> partition_schema_;
```
wgtmac: Same as above.

src/iceberg/partition_spec.h
wgtmac: BTW, we need to add test case for this. Perhaps add a TODO comment in the partition_spec_test.cc?

src/iceberg/manifest_adapter.cc
dongxiao1198: there is some different between them. I put them together like:
#define ICEBERG_NANOARROW_RETURN_IF_NOT_OK(status)          \
  if (status != NANOARROW_OK) [[unlikely]] {                \
    return InvalidArrowData("nanoarrow error: {}", status); \
  }

#define ICEBERG_NANOARROW_RETURN_IF_NOT_OK_WITH_ERROR(status, error)              \
  if (status != NANOARROW_OK) [[unlikely]] {                                      \
    return InvalidArrowData("nanoarrow error: {} msg:{}", status, error.message); \
  }

src/iceberg/test/manifest_reader_writer_test.cc
dongxiao1198: Now we have to ManifestReader::Make(), the one which use manifest file path to create write will not set the partition_spec_id to it, I may use the other creator instead to avoid this.

src/iceberg/manifest_adapter.cc
dongxiao1198: I will add a notice here and keep this protect as unlikely

src/iceberg/manifest_adapter.cc
dongxiao1198: this field is the partition summary's schema field, not the table partition field

src/iceberg/manifest_adapter.cc
dongxiao1198: current case was based on the previous reader case. I have added a TODO to add more reader writer case together

src/iceberg/partition_spec.h
dongxiao1198: Case added

src/iceberg/manifest_adapter.h
dongxiao1198: metadata_ moved to base.  manifest_list_schema_ and manifest_schema_ has different means so leaved here

PR #223, opened 2025-09-11
fix: address feedback from initial release

wgtmac:
- suppress warnings from verify_rc.sh
- fix release.sh not to rename artifacts
- refactor wordings in the email templates

### DIFF
--- a/dev/release/release.sh
+++ b/dev/release/release.sh
@@ -63,13 +63,6 @@ svn \
 svn co "${dist_url}/${release_id}"
 pushd "${release_id}"
 
-echo "Renaming artifacts to their final release names..."
-for fname in ./*; do
-  mv "${fname}" "${fname//-rc${rc}/}"
-done
-echo "Renamed files:"
-ls -l
-
 gh release create "${tag}" \
   --repo "${repository}" \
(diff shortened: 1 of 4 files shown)


dev/release/verify_rc.sh
```diff
@@ -118,9 +118,11 @@ test_source_distribution() {
 
   # Configure build
   cmake -S . -B build \
+    -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations" \
```
wgtmac, line 121: I have added `--compile-no-warning-as-error` below and this one is also required. @lidavidm

dev/release/release.sh
```diff
@@ -63,13 +63,6 @@ svn \
 svn co "${dist_url}/${release_id}"
 pushd "${release_id}"
 
-echo "Renaming artifacts to their final release names..."
```
wgtmac, line 66: We don't need this hack any more. Now artifact names do not have `-rcN` suffix. @HeartLinked 

cc @liurenjie1024

HeartLinked: LGTM

Fokko (approved):

PR #227, opened 2025-09-15
feat: lazy init for Schema and StructType

nullccxsy:
…Type

- Added move and copy constructors and assignment operators for Schema and StructType to manage resource ownership and improve performance.
- Refactored field lookup methods to utilize lazy initialization with thread safety, ensuring safe concurrent access.
- Introduced unit tests for thread safety in Schema and StructType, validating concurrent operations and access patterns.

### DIFF
--- a/src/iceberg/schema.h
+++ b/src/iceberg/schema.h
@@ -83,7 +83,7 @@ class ICEBERG_EXPORT Schema : public StructType {
   Status InitNameToIdMap() const;
   Status InitLowerCaseNameToIdMap() const;
 
-  std::optional<int32_t> schema_id_;
+  const std::optional<int32_t> schema_id_;
   /// Mapping from field id to field.
   mutable std::unordered_map<int32_t, std::reference_wrapper<const SchemaField>>
       id_to_field_;
@@ -94,9 +94,9 @@ class ICEBERG_EXPORT Schema : public StructType {
   mutable std::unordered_map<std::string, int32_t, StringHash, std::equal_to<>>
       lowercase_name_to_id_;
 
-  mutable std::once_flag id_flag_;
-  mutable std::once_flag name_flag_;
-  mutable std::once_flag lowercase_name_flag_;
+  mutable std::once_flag id_to_field_flag_;
+  mutable std::once_flag name_to_id_flag_;
+  mutable std::once_flag lowercase_name_to_id_flag_;
 };
 
 }  // namespace iceberg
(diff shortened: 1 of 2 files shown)


src/iceberg/schema.h
```diff
@@ -48,6 +49,10 @@ class ICEBERG_EXPORT Schema : public StructType {
   explicit Schema(std::vector<SchemaField> fields,
                   std::optional<int32_t> schema_id = std::nullopt);
 
+  Schema(const Schema& other);
```
wgtmac: What about marking them `= delete`? Or simply ignore them as they are not auto-generated any more?

nullccxsy: Yes, if we don't use the operation of copy, there will no problem. I have modified the parts where copy operations are needed.

src/iceberg/schema.h
```diff
@@ -78,13 +79,11 @@ class ICEBERG_EXPORT Schema : public StructType {
...
   [[nodiscard]] bool Equals(const Schema& other) const;
 
-  // TODO(nullccxsy): Address potential concurrency issues in lazy initialization (e.g.,
-  // use std::call_once)
   Status InitIdToFieldMap() const;
   Status InitNameToIdMap() const;
   Status InitLowerCaseNameToIdMap() const;
 
-  const std::optional<int32_t> schema_id_;
+  std::optional<int32_t> schema_id_;
```
wgtmac: This line can be reverted.

src/iceberg/schema.h
```diff
@@ -94,6 +93,10 @@ class ICEBERG_EXPORT Schema : public StructType {
   /// Mapping from lowercased field name to field id
   mutable std::unordered_map<std::string, int32_t, StringHash, std::equal_to<>>
       lowercase_name_to_id_;
+
+  mutable std::once_flag id_flag_;
+  mutable std::once_flag name_flag_;
+  mutable std::once_flag lowercase_name_flag_;
```
wgtmac:
```suggestion
  mutable std::once_flag id_to_field_flag_;
  mutable std::once_flag name_to_id_flag_;
  mutable std::once_flag lowercase_name_to_id_flag_;
```

Fokko (approved):
Looks good, thanks @nullccxsy for working on this. In practice, we don't look up by (lowercase) name, and most of the lookups are by Field-ID. So it is good to have these initialize in a lazy fashion 👍

Fokko:
Thanks @wgtmac for the review 🚀

PR #230, opened 2025-09-16
feat: Add rest catalog http request dependency for poc

HeartLinked:


src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -22,19 +22,28 @@ set(ICEBERG_REST_SHARED_BUILD_INTERFACE_LIBS)
...
 
-list(APPEND ICEBERG_REST_STATIC_BUILD_INTERFACE_LIBS
-     "$<IF:$<TARGET_EXISTS:iceberg_static>,iceberg_static,iceberg_shared>" cpr::cpr)
-list(APPEND ICEBERG_REST_SHARED_BUILD_INTERFACE_LIBS
-     "$<IF:$<TARGET_EXISTS:iceberg_shared>,iceberg_shared,iceberg_static>" cpr::cpr)
+list(APPEND
+     ICEBERG_REST_STATIC_BUILD_INTERFACE_LIBS
+     "$<IF:$<TARGET_EXISTS:iceberg_static>,iceberg_static,iceberg_shared>"
+     cpr::cpr
+     nlohmann_json::nlohmann_json)
```
wgtmac: I don't think we need to explicitly add this. It is transitive from `libiceberg`

wgtmac: Same for below.

src/iceberg/test/rest_catalog_test.cc
```diff
@@ -0,0 +1,111 @@
...
+#include "iceberg/catalog/rest/rest_catalog.h"
+
+#include <httplib.h>
+
+#include <memory>
+#include <thread>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
```
wgtmac, line 29:
```suggestion
#include <memory>
#include <thread>

#include <httplib.h>
#include <gmock/gmock.h>
#include <gtest/gtest.h>
#include <nlohmann/json.hpp>
```

src/iceberg/test/CMakeLists.txt
```diff
@@ -15,13 +15,23 @@
 # specific language governing permissions and limitations
 # under the License.
 
+fetchcontent_declare(cpp-httplib
```
wgtmac, line 18: @WillAyd Do you have any comment?

src/iceberg/catalog/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@
 # under the License.
 
 iceberg_install_all_headers(iceberg/catalog)
+add_subdirectory(memory)
```
wgtmac, line 20:
```suggestion

add_subdirectory(memory)
```

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -34,7 +34,6 @@ list(APPEND
      ICEBERG_REST_SHARED_INSTALL_INTERFACE_LIBS
      "$<IF:$<TARGET_EXISTS:Iceberg::iceberg_shared>,Iceberg::iceberg_shared,Iceberg::iceberg_static>"
      "$<IF:$<BOOL:${CPR_VENDORED}>,Iceberg::cpr,cpr::cpr>")
-
```
wgtmac, line 37: Please revert this line

src/iceberg/test/CMakeLists.txt
WillAyd: Assuming this is in the context of Meson then no - this package is available in the wrapdb so it's easy to integrate

src/iceberg/catalog/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@
 
 iceberg_install_all_headers(iceberg/catalog)
```
WillAyd, line 18: FYI since the in_memory_catalog.h header was moved, this no longer installs anything. Maybe worth removing?

wgtmac: Good catch!

PR #235, opened 2025-09-17
feat: add Metadata method to Reader and its implementations

nullccxsy:
- Introduced a new virtual method `Metadata()` in the `Reader` class to retrieve file metadata.
- Implemented `Metadata()` in `AvroReader` to return key-value pairs from the Avro file's metadata.
- Implemented `Metadata()` in `ParquetReader` to extract and return key-value pairs from the Parquet file's metadata.

### DIFF
--- a/src/iceberg/file_reader.h
+++ b/src/iceberg/file_reader.h
@@ -55,7 +55,7 @@ class ICEBERG_EXPORT Reader {
   /// \brief Get the schema of the data.
   virtual Result<ArrowSchema> Schema() = 0;
 
-  /// \brief Get the metadata of the file
+  /// \brief Get the metadata of the file.
   virtual Result<std::unordered_map<std::string, std::string>> Metadata() = 0;
 };
 
--- a/src/iceberg/parquet/parquet_reader.cc
+++ b/src/iceberg/parquet/parquet_reader.cc
@@ -202,11 +202,7 @@ class ParquetReader::Impl {
     }
 
     std::unordered_map<std::string, std::string> metadata_map;
-    metadata_map.reserve(kv_metadata->size());
-
-    for (int i = 0; i < kv_metadata->size(); ++i) {
-      metadata_map.insert_or_assign(kv_metadata->key(i), kv_metadata->value(i));
-    }
+    kv_metadata->ToUnorderedMap(&metadata_map);
 
     return metadata_map;
   }


src/iceberg/avro/avro_reader.cc
```diff
@@ -173,6 +173,24 @@ class AvroReader::Impl {
     return arrow_schema;
   }
 
+  Result<std::unordered_map<std::string, std::string>> Metadata() {
+    if (reader_ == nullptr) {
+      return InvalidArgument("Reader is not opened");
```
wgtmac: Usually `InvalidArgument` means an invalid function parameter. Perhaps just `Invalid`?

src/iceberg/avro/avro_reader.cc
```diff
@@ -173,6 +173,24 @@ class AvroReader::Impl {
...
+    }
+
+    auto metadata = reader_->metadata();
```
```suggestion
    const auto& metadata = reader_->metadata();
```

We don't want to copy the whole map here

src/iceberg/avro/avro_reader.cc
```diff
@@ -173,6 +173,24 @@ class AvroReader::Impl {
...
+
+    std::unordered_map<std::string, std::string> metadata_map;
+    metadata_map.reserve(metadata.size());
+
+    for (const auto& pair : metadata) {
+      metadata_map.try_emplace(pair.first,
```
wgtmac: nit: can we use ranges with transform?

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -185,6 +186,31 @@ class ParquetReader::Impl {
     return arrow_schema;
   }
 
+  Result<std::unordered_map<std::string, std::string>> Metadata() {
+    if (reader_ == nullptr) {
+      return InvalidArgument("Reader is not opened");
```
wgtmac: ditto

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -185,6 +186,31 @@ class ParquetReader::Impl {
...
+    auto kv_metadata = metadata->key_value_metadata();
+    if (!kv_metadata) {
+      return std::unordered_map<std::string, std::string>{};
+    }
+
+    std::unordered_map<std::string, std::string> metadata_map;
+    metadata_map.reserve(kv_metadata->size());
+
+    for (int i = 0; i < kv_metadata->size(); ++i) {
+      metadata_map.try_emplace(kv_metadata->key(i), kv_metadata->value(i));
```
wgtmac: Do we need to check the return value?

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -185,6 +186,35 @@ class ParquetReader::Impl {
...
+      return std::unordered_map<std::string, std::string>{};
+    }
+
+    std::unordered_map<std::string, std::string> metadata_map;
+    metadata_map.reserve(kv_metadata->size());
+
+    for (int i = 0; i < kv_metadata->size(); ++i) {
+      auto [it, inserted] =
+          metadata_map.try_emplace(kv_metadata->key(i), kv_metadata->value(i));
+      if (!inserted) {
```
wgtmac: Can we just use the latter to override the former one to avoid error?

src/iceberg/avro/avro_reader.cc
```diff
@@ -173,6 +173,27 @@ class AvroReader::Impl {
...
+
+    const auto& metadata = reader_->metadata();
+
+    std::unordered_map<std::string, std::string> metadata_map;
+    metadata_map.reserve(metadata.size());
+
+    for (const auto& pair : metadata) {
+      auto [it, inserted] = metadata_map.try_emplace(
+          pair.first, std::string(pair.second.begin(), pair.second.end()));
+      if (!inserted) {
```
wgtmac: Same as the other comment.

src/iceberg/file_reader.h
```diff
@@ -54,6 +54,9 @@ class ICEBERG_EXPORT Reader {
 
   /// \brief Get the schema of the data.
   virtual Result<ArrowSchema> Schema() = 0;
+
+  /// \brief Get the metadata of the file
```
HuaHuaY:
```suggestion
  /// \brief Get the metadata of the file.
```

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -185,6 +186,31 @@ class ParquetReader::Impl {
...
+    if (!kv_metadata) {
+      return std::unordered_map<std::string, std::string>{};
+    }
+
+    std::unordered_map<std::string, std::string> metadata_map;
+    metadata_map.reserve(kv_metadata->size());
+
+    for (int i = 0; i < kv_metadata->size(); ++i) {
+      metadata_map.insert_or_assign(kv_metadata->key(i), kv_metadata->value(i));
+    }
```
HuaHuaY:
```suggestion
    kv_metadata->ToUnorderedMap(&metadata_map);
```

Issue #232, opened 2025-09-16
[DISCUSS] Selecting a C++ HTTP Client Library for REST


HeartLinked:
I'm proposing we select an HTTP client library for rest catalog. I've reviewed `libcurl`, `cpr`, and `cpp-httplib`.

I suggest we rule out `libcurl` for direct use. Its C-style API is too complex and error-prone for a modern C++23 project.This leaves us with two primary contenders that offer modern C++ wrappers: `cpr` and `cpp-httplib`.The main advantage of cpp-httplib is its simplicity. It's a header-only library with virtually zero dependencies (unless HTTPS is needed), which makes integration extremely straightforward.

However, after a deeper comparison, I believe `cpr` is the superior choice for our project for its modern and usable API, rich features. Because `cpr` is a wrapper around `libcurl`, it inherits its robustness and extensive feature set. This includes superior performance characteristics, fine-grained control over things like timeouts and authentication, and asynchronous request capabilities.

<img width="832" height="303" alt="Image" src="https://github.com/user-attachments/assets/d84ec4a3-0934-45cd-965d-b01b033a5b5b" />

I've conducted a POC with a demo using integrated `cpr` here: #230   . WDYT?

dongxiao1198:
Libcurl is a low-level common lib,has a lot of settings, but it is stable & small & no-extra dependency.
CPR is much more modern and easy to use, but itself is a 3rd lib, we need to keep upgrading it in the future(libcurl is much more stable)
Poco is also a good choice lots of rest api client use it.(but it is too heavy for us?)

Anyway, since we just play as a simple rest client, I prefer to use a small & lightweight lib such as libcurl or cpr.

PS: Can we make it pluggable？

lidavidm:
Is there any benefit to pluggability here (is there any real alternative? cronet? OS APIs?)

acking-you:
>  CPR is much more modern and easy to use, but itself is a 3rd lib, we need to keep upgrading it in the future(libcurl is much more stable) 

If package management is convenient enough, I don't think continuous upgrades are a bad thing.

zhjwpku:
Based on your description and a brief search, I'm +1 for adopting cpr, since its API is easier to use compared to libcurl's C-only interface, at this early stage, I think we should rapidly get things working, we can always revisit and refactor later if needed.

HeartLinked:
The POC for integrating cpr #236 has now passed all CI checks, successfully completing the integration work.

PR #236, opened 2025-09-18
feat: add cpr dependency for rest catalog client

HeartLinked:
Add cpr and curl dependency for REST catalog. This is related to #232 .

The current code in the REST package is just a demo example to verify the feasibility of cpr, which will be replaced in future commits. The actual content will be added in subsequent PRs so that we can keep the size of each PR reasonable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ option(ICEBERG_BUILD_STATIC "Build static library" ON)
 option(ICEBERG_BUILD_SHARED "Build shared library" OFF)
 option(ICEBERG_BUILD_TESTS "Build tests" ON)
 option(ICEBERG_BUILD_BUNDLE "Build the battery included library" ON)
+option(ICEBERG_BUILD_REST "Build rest catalog client" ON)
 option(ICEBERG_ENABLE_ASAN "Enable Address Sanitizer" OFF)
 option(ICEBERG_ENABLE_UBSAN "Enable Undefined Behavior Sanitizer" OFF)
 
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -431,87 +431,31 @@ function(resolve_zlib_dependency)
 endfunction()
 
 # ----------------------------------------------------------------------
-# CURL
+# CURL (for cpr)
 
 function(resolve_curl_dependency)
   prepare_fetchcontent()
 
-  set(BUILD_CURL_EXE
-      OFF
-      CACHE BOOL "" FORCE)
-  set(BUILD_TESTING
-      OFF
-      CACHE BOOL "" FORCE)
-  set(CURL_ENABLE_EXPORT_TARGET
-      OFF
-      CACHE BOOL "" FORCE)
-  set(BUILD_SHARED_LIBS
-      OFF
-      CACHE BOOL "" FORCE)
-  set(CURL_STATICLIB
-      ON
-      CACHE BOOL "" FORCE)
-  set(HTTP_ONLY
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_LDAP
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_LDAPS
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_RTSP
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_FTP
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_FILE
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_TELNET
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_DICT
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_TFTP
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_GOPHER
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_POP3
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_IMAP
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_SMTP
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_DISABLE_SMB
-      ON
-      CACHE BOOL "" FORCE)
-  set(CURL_CA_BUNDLE
-      "auto"
-      CACHE STRING "" FORCE)
-  set(USE_LIBIDN2
-      OFF
-      CACHE BOOL "" FORCE)
+  set(BUILD_CURL_EXE OFF)
+  set(BUILD_SHARED_LIBS OFF)
+  set(BUILD_TESTING OFF)
+  set(CURL_CA_BUNDLE "auto")
+  set(CURL_ENABLE_EXPORT_TARGET OFF)
+  set(CURL_STATICLIB ON)
+  set(HTTP_ONLY ON)
+  set(USE_LIBIDN2 OFF)
 
-  fetchcontent_declare(CURL
+  fetchcontent_declare(curl
                        ${FC_DECLARE_COMMON_OPTIONS}
-                       GIT_REPOSITORY https://github.com/curl/curl.git
-                       GIT_TAG curl-8_11_0
-                       FIND_PACKAGE_ARGS
-                       NAMES
-                       CURL
-                       CONFIG)
+                       URL https://curl.se/download/curl-8.11.0.tar.gz FIND_PACKAGE_ARGS
+                           NAMES curl)
 
   fetchcontent_makeavailable(CURL)
 
   if(curl_SOURCE_DIR)
+    set(CURL_VERSION_STRING
+        "8.11.0"
+        PARENT_SCOPE)
     if(NOT TARGET CURL::libcurl)
       add_library(CURL::libcurl INTERFACE IMPORTED)
       target_link_libraries(CURL::libcurl INTERFACE libcurl_static)
@@ -529,6 +473,9 @@ function(resolve_curl_dependency)
             ARCHIVE DESTINATION "${ICEBERG_INSTALL_LIBDIR}"
             LIBRARY DESTINATION "${ICEBERG_INSTALL_LIBDIR}")
     message(STATUS "Use vendored CURL")
+
+    # curl depends on the system installed OpenSSL
+    list(APPEND ICEBERG_SYSTEM_DEPENDENCIES OpenSSL)
   else()
     set(CURL_VENDORED FALSE)
     list(APPEND ICEBERG_SYSTEM_DEPENDENCIES CURL)
@@ -544,52 +491,26 @@ function(resolve_curl_dependency)
 endfunction()
 
 # ----------------------------------------------------------------------
-# cpr
+# cpr (C++ Requests)
 
 function(resolve_cpr_dependency)
   prepare_fetchcontent()
 
-  set(CPR_BUILD_TESTS
-      OFF
-      CACHE BOOL "" FORCE)
-  set(CPR_BUILD_TESTS_SSL
-      OFF
-      CACHE BOOL "" FORCE)
-  set(CPR_ENABLE_SSL
-      ON
-      CACHE BOOL "" FORCE)
-  set(CPR_USE_SYSTEM_CURL
-      ON
-      CACHE BOOL "" FORCE)
-  set(CPR_CURL_NOSIGNAL
-      ON
-      CACHE BOOL "" FORCE)
-
-  set(CURL_VERSION_STRING
-      "8.11.0"
-      CACHE STRING "" FORCE)
-  set(CURL_LIB
-      "CURL::libcurl"
-      CACHE STRING "" FORCE)
+  set(CPR_BUILD_TESTS OFF)
+  set(CPR_BUILD_TESTS_SSL OFF)
+  set(CPR_ENABLE_SSL ON)
+  set(CPR_USE_SYSTEM_CURL ON)
+  set(CPR_CURL_NOSIGNAL ON)
 
   fetchcontent_declare(cpr
                        ${FC_DECLARE_COMMON_OPTIONS}
-                       GIT_REPOSITORY https://github.com/libcpr/cpr.git
-                       GIT_TAG 1.11.0
-                       FIND_PACKAGE_ARGS
-                       NAMES
-                       cpr
-                       CONFIG)
-
-  # Create a custom install command that does nothing
-  # function(install)
-  #   # Do nothing - effectively disables install
-  # endfunction()
+                       URL https://github.com/libcpr/cpr/archive/refs/tags/1.12.0.tar.gz
+                           FIND_PACKAGE_ARGS
+                           NAMES
+                           cpr
+                           CONFIG)
 
   fetchcontent_makeavailable(cpr)
-
-  # Restore the original install function
-  # unset(install)
 
   if(cpr_SOURCE_DIR)
     if(NOT TARGET cpr::cpr)
@@ -641,11 +562,13 @@ resolve_croaring_dependency()
 resolve_nlohmann_json_dependency()
 resolve_spdlog_dependency()
 
-resolve_curl_dependency()
-resolve_cpr_dependency()
-
 if(ICEBERG_BUILD_BUNDLE)
   resolve_arrow_dependency()
   resolve_avro_dependency()
   resolve_zstd_dependency()
+endif()
+
+if(ICEBERG_BUILD_REST)
+  resolve_curl_dependency()
+  resolve_cpr_dependency()
 endif()
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -431,81 +431,6 @@ function(resolve_zlib_dependency)
 endfunction()
 
 # ----------------------------------------------------------------------
-# CURL (for cpr)
-
-# function(resolve_curl_dependency)
-# prepare_fetchcontent()
-
-# find_package(CURL QUIET COMPONENTS HTTPS SSL)
-
-# if(CURL_FOUND)
-#   if(CURL_VERSION VERSION_LESS "7.61.0")
-#     message(STATUS "System CURL version ${CURL_VERSION} too old, will download")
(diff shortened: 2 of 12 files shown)


cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,196 @@ function(resolve_zlib_dependency)
 
 endfunction()
 
+# ----------------------------------------------------------------------
+# CURL (for cpr)
+
+function(resolve_curl_dependency)
+  prepare_fetchcontent()
+
+  set(BUILD_CURL_EXE
```
zhjwpku: I'm not sure if a normal variable is enough here, but @wgtmac has changed the arrow dependency from cache entries to normal variables.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,196 @@ function(resolve_zlib_dependency)
...
+      ON
+      CACHE BOOL "" FORCE)
+  set(CURL_CA_BUNDLE
+      "auto"
+      CACHE STRING "" FORCE)
+  set(USE_LIBIDN2
+      OFF
+      CACHE BOOL "" FORCE)
+
+  fetchcontent_declare(CURL
```
zhjwpku: Do we need to vendor curl? Isn't that supplied by most os distributions?

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,196 @@ function(resolve_zlib_dependency)
...
+  set(CURL_VERSION_STRING
+      "8.11.0"
+      CACHE STRING "" FORCE)
+  set(CURL_LIB
+      "CURL::libcurl"
+      CACHE STRING "" FORCE)
+
+  fetchcontent_declare(cpr
+                       ${FC_DECLARE_COMMON_OPTIONS}
+                       GIT_REPOSITORY https://github.com/libcpr/cpr.git
```
zhjwpku: We might want to use the release package rather than the git repo.

cmake_modules/IcebergThirdpartyToolchain.cmake
HeartLinked: This is because we don't know whether the version supplied by os fit for our cpr verison. For example, in our cpr v1.12.0,  the minimum version requirement is 7.64.0.

HeartLinked: We use `FetchContent_Declare` with `FIND_PACKAGE_ARGS` to ensure that if a suitable version of the dependency is already installed on the system, it will be used; otherwise, FetchContent will download it as a fallback.

zhjwpku: Sure, make sense to me.

cmake_modules/IcebergThirdpartyToolchain.cmake
HeartLinked: @wgtmac WDYT?

cmake_modules/IcebergThirdpartyToolchain.cmake
HeartLinked: Thanks for the feedback. While I still believe that both methods are identical from a functional and code-consistency standpoint, I'm happy to align with the recommended best practice. I've already updated the code.

zhjwpku: Yes, code is the same, but using a release package requires less network bandwidth compared to the git repo.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,156 @@ function(resolve_zlib_dependency)
...
 
+# ----------------------------------------------------------------------
+# CURL (for cpr)
+
+function(resolve_curl_dependency)
+  prepare_fetchcontent()
+
+  set(BUILD_CURL_EXE
+      OFF
+      CACHE BOOL "" FORCE)
```
wgtmac:
```suggestion
  set(BUILD_CURL_EXE OFF)
```

This is enough. I have optimized this to use CMP0077: https://github.com/apache/iceberg-cpp/blob/main/cmake_modules/IcebergThirdpartyToolchain.cmake#L59-L61

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,156 @@ function(resolve_zlib_dependency)
...
+                       ${FC_DECLARE_COMMON_OPTIONS}
+                       URL https://curl.se/download/curl-8.11.0.tar.gz
+                           FIND_PACKAGE_ARGS
+                           NAMES
+                           CURL
+                           CONFIG)
+
+  fetchcontent_makeavailable(CURL)
+
+  if(TARGET OpenSSL::SSL)
```
wgtmac: Why do we need this? Is it vendored by curl? If this is a required library, we need to also support vendored and installed mode, otherwise it may conflict with user installed openssl.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,156 @@ function(resolve_zlib_dependency)
...
+                                                    POSITION_INDEPENDENT_CODE ON)
+    add_library(Iceberg::libcurl_static ALIAS libcurl_static)
+    install(TARGETS libcurl_static
+            EXPORT iceberg_targets
+            RUNTIME DESTINATION "${ICEBERG_INSTALL_BINDIR}"
+            ARCHIVE DESTINATION "${ICEBERG_INSTALL_LIBDIR}"
+            LIBRARY DESTINATION "${ICEBERG_INSTALL_LIBDIR}")
+    message(STATUS "Use vendored CURL")
+
+    list(APPEND ICEBERG_SYSTEM_DEPENDENCIES OpenSSL)
```
wgtmac: This duplicates with line 476 above.

cmake_modules/IcebergThirdpartyToolchain.cmake
wgtmac: It is more preferable to use the official release tarball than a git tag.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -450,6 +600,9 @@ resolve_croaring_dependency()
 resolve_nlohmann_json_dependency()
 resolve_spdlog_dependency()
 
+resolve_curl_dependency()
```
wgtmac: Please add an option to enable it, like `ICEBERG_BUILD_REST_CATALOG`.

src/iceberg/CMakeLists.txt
```diff
@@ -133,13 +133,17 @@ if(ICEBERG_BUILD_BUNDLE)
   list(APPEND
        ICEBERG_BUNDLE_STATIC_BUILD_INTERFACE_LIBS
        "$<IF:$<TARGET_EXISTS:iceberg_static>,iceberg_static,iceberg_shared>"
+       "$<IF:$<TARGET_EXISTS:CURL::libcurl>,CURL::libcurl,CURL::libcurl>"
```
wgtmac: Can we add a separate library to link with and use `ICEBERG_BUILD_REST_CATALOG` to control whether to build it? The bundled library has nothing to do with cpr and curl.

cmake_modules/IcebergThirdpartyToolchain.cmake
HeartLinked: curl depends on the system's OpenSSL, introducing a transitive dependency to our library.
We must add OpenSSL here to ensure our installed package configuration correctly tells downstream consumers to find it.

cmake_modules/IcebergThirdpartyToolchain.cmake
HeartLinked: removed

CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ option(ICEBERG_BUILD_STATIC "Build static library" ON)
 option(ICEBERG_BUILD_SHARED "Build shared library" OFF)
 option(ICEBERG_BUILD_TESTS "Build tests" ON)
 option(ICEBERG_BUILD_BUNDLE "Build the battery included library" ON)
+option(ICEBERG_BUILD_REST_CATALOG "Build REST catalog library" ON)
```
wgtmac:
```suggestion
option(ICEBERG_BUILD_REST "Build rest catalog client" ON)
```

LICENSE
```diff
@@ -316,3 +316,59 @@ freely, subject to the following restrictions:
 
 Jean-loup Gailly        Mark Adler
 jloup@gzip.org          madler@alumni.caltech.edu
+
+--------------------------------------------------------------------------------
+
+3rdparty dependency cpr is statically linked in certain binary
```
wgtmac: @jbonofre Could you help review LICENSE and NOTICE files for adding `cpr` and `curl`? BTW, do we need to add `OpenSSL` which is an indirect dependency (required by curl)?

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,149 @@ function(resolve_zlib_dependency)
...
+      CACHE BOOL "" FORCE)
+  set(HTTP_ONLY
+      ON
+      CACHE BOOL "" FORCE)
+  set(CURL_CA_BUNDLE
+      "auto"
+      CACHE STRING "" FORCE)
+  set(USE_LIBIDN2
+      OFF
+      CACHE BOOL "" FORCE)
```
wgtmac:
```suggestion
  set(BUILD_TESTING OFF)
  set(CURL_ENABLE_EXPORT_TARGET OFF)
  set(BUILD_SHARED_LIBS OFF)
  set(CURL_STATICLIB ON)
  set(HTTP_ONLY ON)
  set(CURL_CA_BUNDLE "auto")
  set(USE_LIBIDN2 OFF)
```

wgtmac: BTW, can we sort them alphabetically?

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,149 @@ function(resolve_zlib_dependency)
...
+                                                    POSITION_INDEPENDENT_CODE ON)
+    add_library(Iceberg::libcurl_static ALIAS libcurl_static)
+    install(TARGETS libcurl_static
+            EXPORT iceberg_targets
+            RUNTIME DESTINATION "${ICEBERG_INSTALL_BINDIR}"
+            ARCHIVE DESTINATION "${ICEBERG_INSTALL_LIBDIR}"
+            LIBRARY DESTINATION "${ICEBERG_INSTALL_LIBDIR}")
+    message(STATUS "Use vendored CURL")
+
+    # curl depends on the system's OpenSSL
```
wgtmac:
```suggestion
    # curl depends on the system installed OpenSSL
```

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,149 @@ function(resolve_zlib_dependency)
...
+    list(APPEND ICEBERG_SYSTEM_DEPENDENCIES OpenSSL)
+  else()
+    set(CURL_VENDORED FALSE)
+    list(APPEND ICEBERG_SYSTEM_DEPENDENCIES CURL)
```
wgtmac: Isn't `OpenSSL` required too? Could you enforce using system CURL to test building the demo example?

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,149 @@ function(resolve_zlib_dependency)
...
+      PARENT_SCOPE)
+endfunction()
+
+# ----------------------------------------------------------------------
+# cpr (C++ Requests)
+
+function(resolve_cpr_dependency)
+  prepare_fetchcontent()
+
+  set(CPR_BUILD_TESTS
```
wgtmac: ditto, we don't need CACHE and FORCE now.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,149 @@ function(resolve_zlib_dependency)
...
+      ON
+      CACHE BOOL "" FORCE)
+  set(CPR_USE_SYSTEM_CURL
+      ON
+      CACHE BOOL "" FORCE)
+  set(CPR_CURL_NOSIGNAL
+      ON
+      CACHE BOOL "" FORCE)
+
+  set(CURL_VERSION_STRING
```
wgtmac: This is an anti-pattern as we should set this in the function `resolve_curl_dependency`.

To be clear, we should set the following when `CURL_VENDORED` is true

```cmake
  set(CURL_VERSION_STRING
      "8.11.0"
      PARENT_SCOPE)
```

and use detected version of curl when using system curl.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -430,6 +430,149 @@ function(resolve_zlib_dependency)
...
+      "8.11.0"
+      CACHE STRING "" FORCE)
+  set(CURL_LIB
+      "CURL::libcurl"
+      CACHE STRING "" FORCE)
```
wgtmac:
```suggestion
```

I guess this is not required

src/iceberg/CMakeLists.txt
```diff
@@ -137,13 +137,23 @@ if(ICEBERG_BUILD_BUNDLE)
        "$<IF:$<TARGET_EXISTS:Parquet::parquet_static>,Parquet::parquet_static,Parquet::parquet_shared>"
        "$<IF:$<TARGET_EXISTS:avro-cpp::avrocpp_static>,avro-cpp::avrocpp_static,avro-cpp::avrocpp_shared>"
   )
+  if(ICEBERG_BUILD_REST_CATALOG)
+    list(APPEND ICEBERG_BUNDLE_STATIC_BUILD_INTERFACE_LIBS
+         "$<IF:$<TARGET_EXISTS:CURL::libcurl>,CURL::libcurl,CURL::libcurl>"
+         "$<IF:$<TARGET_EXISTS:cpr::cpr>,cpr::cpr,cpr::cpr>")
+  endif()
```
wgtmac:
```suggestion
```

`libiceberg-bundle` does not interfere with rest catalog so it should not depend on curl or cpr.

wgtmac: Same for below.

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+list(APPEND
+     ICEBERG_REST_CATALOG_STATIC_INSTALL_INTERFACE_LIBS
+     "$<IF:$<TARGET_EXISTS:Iceberg::iceberg_static>,Iceberg::iceberg_static,Iceberg::iceberg_shared>"
+     "$<IF:$<BOOL:${CPR_VENDORED}>,Iceberg::cpr,cpr::cpr>")
+list(APPEND
+     ICEBERG_REST_CATALOG_SHARED_INSTALL_INTERFACE_LIBS
+     "$<IF:$<TARGET_EXISTS:Iceberg::iceberg_shared>,Iceberg::iceberg_shared,Iceberg::iceberg_static>"
+     "$<IF:$<BOOL:${CPR_VENDORED}>,Iceberg::cpr,cpr::cpr>")
+
+add_iceberg_lib(iceberg_rest_catalog
```
```suggestion
add_iceberg_lib(iceberg_rest
```

Let's make it shorter

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+if(NOT ICEBERG_BUILD_REST_CATALOG)
+  return()
+endif()
+
+set(ICEBERG_REST_CATALOG_SOURCES rest_catalog.cc)
+
+set(ICEBERG_REST_CATALOG_STATIC_BUILD_INTERFACE_LIBS)
+set(ICEBERG_REST_CATALOG_SHARED_BUILD_INTERFACE_LIBS)
+set(ICEBERG_REST_CATALOG_STATIC_INSTALL_INTERFACE_LIBS)
+set(ICEBERG_REST_CATALOG_SHARED_INSTALL_INTERFACE_LIBS)
```
```suggestion
set(ICEBERG_REST_SOURCES rest_catalog.cc)

set(ICEBERG_REST_STATIC_BUILD_INTERFACE_LIBS)
set(ICEBERG_REST_SHARED_BUILD_INTERFACE_LIBS)
set(ICEBERG_REST_STATIC_INSTALL_INTERFACE_LIBS)
set(ICEBERG_REST_SHARED_INSTALL_INTERFACE_LIBS)
```

wgtmac (commented):
Could you try apply this change locally and build the example?

```
diff --git a/example/CMakeLists.txt b/example/CMakeLists.txt
index 3e178a8..4fc05c7 100644
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -26,4 +26,4 @@ find_package(Iceberg CONFIG REQUIRED)
 
 add_executable(demo_example demo_example.cc)
 
-target_link_libraries(demo_example PRIVATE Iceberg::iceberg_bundle_static)
+target_link_libraries(demo_example PRIVATE Iceberg::iceberg_bundle_static Iceberg::iceberg_rest)
```

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+if(NOT ICEBERG_BUILD_REST)
+  return()
+endif()
+
```
zhjwpku: These are not necessary.
```suggestion
```

HeartLinked: ok I' ll fix this.

LICENSE
HeartLinked: > @jbonofre Could you help review LICENSE and NOTICE files for adding `cpr` and `curl`? BTW, do we need to add `OpenSSL` which is an indirect dependency (required by curl)?

I have split the task of modifying the license into a separate PR #243 . @jbonofre Could you help review it? In this case, the current PR will not involve license modifications and should be able to be merged quickly.

wgtmac:
The PR title is out of date. Perhaps `feat: add cpr dependency for rest catalog client`?

Fokko:
Thanks @HeartLinked for working on this and thanks for the review 🙌

Fokko:
@wgtmac LICENSE and NOTICE can be done in a separate PR, but must not be forgotten. Do you know if there are any CPP specific tools for checking this?

wgtmac:
@Fokko The followup PR is https://github.com/apache/iceberg-cpp/pull/243 :)

PR #238, opened 2025-09-19
feat: Literal support decimal & Literal serde

zhjwpku:


HeartLinked:
Thanks for your contribution! I noticed part of the code logic in your PR overlaps with mine (https://github.com/apache/iceberg-cpp/pull/185). My PR has been in review for a long time, and it’s now almost ready for merge. It shouldn’t take long to finalize.​
To avoid unnecessary code conflicts and make our collaboration more efficient, could you wait until my PR is merged? Then you can rebase your code on the merged version and only focus on adding the new Literal support for decimal. This way, your core contribution is preserved, and we reduce maintenance costs from duplicate code. Thanks!

zhjwpku:
> Thanks for your contribution! I noticed part of the code logic in your PR overlaps with mine (#185). My PR has been in review for a long time, and it’s now almost ready for merge. It shouldn’t take long to finalize.​To avoid unnecessary code conflicts and make our collaboration more efficient, could you wait until my PR is merged? Then you can rebase your code on the merged version and only focus on adding the new Literal support for decimal. This way, your core contribution is preserved, and we reduce maintenance costs from duplicate code. Thanks!

Sure thing, I will take a look of your PR, thanks.

src/iceberg/expression/literal.h
```diff
@@ -56,7 +60,8 @@ class ICEBERG_EXPORT Literal {
                              double,          // for double
                              std::string,     // for string
                              std::vector<uint8_t>,     // for binary, fixed
-                             std::array<uint8_t, 16>,  // for uuid and decimal
+                             std::array<uint8_t, 16>,  // for uuid
+                             int128_t,                 // for decimal
```
HuaHuaY: See https://github.com/apache/iceberg-cpp/pull/206#discussion_r2371280533. I think it's better to use the wrapper class `Decimal` here. And use `Decimal(Decimal value, int32_t precision, int32_t scale)` to create a literal.

zhjwpku: OK, I will consider this suggestion, thanks

zhjwpku: I changed to the way you suggested, the same for Uuid.

src/iceberg/expression/literal.h
```diff
@@ -76,6 +78,8 @@ class ICEBERG_EXPORT Literal : public util::Formattable {
   static Literal UUID(Uuid value);
   static Literal Binary(std::vector<uint8_t> value);
   static Literal Fixed(std::vector<uint8_t> value);
+  static Literal MakeDecimal(Decimal value, int32_t precision, int32_t scale);
```
```suggestion
  static Literal Decimal(int128_t value, int32_t precision, int32_t scale);
```

I prefer `int128_t` over `Decimal` because it is no easier to create a `Decimal` than `int128_t`. In this way, we also don't need to make this function name a special one.

wgtmac: We can also use `static Literal Decimal_(Decimal value, int32_t precision, int32_t scale);`

or

`static Literal Decimal(::iceberg::Decimal value, int32_t precision, int32_t scale);`

src/iceberg/expression/literal.cc
```diff
@@ -193,12 +205,44 @@ std::strong_ordering CompareFloat(T lhs, T rhs) {
...
+  }
+}
+
 bool Literal::operator==(const Literal& other) const { return (*this <=> other) == 0; }
 
 // Three-way comparison operator
 std::partial_ordering Literal::operator<=>(const Literal& other) const {
   // If types are different, comparison is unordered
-  if (*type_ != *other.type_) {
+  if (type_->type_id() != other.type_->type_id()) {
```
wgtmac: We shouldn't change this. Otherwise the assumption no longer holds for types including fixed and timestamp.

src/iceberg/expression/literal.cc
```diff
@@ -307,6 +355,14 @@ std::string Literal::ToString() const {
     case TypeId::kDouble: {
       return std::to_string(std::get<double>(value_));
     }
+    case TypeId::kDecimal: {
+      auto decimal = std::get<Decimal>(value_);
+      auto decimal_type = std::dynamic_pointer_cast<DecimalType>(type_);
+      ICEBERG_DCHECK(decimal_type != nullptr, "Type must be DecimalType");
+      auto result = decimal.ToString(decimal_type->scale());
+      ICEBERG_CHECK(result, "Decimal ToString failed");
+      return *result;
```
```suggestion
      const auto& decimal_type = internal::checked_cast<DecimalType>(*type_);
      auto decimal = std::get<Decimal>(value_);
      return decimal.ToString(decimal_type.scale());
```

src/iceberg/expression/literal.cc
```diff
@@ -193,12 +205,44 @@ std::strong_ordering CompareFloat(T lhs, T rhs) {
...
+  ICEBERG_DCHECK(rhs_type != nullptr, "RHS type must be DecimalType");
+  auto lhs_decimal = std::get<Decimal>(lhs.value());
+  auto rhs_decimal = std::get<Decimal>(rhs.value());
+  if (lhs_type->scale() == rhs_type->scale()) {
+    return lhs_decimal <=> rhs_decimal;
+  } else if (lhs_type->scale() > rhs_type->scale()) {
+    // Rescale to larger scale
+    auto rhs_res = rhs_decimal.Rescale(rhs_type->scale(), lhs_type->scale());
+    if (!rhs_res) {
+      // Rescale would cause data loss, so lhs is definitely less than rhs
```
wgtmac: This is not future-proof thus looks unsafe. It is better to explicitly call `RescaleWouldCauseDataLoss` to know this, then return unordered if any error happens after calling `Rescale`.

src/iceberg/test/bucket_util_test.cc
```diff
@@ -0,0 +1,84 @@
...
+  auto decimal = Decimal::FromString("14.20");
+  ASSERT_TRUE(decimal.has_value());
+  EXPECT_EQ(BucketUtils::HashBytes(Decimal::ToBigEndian(decimal->value())), -500754589);
+
+  // date hash
+  // 2017-11-16
+  std::chrono::sys_days sd = std::chrono::year{2017} / 11 / 16;
+  std::chrono::sys_days epoch{std::chrono::year{1970} / 1 / 1};
+  int32_t days = (sd - epoch).count();
+  std::cout << "days: " << days << std::endl;
```
wgtmac:
```suggestion
```

src/iceberg/test/bucket_util_test.cc
```diff
@@ -0,0 +1,84 @@
...
+  EXPECT_EQ(BucketUtils::HashInt(34), 2017239379);
+  EXPECT_EQ(BucketUtils::HashLong(34L), 2017239379);
+
+  // decimal hash
+  auto decimal = Decimal::FromString("14.20");
+  ASSERT_TRUE(decimal.has_value());
+  EXPECT_EQ(BucketUtils::HashBytes(Decimal::ToBigEndian(decimal->value())), -500754589);
+
+  // date hash
+  // 2017-11-16
```
wgtmac:
```suggestion
```

This looks unnecessary

src/iceberg/test/bucket_util_test.cc
```diff
@@ -0,0 +1,84 @@
...
+  std::chrono::sys_days sd = std::chrono::year{2017} / 11 / 16;
+  std::chrono::sys_days epoch{std::chrono::year{1970} / 1 / 1};
+  int32_t days = (sd - epoch).count();
+  std::cout << "days: " << days << std::endl;
+  EXPECT_EQ(BucketUtils::HashInt(days), -653330422);
+
+  // time
+  // 22:31:08 in microseconds
+  int64_t time_micros = (22 * 3600 + 31 * 60 + 8) * 1000000LL;
+  std::cout << "time micros: " << time_micros << std::endl;
```
wgtmac:
```suggestion
```

src/iceberg/test/bucket_util_test.cc
```diff
@@ -0,0 +1,84 @@
...
+
+  // string
+  std::string str = "iceberg";
+  EXPECT_EQ(BucketUtils::HashBytes(std::span<const uint8_t>(
+                reinterpret_cast<const uint8_t*>(str.data()), str.size())),
+            1210000089);
+
+  // uuid
+  // f79c3e09-677c-4bbd-a479-3f349cb785e7
+  std::array<uint8_t, 16> uuid = {0xf7, 0x9c, 0x3e, 0x09, 0x67, 0x7c, 0x4b, 0xbd,
```
wgtmac: It looks better if we use `Uuid::FromString` here.

src/iceberg/test/bucket_util_test.cc
```diff
@@ -0,0 +1,84 @@
...
+
+  // timestamp
+  // 2017-11-16T22:31:08 in microseconds
+  std::chrono::system_clock::time_point tp =
+      std::chrono::sys_days{std::chrono::year{2017} / 11 / 16} + std::chrono::hours{22} +
+      std::chrono::minutes{31} + std::chrono::seconds{8};
+  int64_t timestamp_micros =
+      std::chrono::duration_cast<std::chrono::microseconds>(tp.time_since_epoch())
+          .count();
+  std::cout << "timestamp micros: " << timestamp_micros << std::endl;
```
wgtmac:
```suggestion
```

src/iceberg/test/bucket_util_test.cc
```diff
@@ -0,0 +1,84 @@
...
+
+  // time
+  // 22:31:08 in microseconds
+  int64_t time_micros = (22 * 3600 + 31 * 60 + 8) * 1000000LL;
+  std::cout << "time micros: " << time_micros << std::endl;
+  EXPECT_EQ(BucketUtils::HashLong(time_micros), -662762989);
+
+  // timestamp
+  // 2017-11-16T22:31:08 in microseconds
+  std::chrono::system_clock::time_point tp =
```
wgtmac, line 56: Perhaps we can add some utility functions in the temporal_util.h to create date/time/timestamp from structural inputs.

src/iceberg/test/literal_test.cc
```diff
@@ -204,6 +204,33 @@ TEST(LiteralTest, DoubleComparison) {
   EXPECT_EQ(double2 <=> double1, std::partial_ordering::greater);
 }
 
+// Decimal type tests
+TEST(LiteralTest, DecimalBasics) {
```
wgtmac: Please rebase this PR. I think a recent commit has improved this file a lot. Please follow the similar pattern to use parameterized test.

src/iceberg/test/transform_test.cc
```diff
@@ -245,6 +246,14 @@ TEST(TransformLiteralTest, BucketTransform) {
   constexpr int32_t num_buckets = 4;
   auto transform = Transform::Bucket(num_buckets);
 
+  // uuid
+  // f79c3e09-677c-4bbd-a479-3f349cb785e7
+  std::array<uint8_t, 16> uuid = {0xf7, 0x9c, 0x3e, 0x09, 0x67, 0x7c, 0x4b, 0xbd,
```
wgtmac: nit: use `Uuid` here?

src/iceberg/test/transform_test.cc
```diff
@@ -253,23 +262,43 @@ TEST(TransformLiteralTest, BucketTransform) {
...
       {.source_type = iceberg::timestamp(),
-       .source = Literal::Timestamp(1622547800000000),
-       .expected = Literal::Int(1)},
+       .source = Literal::Timestamp(1510871468000000),
+       .expected = Literal::Int(3)},
+      // // 2017-11-16T22:31:08.000001 in microseconds
       {.source_type = iceberg::timestamp_tz(),
-       .source = Literal::TimestampTz(1622547800000000),
-       .expected = Literal::Int(1)},
+       .source = Literal::TimestampTz(1510871468000001),
```
wgtmac: Why changing both values here?

src/iceberg/util/truncate_util.h
```diff
@@ -19,10 +19,13 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 #include <utility>
 
+#include "iceberg/expression/literal.h"
```
wgtmac: Use forward declaration for Literal?

src/iceberg/util/truncate_util.cc
```diff
@@ -0,0 +1,113 @@
...
+Literal TruncateLiteralImpl<TypeId::kLong>(const Literal& literal, int32_t width) {
+  int64_t v = std::get<int64_t>(literal.value());
+  return Literal::Long(TruncateUtils::TruncateInteger(v, width));
+}
+
+template <>
+Literal TruncateLiteralImpl<TypeId::kDecimal>(const Literal& literal, int32_t width) {
+  const auto& decimal = std::get<Decimal>(literal.value());
+  const auto& type = std::dynamic_pointer_cast<DecimalType>(literal.type());
+  ICEBERG_DCHECK(type != nullptr, "Literal type must be DecimalType");
```
wgtmac:
```suggestion
  const auto& type = internal::checked_cast<DecimalType>(*literal.type());
```

src/iceberg/util/truncate_util.cc
```diff
@@ -0,0 +1,113 @@
...
+  return Literal::MakeDecimal(TruncateUtils::TruncateDecimal(decimal, width),
+                              type->precision(), type->scale());
```
wgtmac:
```suggestion
  return Literal::MakeDecimal(TruncateUtils::TruncateDecimal(decimal, width),
                              type.precision(), type.scale());
```

src/iceberg/util/truncate_util.cc
```diff
@@ -0,0 +1,113 @@
...
+}
+
+template <>
+Literal TruncateLiteralImpl<TypeId::kString>(const Literal& literal, int32_t width) {
+  // Strings are truncated to a valid UTF-8 string with no more than L code points.
```
wgtmac:
```suggestion
  // Strings are truncated to a valid UTF-8 string with no more than `width` code points.
```

src/iceberg/util/truncate_util.cc
```diff
@@ -0,0 +1,113 @@
...
+  const auto& str = std::get<std::string>(literal.value());
+  return Literal::String(TruncateUtils::TruncateUTF8(str, width));
+}
+
+template <>
+Literal TruncateLiteralImpl<TypeId::kBinary>(const Literal& literal, int32_t width) {
+  /// In contrast to strings, binary values do not have an assumed encoding and are
+  /// truncated to L bytes.
```
wgtmac:
```suggestion
  // In contrast to strings, binary values do not have an assumed encoding and are
  // truncated to `width` bytes.
```

We only need `///` for docstrings so here using `//` is enough.

src/iceberg/util/truncate_util.cc
```diff
@@ -0,0 +1,113 @@
...
+    // Return null as is
+    return literal;
+  }
+
+  if (literal.IsAboveMax()) [[unlikely]] {
+    return NotSupported("Cannot truncate AboveMax");
+  }
+
+  if (literal.IsBelowMin()) [[unlikely]] {
+    return NotSupported("Cannot truncate BelowMin");
```
wgtmac:
```suggestion
  if (literal.IsAboveMax() || literal.IsBelowMin()) [[unlikely]] {
    return NotSupported("Cannot truncate {}", literal->ToString());
```

src/iceberg/util/temporal_util.cc
```diff
@@ -0,0 +1,253 @@
...
+ * under the License.
+ */
+
+#include "iceberg/util/temporal_util.h"
+
+#include <chrono>
+#include <utility>
+
+#include "iceberg/result.h"
+
```
wgtmac:
```suggestion
```

src/iceberg/util/temporal_util.cc
```diff
@@ -0,0 +1,253 @@
...
+template <TypeId type_id>
+Result<Literal> ExtractYearImpl(const Literal& literal) {
+  std::unreachable();
+}
+
+template <>
+Result<Literal> ExtractYearImpl<TypeId::kDate>(const Literal& literal) {
+  auto value = std::get<int32_t>(literal.value());
+  auto epoch = sys_days(year{1970} / January / 1);
```
wgtmac: Can we add something like below for the boilerplate code?

```cpp
constexpr auto kEpoch = year{1970} / January / 1;
inline const auto kEpochDays = sys_days(kEpoch);
inline const auto kEpochYmd = year_month_day(kEpochDays);

inline year_month_day DateToYmd(int32_t days_since_epoch) {
  return {kEpochDays + days{days_since_epoch}};
}

inline year_month_day TimestampToYmd(int64_t micros_since_epoch) {
  return {floor<days>(sys_time<microseconds>(microseconds{micros_since_epoch}))};
}

inline int32_t CalculateMonthsSinceEpoch(const year_month_day& ymd) {
  auto delta = ymd.year() - kEpochYmd.year();
  // Calculate the month as months from 1970-01
  // Note: January is month 1, so we subtract 1 to get zero-based month count.
  return static_cast<int32_t>(delta.count() * 12 + static_cast<unsigned>(ymd.month()) - 1);
}
```

src/iceberg/util/decimal.h
```diff
@@ -164,6 +165,12 @@ class ICEBERG_EXPORT Decimal : public util::Formattable {
   /// \return error status if the length is an invalid value
   static Result<Decimal> FromBigEndian(const uint8_t* data, int32_t length);
 
+  /// \brief Convert Decimal's unscaled value to two’s-complement big-endian binary, using
+  ///        the minimum number of bytes for the value.
+  /// \param value The unscaled value.
+  /// \return A vector containing the big-endian bytes.
+  static std::vector<uint8_t> ToBigEndian(int128_t value);
```
wgtmac: Why not making it a member function?

src/iceberg/util/decimal.cc
```diff
@@ -505,6 +504,36 @@ Result<Decimal> Decimal::FromBigEndian(const uint8_t* bytes, int32_t length) {
   return Decimal(static_cast<int128_t>(result));
 }
 
+std::vector<uint8_t> Decimal::ToBigEndian(int128_t value) {
+  std::vector<uint8_t> bytes(kMaxDecimalBytes);
+
+  auto uvalue = static_cast<uint128_t>(value);
+  std::memcpy(bytes.data(), &uvalue, 16);
```
wgtmac:
```suggestion
  std::memcpy(bytes.data(), &uvalue, kMaxDecimalBytes);
```

src/iceberg/util/bucket_util.cc
```diff
@@ -0,0 +1,157 @@
...
+  return hash_value;
+}
+
+#define DISPATCH_HASH_LITERAL(TYPE_ID)          \
+  case TYPE_ID:                                 \
+    hash_value = HashLiteral<TYPE_ID>(literal); \
+    break;
+
+Result<Literal> BucketUtils::BucketIndex(const Literal& literal, int32_t num_buckets) {
+  ICEBERG_DCHECK(num_buckets > 0, "Number of buckets must be positive");
```
wgtmac: I'd prefer this to be an error instead of an assert.

src/iceberg/util/bucket_util.h
```diff
@@ -0,0 +1,61 @@
...
+  /// \param bytes The input byte array to hash.
+  /// \return A 32-bit hash value.
+  static int32_t HashBytes(std::span<const uint8_t> bytes);
+
+  /// \brief Compute the bucket index for a given literal and number of buckets.
+  /// \param literal The input literal to hash.
+  /// \param num_buckets The number of buckets to hash into.
+  /// \return A Result containing the bucket index or an error.
+  /// The bucket index is in the range [0, num_buckets - 1].
+  static Result<Literal> BucketIndex(const Literal& literal, int32_t num_buckets);
```
wgtmac: Is it better to return `int32_t`? The caller can wrap it in a Literal.

src/iceberg/expression/literal.cc
zhjwpku: Then we can not compare two Decimal literal with different precision/scale, is that acceptable? I'm unclear about the assumption for fixed and timestamp, can you elaborate?

wgtmac:
```java
  private abstract static class ComparableLiteral<C extends Comparable<C>> extends BaseLiteral<C> {
    @SuppressWarnings("unchecked")
    private static final Comparator<? extends Comparable> CMP =
        Comparators.<Comparable>nullsFirst().thenComparing(Comparator.naturalOrder());

    ComparableLiteral(C value) {
      super(value);
    }

    @Override
    @SuppressWarnings("unchecked")
    public Comparator<C> comparator() {
      return (Comparator<C>) CMP;
    }
  }
```

I just happened to see that Java always compares values in the `nulls-first` order.

src/iceberg/expression/literal.h
zhjwpku: I ended up with `static Literal Decimal(int128_t value, int32_t precision, int32_t scale);`, where `value` represents the unscaled 128-bit integer, this seems more natural.

src/iceberg/util/decimal.h
zhjwpku: Changed to a member function.

src/iceberg/util/temporal_util.cc
zhjwpku: Reasonable, this will make the code cleaner, done.

src/iceberg/util/bucket_util.cc
zhjwpku: `num_buckets` has been validated in `BucketTransform::Make`. I initially assumed this would only be used internally, but since it can also be called by users, I'll update it to return an error when `num_buckets <= 0`.

src/iceberg/util/bucket_util.h
zhjwpku: Reasonable, done.

src/iceberg/expression/literal.h
```diff
@@ -56,10 +58,10 @@ class ICEBERG_EXPORT Literal : public util::Formattable {
                              int64_t,         // for long, timestamp, timestamp_tz, time
                              float,           // for float
                              double,          // for double
-                             std::string,     // for string
-                             Uuid,            // for uuid
-                             std::vector<uint8_t>,     // for binary, fixed
-                             std::array<uint8_t, 16>,  // for decimal
+                             ::iceberg::Decimal,    // for decimal
+                             std::string,           // for string
+                             Uuid,                  // for uuid
```
HuaHuaY, line 64:
```suggestion
                             Decimal,         // for decimal
                             std::string,     // for string
                             Uuid,            // for uuid
```

zhjwpku: I'm using `::iceberg::Decimal` to avoid `changes-meaning` compiling warning, see below:

```
In file included from /home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/expression/literal.cc:20:
/home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/expression/literal.h:84:18: error: declaration of ‘static iceberg::Literal iceberg::Literal::Decimal(int128_t, int32_t, int32_t)’ changes meaning of ‘Decimal’ [-Wchanges-meaning]
   84 |   static Literal Decimal(int128_t value, int32_t precision,
      |                  ^~~~~~~
/home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/expression/literal.h:61:30: note: used here to mean ‘class iceberg::Decimal’
   61 |                              Decimal,         // for decimal
      |                              ^~~~~~~
In file included from /home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/expression/literal.h:30:
/home/runner/work/iceberg-cpp/iceberg-cpp/src/iceberg/util/decimal.h:45:22: note: declared here
   45 | class ICEBERG_EXPORT Decimal : public util::Formattable {
      |                      ^~~~~~~
```

HuaHuaY: It's an interesting problem that these code can be compiled in my environment. It seems that different platforms have different compile requirements. I'm troubled by this.

You can resolve this conversation.

src/iceberg/test/transform_test.cc
zhjwpku: 2017-11-16T22:31:08 is from spec[1], here use it to replace the randomly selected timestamp by me.

[1] https://iceberg.apache.org/spec/#appendix-b-32-bit-hash-requirements

src/iceberg/test/bucket_util_test.cc
zhjwpku: Agreed, do you mind if I add these in a separate PR?

src/iceberg/expression/literal.cc
zhjwpku: I ended up adding a new `kRescaleDataLoss` ErrorKind and returning unordered for all other errors.

src/iceberg/expression/literal.cc
zhjwpku: I plan to look into this later in a separate PR ;)

src/iceberg/expression/literal.cc
```diff
@@ -331,12 +339,50 @@ std::strong_ordering CompareFloat(T lhs, T rhs) {
   return lhs_is_negative <=> rhs_is_negative;
 }
 
+std::partial_ordering CompareDecimal(Literal const& lhs, Literal const& rhs) {
+  ICEBERG_DCHECK(std::holds_alternative<Decimal>(lhs.value()),
+                 "LHS of decimal comparison must hold Decimal");
+  ICEBERG_DCHECK(std::holds_alternative<Decimal>(rhs.value()),
+                 "RHS of decimal comparison must hold decimal");
```
```suggestion
  ICEBERG_DCHECK(std::holds_alternative<Decimal>(lhs.value()) && std::holds_alternative<Decimal>(rhs.value()),
                 "Cannot compare non-decimal literals!");
```

src/iceberg/expression/literal.cc
```diff
@@ -331,12 +339,50 @@ std::strong_ordering CompareFloat(T lhs, T rhs) {
...
+  auto lhs_type = internal::checked_pointer_cast<DecimalType>(lhs.type());
+  auto rhs_type = internal::checked_pointer_cast<DecimalType>(rhs.type());
```
```suggestion
  const auto& lhs_type = internal::checked_cast<DecimalType>(*lhs.type());
  const auto& rhs_type = internal::checked_cast<DecimalType>(*rhs.type());
```

Can we avoid unnecessary shared_ptr operation?

src/iceberg/type_fwd.h
```diff
@@ -115,6 +115,7 @@ class NameMapping;
 enum class SnapshotRefType;
 enum class TransformType;
 
+class Decimal;
```
wgtmac, line 118: What aboud `Uuid`?

src/iceberg/expression/literal.cc
```diff
@@ -331,12 +339,50 @@ std::strong_ordering CompareFloat(T lhs, T rhs) {
...
+  auto lhs_decimal = std::get<Decimal>(lhs.value());
+  auto rhs_decimal = std::get<Decimal>(rhs.value());
+  if (lhs_type->scale() == rhs_type->scale()) {
+    return lhs_decimal <=> rhs_decimal;
+  } else if (lhs_type->scale() > rhs_type->scale()) {
+    // Rescale to larger scale
+    auto rhs_res = rhs_decimal.Rescale(rhs_type->scale(), lhs_type->scale());
+    if (!rhs_res.has_value()) {
+      if (rhs_res.error().kind == ErrorKind::kRescaleDataLoss) {
```
wgtmac: IMHO, I don't think `ErrorKind::kRescaleDataLoss` is general enough to be added as a new error kind. Perhaps we can add a `static std::partial_ordering Decimal::Compare(...)` for this?

src/iceberg/expression/literal.cc
wgtmac: Then perhaps revert this line so all types still require strong type equality?

src/iceberg/expression/literal.cc
```diff
@@ -440,6 +490,13 @@ std::string Literal::ToString() const {
     case TypeId::kDouble: {
       return std::to_string(std::get<double>(value_));
     }
+    case TypeId::kDecimal: {
+      auto decimal_type = internal::checked_pointer_cast<DecimalType>(type_);
```
wgtmac: Same thing here. Let's avoid shared_pointer operation as much as possible.

src/iceberg/expression/literal.cc
```diff
@@ -440,6 +490,13 @@ std::string Literal::ToString() const {
...
+      auto decimal = std::get<::iceberg::Decimal>(value_);
+      auto result = decimal.ToString(decimal_type->scale());
+      ICEBERG_CHECK(result, "Decimal ToString failed");
+      return *result;
```
```suggestion
      return decimal.ToString(decimal_type->scale()).value_or("invalid literal of type decimal");
```

src/iceberg/expression/literal.cc
```diff
@@ -24,9 +24,10 @@
 #include <cstdint>
 #include <string>
 
-#include "iceberg/type_fwd.h"
+#include "iceberg/exception.h"
```
```suggestion
```

Let's fix decimal ToString not to throw and remove this.

src/iceberg/expression/literal.h
```diff
@@ -56,10 +58,10 @@ class ICEBERG_EXPORT Literal : public util::Formattable {
                              int64_t,         // for long, timestamp, timestamp_tz, time
                              float,           // for float
                              double,          // for double
-                             std::string,     // for string
-                             Uuid,            // for uuid
-                             std::vector<uint8_t>,     // for binary, fixed
-                             std::array<uint8_t, 16>,  // for decimal
+                             ::iceberg::Decimal,    // for decimal
```
wgtmac, line 63: What about putting specialized types like Decimal and Uuid to the end before BelowXXX?

src/iceberg/result.h
```diff
@@ -47,6 +47,7 @@ enum class ErrorKind {
   kNotFound,
   kNotImplemented,
   kNotSupported,
+  kRescaleDataLoss,
```
wgtmac: See my other comment to remove this.

src/iceberg/test/bucket_util_test.cc
wgtmac: No problem!

src/iceberg/util/decimal.cc
```diff
@@ -522,8 +552,8 @@ Result<Decimal> Decimal::Rescale(int32_t orig_scale, int32_t new_scale) const {
       RescaleWouldCauseDataLoss(*this, delta_scale, multiplier, &out);
 
   if (rescale_would_cause_data_loss) {
-    return Invalid("Rescale {} from {} to {} would cause data loss", ToIntegerString(),
-                   orig_scale, new_scale);
+    return RescaleDataLoss("Rescale {} from {} to {} would cause data loss",
```
wgtmac: Perhaps we need still use Invalid.

wgtmac (commented):
Mostly looks good! Thanks @zhjwpku!

src/iceberg/expression/literal.cc
zhjwpku: I've done this way, thanks.

src/iceberg/expression/literal.h
zhjwpku: Changed.

src/iceberg/result.h
zhjwpku: Removed

src/iceberg/util/decimal.cc
zhjwpku: Done.

src/iceberg/expression/literal.cc
zhjwpku: Done

PR #241, opened 2025-09-21
fix: disable CRoaring tests and microbenchmarks

zhjwpku:
Previously, we mistakenly used the outdated option BUILD_TESTING for CRoaring, change it to ENABLE_ROARING_TESTS, and also disable ENABLE_ROARING_MICROBENCHMARKS.


wgtmac (approved):
Thanks!

Fokko:
Thanks @zhjwpku for the fix and @wgtmac for the review

PR #242, opened 2025-09-23
feat: add UUID representation

zhjwpku:
UUID representation along with utilities such as generators for v4 and v7.

### DIFF
--- a/src/iceberg/util/uuid.cc
+++ b/src/iceberg/util/uuid.cc
@@ -27,6 +27,7 @@
 
 #include "iceberg/exception.h"
 #include "iceberg/result.h"
+#include "iceberg/util/formatter.h"  // IWYU pragma: keep
 #include "iceberg/util/int128.h"
 #include "iceberg/util/macros.h"
 
@@ -70,7 +71,7 @@ inline Result<Uuid> ParseSimple(std::string_view s) {
     uint8_t h1 = kHexTable[static_cast<uint8_t>(s[i * 2])];
     uint8_t h2 = kHexTable[static_cast<uint8_t>(s[i * 2 + 1])];
 
-    if ((h1 | h2) == 0xFF) {
+    if ((h1 | h2) == 0xFF) [[unlikely]] {
       return InvalidArgument("Invalid UUID string: {}", s);
     }
 
@@ -84,7 +85,7 @@ inline Result<Uuid> ParseHyphenated(std::string_view s) {
   ICEBERG_DCHECK(s.size() == 36, "s must be 36 characters long");
 
   // Check that dashes are in the right places
-  if (!(s[8] == '-' && s[13] == '-' && s[18] == '-' && s[23] == '-')) {
+  if (!(s[8] == '-' && s[13] == '-' && s[18] == '-' && s[23] == '-')) [[unlikely]] {
     return InvalidArgument("Invalid UUID string: {}", s);
   }
 
@@ -98,7 +99,7 @@ inline Result<Uuid> ParseHyphenated(std::string_view s) {
     uint8_t h3 = kHexTable[static_cast<uint8_t>(s[i + 2])];
     uint8_t h4 = kHexTable[static_cast<uint8_t>(s[i + 3])];
 
-    if ((h1 | h2 | h3 | h4) == 0xFF) {
+    if ((h1 | h2 | h3 | h4) == 0xFF) [[unlikely]] {
       return InvalidArgument("Invalid UUID string: {}", s);
     }
 
--- a/src/iceberg/util/uuid.h
+++ b/src/iceberg/util/uuid.h
@@ -26,13 +26,14 @@
 
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
+#include "iceberg/util/formattable.h"
 
 /// \file iceberg/util/uuid.h
 /// \brief UUID (Universally Unique Identifier) representation.
 
 namespace iceberg {
 
-class ICEBERG_EXPORT Uuid {
+class ICEBERG_EXPORT Uuid : public util::Formattable {
  public:
   Uuid() = delete;
   constexpr static size_t kLength = 16;
@@ -71,7 +72,7 @@ class ICEBERG_EXPORT Uuid {
   uint8_t operator[](size_t index) const;
 
   /// \brief Convert the UUID to a string in standard format.
-  std::string ToString() const;
+  std::string ToString() const override;
 
   friend bool operator==(const Uuid& lhs, const Uuid& rhs) {
     return lhs.data_ == rhs.data_;


src/iceberg/util/uuid_util.cc
```diff
@@ -0,0 +1,207 @@
...
+  } else if (str.size() == 36) {
+    return ParseHyphenated(str);
+  } else {
+    return InvalidArgument("Invalid UUID string: {}", str);
+  }
+}
+
+std::string UUIDUtils::ToString(std::span<uint8_t> uuid) {
+  static const char* hex_chars = "0123456789abcdef";
+  ICEBERG_DCHECK(uuid.size() == 16, "uuid must be 16 bytes long");
```
mapleFU: Use check?

src/iceberg/util/uuid_util.cc
```diff
@@ -0,0 +1,207 @@
...
+  std::string str(36, '-');
+
+  for (size_t i = 0; i < 16; i++) {
```
HeartLinked: This can be simplified using std::format?

src/iceberg/util/uuid_util.cc
zhjwpku: Agreed, I think it should throw an exception here, do you think it's necessary to create macro for that? something like

```
#define ICEBERG_CHECK(expr, message) \
  if (!(expr)) [[unlikely]] {        \
    throw IcebergError(message);     \
  }
```

zhjwpku: I just noticed that we already had that ;)

src/iceberg/util/uuid_util.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+
+/// \file iceberg/util/uuid_util.h
+/// \brief UUID (Universally Unique Identifier) utilities.
+
+namespace iceberg {
+
+class ICEBERG_EXPORT UUIDUtils {
```
wgtmac: Should we add a `UUID` wrapper class instead of passing `std::array<uint8_t, 16>` around?

src/iceberg/util/uuid_util.cc
```diff
@@ -0,0 +1,204 @@
...
+
+constexpr std::array<uint8_t, 256> BuildShl4Table() {
+  std::array<uint8_t, 256> buf{};
+  for (int i = 0; i < 256; i++) {
+    buf[i] = static_cast<uint8_t>(i << 4);
+  }
+  return buf;
+}
+
+constexpr auto HEX_TABLE = BuildHexTable();
```
wgtmac:
```suggestion
constexpr auto kHexTable = BuildHexTable();
```

Should we follow this style as it is a constant?

src/iceberg/util/uuid_util.cc
```diff
@@ -0,0 +1,204 @@
...
+// Parse a UUID string without dashes, e.g. "67e5504410b1426f9247bb680e5fe0c8"
+inline Result<std::array<uint8_t, 16>> ParseSimple(std::string_view s) {
+  ICEBERG_DCHECK(s.size() == 32, "s must be 32 characters long");
+
+  std::array<uint8_t, 16> buf{};
+  for (size_t i = 0; i < 16; i++) {
+    uint8_t h1 = HEX_TABLE[static_cast<uint8_t>(s[i * 2])];
+    uint8_t h2 = HEX_TABLE[static_cast<uint8_t>(s[i * 2 + 1])];
+
+    if ((h1 | h2) == 0xff) {
```
wgtmac:
```suggestion
    if ((h1 | h2) == 0xFF) {
```

src/iceberg/util/uuid_util.cc
```diff
@@ -0,0 +1,204 @@
...
+  std::array<uint8_t, 16> buf{};
+
+  for (size_t j = 0; j < 8; j++) {
+    size_t i = positions[j];
+    uint8_t h1 = HEX_TABLE[static_cast<uint8_t>(s[i])];
+    uint8_t h2 = HEX_TABLE[static_cast<uint8_t>(s[i + 1])];
+    uint8_t h3 = HEX_TABLE[static_cast<uint8_t>(s[i + 2])];
+    uint8_t h4 = HEX_TABLE[static_cast<uint8_t>(s[i + 3])];
+
+    if ((h1 | h2 | h3 | h4) == 0xff) {
```
wgtmac:
```suggestion
    if ((h1 | h2 | h3 | h4) == 0xFF) {
```

src/iceberg/util/uuid_util.h
zhjwpku: Agreed, changed to a class with name Uuid, not sure if this respects our naming convention.

src/iceberg/util/uuid_util.cc
zhjwpku: make sense, changed.

src/iceberg/util/uuid_util.cc
zhjwpku: fixed.

src/iceberg/util/uuid.cc
```diff
@@ -0,0 +1,221 @@
...
+  return Uuid(std::move(data));
+}
+
+uint8_t Uuid::operator[](size_t index) const {
+  ICEBERG_CHECK(index < kUuidSize, "UUID index out of range: {}", index);
+  return data_[index];
+}
+
+std::string Uuid::ToString() const {
+  static const char* hex_chars = "0123456789abcdef";
```
wgtmac:
```suggestion
  static const char* kHexChars = "0123456789abcdef";
```

wgtmac: Is it a convention to print lower cased letters for uuid?

src/iceberg/util/uuid.h
```diff
@@ -0,0 +1,81 @@
...
+
+/// \file iceberg/util/uuid.h
+/// \brief UUID (Universally Unique Identifier) representation.
+
+namespace iceberg {
+
+class ICEBERG_EXPORT Uuid {
+ public:
+  Uuid() = delete;
+  constexpr static size_t kUuidSize = 16;
```
wgtmac: nit: maybe `kSize` or `kLength` is enough?

src/iceberg/util/uuid.h
```diff
@@ -0,0 +1,81 @@
...
+  static Uuid GenerateV7(uint64_t unix_ts_ms);
+
+  /// \brief Create a UUID from a string in standard format.
+  static Result<Uuid> FromString(std::string_view str);
+
+  /// \brief Create a UUID from a 16-byte array.
+  static Result<Uuid> FromBytes(std::span<const uint8_t> bytes);
+
+  /// \brief Get the raw bytes of the UUID.
+  std::span<const uint8_t> bytes() const { return data_; }
```
wgtmac, line 66: Perhaps we need a `ToBigEndianBytes` for single value serialization?

src/iceberg/util/uuid.h
```diff
@@ -0,0 +1,81 @@
...
+
+  /// \brief Access individual bytes of the UUID.
+  uint8_t operator[](size_t index) const;
```
wgtmac, line 72: We might need to document if it throws when index is out of range.

src/iceberg/util/uuid.h
```diff
@@ -0,0 +1,81 @@
...
+
+  /// \brief Convert the UUID to a string in standard format.
+  std::string ToString() const;
+
+  friend bool operator==(const Uuid& lhs, const Uuid& rhs) {
```
wgtmac, line 77: Do we need to add `operator<=>` as well?

src/iceberg/util/uuid.cc
```diff
@@ -0,0 +1,221 @@
...
+  // Set magic numbers for a "version 7" (pseudorandom) UUID and variant,
+  // see https://www.rfc-editor.org/rfc/rfc9562#name-version-field
+  uuid[6] = (uuid[6] & 0x0F) | 0x70;
+  // set variant field, top two bits are 1, 0
+  uuid[8] = (uuid[8] & 0x3F) | 0x80;
+
+  return Uuid(std::move(uuid));
+}
+
+namespace {
```
HeartLinked, line 36: I have only one extra suggestion to add: to improve readability, I suggest moving the anonymous namespace to the top or end of the file.

src/iceberg/util/uuid.cc
zhjwpku: hex_chars is no longer needed any more, deleted.

src/iceberg/util/uuid.h
zhjwpku: changed to kLength

src/iceberg/util/uuid.h
zhjwpku: Added

src/iceberg/util/uuid.h
zhjwpku: Added

src/iceberg/util/uuid.h
zhjwpku: I can't think of a scenario where we'd need to compare two UUIDs right now. We can add that functionality later if needed, along with better documentation.

src/iceberg/util/uuid.cc
zhjwpku: moved, thanks.

src/iceberg/util/uuid.cc
```diff
@@ -0,0 +1,221 @@
...
+  std::array<uint8_t, 256> buf{};
+  for (int32_t i = 0; i < 256; i++) {
+    if (i >= '0' && i <= '9') {
+      buf[i] = static_cast<uint8_t>(i - '0');
+    } else if (i >= 'a' && i <= 'f') {
+      buf[i] = static_cast<uint8_t>(i - 'a' + 10);
+    } else if (i >= 'A' && i <= 'F') {
+      buf[i] = static_cast<uint8_t>(i - 'A' + 10);
+    } else {
+      buf[i] = 0xff;
```
HuaHuaY:
```suggestion
      buf[i] = 0xFF;
```

zhjwpku: Fixed, thanks

src/iceberg/util/uuid.h
HuaHuaY: You can remove `bytes()` because it's the same as `ToBigEndianBytes`.

src/iceberg/util/uuid.cc
```diff
@@ -0,0 +1,221 @@
...
+  std::memcpy(data.data(), bytes.data(), kLength);
+  return Uuid(std::move(data));
+}
+
+uint8_t Uuid::operator[](size_t index) const {
+  ICEBERG_CHECK(index < kLength, "UUID index out of range: {}", index);
+  return data_[index];
+}
+
+std::array<uint8_t, Uuid::kLength> Uuid::ToBigEndianBytes() const { return data_; }
```
wgtmac: Is it correct? Shouldn't `data_` use native endianness? Same question to `FromBytes` since we might need `FromBigEndianBytes` as well.

zhjwpku: I don't think byte array hold any endianness, just like ASCII string, there is an interesting stackoverflow question [1], here I quote one answer:

```
Endianness defines the order of bytes within multi-byte values. Character strings are arrays of single-byte values. So each value (character in the string) is the same on both little-endian and big-endian architectures, and endianness does not affect the order of values in a structure.
```

So I'm inclined to remove the methods related to BigEndian, WDYT?

[1] https://stackoverflow.com/questions/1568057/ascii-strings-and-endianness

wgtmac: Sounds good!

src/iceberg/util/uuid.h
```diff
@@ -0,0 +1,84 @@
...
+  std::span<const uint8_t> bytes() const { return data_; }
+
+  /// \brief Access individual bytes of the UUID.
+  /// \param index The index of the byte to access (0-15).
+  /// \return The byte at the specified index.
+  /// \throw IcebergError if index is out of bounds.
+  uint8_t operator[](size_t index) const;
+
+  /// \brief Convert the UUID to a string in standard format.
+  std::string ToString() const;
```
wgtmac: nit: inherit `util::Formattable`

wgtmac (approved):

src/iceberg/util/uuid.cc
```diff
@@ -0,0 +1,219 @@
...
+// Parse a UUID string without dashes, e.g. "67e5504410b1426f9247bb680e5fe0c8"
+inline Result<Uuid> ParseSimple(std::string_view s) {
+  ICEBERG_DCHECK(s.size() == 32, "s must be 32 characters long");
+
+  std::array<uint8_t, 16> uuid{};
+  for (size_t i = 0; i < 16; i++) {
+    uint8_t h1 = kHexTable[static_cast<uint8_t>(s[i * 2])];
+    uint8_t h2 = kHexTable[static_cast<uint8_t>(s[i * 2 + 1])];
+
+    if ((h1 | h2) == 0xFF) {
```
mapleFU: Unlikely?

src/iceberg/util/uuid.cc
```diff
@@ -0,0 +1,219 @@
...
+  std::array<uint8_t, 16> uuid{};
+
+  for (size_t j = 0; j < 8; j++) {
+    size_t i = positions[j];
+    uint8_t h1 = kHexTable[static_cast<uint8_t>(s[i])];
+    uint8_t h2 = kHexTable[static_cast<uint8_t>(s[i + 1])];
+    uint8_t h3 = kHexTable[static_cast<uint8_t>(s[i + 2])];
+    uint8_t h4 = kHexTable[static_cast<uint8_t>(s[i + 3])];
+
+    if ((h1 | h2 | h3 | h4) == 0xFF) {
```
mapleFU: Unlikely?

mapleFU (commented):
General LGTM

src/iceberg/util/uuid.h
zhjwpku: Fixed, thanks

src/iceberg/util/uuid.cc
zhjwpku: Fixed, thanks

zhjwpku:
Hi @Fokko @zeroshade @Xuanwo , this PR is ready for merge, I'd appreciate if you can take a look when available, thanks.

wgtmac (approved):

Fokko:
Looks good @zhjwpku, thanks for working on this, and thanks @mapleFU @wgtmac @HuaHuaY and @HeartLinked for the review!

PR #243, opened 2025-09-23
chore: Align LICENSE and NOTICE with ASF policy

HeartLinked:
Refines the LICENSE and NOTICE files to ensure full compliance with Apache Software Foundation (ASF) guidelines.https://infra.apache.org/licensing-howto.html

+ Simplify NOTICE File: Dependencies with permissive licenses (such as MIT, zlib, and Public Domain) have been removed from the NOTICE file. According to ASF policy, the copyright notices embedded within these types of licenses do not need to be duplicated in the project's NOTICE file. 

+ Reformat LICENSE File: The appendix of the LICENSE file, which lists third-party dependencies, has been reformatted for clarity and consistency. The new style is based on the Apache Iceberg repository's LICENSE file. https://github.com/apache/iceberg/blob/main/LICENSE

+ Add cpr license for rest catalog.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -214,9 +214,13 @@ License: Public Domain
 
 This product includes code from Apache Arrow.
 
-* checked_cast utility in src/iceberg/util/checked_cast.h.
-* visit_type utility in src/iceberg/util/visit_type.h.
-* Decimal128 implementation details in src/iceberg/util/decimal files.
+* Core utilities:
+  * checked_cast utility in src/iceberg/util/checked_cast.h.
+  * visit_type utility in src/iceberg/util/visit_type.h.
+  * Decimal128 implementation details in src/iceberg/util/decimal files.
+* Build system modules:
+  * cmake_modules/IcebergBuildUtils.cmake.
+  * cmake_modules/IcebergThirdpartyToolchain.cmake.
 
 Copyright: 2016-2025 The Apache Software Foundation.
 Home page: https://arrow.apache.org/


LICENSE
```diff
@@ -202,117 +202,63 @@
...
+Copyright: Austin Appleby (placed in the public domain).
+Home page: https://github.com/aappleby/smhasher
+License: Public Domain
 
 --------------------------------------------------------------------------------
 
-The file src/iceberg/murmur3_internal.h contains code adapted from
-
-https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.h
+This product bundles nlohmann-json.
```
jbonofre: By bundle, does it mean code copy or binary (so files for instance) shipped in the binary package.
The `LICENSE` here is for source distribution, so we should only mention code copied from other projects (not binary artifact bundles in the binary distribution).
If we distribute a binary artifact, I recommend to use a `LICENSE` file dedicated for the binary distribution.

LICENSE
```diff
@@ -202,117 +202,63 @@
...
+* JSON serialization/deserialization functionality in src/iceberg/json_internal.h and src/iceberg/json_internal.cc.
+* Used throughout the codebase for Iceberg table metadata processing.
 
-https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp
-
-MurmurHash3 was written by Austin Appleby, and is placed in the public
-domain. The author disclaims copyright to this source code.
+Copyright: 2013-2022 Niels Lohmann.
+Home page: https://github.com/nlohmann/json
+License: https://mit-license.org/
```
jbonofre: MIT license content should be copied here inline.

LICENSE
```diff
@@ -202,117 +202,63 @@
...
 Home page: https://arrow.apache.org/
 License: https://www.apache.org/licenses/LICENSE-2.0
 
 --------------------------------------------------------------------------------
 
-3rdparty dependency spdlog is statically linked in certain binary
-distributions. spdlog has the following license:
-
-MIT License
+This product bundles spdlog.
```
jbonofre: Same question: is it code copy or binary bundle ?

LICENSE
```diff
@@ -202,117 +202,63 @@
...
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
+Copyright: 2016 Gabi Melman.
+Home page: https://github.com/gabime/spdlog
+License: https://mit-license.org/
```
jbonofre: The MIT license content should be copied in-line here.

LICENSE
```diff
@@ -202,117 +202,63 @@
...
 
 --------------------------------------------------------------------------------
 
-3rdparty dependency zlib is used by certain binary distributions. zlib has
-the following license:
+This product bundles zlib.
```
jbonofre: Same question here: is it code copy or binary bundle ?

LICENSE
```diff
@@ -202,117 +202,63 @@
...
 
-zlib License
+* gzip compression/decompression functionality in src/iceberg/util/gzip_internal.cc
 
-Copyright (c) 1995-2024 Jean-loup Gailly and Mark Adler
+Copyright: 1995-2024 Jean-loup Gailly and Mark Adler.
+Home page: https://zlib.net/
+License: https://www.zlib.net/zlib_license.html
```
jbonofre: This license looks pretty close to MIT/BSD one. So I would sugest to copy the license content inline.

LICENSE
```diff
@@ -202,117 +202,63 @@
...
 
-This software is provided 'as-is', without any express or implied
```
jbonofre, line 301: This is actually better: for non Apache license (BSD/MIT, etc), the license content should be in this `LICENSE` or in folder `licenses` containing 3rd party license.

NOTICE
```diff
@@ -3,22 +3,3 @@ Copyright 2024-2025 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
```
jbonofre, line 5: This is correct. The `NOTICE` file should only contains the `NOTICE` of Apache licensed projects where the code has been copied (`NOTICE` should not be used for other license types), if provided, and only the relevant part of our project.
In the case of ASF projects (like Apache Arrow), it's not needed to copy `NOTICE` file as it's already covered by `This product includes software developed at The Apache Software Foundation (http://www.apache.org/)`.

jbonofre:
The `NOTICE` file is better and correct. Thanks for that.

Imho, the `LICENSE` file is not fully correct because:
1. The MIT/BSD license content is missing (I don't see a separate folder containing these licenses, and the license content is not inline).
2. It's not clear to me what do you mean by "bundle". Can you please clarify ?

LICENSE
wgtmac: I think the word `bundles` here is incorrect. We do not copy source files from `nlohmann-json`. Instead, we directly use system installed `nlohmann-json` or vendor it by building the downloaded tarball. Almost 3rd party libraries (nlohmann-json, nanoarrow, croaring, arrow-cpp, avro-cpp, cpr, spdlog) that we use follow the same pattern.

LICENSE
wgtmac: We use system installed one or vendor built binaries (if not installed).

LICENSE
wgtmac: For now, we only use the system installed zlib via CMake find_package.

LICENSE
jbonofre: Exactly. I don't think we need mention anything here as it's not code copy or actually bundling.

HeartLinked:
> The `NOTICE` file is better and correct. Thanks for that.
> 
> Imho, the `LICENSE` file is not fully correct because:
> 
> 1. The MIT/BSD license content is missing (I don't see a separate folder containing these licenses, and the license content is not inline).
> 2. It's not clear to me what do you mean by "bundle". Can you please clarify ?

Thanks for your review! Actually, our project is distributed as a source package that users compile themselves. We do not copy source files from `zlib`, `spdlog`, `cpr`, etc. Instead, we directly use system installed them or vendor it by building the downloaded tarball.  Previously, our LICENSE file contained sentences like "3rdparty dependency spdlog is statically linked in certain binary distributions". Based on this, I had mistakenly concluded that these dependencies should always be treated as "bundled".

Following our discussion, I have now removed all related third-party license content from the file, as our project is a source-only distribution where bundling is conditional. Do you think this revised version is more accurate?

LICENSE
```diff
@@ -202,117 +202,24 @@
...
-
-The file src/iceberg/murmur3_internal.cc contains code adapted from
-
-https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp
-
-MurmurHash3 was written by Austin Appleby, and is placed in the public
-domain. The author disclaims copyright to this source code.
+Copyright: Austin Appleby (placed in the public domain).
+Home page: https://github.com/aappleby/smhasher
+License: Public Domain
```
jbonofre, line 211: SMHasher is released under the MIT license. All MurmurHash versions are public domain software, and the author disclaims all copyright to their code.
As there's nothing special regarding the license on https://github.com/aappleby/smhasher, this is good to me.

jbonofre (approved):
It's correct now. Thanks !

LICENSE
HeartLinked: thanks!

LICENSE
```diff
@@ -202,117 +202,24 @@
...
-
-https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/decimal.h
-
-The file src/iceberg/util/decimal.cc contains code adapted from
+This product includes code from Apache Arrow.
 
-https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/decimal.cc
+* checked_cast utility in src/iceberg/util/checked_cast.h.
+* visit_type utility in src/iceberg/util/visit_type.h.
+* Decimal128 implementation details in src/iceberg/util/decimal files.
```
wgtmac: nit: cmake_modules/IcebergBuildUtils.cmake and cmake_modules/IcebergThirdpartyToolchain.cmake are also adapted from arrow-cpp

HeartLinked: I've added these.  :)

PR #244, opened 2025-09-23
refactor: use ICEBERG_CHECK where possible

zhjwpku:
This will simplify the code and potentially optimize the binary due to the [[unlikely]] attribute in ICEBERG_CHECK.


mapleFU (approved):
This looks good to me, but `throw` originally might already have internal unlikely in most compilers, so I think it just make code shorter

zhjwpku:
> This looks good to me, but `throw` originally might already have internal unlikely in most compilers, so I think it just make code shorter

Fair enough, thanks for the review.

PR #249, opened 2025-09-29
refactor: extract JSON utilities for code reuse

HeartLinked:

PR #248, opened 2025-09-29
refactor: unify enum string conversion

HeartLinked:

PR #251, opened 2025-09-29
feat: add json serde for REST Catalog request/response models

HeartLinked:
This PR depends on and should be merged after https://github.com/apache/iceberg-cpp/pull/230. #249 #245 #248

### DIFF
--- a/src/iceberg/catalog/rest/json_internal.cc
+++ b/src/iceberg/catalog/rest/json_internal.cc
@@ -19,7 +19,6 @@
 
 #include "iceberg/catalog/rest/json_internal.h"
 
-#include <ranges>
 #include <string>
 #include <unordered_map>
 #include <utility>
@@ -29,10 +28,7 @@
 
 #include "iceberg/catalog/rest/types.h"
 #include "iceberg/json_internal.h"
-#include "iceberg/partition_spec.h"
-#include "iceberg/schema.h"
-#include "iceberg/sort_order.h"
-#include "iceberg/table_metadata.h"
+#include "iceberg/table_identifier.h"
 #include "iceberg/util/json_util_internal.h"
 #include "iceberg/util/macros.h"
 
@@ -66,7 +62,6 @@ constexpr std::string_view kIdentifiers = "identifiers";
 
 }  // namespace
 
-// CreateNamespaceRequest
 nlohmann::json ToJson(const CreateNamespaceRequest& request) {
   nlohmann::json json;
   json[kNamespace] = request.namespace_.levels;
@@ -83,12 +78,10 @@ Result<CreateNamespaceRequest> CreateNamespaceRequestFromJson(
                           GetJsonValue<std::vector<std::string>>(json, kNamespace));
   ICEBERG_ASSIGN_OR_RAISE(
       request.properties,
-      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
-                                                                           kProperties)));
+      GetJsonValueOrDefault<decltype(request.properties)>(json, kProperties));
   return request;
 }
 
-// UpdateNamespacePropertiesRequest
 nlohmann::json ToJson(const UpdateNamespacePropertiesRequest& request) {
   // Initialize as an empty object so that when all optional fields are absent we return
   // {} instead of null
@@ -108,65 +101,10 @@ Result<UpdateNamespacePropertiesRequest> UpdateNamespacePropertiesRequestFromJso
   ICEBERG_ASSIGN_OR_RAISE(
       request.removals, GetJsonValueOrDefault<std::vector<std::string>>(json, kRemovals));
   ICEBERG_ASSIGN_OR_RAISE(
-      request.updates,
-      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
-                                                                           kUpdates)));
+      request.updates, GetJsonValueOrDefault<decltype(request.updates)>(json, kUpdates));
   return request;
 }
 
-// CreateTableRequest
-nlohmann::json ToJson(const CreateTableRequest& request) {
-  nlohmann::json json;
-  json[kName] = request.name;
-  if (!request.location.empty()) {
-    json[kLocation] = request.location;
-  }
-  json[kSchema] = ToJson(*request.schema);
-  if (request.partition_spec) {
-    json[kPartitionSpec] = ToJson(*request.partition_spec);
-  }
-  if (request.write_order) {
-    json[kWriteOrder] = ToJson(*request.write_order);
-  }
-  SetOptionalField(json, kStageCreate, request.stage_create);
-  if (!request.properties.empty()) {
-    json[kProperties] = request.properties;
-  }
-  return json;
-}
-
-Result<CreateTableRequest> CreateTableRequestFromJson(const nlohmann::json& json) {
-  CreateTableRequest request;
-  ICEBERG_ASSIGN_OR_RAISE(request.name, GetJsonValue<std::string>(json, kName));
-  ICEBERG_ASSIGN_OR_RAISE(request.location,
-                          GetJsonValueOrDefault<std::string>(json, kLocation, ""));
-  ICEBERG_ASSIGN_OR_RAISE(auto schema_json, GetJsonValue<nlohmann::json>(json, kSchema));
-  ICEBERG_ASSIGN_OR_RAISE(request.schema, SchemaFromJson(schema_json));
-  if (json.contains(kPartitionSpec)) {
-    ICEBERG_ASSIGN_OR_RAISE(auto partition_spec_json,
-                            GetJsonValue<nlohmann::json>(json, kPartitionSpec));
-    ICEBERG_ASSIGN_OR_RAISE(request.partition_spec,
-                            PartitionSpecFromJson(request.schema, partition_spec_json));
-  } else {
-    request.partition_spec = nullptr;
-  }
-  if (json.contains(kWriteOrder)) {
-    ICEBERG_ASSIGN_OR_RAISE(auto write_order_json,
-                            GetJsonValue<nlohmann::json>(json, kWriteOrder));
-    ICEBERG_ASSIGN_OR_RAISE(request.write_order, SortOrderFromJson(write_order_json));
-  } else {
-    request.write_order = nullptr;
-  }
-  ICEBERG_ASSIGN_OR_RAISE(request.stage_create,
-                          GetJsonValueOptional<bool>(json, kStageCreate));
-  ICEBERG_ASSIGN_OR_RAISE(
-      request.properties,
-      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
-                                                                           kProperties)));
-  return request;
-}
-
-// RegisterTableRequest
 nlohmann::json ToJson(const RegisterTableRequest& request) {
   nlohmann::json json;
   json[kName] = request.name;
@@ -182,14 +120,11 @@ Result<RegisterTableRequest> RegisterTableRequestFromJson(const nlohmann::json& 
   ICEBERG_ASSIGN_OR_RAISE(request.name, GetJsonValue<std::string>(json, kName));
   ICEBERG_ASSIGN_OR_RAISE(request.metadata_location,
                           GetJsonValue<std::string>(json, kMetadataLocation));
-  // Default to false if not present
-  ICEBERG_ASSIGN_OR_RAISE(auto overwrite_opt,
-                          GetJsonValueOptional<bool>(json, kOverwrite));
-  request.overwrite = overwrite_opt.value_or(false);
+  ICEBERG_ASSIGN_OR_RAISE(request.overwrite,
+                          GetJsonValueOrDefault<bool>(json, kOverwrite, false));
   return request;
 }
 
-// RenameTableRequest
 nlohmann::json ToJson(const RenameTableRequest& request) {
   nlohmann::json json;
   json[kSource] = ToJson(request.source);
@@ -222,8 +157,8 @@ nlohmann::json ToJson(const LoadTableResult& result) {
 
 Result<LoadTableResult> LoadTableResultFromJson(const nlohmann::json& json) {
   LoadTableResult result;
-  ICEBERG_ASSIGN_OR_RAISE(result.metadata_location, GetJsonValueOrDefault<std::string>(
-                                                        json, kMetadataLocation, ""));
+  ICEBERG_ASSIGN_OR_RAISE(result.metadata_location,
+                          GetJsonValueOrDefault<std::string>(json, kMetadataLocation));
   ICEBERG_ASSIGN_OR_RAISE(auto metadata_json,
                           GetJsonValue<nlohmann::json>(json, kMetadata));
   ICEBERG_ASSIGN_OR_RAISE(result.metadata, TableMetadataFromJson(metadata_json));
@@ -233,7 +168,6 @@ Result<LoadTableResult> LoadTableResultFromJson(const nlohmann::json& json) {
   return result;
 }
 
-// ListNamespacesResponse
 nlohmann::json ToJson(const ListNamespacesResponse& response) {
   nlohmann::json json;
   if (!response.next_page_token.empty()) {
@@ -251,7 +185,7 @@ Result<ListNamespacesResponse> ListNamespacesResponseFromJson(
     const nlohmann::json& json) {
   ListNamespacesResponse response;
   ICEBERG_ASSIGN_OR_RAISE(response.next_page_token,
-                          GetJsonValueOrDefault<std::string>(json, kNextPageToken, ""));
+                          GetJsonValueOrDefault<std::string>(json, kNextPageToken));
   ICEBERG_ASSIGN_OR_RAISE(auto namespaces_json,
                           GetJsonValue<nlohmann::json>(json, kNamespaces));
   for (const auto& ns_json : namespaces_json) {
@@ -261,7 +195,6 @@ Result<ListNamespacesResponse> ListNamespacesResponseFromJson(
   return response;
 }
 
-// CreateNamespaceResponse
 nlohmann::json ToJson(const CreateNamespaceResponse& response) {
   nlohmann::json json;
   json[kNamespace] = response.namespace_.levels;
@@ -278,12 +211,10 @@ Result<CreateNamespaceResponse> CreateNamespaceResponseFromJson(
                           GetJsonValue<std::vector<std::string>>(json, kNamespace));
   ICEBERG_ASSIGN_OR_RAISE(
       response.properties,
-      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
-                                                                           kProperties)));
+      GetJsonValueOrDefault<decltype(response.properties)>(json, kProperties));
   return response;
 }
 
-// GetNamespaceResponse
 nlohmann::json ToJson(const GetNamespaceResponse& response) {
   nlohmann::json json;
   json[kNamespace] = response.namespace_.levels;
@@ -299,12 +230,10 @@ Result<GetNamespaceResponse> GetNamespaceResponseFromJson(const nlohmann::json& 
                           GetJsonValue<std::vector<std::string>>(json, kNamespace));
   ICEBERG_ASSIGN_OR_RAISE(
       response.properties,
-      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
-                                                                           kProperties)));
+      GetJsonValueOrDefault<decltype(response.properties)>(json, kProperties));
   return response;
 }
 
-// UpdateNamespacePropertiesResponse
 nlohmann::json ToJson(const UpdateNamespacePropertiesResponse& response) {
   nlohmann::json json;
   json[kUpdated] = response.updated;
@@ -327,7 +256,6 @@ Result<UpdateNamespacePropertiesResponse> UpdateNamespacePropertiesResponseFromJ
   return response;
 }
 
-// ListTablesResponse
 nlohmann::json ToJson(const ListTablesResponse& response) {
   nlohmann::json json;
   if (!response.next_page_token.empty()) {
@@ -344,7 +272,7 @@ nlohmann::json ToJson(const ListTablesResponse& response) {
 Result<ListTablesResponse> ListTablesResponseFromJson(const nlohmann::json& json) {
   ListTablesResponse response;
   ICEBERG_ASSIGN_OR_RAISE(response.next_page_token,
-                          GetJsonValueOrDefault<std::string>(json, kNextPageToken, ""));
+                          GetJsonValueOrDefault<std::string>(json, kNextPageToken));
   ICEBERG_ASSIGN_OR_RAISE(auto identifiers_json,
                           GetJsonValue<nlohmann::json>(json, kIdentifiers));
   for (const auto& id_json : identifiers_json) {
--- a/src/iceberg/catalog/rest/json_internal.h
+++ b/src/iceberg/catalog/rest/json_internal.h
@@ -77,13 +77,6 @@ ICEBERG_REST_EXPORT nlohmann::json ToJson(const ListTablesResponse& response);
 ICEBERG_REST_EXPORT Result<ListTablesResponse> ListTablesResponseFromJson(
     const nlohmann::json& json);
 
-/// \brief Serializes a `CreateTableRequest` object to JSON.
-ICEBERG_REST_EXPORT nlohmann::json ToJson(const CreateTableRequest& request);
-
-/// \brief Deserializes a JSON object into a `CreateTableRequest` object.
-ICEBERG_REST_EXPORT Result<CreateTableRequest> CreateTableRequestFromJson(
-    const nlohmann::json& json);
-
 /// \brief Serializes a `LoadTableResult` object to JSON.
 ICEBERG_REST_EXPORT nlohmann::json ToJson(const LoadTableResult& result);
 
--- a/src/iceberg/catalog/rest/meson.build
+++ b/src/iceberg/catalog/rest/meson.build
@@ -46,7 +46,4 @@ iceberg_rest_dep = declare_dependency(
 meson.override_dependency('iceberg-rest', iceberg_rest_dep)
 pkg.generate(iceberg_rest_lib)
 
-install_headers(
-    ['rest_catalog.h', 'types.h', 'json_internal.h'],
-    subdir: 'iceberg/catalog/rest',
-)
+install_headers(['rest_catalog.h', 'types.h'], subdir: 'iceberg/catalog/rest')
--- a/src/iceberg/catalog/rest/types.h
+++ b/src/iceberg/catalog/rest/types.h
@@ -46,17 +46,6 @@ struct ICEBERG_REST_EXPORT UpdateNamespacePropertiesRequest {
   std::unordered_map<std::string, std::string> updates;
 };
 
-/// \brief Request to create a table.
-struct ICEBERG_REST_EXPORT CreateTableRequest {
-  std::string name;  // required
-  std::string location;
-  std::shared_ptr<Schema> schema;  // required
-  std::shared_ptr<PartitionSpec> partition_spec;
-  std::shared_ptr<SortOrder> write_order;
-  std::optional<bool> stage_create;
-  std::unordered_map<std::string, std::string> properties;
-};
-
 /// \brief Request to register a table.
 struct ICEBERG_REST_EXPORT RegisterTableRequest {
   std::string name;               // required
--- a/src/iceberg/test/rest_json_internal_test.cc
+++ b/src/iceberg/test/rest_json_internal_test.cc
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-#include <memory>
+#include <algorithm>
 #include <string>
 #include <unordered_map>
 #include <vector>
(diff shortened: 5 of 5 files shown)


src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -15,25 +15,35 @@
...
+list(APPEND
+     ICEBERG_REST_STATIC_BUILD_INTERFACE_LIBS
+     "$<IF:$<TARGET_EXISTS:iceberg_static>,iceberg_static,iceberg_shared>"
+     cpr::cpr
+     nlohmann_json::nlohmann_json)
+list(APPEND
+     ICEBERG_REST_SHARED_BUILD_INTERFACE_LIBS
+     "$<IF:$<TARGET_EXISTS:iceberg_shared>,iceberg_shared,iceberg_static>"
+     cpr::cpr
+     nlohmann_json::nlohmann_json)
```
```suggestion
list(APPEND
     ICEBERG_REST_STATIC_BUILD_INTERFACE_LIBS
     "$<IF:$<TARGET_EXISTS:iceberg_static>,iceberg_static,iceberg_shared>"
     cpr::cpr)
list(APPEND
     ICEBERG_REST_SHARED_BUILD_INTERFACE_LIBS
     "$<IF:$<TARGET_EXISTS:iceberg_shared>,iceberg_shared,iceberg_static>"
     cpr::cpr)
```

We don't need `nlohmann_json::nlohmann_json` explicitly because it is already a `BUILD_INTERFACE` of the `iceberg` library. Could you double check?

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -15,25 +15,35 @@
...
 list(APPEND
      ICEBERG_REST_STATIC_INSTALL_INTERFACE_LIBS
      "$<IF:$<TARGET_EXISTS:iceberg::iceberg_static>,iceberg::iceberg_static,iceberg::iceberg_shared>"
-     "$<IF:$<BOOL:${CPR_VENDORED}>,iceberg::cpr,cpr::cpr>")
+     "$<IF:$<BOOL:${CPR_VENDORED}>,iceberg::cpr,cpr::cpr>"
```
wgtmac: Same for here and below.

src/iceberg/catalog/rest/types.h
```diff
@@ -0,0 +1,112 @@
...
+
+#include "iceberg/partition_spec.h"
+#include "iceberg/schema.h"
+#include "iceberg/sort_order.h"
+#include "iceberg/table_identifier.h"
+#include "iceberg/table_metadata.h"
+
+namespace iceberg::rest {
+
+struct ListNamespaceResponse {
```
wgtmac: We need to include `iceberg/catalog/rest/iceberg_rest_export.h` and add `ICEBERG_REST_EXPORT` to all classes to export them in the shared library.

```suggestion
struct ICEBERG_REST_EXPORT ListNamespaceResponse {
```

wgtmac: BTW, this should be `ListNamespacesResponse` to be consistent with the rest spec.

src/iceberg/catalog/rest/types.h
```diff
@@ -0,0 +1,112 @@
...
+ */
+#pragma once
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "iceberg/partition_spec.h"
```
wgtmac: Try to include `iceberg/type_fwd.h` (and remove some other header files below) to use forward declaration as much as possible. The main point is to make the compiler run faster.

Perhaps we can also add `iceberg/catalog/rest/type_fwd.h` to make it easy to use forward declaration of the rest catalog library. It can be a separate PR later.

src/iceberg/catalog/rest/types.h
```diff
@@ -0,0 +1,112 @@
...
+namespace iceberg::rest {
+
+struct ListNamespaceResponse {
+  std::vector<std::vector<std::string>> namespaces;
+  // TODO(Li Feiyang): Add next_page_token
+};
+
+struct CreateNamespaceRequest {
+  std::vector<std::string> namespaces;
+  std::optional<std::unordered_map<std::string, std::string>> properties;
```
```suggestion
  std::unordered_map<std::string, std::string> properties;
```

We don't need an extra `std::optional` here because an empty map can be considered empty.

src/iceberg/catalog/rest/types.h
```diff
@@ -0,0 +1,112 @@
...
+#include "iceberg/partition_spec.h"
+#include "iceberg/schema.h"
+#include "iceberg/sort_order.h"
+#include "iceberg/table_identifier.h"
+#include "iceberg/table_metadata.h"
+
+namespace iceberg::rest {
+
+struct ListNamespaceResponse {
+  std::vector<std::vector<std::string>> namespaces;
```
```suggestion
  std::vector<Namespace> namespaces;
```

wgtmac: Let's reuse defined classes as many as possible.

src/iceberg/catalog/rest/types.h
wgtmac: Same for below

src/iceberg/catalog/rest/types.h
```diff
@@ -0,0 +1,112 @@
...
+  // TODO(Li Feiyang): Add next_page_token
+};
+
+struct CreateNamespaceRequest {
```
wgtmac: I'd suggest to add a file like [iceberg_rest_api_todo.md](https://github.com/user-attachments/files/23044946/iceberg_rest_api_todo.md) to the `rest` directory to keep track of the progress.

wgtmac: For each class, please add some docstring. We can simply copy the description from the rest api spec.

src/iceberg/catalog/rest/types.h
```diff
@@ -0,0 +1,112 @@
...
+};
+
+struct ListTableResponse {
+  std::optional<std::vector<TableIdentifier>> identifiers;
+  // TODO(Li Feiyang): Add next_page_token
+};
+
+struct CreateTableRequest {
+  std::string name;
+  std::optional<std::string> location;
```
```suggestion
  std::string location;
```

An empty string is enough to represent an optional one.

src/iceberg/catalog/rest/types.h
```diff
@@ -0,0 +1,112 @@
...
+  std::shared_ptr<Schema> schema;
+  std::shared_ptr<PartitionSpec> partition_spec;  // optional
```
wgtmac: Instead of adding `// optional` comment, can we add `// required` to those required fields?

src/iceberg/catalog/rest/json_internal.h
```diff
@@ -0,0 +1,190 @@
...
+///
+/// \param request The `CreateNamespaceRequest` object to be serialized.
+/// \return A JSON object representing the `CreateNamespaceRequest`.
+nlohmann::json ToJson(const CreateNamespaceRequest& request);
+
+/// \brief Deserializes a JSON object into a `CreateNamespaceRequest` object.
+///
+/// \param json The JSON object representing a `CreateNamespaceRequest`.
+/// \return A `CreateNamespaceRequest` object or an error if the conversion fails.
+Result<CreateNamespaceRequest> CreateNamespaceRequestFromJson(const nlohmann::json& json);
```
wgtmac: The signature of these `FromJson` functions is determined by how we will use request and response objects. If we use them without smart pointers, we can keep the current ones. If we use them with smart pointers, we need to follow `iceberg/json_internal.h` to return `Result<std::unique_ptr<X>` for them. I think the current ones are fine.

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-set(ICEBERG_REST_SOURCES rest_catalog.cc)
+set(ICEBERG_REST_SOURCES rest_catalog.cc json_internal.cc)
```
wgtmac, line 18: Please update the corresponding `meson.build` file

src/iceberg/test/CMakeLists.txt
```diff
@@ -150,7 +150,8 @@ if(ICEBERG_BUILD_BUNDLE)
 endif()
 
 if(ICEBERG_BUILD_REST)
-  add_iceberg_test(rest_catalog_test SOURCES rest_catalog_test.cc)
+  add_iceberg_test(rest_catalog_test SOURCES rest_catalog_test.cc
```
wgtmac, line 153: Please update `meson.build` as well.

src/iceberg/catalog/rest/json_internal.h
```diff
@@ -0,0 +1,166 @@
...
+#include "iceberg/catalog/rest/types.h"
+#include "iceberg/result.h"
+
+namespace iceberg::rest {
+
+/// \brief Serializes a `CreateNamespaceRequest` object to JSON.
+///
+/// \param request The `CreateNamespaceRequest` object to be serialized.
+/// \return A JSON object representing the `CreateNamespaceRequest`.
+nlohmann::json ToJson(const CreateNamespaceRequest& request);
```
wgtmac: Add `ICEBERG_REST_EXPORT` to all these functions.

src/iceberg/catalog/rest/json_internal.h
```diff
@@ -0,0 +1,166 @@
...
+
+#include "iceberg/catalog/rest/types.h"
+#include "iceberg/result.h"
+
+namespace iceberg::rest {
+
+/// \brief Serializes a `CreateNamespaceRequest` object to JSON.
+///
+/// \param request The `CreateNamespaceRequest` object to be serialized.
+/// \return A JSON object representing the `CreateNamespaceRequest`.
```
wgtmac:
```suggestion
/// \brief Serializes a `CreateNamespaceRequest` object to JSON.
```

Let's simplify the docstring because all of them look similar. It is also acceptable to remove all comments in this file because function signatures are self-explanatory.

BTW, please reorder the functions to put the respective request and response close to each other.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+constexpr std::string_view kSource = "source";
+constexpr std::string_view kDestination = "destination";
+constexpr std::string_view kMetadata = "metadata";
+constexpr std::string_view kConfig = "config";
+constexpr std::string_view kIdentifiers = "identifiers";
+
+using MapType = std::unordered_map<std::string, std::string>;
+
+/// Helper function to convert TableIdentifier to JSON
+nlohmann::json ToJson(const TableIdentifier& identifier) {
```
wgtmac: I think this and the below one are better to be in the `iceberg/json_internal.h/.cc`

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  return json;
+}
+
+Result<CreateNamespaceRequest> CreateNamespaceRequestFromJson(
+    const nlohmann::json& json) {
+  CreateNamespaceRequest request;
+
+  ICEBERG_ASSIGN_OR_RAISE(auto levels,
+                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
+  request.namespace_.levels = std::move(levels);
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(request.namespace_.levels,
                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+constexpr std::string_view kStageCreate = "stage-create";
+constexpr std::string_view kMetadataLocation = "metadata-location";
+constexpr std::string_view kOverwrite = "overwrite";
+constexpr std::string_view kSource = "source";
+constexpr std::string_view kDestination = "destination";
+constexpr std::string_view kMetadata = "metadata";
+constexpr std::string_view kConfig = "config";
+constexpr std::string_view kIdentifiers = "identifiers";
+
+using MapType = std::unordered_map<std::string, std::string>;
```
wgtmac: This is not a good alias pattern. It leads to confusion with `iceberg::MapType` and conceals information like key/value type. I'd suggest directly using the original type during json serialization.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+
+  ICEBERG_ASSIGN_OR_RAISE(request.properties,
+                          GetJsonValueOrDefault<MapType>(json, kProperties));
+
+  return request;
+}
+
+// UpdateNamespacePropertiesRequest
+nlohmann::json ToJson(const UpdateNamespacePropertiesRequest& request) {
+  nlohmann::json json = nlohmann::json::object();
```
wgtmac, line 88:
```suggestion
  nlohmann::json json;
```

Please be consistent with others.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+nlohmann::json ToJson(const CreateTableRequest& request) {
+  nlohmann::json json;
+
+  json[kName] = request.name;
+
+  if (!request.location.empty()) {
+    json[kLocation] = request.location;
+  }
+
+  json[kSchema] = iceberg::ToJson(*request.schema);
```
wgtmac:
```suggestion
  json[kSchema] = ToJson(*request.schema);
```

Same for below.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(
+      request.removals, GetJsonValueOrDefault<std::vector<std::string>>(json, kRemovals));
+  ICEBERG_ASSIGN_OR_RAISE(request.updates,
+                          GetJsonValueOrDefault<MapType>(json, kUpdates));
+
+  return request;
+}
+
+// CreateTableRequest
+nlohmann::json ToJson(const CreateTableRequest& request) {
```
wgtmac: Both spec and order should use unbound form. I think we need to remove `CreateTableRequest` at the moment because we don't have them yet. FTR, Java implements it like the below:

```java
public class CreateTableRequest implements RESTRequest {

  private UnboundPartitionSpec partitionSpec;
  private UnboundSortOrder writeOrder;

  public PartitionSpec spec() {
    return partitionSpec != null ? partitionSpec.bind(schema) : null;
  }

  public SortOrder writeOrder() {
    return writeOrder != null ? writeOrder.bind(schema) : null;
  }

}
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  SetOptionalField(json, kStageCreate, request.stage_create);
+
+  if (!request.properties.empty()) {
+    json[kProperties] = request.properties;
+  }
+
+  return json;
+}
+
+Result<CreateTableRequest> CreateTableRequestFromJson(const nlohmann::json& json) {
```
wgtmac: Same as above. We need to remove it ATM.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  return request;
+}
+
+// RegisterTableRequest
+nlohmann::json ToJson(const RegisterTableRequest& request) {
+  nlohmann::json json;
+  json[kName] = request.name;
+  json[kMetadataLocation] = request.metadata_location;
+  if (request.overwrite) {
+    json[kOverwrite] = request.overwrite;
```
wgtmac, line 113:
```suggestion
    json[kOverwrite] = *request.overwrite;
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(request.destination, TableIdentifierFromJson(dest_json));
+
+  return request;
+}
+
+// LoadTableResult (used by CreateTableResponse, LoadTableResponse)
+nlohmann::json ToJson(const LoadTableResult& result) {
+  nlohmann::json json;
+
+  if (result.metadata_location.has_value() && !result.metadata_location->empty()) {
```
wgtmac: Sorry I may have missed the definition of `LoadTableResult`. Please fix it like below

```
diff --git a/src/iceberg/catalog/rest/types.h b/src/iceberg/catalog/rest/types.h
index 4c50ab2..62b9f71 100644
--- a/src/iceberg/catalog/rest/types.h
+++ b/src/iceberg/catalog/rest/types.h
@@ -75,8 +75,8 @@ using PageToken = std::string;
 
 /// \brief Result body for table create/load/register APIs.
 struct ICEBERG_REST_EXPORT LoadTableResult {
-  std::optional<std::string> metadata_location;
-  std::shared_ptr<TableMetadata> metadata;  // required  // required
+  std::string metadata_location;
+  std::shared_ptr<TableMetadata> metadata;  // required
   std::unordered_map<std::string, std::string> config;
   // TODO(Li Feiyang): Add std::shared_ptr<StorageCredential> storage_credential;
 };
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+    json[kMetadataLocation] = result.metadata_location.value();
+  }
+
```
wgtmac: It is unnecessary to insert blank lines between them.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  return json;
+}
+
+Result<LoadTableResult> LoadTableResultFromJson(const nlohmann::json& json) {
+  LoadTableResult result;
+
+  ICEBERG_ASSIGN_OR_RAISE(auto metadata_location,
+                          GetJsonValueOptional<std::string>(json, kMetadataLocation));
+  result.metadata_location = metadata_location;
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(result.metadata_location,
                          GetJsonValueOrDefault<std::string>(json, kMetadataLocation));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+
+  ICEBERG_ASSIGN_OR_RAISE(auto metadata_json,
+                          GetJsonValue<nlohmann::json>(json, kMetadata));
+  ICEBERG_ASSIGN_OR_RAISE(auto metadata_ptr,
+                          iceberg::TableMetadataFromJson(metadata_json));
+  result.metadata = std::move(metadata_ptr);
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(result.metadata,
                          TableMetadataFromJson(metadata_json));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  return json;
+}
+
+Result<ListNamespacesResponse> ListNamespacesResponseFromJson(
+    const nlohmann::json& json) {
+  ListNamespacesResponse response;
+
+  ICEBERG_ASSIGN_OR_RAISE(auto next_page_token,
+                          GetJsonValueOptional<std::string>(json, kNextPageToken));
+  response.next_page_token = next_page_token.value_or("");
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(response.next_page_token,
                          GetJsonValueOrDefault<std::string>(json, kNextPageToken));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto namespace_levels,
+      GetJsonValue<std::vector<std::vector<std::string>>>(json, kNamespaces));
+  response.namespaces.reserve(namespace_levels.size());
```
wgtmac: nit: you may use std::ranges with transfrom

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+}
+
+Result<CreateNamespaceResponse> CreateNamespaceResponseFromJson(
+    const nlohmann::json& json) {
+  CreateNamespaceResponse response;
+
+  ICEBERG_ASSIGN_OR_RAISE(auto levels,
+                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
+  response.namespace_.levels = std::move(levels);
+
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(response.namespace_.levels,
                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  }
+  return json;
+}
+
+Result<GetNamespaceResponse> GetNamespaceResponseFromJson(const nlohmann::json& json) {
+  GetNamespaceResponse response;
+
+  ICEBERG_ASSIGN_OR_RAISE(auto levels,
+                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
+  response.namespace_.levels = std::move(levels);
```
wgtmac: ditto

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  nlohmann::json json;
+  json[kNamespace] = response.namespace_.levels;
+  if (!response.properties.empty()) {
+    json[kProperties] = response.properties;
+  }
+  return json;
+}
+
+Result<GetNamespaceResponse> GetNamespaceResponseFromJson(const nlohmann::json& json) {
+  GetNamespaceResponse response;
```
wgtmac, line 228: I don't think we need to add blank lines in between like this function.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,433 @@
...
+  }
+  json[kIdentifiers] = identifiers_json;
+
+  return json;
+}
+
+Result<ListTablesResponse> ListTablesResponseFromJson(const nlohmann::json& json) {
+  ListTablesResponse response;
+
+  ICEBERG_ASSIGN_OR_RAISE(auto next_page_token,
```
wgtmac: ditto

src/iceberg/test/rest_json_internal_test.cc
```diff
@@ -0,0 +1,664 @@
...
+
+#include "iceberg/catalog/rest/json_internal.h"
+#include "iceberg/partition_spec.h"
+#include "iceberg/schema.h"
+#include "iceberg/snapshot.h"
+#include "iceberg/sort_order.h"
+#include "iceberg/table_identifier.h"
+#include "iceberg/table_metadata.h"
+#include "iceberg/transform.h"
+#include "matchers.h"
```
wgtmac:
```suggestion
#include "iceberg/test/matchers.h"
```

wgtmac (changes requested):
Thanks for working on this! I mainly have two feedbacks other than those inline comments:

1. All Java request and reposes classes inherit from `RESTMessage` below with a `validate` method. I think this is important because these objects may be created out of our control and must be validated before use.

```java
/** Interface to mark both REST requests and responses. */
public interface RESTMessage {

  /**
   * Ensures that a constructed instance of a REST message is valid according to the REST spec.
   *
   * <p>This is needed when parsing data that comes from external sources and the object might have
   * been constructed without all the required fields present.
   */
  void validate();
}
```

I'd propose to add a `iceberg/catalog/rest/validator` with interface like below for every request and response:

```cpp
struct Validator {
  static Status Validate(const ListNamespacesResponse& resp);
  ...
};
```

2. Interoperability is our top priority. I have followed all Java test cases for TableMetadata serde by adding them to `src/iceberg/test/metadata_serde_test.cc`. There are a lot of test in the Java repo like `org/apache/iceberg/rest/responses/TestListNamespacesResponse.java` which contains both good and bad cases. It would be good to include them while implementing each objects.

src/iceberg/test/rest_json_internal_test.cc
```diff
@@ -0,0 +1,664 @@
...
+void TestJsonRoundTrip(const T& obj) {
+  auto j = ToJson(obj);
+  auto parsed = FromJsonHelper<T>(j);
+  ASSERT_TRUE(parsed.has_value()) << parsed.error().message;
+  auto j2 = ToJson(parsed.value());
+  EXPECT_EQ(j, j2) << "Round-trip JSON mismatch.";
+}
+
+template <typename T>
+void TestJsonConversion(const T& obj, const nlohmann::json& expected_json) {
```
wgtmac: What's the difference of this function compared to the above?

src/iceberg/test/rest_json_internal_test.cc
```diff
@@ -0,0 +1,664 @@
...
+                             .location = "",
+                             .schema = schema,
+                             .partition_spec = nullptr,
+                             .write_order = nullptr,
+                             .stage_create = std::nullopt,
+                             .properties = {}};
+
+  TestJsonRoundTrip(request);
+
+  auto json = ToJson(request);
```
wgtmac: Why do we need to test the below if the round trip test has already covered this?

src/iceberg/catalog/rest/meson.build
```diff
@@ -46,4 +46,7 @@ iceberg_rest_dep = declare_dependency(
 meson.override_dependency('iceberg-rest', iceberg_rest_dep)
 pkg.generate(iceberg_rest_lib)
 
-install_headers(['rest_catalog.h', 'types.h'], subdir: 'iceberg/catalog/rest')
+install_headers(
+    ['rest_catalog.h', 'types.h', 'json_internal.h'],
```
wgtmac:
```suggestion
    ['rest_catalog.h', 'types.h'],
```

Internal headers are not supposed to be installed.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+
+Result<CreateNamespaceRequest> CreateNamespaceRequestFromJson(
+    const nlohmann::json& json) {
+  CreateNamespaceRequest request;
+  ICEBERG_ASSIGN_OR_RAISE(request.namespace_.levels,
+                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
+  ICEBERG_ASSIGN_OR_RAISE(
+      request.properties,
+      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
+                                                                           kProperties)));
```
wgtmac:
```suggestion
      GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
                                                                          kProperties));
```

Why adding an extra pair of parenthesis?

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+constexpr std::string_view kOverwrite = "overwrite";
+constexpr std::string_view kSource = "source";
+constexpr std::string_view kDestination = "destination";
+constexpr std::string_view kMetadata = "metadata";
+constexpr std::string_view kConfig = "config";
+constexpr std::string_view kIdentifiers = "identifiers";
+
+}  // namespace
+
+// CreateNamespaceRequest
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(request.namespace_.levels,
+                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
+  ICEBERG_ASSIGN_OR_RAISE(
+      request.properties,
+      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
+                                                                           kProperties)));
+  return request;
+}
+
+// UpdateNamespacePropertiesRequest
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+}
+
+Result<UpdateNamespacePropertiesRequest> UpdateNamespacePropertiesRequestFromJson(
+    const nlohmann::json& json) {
+  UpdateNamespacePropertiesRequest request;
+  ICEBERG_ASSIGN_OR_RAISE(
+      request.removals, GetJsonValueOrDefault<std::vector<std::string>>(json, kRemovals));
+  ICEBERG_ASSIGN_OR_RAISE(
+      request.updates,
+      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
```
wgtmac: ditto

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// CreateTableRequest
+nlohmann::json ToJson(const CreateTableRequest& request) {
+  nlohmann::json json;
+  json[kName] = request.name;
+  if (!request.location.empty()) {
+    json[kLocation] = request.location;
+  }
+  json[kSchema] = ToJson(*request.schema);
```
wgtmac: request.schema is a pointer and we need to check if it is nullptr before dereferencing it.

wgtmac: Perhaps you can also add `SetNullableField` and `SetArrayField` just like `SetOptionalField`.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(request.stage_create,
+                          GetJsonValueOptional<bool>(json, kStageCreate));
+  ICEBERG_ASSIGN_OR_RAISE(
+      request.properties,
+      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
+                                                                           kProperties)));
+  return request;
+}
+
+// RegisterTableRequest
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+    json[kProperties] = request.properties;
+  }
+  return json;
+}
+
+Result<CreateTableRequest> CreateTableRequestFromJson(const nlohmann::json& json) {
+  CreateTableRequest request;
+  ICEBERG_ASSIGN_OR_RAISE(request.name, GetJsonValue<std::string>(json, kName));
+  ICEBERG_ASSIGN_OR_RAISE(request.location,
+                          GetJsonValueOrDefault<std::string>(json, kLocation, ""));
```
wgtmac:
```suggestion
                          GetJsonValueOrDefault<std::string>(json, kLocation));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+    ICEBERG_ASSIGN_OR_RAISE(request.write_order, SortOrderFromJson(write_order_json));
+  } else {
+    request.write_order = nullptr;
+  }
+  ICEBERG_ASSIGN_OR_RAISE(request.stage_create,
+                          GetJsonValueOptional<bool>(json, kStageCreate));
+  ICEBERG_ASSIGN_OR_RAISE(
+      request.properties,
+      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
+                                                                           kProperties)));
```
wgtmac:
```suggestion
      GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
                                                                          kProperties));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+
+Result<RegisterTableRequest> RegisterTableRequestFromJson(const nlohmann::json& json) {
+  RegisterTableRequest request;
+  ICEBERG_ASSIGN_OR_RAISE(request.name, GetJsonValue<std::string>(json, kName));
+  ICEBERG_ASSIGN_OR_RAISE(request.metadata_location,
+                          GetJsonValue<std::string>(json, kMetadataLocation));
+  // Default to false if not present
+  ICEBERG_ASSIGN_OR_RAISE(auto overwrite_opt,
+                          GetJsonValueOptional<bool>(json, kOverwrite));
+  request.overwrite = overwrite_opt.value_or(false);
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(request.overwrite,
                          GetJsonValueOrDefault<bool>(json, kOverwrite, false));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  return request;
+}
+
+// RenameTableRequest
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+Result<LoadTableResult> LoadTableResultFromJson(const nlohmann::json& json) {
+  LoadTableResult result;
+  ICEBERG_ASSIGN_OR_RAISE(result.metadata_location, GetJsonValueOrDefault<std::string>(
+                                                        json, kMetadataLocation, ""));
+  ICEBERG_ASSIGN_OR_RAISE(auto metadata_json,
+                          GetJsonValue<nlohmann::json>(json, kMetadata));
+  ICEBERG_ASSIGN_OR_RAISE(result.metadata, TableMetadataFromJson(metadata_json));
+  ICEBERG_ASSIGN_OR_RAISE(
+      result.config, (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(
+                         json, kConfig)));
```
wgtmac, line 167:
```suggestion
      result.config, GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(
                         json, kConfig));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+    json[kMetadataLocation] = result.metadata_location;
+  }
+  json[kMetadata] = ToJson(*result.metadata);
+  if (!result.config.empty()) {
+    json[kConfig] = result.config;
+  }
+  return json;
+}
+
+Result<LoadTableResult> LoadTableResultFromJson(const nlohmann::json& json) {
```
wgtmac, line 158: We need `StorageCredential`. At least you need to add a TODO to `types.h`, `ToJson` and `LoadTableResultFromJson`.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto metadata_json,
+                          GetJsonValue<nlohmann::json>(json, kMetadata));
+  ICEBERG_ASSIGN_OR_RAISE(result.metadata, TableMetadataFromJson(metadata_json));
+  ICEBERG_ASSIGN_OR_RAISE(
+      result.config, (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(
+                         json, kConfig)));
+  return result;
+}
+
+// ListNamespacesResponse
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
wgtmac: Please also carefully check if any missing field that is in the spec but not added on our side.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto namespaces_json,
+                          GetJsonValue<nlohmann::json>(json, kNamespaces));
+  for (const auto& ns_json : namespaces_json) {
+    ICEBERG_ASSIGN_OR_RAISE(auto ns, NamespaceFromJson(ns_json));
+    response.namespaces.push_back(std::move(ns));
+  }
+  return response;
+}
+
+// CreateNamespaceResponse
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+
+Result<CreateNamespaceResponse> CreateNamespaceResponseFromJson(
+    const nlohmann::json& json) {
+  CreateNamespaceResponse response;
+  ICEBERG_ASSIGN_OR_RAISE(response.namespace_.levels,
+                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
+  ICEBERG_ASSIGN_OR_RAISE(
+      response.properties,
+      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
+                                                                           kProperties)));
```
wgtmac:
```suggestion
      GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
                                                                           kProperties));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  return response;
+}
+
+// GetNamespaceResponse
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+}
+
+Result<GetNamespaceResponse> GetNamespaceResponseFromJson(const nlohmann::json& json) {
+  GetNamespaceResponse response;
+  ICEBERG_ASSIGN_OR_RAISE(response.namespace_.levels,
+                          GetJsonValue<std::vector<std::string>>(json, kNamespace));
+  ICEBERG_ASSIGN_OR_RAISE(
+      response.properties,
+      (GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
+                                                                           kProperties)));
```
wgtmac:
```suggestion
      GetJsonValueOrDefault<std::unordered_map<std::string, std::string>>(json,
                                                                           kProperties));
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  return response;
+}
+
+// UpdateNamespacePropertiesResponse
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -0,0 +1,357 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(response.updated,
+                          GetJsonValue<std::vector<std::string>>(json, kUpdated));
+  ICEBERG_ASSIGN_OR_RAISE(response.removed,
+                          GetJsonValue<std::vector<std::string>>(json, kRemoved));
+  ICEBERG_ASSIGN_OR_RAISE(
+      response.missing, GetJsonValueOrDefault<std::vector<std::string>>(json, kMissing));
+  return response;
+}
+
+// ListTablesResponse
```
wgtmac:
```suggestion
```

src/iceberg/test/rest_json_internal_test.cc
```diff
@@ -0,0 +1,801 @@
...
+  return a.ns.levels == b.ns.levels && a.name == b.name;
+}
+
+// Helper to compare string vectors (order-independent)
+bool StringVectorEqual(const std::vector<std::string>& a,
+                       const std::vector<std::string>& b) {
+  return std::ranges::is_permutation(a, b);
+}
+
+}  // namespace
```
wgtmac:
```suggestion
```

We don't need them. They are auto-generated by compilers.

src/iceberg/test/rest_json_internal_test.cc
```diff
@@ -0,0 +1,801 @@
...
+    auto expected_json = nlohmann::json::parse(param.expected_json_str);
+    EXPECT_EQ(json, expected_json) << "ToJson mismatch";
+
+    // FromJson and verify object equality
+    auto result = CreateNamespaceRequestFromJson(expected_json);
+    ASSERT_TRUE(result.has_value()) << result.error().message;
+    auto& parsed = result.value();
+
+    EXPECT_EQ(parsed.namespace_.levels, original.namespace_.levels);
+    EXPECT_EQ(parsed.properties, original.properties);
```
wgtmac:
```suggestion
    EXPECT_EQ(parsed, original);
```

Can we directly compare `CreateNamespaceRequest`s?

src/iceberg/test/rest_json_internal_test.cc
```diff
@@ -0,0 +1,801 @@
...
+  EXPECT_TRUE(result2.value().properties.empty());
+}
+
+TEST(CreateNamespaceRequestTest, InvalidRequests) {
+  // Incorrect type for namespace
+  std::string json_wrong_ns_type =
+      R"({"namespace":"accounting%1Ftax","properties":null})";
+  auto result1 =
+      CreateNamespaceRequestFromJson(nlohmann::json::parse(json_wrong_ns_type));
+  EXPECT_FALSE(result1.has_value());
```
wgtmac, line 182: Can you check the expected error message as well?

```
  EXPECT_THAT(result, IsError(ErrorKind::kJsonParseError));
  EXPECT_THAT(result, HasErrorMessage("..."));
```

PR #255, opened 2025-10-10
refactor: remove unnecessary nodiscard attributes

HeartLinked:


src/iceberg/type.h
```diff
@@ -83,28 +83,27 @@ class ICEBERG_EXPORT NestedType : public Type {
...
 
   /// \brief Get a view of the child fields.
-  [[nodiscard]] virtual std::span<const SchemaField> fields() const = 0;
+  virtual std::span<const SchemaField> fields() const = 0;
   using SchemaFieldConstRef = std::reference_wrapper<const SchemaField>;
   /// \brief Get a field by field ID.
   ///
   /// \note This is O(1) complexity.
-  [[nodiscard]] virtual Result<std::optional<SchemaFieldConstRef>> GetFieldById(
+  virtual Result<std::optional<SchemaFieldConstRef>> GetFieldById(
```
HuaHuaY, line 91: It may be better to reserve `[[nodiscard]]` before `Result<>`.

wgtmac: Agreed.

zhjwpku: It seems the comment suggested keeping `[[nodiscard]]` before `Result<>`, but the issue was marked as resolved without any changes.

I'd suggest removing `[[nodiscard]]` altogether. We already have many `Result<>` and `Status` methods that don't carry the [[nodiscard]] attribute, so removing it would improve consistency.

PR #257, opened 2025-10-12
feat: add metadata updates and requirements interface

gty404:
This commit introduces the metadata update and update requirements system for table metadata modifications.

### DIFF
--- a/src/iceberg/table_requirement.cc
+++ b/src/iceberg/table_requirement.cc
@@ -21,9 +21,7 @@
 
 #include "iceberg/table_metadata.h"
 
-namespace iceberg {
-
-namespace table {
+namespace iceberg::table {
 
 Status AssertDoesNotExist::Validate(const TableMetadata* base) const {
   return NotImplemented("AssertTableDoesNotExist::Validate not implemented");
@@ -59,6 +57,4 @@ Status AssertDefaultSortOrderID::Validate(const TableMetadata* base) const {
   return NotImplemented("AssertDefaultTableSortOrderID::Validate not implemented");
 }
 
-}  // namespace table
-
-}  // namespace iceberg
+}  // namespace iceberg::table
--- a/src/iceberg/table_update.cc
+++ b/src/iceberg/table_update.cc
@@ -23,9 +23,7 @@
 #include "iceberg/table_metadata.h"
 #include "iceberg/table_requirements.h"
 
-namespace iceberg {
-
-namespace table {
+namespace iceberg::table {
 
 // AssignUUID
 
@@ -198,6 +196,4 @@ Status SetLocation::GenerateRequirements(TableUpdateContext& context) const {
   return NotImplemented("SetTableLocation::GenerateRequirements not implemented");
 }
 
-}  // namespace table
-
-}  // namespace iceberg
+}  // namespace iceberg::table


src/iceberg/util/string_util.h
```diff
@@ -44,6 +44,13 @@ class ICEBERG_EXPORT StringUtils {
                    [](char c) { return std::toupper(c); });    // NOLINT
     return input;
   }
+
+  static bool EqualsIgnoreCase(const std::string& a, const std::string& b) {
+    return a.size() == b.size() &&
+           std::equal(a.begin(), a.end(), b.begin(), [](char ca, char cb) {
```
HuaHuaY: We can use `std::ranges::equal` here.

nullccxsy: There we can use `ToLower()` in `util\string_util.h`

HuaHuaY: `ToLower` will allocate heap memory. I don't think it's a good idea.

src/iceberg/table_metadata.h
```diff
@@ -29,6 +29,7 @@
 #include <vector>
 
 #include "iceberg/iceberg_export.h"
+#include "iceberg/metadata_update.h"
```
wgtmac: Use forward declaration and remove this?

src/iceberg/table_metadata.h
```diff
@@ -144,6 +145,203 @@ ICEBERG_EXPORT std::string ToString(const SnapshotLogEntry& entry);
...
+/// in a changes list. This allows the builder to maintain a complete history of all
+/// modifications made to the table metadata, which is important for tracking table
+/// evolution and for serialization purposes.
+class ICEBERG_EXPORT TableMetadataBuilder {
+ public:
+  /// \brief Create a builder for a new table
+  ///
+  /// \param format_version The format version for the table
+  /// \return A new TableMetadataBuilder instance
+  static TableMetadataBuilder BuildFromEmpty(
```
wgtmac: Should we return `std::unique_ptr<TableMetadataBuilder>`? I'm also fine with the current one.

src/iceberg/table_metadata.h
```diff
@@ -144,6 +145,203 @@ ICEBERG_EXPORT std::string ToString(const SnapshotLogEntry& entry);
...
+  ///
+  /// If no UUID is provided, a random UUID will be generated.
+  /// \return Reference to this builder for method chaining
+  TableMetadataBuilder& AssignUUID();
+
+  /// \brief Assign a specific UUID to the table
+  ///
+  /// \param uuid The UUID string to assign
+  /// \return Reference to this builder for method chaining
+  TableMetadataBuilder& AssignUUID(const std::string& uuid);
```
```suggestion
  TableMetadataBuilder& AssignUUID(std::string_view uuid);
```

src/iceberg/table_metadata.h
```diff
@@ -144,6 +145,203 @@ ICEBERG_EXPORT std::string ToString(const SnapshotLogEntry& entry);
...
+  /// \param new_format_version The new format version (must be >= current version)
+  /// \return Reference to this builder for method chaining
+  TableMetadataBuilder& UpgradeFormatVersion(int8_t new_format_version);
+
+  /// \brief Set the current schema for the table
+  ///
+  /// \param schema The schema to set as current
+  /// \param new_last_column_id The highest column ID in the schema
+  /// \return Reference to this builder for method chaining
+  TableMetadataBuilder& SetCurrentSchema(std::shared_ptr<iceberg::Schema> schema,
```
```suggestion
  TableMetadataBuilder& SetCurrentSchema(std::shared_ptr<Schema> schema,
```

wgtmac: Same for below to remove redundant namespace prefix.

src/iceberg/table_metadata.h
```diff
@@ -144,6 +145,203 @@ ICEBERG_EXPORT std::string ToString(const SnapshotLogEntry& entry);
...
+  ///
+  /// \param base The base table metadata to build from
+  /// \return A new TableMetadataBuilder instance initialized with base metadata
+  static TableMetadataBuilder BuildFrom(const std::shared_ptr<const TableMetadata>& base);
+
+  /// \brief Assign a UUID to the table
+  ///
+  /// If no UUID is provided, a random UUID will be generated.
+  /// \return Reference to this builder for method chaining
+  TableMetadataBuilder& AssignUUID();
```
wgtmac, line 192: Any specific reason to choose this subset of functions? Should we add all Java equivalents to not miss any?

src/iceberg/table_metadata.cc
```diff
@@ -196,4 +197,229 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
   return io.WriteFile(location, json_string);
 }
 
+// TableMetadataBuilder implementation
+
+struct TableMetadataBuilder::Impl {
+  // Base metadata (if building from existing metadata)
+  std::shared_ptr<const TableMetadata> base;
```
wgtmac: Is it better to make it `const TableMetadata*` so it can accept broader types?

src/iceberg/table_metadata.cc
```diff
@@ -196,4 +197,229 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+
+  // Mutable fields that will be used to build the final TableMetadata
+  int8_t format_version;
```
wgtmac: Should we just use `std::unique_ptr<TableMetadata>` and set everything to it?

src/iceberg/table_metadata.cc
```diff
@@ -196,4 +197,229 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+  std::vector<std::shared_ptr<StatisticsFile>> statistics;
+  std::vector<std::shared_ptr<PartitionStatisticsFile>> partition_statistics;
+  int64_t next_row_id;
+
+  // List of changes (MetadataUpdate objects)
+  std::vector<std::unique_ptr<MetadataUpdate>> changes;
+
+  explicit Impl(int8_t fmt_version)
+      : format_version(fmt_version),
+        last_sequence_number(TableMetadata::kInitialSequenceNumber),
```
wgtmac: Should we use in-place initialization?

src/iceberg/metadata_update.h
```diff
@@ -0,0 +1,431 @@
...
+  void GenerateRequirements(UpdateRequirementsContext& context) const override;
+
+ private:
+  int8_t format_version_;
+};
+
+/// \brief Represents adding a new schema to the table
+class ICEBERG_EXPORT AddSchema : public MetadataUpdate {
+ public:
+  explicit AddSchema(std::shared_ptr<iceberg::Schema> schema, int32_t last_column_id)
```
wgtmac:
```suggestion
  explicit AddSchema(std::shared_ptr<Schema> schema, int32_t last_column_id)
```

Remove redundant namespace prefix. Same for below.

src/iceberg/table_metadata.cc
gty404: I consider making TableMetadata read-only, once TableMetadata is constructed, it cannot be modified anymore.

src/iceberg/table_metadata.h
gty404: I have checked the corresponding interface for Java, and all the fields supported in the current C++ have corresponding functions.

src/iceberg/util/string_util.h
gty404: I have changed to use std::ranges::equal

src/iceberg/util/string_util.h
```diff
@@ -44,6 +44,12 @@ class ICEBERG_EXPORT StringUtils {
                    [](char c) { return std::toupper(c); });    // NOLINT
     return input;
   }
+
+  static bool EqualsIgnoreCase(const std::string& a, const std::string& b) {
+    return a.size() == b.size() && std::ranges::equal(a, b, [](char ca, char cb) {
```
HuaHuaY: `std::ranges::equal` will compare length so we don't need to write `a.size() == b.size()`

src/iceberg/table_metadata.h
wgtmac: I just took a quick glance and it seems that `Builder withMetadataLocation(String newMetadataLocation)` is missing?

src/iceberg/table_metadata.h
```diff
@@ -144,6 +144,202 @@ ICEBERG_EXPORT std::string ToString(const SnapshotLogEntry& entry);
...
+  ///
+  /// \param schema The schema to add
+  /// \return Reference to this builder for method chaining
+  TableMetadataBuilder& AddSchema(std::shared_ptr<Schema> schema);
+
+  /// \brief Set the default partition spec for the table
+  ///
+  /// \param spec The partition spec to set as default
+  /// \return Reference to this builder for method chaining
+  TableMetadataBuilder& SetDefaultPartitionSpec(std::shared_ptr<PartitionSpec> spec);
```
wgtmac, line 230: We cannot throw from these setters. How do you plan to deal with invalid inputs?

1. Change the setters to return `Result<std::reference_wrapper<TableMetadataBuilder>>`, but this makes return value chaining unappealing.
2. Change the setters to return `Status`.
3. Postpone any error reporting. Add a state to record most recent error and report it in the `Build` call.
4. ...

src/iceberg/table_metadata.cc
```diff
@@ -196,4 +198,222 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+  int8_t format_version;
+  std::string table_uuid;
+  std::string location;
+  int64_t last_sequence_number{TableMetadata::kInitialSequenceNumber};
+  TimePointMs last_updated_ms{std::chrono::milliseconds(0)};
+  int32_t last_column_id{0};
+  std::vector<std::shared_ptr<Schema>> schemas;
+  std::optional<int32_t> current_schema_id;
+  std::vector<std::shared_ptr<PartitionSpec>> partition_specs;
+  int32_t default_spec_id{0};
```
wgtmac: Do we have predefined constants for this and below? For example, `PartitionSpec::kInitialSpecId`?

src/iceberg/table_metadata.cc
```diff
@@ -196,4 +198,222 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+  std::vector<std::shared_ptr<PartitionStatisticsFile>> partition_statistics;
+  int64_t next_row_id{TableMetadata::kInitialRowId};
+
+  // List of changes (MetadataUpdate objects)
+  std::vector<std::unique_ptr<MetadataUpdate>> changes;
+
+  explicit Impl(int8_t fmt_version) : format_version(fmt_version) {}
+
+  explicit Impl(const TableMetadata* base_metadata)
+      : base(base_metadata),
```
wgtmac: What if `base_metadata` is nullptr?

src/iceberg/metadata_update.h
```diff
@@ -0,0 +1,431 @@
...
+#include <vector>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/snapshot.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+class TableMetadataBuilder;
+class UpdateRequirementsContext;
```
wgtmac:
```suggestion
```

They are already declared by type_fwd.h

src/iceberg/metadata_update.h
```diff
@@ -0,0 +1,431 @@
...
+  virtual void ApplyTo(TableMetadataBuilder& builder) const = 0;
+
+  /// \brief Generate update requirements for this metadata update
+  ///
+  /// This method generates the appropriate UpdateRequirement instances
+  /// that must be validated before this update can be applied. The context
+  /// provides information about the base metadata and operation mode.
+  ///
+  /// \param context The context containing base metadata and operation state
+  virtual void GenerateRequirements(UpdateRequirementsContext& context) const = 0;
```
wgtmac: Why do we need this?

src/iceberg/update_requirement.cc
```diff
@@ -0,0 +1,149 @@
...
+
+#include "iceberg/update_requirement.h"
+
+#include "iceberg/snapshot.h"
+#include "iceberg/table_metadata.h"
+#include "iceberg/util/string_util.h"
+
+namespace iceberg {
+
+Status AssertTableDoesNotExist::Validate(const TableMetadata* base) const {
```
wgtmac: I haven't reviewed all the function definitions yet but I strongly recommend to leave implementation out in this PR. We can add individual followup PRs to make it easier to review. Let's focus on the interface design here.

src/iceberg/metadata_update.h
wgtmac: What about renaming `UpdateRequirementsContext` to `MetadataUpdateContext` or `TableMetadataUpdateContext` (same as iceberg-python) ?

wgtmac: Perhaps like below:

```
virtual void AddTo(MetadataUpdateContext& context) const = 0;
```

src/iceberg/update_requirement.h
```diff
@@ -0,0 +1,220 @@
...
+///
+/// Represents a requirement that must be validated before applying
+/// metadata updates to a table. Each concrete subclass represents
+/// a specific type of requirement check.
+class ICEBERG_EXPORT UpdateRequirement {
+ public:
+  virtual ~UpdateRequirement() = default;
+
+  /// \brief Clone this update requirement
+  virtual std::unique_ptr<UpdateRequirement> Clone() const = 0;
```
wgtmac: Is it actually used? If not, remove it for now?

src/iceberg/update_requirements.h
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  /// \brief Get the base table metadata
+  const TableMetadata* base() const { return base_; }
+
+  /// \brief Check if this is a replace operation
+  bool is_replace() const { return is_replace_; }
+
+  /// \brief Build and return the list of requirements
+  std::vector<std::unique_ptr<UpdateRequirement>> Build() {
```
wgtmac: How do we handle errors? Should we return a `Result` wrapper?

wgtmac (changes requested):

src/iceberg/table_metadata.h
gty404: I prefer the third option

src/iceberg/table_metadata.h
gty404: There is a SetLocation similar to it, and I mistakenly thought it was the same method. I am adding a new interface.

src/iceberg/util/string_util.h
```diff
@@ -44,6 +44,11 @@ class ICEBERG_EXPORT StringUtils {
                    [](char c) { return std::toupper(c); });    // NOLINT
     return input;
   }
+
+  static bool EqualsIgnoreCase(const std::string& a, const std::string& b) {
```
zhjwpku: nit: We have some `operator==` definitions that use `lhs`/`rhs` as the parameters names, should we follow the same convention here?

src/iceberg/CMakeLists.txt
```diff
@@ -47,7 +47,10 @@ set(ICEBERG_SOURCES
     table.cc
     table_metadata.cc
     table_properties.cc
+    table_requirement.cc
```
zhjwpku, line 55: I guess you should also add these files to meson.build.

zhjwpku (approved):
I was initially confused about renaming `MetadataUpdate` to `TableUpdate` and `UpdateRequirement` to `TableRequirement`, but after checking the iceberg-rust and iceberg-python implementations, it now makes sense to me.

src/iceberg/CMakeLists.txt
wgtmac: +1

Perhaps the CI does not fail because the file is not referenced by others.

wgtmac: @gty404 Please add new header and source files to the corresponding `meson.build` file.

src/iceberg/table_metadata.cc
```diff
@@ -196,4 +197,185 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+}
+
+std::unique_ptr<TableMetadataBuilder> TableMetadataBuilder::BuildFrom(
+    const TableMetadata* base) {
+  return std::unique_ptr<TableMetadataBuilder>(new TableMetadataBuilder(base));  // NOLINT
+}
+
+TableMetadataBuilder& TableMetadataBuilder::SetMetadataLocation(
+    std::string_view metadata_location) {
+  return *this;
```
wgtmac: We should throw for unimplemented functions.

src/iceberg/table_requirements.h
```diff
@@ -0,0 +1,118 @@
...
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Context for generating update requirements
+///
+/// This context is passed to each MetadataUpdate's GenerateRequirements method
+/// and maintains state about what requirements have already been added to avoid
+/// duplicates.
+class ICEBERG_EXPORT MetadataUpdateContext {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT TableUpdateContext {
```

Also fix the comments after renaming

src/iceberg/table_update.h
```diff
@@ -0,0 +1,356 @@
...
+  /// that must be validated before this update can be applied. The context
+  /// provides information about the base metadata and operation mode.
+  ///
+  /// \param context The context containing base metadata and operation state
+  /// \return Status indicating success or failure with error details
+  virtual Status GenerateRequirements(MetadataUpdateContext& context) const = 0;
+};
+
+/// \brief Represents an assignment of a UUID to the table
+class ICEBERG_EXPORT AssignUUID : public TableUpdate {
```
wgtmac, line 68: Will it cause name conflicts once we add `class AssignUUID : public ViewUpdate`? Or do we want to modify it as `class AssignUUID : public TableUpdate, public ViewUpdate` ? If latter, we may need to rename the file to `metadata_update.h/cc`?

src/iceberg/table_requirements.h
```diff
@@ -0,0 +1,118 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/update_requirements.h
```
wgtmac: The comment looks inconsistent now.

src/iceberg/util/string_util.h
wgtmac: +1

wgtmac (commented):
Generally looks good. I left some minor comments.

src/iceberg/table_update.h
gty404: The View is expected to have a separate ViewMetadata, and the ViewUpdate interface will differ from TableUpdate. Therefore, I plan to add the 'table' keyword to these updates/requirements to distinguish them.

src/iceberg/table_update.h
```diff
@@ -0,0 +1,358 @@
...
+
+  Status GenerateRequirements(TableUpdateContext& context) const override;
+
+ private:
+  std::shared_ptr<Schema> schema_;
+  int32_t last_column_id_;
+};
+
+/// \brief Represents setting the current schema
+class ICEBERG_EXPORT SetCurrentTableSchema : public TableUpdate {
```
wgtmac: These names look a little bit weird. What about this?

```cpp
class ICEBERG_EXPORT TableUpdate {
  // ...
};

namespace table {

class ICEBERG_EXPORT AssignUUID : public TableUpdate {
  // ...
};

class ICEBERG_EXPORT UpgradeFormatVersion : public TableUpdate {
  // ...
};

}  // namespace table
```

wgtmac (approved):
LGTM. I've just left a comment w.r.t. class names.

src/iceberg/table_update.h
wgtmac: Same for `TableRequirement`.

gty404: Alright, I'll add a namespace.

wgtmac (approved):

PR #258, opened 2025-10-15
fix: Handle null values in GetJsonValueXXX

HeartLinked:
The current JSON parsing logic in GetJsonValueOrDefault does not correctly handle cases where an optional field is present with an explicit null value. Now, if a key is present but its value is null, the function correctly returns the default value, treating it the same as a missing key.

### DIFF
--- a/src/iceberg/util/json_util_internal.h
+++ b/src/iceberg/util/json_util_internal.h
@@ -57,7 +57,7 @@ Result<T> GetJsonValueImpl(const nlohmann::json& json, std::string_view key) {
 template <typename T>
 Result<std::optional<T>> GetJsonValueOptional(const nlohmann::json& json,
                                               std::string_view key) {
-  if (!json.contains(key)) {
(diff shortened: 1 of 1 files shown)


zhjwpku (approved):
+1

src/iceberg/util/json_util_internal.h
```diff
@@ -75,7 +75,7 @@ Result<T> GetJsonValue(const nlohmann::json& json, std::string_view key) {
 template <typename T>
 Result<T> GetJsonValueOrDefault(const nlohmann::json& json, std::string_view key,
                                 T default_value = T{}) {
-  if (!json.contains(key)) {
+  if (!json.contains(key) || json.at(key).is_null()) {
```
wgtmac, line 78: Why not fixing all `json.contains(key)` at once? Similar functions are `GetJsonValueOptional` and `GetJsonValue`.

PR #261, opened 2025-10-17
feat: write manifest avro metadata

dongxiao1198:
1 fix TODOs in manifest adaptor
2 pass metadata to avro writer

### DIFF
--- a/src/iceberg/v1_metadata.cc
+++ b/src/iceberg/v1_metadata.cc
@@ -19,8 +19,6 @@
 
 #include "iceberg/v1_metadata.h"
 
-#include <nlohmann/json.hpp>
-
 #include "iceberg/json_internal.h"
 #include "iceberg/manifest_entry.h"
 #include "iceberg/manifest_list.h"
@@ -51,9 +49,9 @@ Status ManifestEntryAdapterV1::Init() {
       DataFile::kSortOrderId.field_id(),
   };
   ICEBERG_RETURN_UNEXPECTED(InitSchema(kManifestEntryFieldIds));
-  metadata_["schema"] = ToJson(*manifest_schema_).dump();
+  ICEBERG_ASSIGN_OR_RAISE(metadata_["schema"], ToJsonString(*manifest_schema_))
   if (partition_spec_ != nullptr) {
-    metadata_["partition-spec"] = ToJson(*partition_spec_).dump();
+    ICEBERG_ASSIGN_OR_RAISE(metadata_["partition-spec"], ToJsonString(*partition_spec_));
     metadata_["partition-spec-id"] = std::to_string(partition_spec_->spec_id());
   }
   metadata_["format-version"] = "1";
--- a/src/iceberg/v2_metadata.cc
+++ b/src/iceberg/v2_metadata.cc
@@ -19,8 +19,6 @@
 
 #include "iceberg/v2_metadata.h"
 
-#include <nlohmann/json.hpp>
-
 #include "iceberg/json_internal.h"
 #include "iceberg/manifest_entry.h"
 #include "iceberg/manifest_list.h"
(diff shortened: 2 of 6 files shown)


src/iceberg/v1_metadata.cc
```diff
@@ -19,6 +19,9 @@
 
 #include "iceberg/v1_metadata.h"
 
+#include <nlohmann/json.hpp>
+
```
wgtmac:
```suggestion
```

I don't think we need this.

src/iceberg/v1_metadata.cc
```diff
@@ -47,15 +50,14 @@ Status ManifestEntryAdapterV1::Init() {
...
       DataFile::kSortOrderId.field_id(),
   };
-  // TODO(xiao.dong) schema to json
-  metadata_["schema"] = "{}";
-  // TODO(xiao.dong) partition spec to json
-  metadata_["partition-spec"] = "{}";
+  ICEBERG_RETURN_UNEXPECTED(InitSchema(kManifestEntryFieldIds));
+  metadata_["schema"] = ToJson(*manifest_schema_).dump();
   if (partition_spec_ != nullptr) {
+    metadata_["partition-spec"] = ToJson(*partition_spec_).dump();
```
wgtmac:
```suggestion
    metadata_["partition-spec"] = ToJsonString(ToJson(*partition_spec_));
```

src/iceberg/v1_metadata.cc
```diff
@@ -47,15 +50,14 @@ Status ManifestEntryAdapterV1::Init() {
       DataFile::kSplitOffsets.field_id(),
       DataFile::kSortOrderId.field_id(),
   };
-  // TODO(xiao.dong) schema to json
-  metadata_["schema"] = "{}";
-  // TODO(xiao.dong) partition spec to json
-  metadata_["partition-spec"] = "{}";
+  ICEBERG_RETURN_UNEXPECTED(InitSchema(kManifestEntryFieldIds));
+  metadata_["schema"] = ToJson(*manifest_schema_).dump();
```
wgtmac:
```suggestion
  metadata_["schema"] = ToJsonString(ToJson(*manifest_schema_));
```

wgtmac: Or we can add the following to `json_internal.h`

```cpp
template <typename T>
ICEBERG_EXPORT Result<std::string> ToJsonString(const T& value)
  requires requires { ToJson(value); }
{
  return ToJsonString(ToJson(value));
}
```

src/iceberg/v2_metadata.cc
```diff
@@ -19,9 +19,13 @@
 
 #include "iceberg/v2_metadata.h"
 
+#include <nlohmann/json.hpp>
```
wgtmac: ditto

src/iceberg/v3_metadata.cc
```diff
@@ -19,6 +19,9 @@
 
 #include "iceberg/v3_metadata.h"
 
+#include <nlohmann/json.hpp>
```
wgtmac: ditto

wgtmac (changes requested):

src/iceberg/v1_metadata.cc
dongxiao1198: json_internal only include #include <nlohmann/json_fwd.hpp>, so this cc must include this header since `ToJson` returns a  nlohmann::json

wgtmac: OK, let's include it in the cc file.

src/iceberg/v1_metadata.cc
dongxiao1198: I really think we should make toJson like this. This way we can ignore the nlohmann:: outside.

dongxiao1198: I just add a new function: 
ICEBERG_EXPORT Result<std::string> ToJsonString(const Schema& schema);

PR #266, opened 2025-10-20
refactor: package once_flag and LazyInitWithCallOnce

HuaHuaY:


src/iceberg/util/lazy.h
```diff
@@ -0,0 +1,66 @@
...
+        this->value = std::move(result.value());
+      }
+    });
+    ICEBERG_RETURN_UNEXPECTED(result);
+    return &value;
+  }
+
+ private:
+  mutable T value;
+  mutable std::once_flag flag;
```
wgtmac:
```suggestion
  Result<T*> Get(Args&&... args) const {
    Result<T> result;
    std::call_once(flag_, [&result, this, &args...]() {
      result = InitFunc(std::forward<Args>(args)...);
      if (result) {
        this->value_ = std::move(result.value());
      }
    });
    ICEBERG_RETURN_UNEXPECTED(result);
    return &value_;
  }

 private:
  mutable T value_;
  mutable std::once_flag flag_;
```

Just follow the coding style across the repo.

PR #267, opened 2025-10-20
test: construct temporal values from structural inputs

zhjwpku:


src/iceberg/util/temporal_util.cc
```diff
@@ -20,6 +20,7 @@
 #include "iceberg/util/temporal_util.h"
 
 #include <chrono>
+#include <cstdint>
```
gty404, line 23: <cstdint> has already been added by the header

src/iceberg/util/temporal_util.h
```diff
@@ -19,12 +19,27 @@
...
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
 #include "iceberg/type_fwd.h"
 
 namespace iceberg {
 
+struct ICEBERG_EXPORT TemporalParts {
+  int32_t year{0};
+  uint16_t month{0};
+  uint16_t day{0};
```
gty404: consider uint8_t or std::chrono::month/std::chrono::day as the type of month/day ？

src/iceberg/util/temporal_util.h
```diff
@@ -38,6 +53,24 @@ class ICEBERG_EXPORT TemporalUtils {
 
   /// \brief Extract a timestamp hour, as hours from 1970-01-01 00:00:00
   static Result<Literal> ExtractHour(const Literal& literal);
+
+  /// \brief Construct a Calendar date without timezone or time
+  static int32_t CreateDate(const TemporalParts& parts);
```
gty404: Should these functions return a `Result<>`, returning an error message when the parameters in TemporalParts are invalid?

src/iceberg/util/temporal_util.h
```diff
@@ -19,12 +19,27 @@
...
+  int32_t year{0};
+  uint16_t month{0};
+  uint16_t day{0};
+  int32_t hour{0};
+  int32_t minute{0};
+  int32_t second{0};
+  int32_t microsecond{0};
+  int32_t nanosecond{0};
+  // e.g. -480 for PST (UTC-8:00), +480 for Asia/Shanghai (UTC+8:00)
+  int32_t tz_offset_minutes;
```
gty404: Initialize with a default value or use optional to avoid UB ？

src/iceberg/util/temporal_util.h
zhjwpku: I avoided using std::chrono::month and std::chrono::day to not including <chrono>, but I might be overthinking it. Will try, thanks.

src/iceberg/util/temporal_util.h
```diff
@@ -19,12 +19,27 @@
...
 
+#include <cstdint>
+
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
 #include "iceberg/type_fwd.h"
 
 namespace iceberg {
 
+struct ICEBERG_EXPORT TemporalParts {
```
wgtmac: I was thinking that only adding these structures to the test directory to make it easy for us to create `Date`, `Time` and `Timestamp` values in the test cases. Temporal values are notorious to deal with, especially when it comes to timezone and daylight saving rules. We don't want to expose such APIs externally and maintain it hereafter.

wgtmac: Perhaps we should not reuse a single structure to create different temporal types, which leads to unused fields.

src/iceberg/util/temporal_util.h
zhjwpku: I just moved these constructors into the test directory, not user-facing API any more, so I think returning the value directly should be fine.

src/iceberg/util/temporal_util.cc
zhjwpku: This hasn't been directly added by the temporal_util.h, there are uses of `int32_t` in this file, so I think include the header is fine?

wgtmac (approved):

PR #268, opened 2025-10-20
feat: implement TableMetadataBuilder with AssignUUID

gty404:
This commit implements the core TableMetadataBuilder pattern following the established design from the previous session.

Pattern established:
1. Builder validates input and collects errors (no exceptions)
2. Changes are tracked for serialization
3. Requirements generated for optimistic concurrency
4. Build() validates metadata consistency before returning

### DIFF
--- a/src/iceberg/table_metadata.cc
+++ b/src/iceberg/table_metadata.cc
@@ -41,6 +41,10 @@
 
 namespace iceberg {
 
+namespace {
+const TimePointMs kInvalidLastUpdatedMs = TimePointMs::min();
+}
+
 std::string ToString(const SnapshotLogEntry& entry) {
   return std::format("SnapshotLogEntry[timestampMillis={},snapshotId={}]",
                      entry.timestamp_ms, entry.snapshot_id);
@@ -224,7 +228,7 @@ struct TableMetadataBuilder::Impl {
   explicit Impl(int8_t format_version) : base(nullptr), metadata{} {
     metadata.format_version = format_version;
     metadata.last_sequence_number = TableMetadata::kInitialSequenceNumber;
-    metadata.last_updated_ms = TableMetadata::kInvalidLastUpdatedMs;
+    metadata.last_updated_ms = kInvalidLastUpdatedMs;
     metadata.last_column_id = Schema::kInvalidColumnId;
     metadata.default_spec_id = PartitionSpec::kInitialSpecId;
     metadata.last_partition_id = PartitionSpec::kInvalidPartitionFieldId;
@@ -448,7 +452,7 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataBuilder::Build() {
   // 2. Validate metadata consistency through TableMetadata#Validate
 
   // 3. Update last_updated_ms if there are changes
-  if (impl_->metadata.last_updated_ms == TableMetadata::kInvalidLastUpdatedMs) {
+  if (impl_->metadata.last_updated_ms == kInvalidLastUpdatedMs) {
     impl_->metadata.last_updated_ms =
         TimePointMs{std::chrono::duration_cast<std::chrono::milliseconds>(
             std::chrono::system_clock::now().time_since_epoch())};
--- a/src/iceberg/table_metadata.h
+++ b/src/iceberg/table_metadata.h
@@ -73,7 +73,6 @@ struct ICEBERG_EXPORT TableMetadata {
   static constexpr int64_t kInitialSequenceNumber = 0;
   static constexpr int64_t kInvalidSequenceNumber = -1;
   static constexpr int64_t kInitialRowId = 0;
-  static constexpr TimePointMs kInvalidLastUpdatedMs = TimePointMs::min();
 
   /// An integer version number for the format
   int8_t format_version;


src/iceberg/table_metadata.h
```diff
@@ -73,6 +73,9 @@ struct ICEBERG_EXPORT TableMetadata {
   static constexpr int64_t kInitialSequenceNumber = 0;
   static constexpr int64_t kInvalidSequenceNumber = -1;
   static constexpr int64_t kInitialRowId = 0;
+  static constexpr int32_t kInitialSpecId = 0;
```
wgtmac: We already have `PartitionSpec::kInitialSpecId`. Should we reuse it?

src/iceberg/table_metadata.h
```diff
@@ -73,6 +73,9 @@ struct ICEBERG_EXPORT TableMetadata {
...
+  static constexpr int32_t kInitialSortOrderId = 1;
```
wgtmac: ditto, we have `SortOrder::kInitialSortOrderId`.

src/iceberg/table_metadata.h
```diff
@@ -73,6 +73,9 @@ struct ICEBERG_EXPORT TableMetadata {
...
+  static constexpr int64_t kInvalidSnapshotId = -1;
```
wgtmac: We have `Snapshot::kInvalidSnapshotId` too.

wgtmac: An alternative is to move these commonly used constants to `iceberg/constant.h`

src/iceberg/table_metadata.cc
```diff
@@ -201,13 +203,46 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+  const TableMetadata* base;
+
+  // Working metadata copy
+  TableMetadata metadata;
+
+  // Change tracking
+  std::vector<std::unique_ptr<TableUpdate>> changes;
+
+  // Error collection (since methods return *this and cannot throw)
+  std::vector<Status> errors;
```
```suggestion
  std::vector<Error> errors;
```

src/iceberg/table_metadata.cc
```diff
@@ -238,12 +273,35 @@ TableMetadataBuilder& TableMetadataBuilder::SetPreviousMetadataLocation(
...
+  std::string uuid_str(uuid);
+
+  // Validation: UUID cannot be null or empty
+  if (uuid_str.empty()) {
+    impl_->errors.emplace_back(InvalidArgument("Cannot assign null or empty UUID"));
+    return *this;
+  }
+
+  // Check if UUID is already set to the same value (no-op)
+  if (StringUtils::EqualsIgnoreCase(impl_->metadata.table_uuid, uuid_str)) {
```
wgtmac, line 298: nit: is it possible to refactor it to use std::string_view as input?

src/iceberg/table_metadata.cc
```diff
@@ -238,12 +273,35 @@ TableMetadataBuilder& TableMetadataBuilder::SetPreviousMetadataLocation(
...
+    return *this;
+  }
+
+  // Update the metadata
+  impl_->metadata.table_uuid = uuid_str;
+
+  // Record the change
+  impl_->changes.push_back(std::make_unique<table::AssignUUID>(uuid_str));
```
```suggestion
  impl_->changes.push_back(std::make_unique<table::AssignUUID>(std::move(uuid_str)));
```

src/iceberg/table_requirement.cc
```diff
@@ -20,15 +20,28 @@
 #include "iceberg/table_requirement.h"
 
 #include "iceberg/table_metadata.h"
+#include "util/string_util.h"
```
wgtmac:
```suggestion
#include "iceberg/util/string_util.h"
```

src/iceberg/test/CMakeLists.txt
```diff
@@ -84,6 +84,8 @@ add_iceberg_test(table_test
                  table_test.cc
                  schema_json_test.cc)
 
+add_iceberg_test(table_metadata_builder_test SOURCES table_metadata_builder_test.cc)
```
wgtmac: How about adding it to the `table_test` above?

src/iceberg/table_metadata.cc
```diff
@@ -238,12 +273,35 @@ TableMetadataBuilder& TableMetadataBuilder::SetPreviousMetadataLocation(
...
 }
 
 TableMetadataBuilder& TableMetadataBuilder::AssignUUID(std::string_view uuid) {
-  throw IcebergError(std::format("{} not implemented", __FUNCTION__));
-  ;
+  std::string uuid_str(uuid);
+
+  // Validation: UUID cannot be null or empty
+  if (uuid_str.empty()) {
+    impl_->errors.emplace_back(InvalidArgument("Cannot assign null or empty UUID"));
```
```suggestion
    impl_->errors.emplace_back(InvalidArgument("Cannot assign empty UUID"));
```

src/iceberg/table_metadata.cc
```diff
@@ -201,13 +203,46 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+
+  // Error collection (since methods return *this and cannot throw)
+  std::vector<Status> errors;
+
+  // Metadata location tracking
+  std::optional<std::string> metadata_location;
+  std::optional<std::string> previous_metadata_location;
+
+  // Constructor for new table
+  explicit Impl(int8_t format_version) : base(nullptr), metadata{} {
```
wgtmac: An unrelated comment: we might need to add function `Status TableMetadata::Validate() const` to make sure all fields are populated correctly. We can add this in a separate PR.

src/iceberg/table_metadata.cc
```diff
@@ -201,13 +203,46 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+    metadata.format_version = format_version;
+    metadata.last_sequence_number = TableMetadata::kInitialSequenceNumber;
+    metadata.last_updated_ms = TimePointMs{std::chrono::milliseconds(0)};
```
wgtmac: The Java version only initializes these fields:

```java
    private Builder(int formatVersion) {
      this.base = null;
      this.formatVersion = formatVersion;
      this.lastSequenceNumber = INITIAL_SEQUENCE_NUMBER;
      this.uuid = UUID.randomUUID().toString();
      this.currentSnapshotId = -1;
      this.nextRowId = INITIAL_ROW_ID;
    }
```

There are some inconsistencies from the Java impl. For instance `last_updated_ms`, the Java impl declares it as `long` in the TableMetadata but `Long` in the TableMetadataBuilder to make it optional. In the Java ctor, `lastUpdatedMillis` is initialized to null.

If we want to follow the same logic, we may not directly use `TableMetadata metadata` directly since we don't know whether some fields have been updated or not.

wgtmac: We may either follow the same pattern of Java to explicitly declare fields in the builder, or initialize fields to invalid values?

src/iceberg/table_metadata.cc
```diff
@@ -378,11 +436,50 @@ TableMetadataBuilder& TableMetadataBuilder::RemoveEncryptionKey(std::string_view
...
+    // Reset to initial state for new table
+    *impl_ = Impl(impl_->metadata.format_version);
+  }
+
+  return *this;
 }
 
 Result<std::unique_ptr<TableMetadata>> TableMetadataBuilder::Build() {
-  return NotImplemented("TableMetadataBuilder::Build not implemented");
+  // 1. Check for accumulated errors
```
wgtmac, line 443: An unrelated comment: we might need to add function `Status TableMetadata::Validate() const` to make sure all fields are populated correctly. We can add this in a separate PR.

src/iceberg/table_metadata.cc
```diff
@@ -378,11 +436,50 @@ TableMetadataBuilder& TableMetadataBuilder::RemoveEncryptionKey(std::string_view
 }
 
 TableMetadataBuilder& TableMetadataBuilder::DiscardChanges() {
-  throw IcebergError(std::format("{} not implemented", __FUNCTION__));
+  // Clear all changes and errors
+  impl_->changes.clear();
+  impl_->errors.clear();
+
+  // Reset metadata to base state
```
wgtmac: I think this is inconsistent with the Java impl where `changes` list is ignored but the updated fields are preserved to create the `TableMetadata`.

src/iceberg/table_metadata.cc
```diff
@@ -378,11 +436,50 @@ TableMetadataBuilder& TableMetadataBuilder::RemoveEncryptionKey(std::string_view
...
+    return InvalidArgument("UUID is required for format version {}",
+                           impl_->metadata.format_version);
+  }
+
+  // 3. Update last_updated_ms if there are changes
+  if (!impl_->changes.empty() && impl_->base != nullptr) {
+    impl_->metadata.last_updated_ms =
+        TimePointMs{std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now().time_since_epoch())};
+  }
```
wgtmac: Why do we need this? I didn't see these on the Java side.

src/iceberg/table_requirements.cc
```diff
@@ -26,11 +26,11 @@
 namespace iceberg {
 
 void TableUpdateContext::AddRequirement(std::unique_ptr<TableRequirement> requirement) {
-  throw IcebergError("TableUpdateContext::AddRequirement not implemented");
+  requirements_.emplace_back(std::move(requirement));
```
wgtmac, line 29: Should we add `AssertUUID` to ForReplaceTable/ForUpdateTable below?

src/iceberg/table_update.cc
```diff
@@ -21,18 +21,32 @@
...
 namespace iceberg::table {
 
 // AssignUUID
 
 void AssignUUID::ApplyTo(TableMetadataBuilder& builder) const {
-  throw IcebergError(std::format("{} not implemented", __FUNCTION__));
+  builder.AssignUUID(uuid_);
 }
 
 Status AssignUUID::GenerateRequirements(TableUpdateContext& context) const {
```
wgtmac, line 35: It seems that this function is not called any where?

src/iceberg/test/table_metadata_builder_test.cc
```diff
@@ -0,0 +1,311 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <memory>
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include "gmock/gmock-matchers.h"
```
wgtmac:
```suggestion
#include <gmock/gmock-matchers.h>
```

src/iceberg/test/table_metadata_builder_test.cc
```diff
@@ -0,0 +1,311 @@
...
+#include "iceberg/table_update.h"
+#include "iceberg/test/matchers.h"
+
+namespace iceberg {
+
+// Helper functions to reduce test boilerplate
+namespace {
+
+// Build and assert success, returning the metadata
+std::unique_ptr<TableMetadata> BuildAndExpectSuccess(TableMetadataBuilder& builder) {
```
wgtmac: Why not directly use `ICEBERG_ASSIGN_OR_RAISE` macro or write a similar one?

src/iceberg/test/table_metadata_builder_test.cc
```diff
@@ -0,0 +1,311 @@
...
+  }
+  return std::move(result.value());
+}
+
+// Build and expect failure with message containing substring
+void BuildAndExpectFailure(TableMetadataBuilder& builder,
+                           const std::string& expected_message_substr) {
+  auto result = builder.Build();
+  ASSERT_FALSE(result.has_value()) << "Build should have failed";
+  EXPECT_TRUE(result.error().message.find(expected_message_substr) != std::string::npos)
```
wgtmac: Use `EXPECT_THAT` with `HasErrorMessage`?

src/iceberg/test/table_metadata_builder_test.cc
```diff
@@ -0,0 +1,311 @@
...
+      << "Expected error message to contain: " << expected_message_substr
+      << "\nActual error: " << result.error().message;
+}
+
+// Generate requirements and return them
+std::vector<std::unique_ptr<TableRequirement>> GenerateRequirements(
+    const TableUpdate& update, const TableMetadata* base) {
+  TableUpdateContext context(base, false);
```
wgtmac:
```suggestion
  TableUpdateContext context(base, /*is_replace=*/false);
```

src/iceberg/table_metadata.cc
gty404: Let's follow the Java pattern and directly define the metadata fields in the builder, rather than the metadata object, although it doesn't look as concise.

src/iceberg/table_metadata.cc
gty404: Maybe I misunderstood the semantics of discard, and I think the implementation of java is problematic. I'll read the context of this piece carefully again to get more understanding

src/iceberg/table_metadata.cc
gty404: Okay, so the logic for Build might be to construct the final TableMetadata and then call Validate to verify its validity.

src/iceberg/table_metadata.cc
gty404: The check exists in the constructor of TableMetadata.

src/iceberg/table_requirements.cc
gty404: Yes, when updating an existing table, it is necessary to verify if the uuid has changed. This assertion is actively added by ForReplaceTable/ForUpdateTable, not generated from TableUpdate.

src/iceberg/table_update.cc
gty404: It is not currently in use. The expected usage scenario is to use TableMetadataBuilder to construct the final TableMetadata object and obtain a list of TableUpdate, then call GenerateRequirements to generate a list of TableRequirement. Finally, use the list of TableUpdate and TableRequirement to call the UpdateTable interface of the catalog.

src/iceberg/table_metadata.cc
```diff
@@ -201,13 +203,46 @@ Status TableMetadataUtil::Write(FileIO& io, const std::string& location,
...
+
+  // Metadata location tracking
+  std::optional<std::string> metadata_location;
+  std::optional<std::string> previous_metadata_location;
+
+  // Constructor for new table
+  explicit Impl(int8_t format_version) : base(nullptr), metadata{} {
+    metadata.format_version = format_version;
+    metadata.last_sequence_number = TableMetadata::kInitialSequenceNumber;
+    metadata.last_updated_ms = TimePointMs::min();
```
wgtmac: nit: define a `kInvalidLastUpdatedMs` at the top of this file

PR #270, opened 2025-10-21
feat: add literal cast to date

zhjwpku:


src/iceberg/test/literal_test.cc
```diff
@@ -689,6 +690,31 @@ INSTANTIATE_TEST_SUITE_P(
                              .source_literal = Literal::Long(42L),
                              .target_type = timestamp_tz(),
                              .expected_literal = Literal::TimestampTz(42L)},
+        CastLiteralTestParam{
+            .test_name = "TimestampToDate",
+            .source_literal =
+                Literal::Timestamp(TemporalTestHelper::CreateTimestamp({.year = 2021,
```
wgtmac, line 696: nit: add a timestamp before UNIX epoch

zhjwpku: added, thanks

PR #275, opened 2025-10-23
feat: fix literal comparison operator

zhjwpku:
Enable comparison for Fixed type with different length and Decimal type with different scale. 
Add some test cases to verify that.

### DIFF
--- a/src/iceberg/expression/literal.h
+++ b/src/iceberg/expression/literal.h
@@ -130,12 +130,12 @@ class ICEBERG_EXPORT Literal : public util::Formattable {
 
   bool operator==(const Literal& other) const;
 
-  /// \brief Compare two PrimitiveLiterals. Both literals must have the same type.
-  /// \param other The other PrimitiveLiteral to compare with.
+  /// \brief Compare two literals of the same primitive type.
+  /// \param other The other literal to compare with.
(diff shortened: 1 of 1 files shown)


src/iceberg/expression/literal.h
```diff
@@ -129,8 +130,12 @@ class ICEBERG_EXPORT Literal : public util::Formattable {
...
   bool operator==(const Literal& other) const;
 
-  /// \brief Compare two PrimitiveLiterals. Both literals must have the same type
-  /// and should not be AboveMax or BelowMin.
+  /// \brief Compare two PrimitiveLiterals. Both literals must have the same type.
+  /// \param other The other PrimitiveLiteral to compare with.
+  /// \return The comparison result as std::partial_ordering. If either side is AboveMax,
+  /// BelowMin or null, the result is unordered.
+  /// Note: This comparison can not be used for sorting literal vectors containing
+  /// AboveMax, BelowMin or null values.
```
```suggestion
  /// \brief Compare two literals of the same primitive type.
  /// \param other The other literal to compare with.
  /// \return The comparison result as std::partial_ordering. If either side is AboveMax,
  /// BelowMin or Null, the result is unordered.
  /// Note: This comparison cannot be used for sorting literals if any literal is
  /// AboveMax, BelowMin or Null.
```

gty404 (approved):

PR #280, opened 2025-10-28
feat: implement all functions of bound predicates

wgtmac:
- Implemented `Negate`, `ToString` and `Test` functions for bound predicate subclasses.
- Added hash support to `Literal`.
- Refactored `BoundSetPredicate` to use unordered set for literals.
- Refactored predicate unit test to be better organized.

### DIFF
--- a/src/iceberg/expression/predicate.cc
+++ b/src/iceberg/expression/predicate.cc
@@ -20,9 +20,9 @@
 #include "iceberg/expression/predicate.h"
 
 #include <algorithm>
+#include <cmath>
 #include <format>
 
-#include "iceberg/exception.h"
 #include "iceberg/expression/expressions.h"
 #include "iceberg/expression/literal.h"
 #include "iceberg/result.h"
@@ -141,6 +141,26 @@ namespace {
 
 bool IsFloatingType(TypeId type) {
   return type == TypeId::kFloat || type == TypeId::kDouble;
+}
+
+bool IsNan(const Literal& literal) {
+  const auto& value = literal.value();
+  if (std::holds_alternative<float>(value)) {
+    return std::isnan(std::get<float>(value));
+  } else if (std::holds_alternative<double>(value)) {
+    return std::isnan(std::get<double>(value));
+  }
+  return false;
+}
+
+bool StartsWith(const Literal& lhs, const Literal& rhs) {
+  const auto& lhs_value = lhs.value();
+  const auto& rhs_value = rhs.value();
+  if (std::holds_alternative<std::string>(lhs_value) &&
+      std::holds_alternative<std::string>(rhs_value)) {
+    return std::get<std::string>(lhs_value).starts_with(std::get<std::string>(rhs_value));
+  }
+  return false;
 }
 
 }  // namespace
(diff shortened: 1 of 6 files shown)


zhjwpku (approved):
LGTM

src/iceberg/expression/predicate.cc
```diff
@@ -287,10 +307,10 @@ BoundPredicate::BoundPredicate(Expression::Operation op, std::shared_ptr<BoundTe
 
 BoundPredicate::~BoundPredicate() = default;
 
-Result<Literal::Value> BoundPredicate::Evaluate(const StructLike& data) const {
+Result<Literal> BoundPredicate::Evaluate(const StructLike& data) const {
```
shangxinli, line 310: Not sure if we have a tool to capture the API signature changes in release notes.

src/iceberg/expression/literal.cc
```diff
@@ -554,4 +554,51 @@ Result<Literal> LiteralCaster::CastTo(const Literal& literal,
...
+          return std::numeric_limits<std::size_t>::max();
+        } else if constexpr (std::is_same_v<T, bool> || std::is_same_v<T, int32_t> ||
+                             std::is_same_v<T, int64_t> || std::is_same_v<T, float> ||
+                             std::is_same_v<T, double> ||
+                             std::is_same_v<T, std::string>) {
+          return std::hash<T>{}(v);
+        } else if constexpr (std::is_same_v<T, std::vector<uint8_t>>) {
+          std::size_t hash = 0;
+          for (size_t i = 0; i < v.size(); ++i) {
+            hash ^= std::hash<uint8_t>{}(v[i]) + kHashPrime + (hash << 6) + (hash >> 2);
```
shangxinli, line 579: Hashing identical bytes at different positions can produce same hash (e.g., [1,2] and [2,1] might collide).
Can we add position i into the hash value to decrease the possibility of collide?

src/iceberg/expression/predicate.cc
```diff
@@ -331,12 +376,91 @@ BoundLiteralPredicate::BoundLiteralPredicate(Expression::Operation op,
 
 BoundLiteralPredicate::~BoundLiteralPredicate() = default;
 
-Result<bool> BoundLiteralPredicate::Test(const Literal::Value& value) const {
-  return NotImplemented("BoundLiteralPredicate::Test not implemented");
+Result<bool> BoundLiteralPredicate::Test(const Literal& value) const {
```
shangxinli, line 379: Can we add null checking? 
    if (value.IsNull() || literal_.IsNull()) {
      return false;
    }

src/iceberg/expression/literal.cc
zhjwpku: > Hashing identical bytes at different positions can produce same hash (e.g., [1,2] and [2,1] might collide). 

I don't think this statement is correct, the left and right shifts should ensure a difference.
I ran a quick demo on Godbolt [1], and as you can see, the hashes of [1,2] and [2,1] are different.

[1] https://godbolt.org/z/4ss9E749q

src/iceberg/expression/predicate.cc
zhjwpku: We don't have that yet, which seems fine at this early stage of iceberg-cpp, but we should follow [semantic versioning](https://semver.org/) once the project becomes stable, i.e. bump up major version for incompatible API changes.

src/iceberg/expression/literal.cc
shangxinli: I traced through the logic again and realized you are right. The hash is position-sensitive because each iteration depends on the accumulated hash state from previous iterations.

src/iceberg/expression/predicate.cc
wgtmac: Comparing nulls returns `std::partial_ordering::unordered` so it is automatically `false`: https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/expression/literal.cc#L358

wgtmac:
Let me merge this. Thanks @zhjwpku @HuaHuaY @yingcai-cy @Fokko @shangxinli for the review!

PR #281, opened 2025-10-28
feat: add rename table interface to catalog

lishuxu:


src/iceberg/transaction.h
```diff
@@ -42,12 +44,23 @@ class ICEBERG_EXPORT Transaction {
...
+  virtual Status UpdateTable(
+      const std::vector<std::unique_ptr<TableRequirement>>& requirements,
+      std::vector<std::unique_ptr<TableUpdate>> updates) = 0;
+
   /// \brief Apply the pending changes from all actions and commit
   ///
-  /// May throw ValidationException if any update cannot be applied to the current table
-  /// metadata. May throw CommitFailedException if the updates cannot be committed due to
-  /// conflicts.
-  virtual void CommitTransaction() = 0;
```
shangxinli, line 50: Do we want to keep backward compatibility with deprecation e.g. 
  [[deprecated("Use CommitTransaction() that returns Status")]]

src/iceberg/catalog.h
```diff
@@ -223,6 +247,16 @@ class ICEBERG_EXPORT Catalog {
...
     virtual std::unique_ptr<Transaction> StageCreate() = 0;
+
+    /// \brief Starts a transaction to replace the table
+    ///
+    /// \return the Transaction to replace the table
+    virtual std::unique_ptr<Transaction> StageReplace() = 0;
+
+    /// \brief Starts a transaction to create or replace the table
+    ///
+    /// \breturn the Transaction to create or replace the table
```
shangxinli: typo? '\return'?

src/iceberg/table.h
```diff
@@ -109,6 +109,11 @@ class ICEBERG_EXPORT Table {
   /// filter data.
   virtual std::unique_ptr<TableScanBuilder> NewScan() const;
 
+  /// \brief Create a new transaction for this table
+  ///
+  /// \return a shared pointer to the new Transaction
+  virtual std::shared_ptr<Transaction> NewTransaction();
```
shangxinli: We don't want to add 'const' just like NewScan() and io()?

src/iceberg/transaction.h
wgtmac: I think we can break any existing API before 1.0.0. We just released 0.1.0 a while ago and I don't think it is ready for production yet.

src/iceberg/catalog.h
lishuxu: Good catch! Fixed the typo.

src/iceberg/table.h
lishuxu: Yes, it should be const.

src/iceberg/transaction.h
```diff
@@ -42,12 +44,23 @@ class ICEBERG_EXPORT Transaction {
...
   virtual std::shared_ptr<AppendFiles> NewAppend() = 0;
 
+  /// \brief Apply multiple metadata updates to this transaction
+  ///
+  /// \param requirements the table requirements to validate
+  /// \param updates the table updates to apply
+  /// \return Status::OK if all updates were queued successfully
+  virtual Status UpdateTable(
+      const std::vector<std::unique_ptr<TableRequirement>>& requirements,
+      std::vector<std::unique_ptr<TableUpdate>> updates) = 0;
```
wgtmac:
```suggestion
      const std::vector<std::unique_ptr<TableUpdate>>& updates) = 0;
```

src/iceberg/table.h
```diff
@@ -109,6 +109,11 @@ class ICEBERG_EXPORT Table {
   /// filter data.
   virtual std::unique_ptr<TableScanBuilder> NewScan() const;
 
+  /// \brief Create a new transaction for this table
+  ///
+  /// \return a shared pointer to the new Transaction
+  virtual std::shared_ptr<Transaction> NewTransaction() const;
```
wgtmac:
```suggestion
  virtual std::unique_ptr<Transaction> NewTransaction() const;
```

src/iceberg/catalog.h
```diff
@@ -223,6 +247,16 @@ class ICEBERG_EXPORT Catalog {
...
+
+    /// \brief Starts a transaction to replace the table
+    ///
+    /// \return the Transaction to replace the table
+    virtual std::unique_ptr<Transaction> StageReplace() = 0;
+
+    /// \brief Starts a transaction to create or replace the table
+    ///
+    /// \return the Transaction to create or replace the table
+    virtual std::unique_ptr<Transaction> StageCreateOrReplace() = 0;
```
wgtmac:
```suggestion
```

src/iceberg/catalog.h
```diff
@@ -140,6 +140,21 @@ class ICEBERG_EXPORT Catalog {
...
+  /// \param location a location for the table; leave empty if unspecified
+  /// \param properties a string map of table properties
+  /// \param orCreate whether to create the table if not exists
+  /// \return a Transaction to replace the table or ErrorKind::kNotFound if the table
+  /// doesn't exist and orCreate is false
+  virtual Result<std::shared_ptr<Transaction>> StageReplaceTable(
+      const TableIdentifier& identifier, const Schema& schema, const PartitionSpec& spec,
+      const std::string& location,
+      const std::unordered_map<std::string, std::string>& properties, bool orCreate) = 0;
+
```
wgtmac:
```suggestion
```

wgtmac (changes requested):

src/iceberg/table.h
lishuxu: OK, that's would be better

src/iceberg/catalog.h
lishuxu: Refer to (https://github.com/apache/iceberg-python/issues/281), the replace table API is currently not planned to be supported.

src/iceberg/transaction.h
```diff
@@ -42,12 +44,23 @@ class ICEBERG_EXPORT Transaction {
   /// \return a new AppendFiles
   virtual std::shared_ptr<AppendFiles> NewAppend() = 0;
 
+  /// \brief Apply multiple metadata updates to this transaction
+  ///
+  /// \param requirements the table requirements to validate
+  /// \param updates the table updates to apply
+  /// \return Status::OK if all updates were queued successfully
+  virtual Status UpdateTable(
```
wgtmac: Why do we need this? I don't think `Transaction` is the entrance to create a single update action and `Catalog` is responsible of this. Let's remove it.

src/iceberg/transaction.h
```diff
@@ -42,12 +44,23 @@ class ICEBERG_EXPORT Transaction {
...
+      const std::vector<std::unique_ptr<TableRequirement>>& requirements,
+      const std::vector<std::unique_ptr<TableUpdate>>& updates) = 0;
+
```
wgtmac:
```suggestion
```

wgtmac (approved):

src/iceberg/transaction.h
```diff
@@ -21,8 +21,10 @@
 #pragma once
 
 #include <memory>
+#include <vector>
```
HeartLinked, line 24: Why include this?

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -415,6 +415,12 @@ Status InMemoryCatalog::DropTable(const TableIdentifier& identifier, bool purge)
   return root_namespace_->UnregisterTable(identifier);
 }
 
+Status InMemoryCatalog::RenameTable(const TableIdentifier& from,
+                                    const TableIdentifier& to) {
+  std::unique_lock lock(mutex_);
```
HuaHuaY, line 420: If we don't need to move `lock`, we can use `std::lock_guard<>` instead. Can you help to modify the other cases in this file? Or we can open a new PR to do this.

PR #284, opened 2025-10-30
feat: add satisfies order for SortField/SortOrder and Transform

zhjwpku:
This PR also makes the `ToString` consistent with Java implementation.

### DIFF
--- a/src/iceberg/sort_order.cc
+++ b/src/iceberg/sort_order.cc
@@ -41,7 +41,7 @@ std::span<const SortField> SortOrder::fields() const { return fields_; }
 
 bool SortOrder::Satisfies(const SortOrder& other) const {
   // any ordering satisfies an unsorted ordering
-  if (other.IsUnsorted()) {
+  if (other.is_unsorted()) {
     return true;
   }
 
@@ -61,17 +61,7 @@ bool SortOrder::Satisfies(const SortOrder& other) const {
 }
 
 bool SortOrder::SameOrder(const SortOrder& other) const {
-  if (fields_.size() != other.fields().size()) {
-    return false;
-  }
-
-  for (const auto& [field, other_field] : std::views::zip(fields_, other.fields())) {
-    if (!(field == other_field)) {
-      return false;
-    }
-  }
-
-  return true;
+  return fields_ == other.fields_;
 }
 
 std::string SortOrder::ToString() const {
--- a/src/iceberg/sort_order.h
+++ b/src/iceberg/sort_order.h
@@ -50,11 +50,11 @@ class ICEBERG_EXPORT SortOrder : public util::Formattable {
   std::span<const SortField> fields() const;
 
   /// \brief Returns true if the sort order is sorted
-  bool IsSorted() const { return !fields_.empty(); }
+  bool is_sorted() const { return !fields_.empty(); }
 
   /// \brief Returns true if the sort order is unsorted
   /// A SortOrder is unsorted if it has no sort fields.
-  bool IsUnsorted() const { return fields_.empty(); }
+  bool is_unsorted() const { return fields_.empty(); }
 
   /// \brief Checks whether this order satisfies another order.
   bool Satisfies(const SortOrder& other) const;
--- a/src/iceberg/transform.cc
+++ b/src/iceberg/transform.cc
@@ -21,6 +21,7 @@
 
 #include <format>
 #include <regex>
+#include <utility>
 
 #include "iceberg/transform_function.h"
 #include "iceberg/type.h"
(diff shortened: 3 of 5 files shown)


src/iceberg/sort_order.cc
```diff
@@ -38,10 +39,48 @@ int32_t SortOrder::order_id() const { return order_id_; }
...
+    return true;
+  }
+
+  // this ordering cannot satisfy an ordering with more sort fields
+  if (fields_.size() < other.fields().size()) {
+    return false;
+  }
+
+  // this ordering has either more or the same number of sort fields
+  for (const auto& [field, other_field] : std::views::zip(fields_, other.fields_)) {
```
shangxinli, line 54: It seems using std::views::zip from C++23 without checking C++ standard compliance. I remember std::views::zip is C++23 only (not available in C++20 or earlier). Not sure if the codebase doesn't appear to enforce C++23. This will cause compilation failures on older compilers. 

// Option 1: Traditional iteration (works in C++17+)
  for (size_t i = 0; i < other.fields().size(); ++i) {
    if (!fields_[i].Satisfies(other.fields()[i])) {
      return false;
    }
  }

  // Option 2: Use std::ranges::zip only if C++23 is available
  #if __cplusplus >= 202302L
    for (const auto& [field, other_field] : std::views::zip(fields_, other.fields())) {
  #else
    for (size_t i = 0; i < other.fields().size(); ++i) {
      const auto& field = fields_[i];
      const auto& other_field = other.fields()[i];
  #endif
      if (!field.Satisfies(other_field)) {
        return false;
      }
    }

src/iceberg/sort_order.cc
```diff
@@ -38,10 +39,48 @@ int32_t SortOrder::order_id() const { return order_id_; }
...
+
+  return true;
+}
+
+bool SortOrder::SameOrder(const SortOrder& other) const {
+  if (fields_.size() != other.fields().size()) {
+    return false;
+  }
+
+  for (const auto& [field, other_field] : std::views::zip(fields_, other.fields())) {
```
shangxinli: same comments as above.

src/iceberg/sort_order.cc
zhjwpku: If you check the README, you'll see that iceberg-cpp targets C++23. We rely on several C++23 features, including std::expected, enhanced ranges, and std::unreachable, among others. So the std::views::zip here should be fine.

shangxinli: OK. Then we are good

shangxinli (approved):

src/iceberg/sort_order.h
```diff
@@ -49,6 +49,20 @@ class ICEBERG_EXPORT SortOrder : public util::Formattable {
   /// \brief Get the list of sort fields.
   std::span<const SortField> fields() const;
 
+  /// \brief Returns true if the sort order is sorted
+  bool IsSorted() const { return !fields_.empty(); }
```
wgtmac:
```suggestion
  bool is_sorted() const { return !fields_.empty(); }
```

src/iceberg/sort_order.h
```diff
@@ -49,6 +49,20 @@ class ICEBERG_EXPORT SortOrder : public util::Formattable {
...
+
+  /// \brief Returns true if the sort order is unsorted
+  /// A SortOrder is unsorted if it has no sort fields.
+  bool IsUnsorted() const { return fields_.empty(); }
```
wgtmac:
```suggestion
  bool is_unsorted() const { return fields_.empty(); }
```

src/iceberg/sort_order.cc
```diff
@@ -38,10 +39,48 @@ int32_t SortOrder::order_id() const { return order_id_; }
...
+    if (!(field == other_field)) {
+      return false;
+    }
+  }
+
+  return true;
```
wgtmac:
```suggestion
  return fields_ == other.fields_;
```

Can we simplify it as above?

src/iceberg/transform.cc
```diff
@@ -124,6 +124,35 @@ Result<std::shared_ptr<TransformFunction>> Transform::Bind(
   }
 }
 
+bool Transform::PreservesOrder() const {
+  switch (transform_type_) {
+    case TransformType::kUnknown:
+    case TransformType::kVoid:
+    case TransformType::kBucket:
+      return false;
+    default:
```
wgtmac: Should we explicitly list all `transform_type_`s just in case of forgetting to add a new one in the future?

src/iceberg/transform.cc
```diff
@@ -124,6 +124,35 @@ Result<std::shared_ptr<TransformFunction>> Transform::Bind(
...
+      return other.PreservesOrder();
+    case TransformType::kHour:
+      return other_type == TransformType::kHour || other_type == TransformType::kDay ||
+             other_type == TransformType::kMonth || other_type == TransformType::kYear;
+    case TransformType::kDay:
+      return other_type == TransformType::kDay || other_type == TransformType::kMonth ||
+             other_type == TransformType::kYear;
+    case TransformType::kMonth:
+      return other_type == TransformType::kMonth || other_type == TransformType::kYear;
+    default:
```
wgtmac: ditto

src/iceberg/transform.cc
```diff
@@ -124,6 +124,35 @@ Result<std::shared_ptr<TransformFunction>> Transform::Bind(
...
+      return true;
+  }
+}
+
+bool Transform::SatisfiesOrderOf(const Transform& other) const {
+  auto other_type = other.transform_type();
+  switch (transform_type_) {
+    case TransformType::kIdentity:
+      return other.PreservesOrder();
```
wgtmac, line 150:
```suggestion
      // ordering by value is the same as long as the other preserves order
      return other.PreservesOrder();
```

src/iceberg/transform.cc
wgtmac: BTW, it seems that this implementation is incomplete? For example, truncate should have special logic.

src/iceberg/sort_order.cc
zhjwpku: reasonable, fixed.

src/iceberg/transform.cc
zhjwpku: Done.

src/iceberg/transform.cc
zhjwpku: Yeah, that's an oversight from my side, added.

src/iceberg/transform.cc
zhjwpku: done.

PR #285, opened 2025-11-01
fix: result type of DayTransform should be date

zhjwpku:
Result type of Day transform should be `int` per the spec, but in the Java impl, it is `DateType`, this discrepancy has been discussed on devlist, see [1]. iceberg-iceberg and iceberg-rust all conform to the java impl, so make iceberg-cpp do the same.

The added comments are borrowed from iceberg-python.

[1] https://lists.apache.org/thread/2gq7b54nvc9q6f1j08l9lnzgm5onkmx5

### DIFF
--- a/src/iceberg/test/partition_spec_test.cc
+++ b/src/iceberg/test/partition_spec_test.cc
@@ -24,7 +24,9 @@
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
 
+#include "iceberg/json_internal.h"
 #include "iceberg/partition_field.h"
 #include "iceberg/schema.h"
 #include "iceberg/schema_field.h"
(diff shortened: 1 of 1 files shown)


src/iceberg/test/transform_test.cc
```diff
@@ -134,7 +134,7 @@ TEST(TransformResultTypeTest, PositiveCases) {
        .expected_result_type = iceberg::int32()},
       {.str = "day",
        .source_type = iceberg::timestamp(),
-       .expected_result_type = iceberg::int32()},
+       .expected_result_type = iceberg::date()},
```
shangxinli, line 137: Adding tests for Wire Format Compatibility would be better. The PR changes the return type from int32() to date() but doesn't verify:
  1. That date() values serialize to int32 on the wire
  2. That existing partitioned data files can still be read
  3. That this change is backward compatible with tables partitioned using the old int32 type

zhjwpku: I'm afraid we don't have fully covered integration test yet. I added a test for PartitionSpec, though I'm not sure if that's sufficient.

shangxinli: SG

wgtmac (approved):

gty404 (approved):

PR #289, opened 2025-11-03
feat: optimize truncate(col) == value to startsWith predicate

shangxinli:
Rewrite truncate(col) == "value" predicates to col STARTS_WITH "value" for string columns. This enables better predicate pushdown to storage formats and efficient use of prefix indexes.

The optimization only applies when:
- Operation is equality
- Term is a truncate transform
- Literal is a string type

Added tests to verify correct application and edge cases.

### DIFF
--- a/src/iceberg/expression/predicate.cc
+++ b/src/iceberg/expression/predicate.cc
@@ -25,7 +25,10 @@
 
 #include "iceberg/expression/expressions.h"
 #include "iceberg/expression/literal.h"
+#include "iceberg/expression/term.h"
 #include "iceberg/result.h"
+#include "iceberg/transform.h"
+#include "iceberg/transform_function.h"
 #include "iceberg/type.h"
 #include "iceberg/util/checked_cast.h"
 #include "iceberg/util/formatter_internal.h"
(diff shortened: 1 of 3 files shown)


wgtmac:
There are some merge conflicts...

shangxinli:
Fixed

src/iceberg/expression/predicate.cc
```diff
@@ -143,6 +145,27 @@ bool IsFloatingType(TypeId type) {
   return type == TypeId::kFloat || type == TypeId::kDouble;
 }
 
+/// \brief Count the number of UTF-8 code points in a string.
+/// This matches the behavior of TruncateUtils::TruncateUTF8.
+///
+/// NOTE: This counts code points, not grapheme clusters (user-perceived characters).
+/// Per the Iceberg spec, combining marks count as separate code points.
```
wgtmac: Iceberg spec does not have this. We should not mislead readers.

src/iceberg/expression/predicate.cc
```diff
@@ -143,6 +145,27 @@ bool IsFloatingType(TypeId type) {
...
+/// Example: "é" as e + combining-acute (U+0065 U+0301) = 2 code points,
+///          but "é" as single precomposed character (U+00E9) = 1 code point.
+///
+/// \param str The UTF-8 encoded string
+/// \return The number of code points (not bytes, not graphemes)
+inline int32_t CountUTF8CodePoints(std::string_view str) {
```
wgtmac: Perhaps this should be moved to `string_util.h` or `truncate_util.h`

src/iceberg/expression/predicate.cc
```diff
@@ -246,7 +269,69 @@ Result<std::shared_ptr<Expression>> UnboundPredicate<B>::BindLiteralOperation(
...
+  //   implies exact match, so STARTS_WITH remains valid (short-string invariance)
+  if (BASE::op() == Expression::Operation::kEq &&
+      bound_term->kind() == Term::Kind::kTransform) {
+    // Safe to cast after kind check confirms it's a transform
+    auto* transform_term = dynamic_cast<BoundTransform*>(bound_term.get());
+    if (!transform_term) {
+      // Should never happen after kind check, but be defensive
+      return std::make_shared<BoundLiteralPredicate>(BASE::op(), std::move(bound_term),
+                                                     std::move(literal));
+    }
```
wgtmac:
```suggestion
    auto* transform_term = internal::checked_cast<BoundTransform*>(bound_term.get());
```

Let's reuse the existing pattern for cast.

src/iceberg/expression/predicate.cc
```diff
@@ -246,7 +269,69 @@ Result<std::shared_ptr<Expression>> UnboundPredicate<B>::BindLiteralOperation(
...
+
+    if (transform_term->transform()->transform_type() == TransformType::kTruncate &&
+        literal.type()->type_id() == TypeId::kString &&
+        !literal.IsNull()) {  // Null safety: skip null literals
+
+      // Extract width parameter using type-safe API
+      auto width_opt = transform_term->transform()->param();
```
wgtmac: I think we can greatly simplify this line and below.

We can add `transform_func` like below.

```cpp
class ICEBERG_EXPORT BoundTransform : public BoundTerm {
 public:
  const std::shared_ptr<TransformFunction>& transform_func() const { return transform_func_; }
};
```

Then we just need to call `transform_term->transform_func()->Transform(literal)` and check if its return value is same as `literal`. We don't even need `CountUTF8CodePoints` above.

src/iceberg/transform.h
```diff
@@ -141,6 +141,19 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
   /// \brief Returns the transform type.
   TransformType transform_type() const;
 
+  /// \brief Returns the optional parameter for parameterized transforms.
+  ///
+  /// For transforms like bucket(N) or truncate(W), returns the parameter value.
+  /// For non-parameterized transforms (identity, year, etc.), returns std::nullopt.
+  ///
+  /// \return The parameter if present, otherwise std::nullopt
+  std::optional<int32_t> param() const {
```
wgtmac: Can we remove this function? This makes it harder to evolve the variant in the future.

src/iceberg/expression/predicate.cc
shangxinli: Removed

src/iceberg/expression/predicate.cc
shangxinli: moved

src/iceberg/expression/predicate.cc
shangxinli: sure

src/iceberg/expression/predicate.cc
shangxinli: sure

src/iceberg/transform.h
shangxinli: sure

src/iceberg/expression/predicate.cc
```diff
@@ -286,7 +289,52 @@ Result<std::shared_ptr<Expression>> UnboundPredicate<B>::BindLiteralOperation(
...
+                                         std::move(literal));
+    }
+
+    if (transform_term->transform()->transform_type() == TransformType::kTruncate &&
+        literal.type()->type_id() == TypeId::kString &&
+        !literal.IsNull()) {  // Null safety: skip null literals
+
+      // Apply truncate transform to the literal and check if result matches
+      // This verifies the literal is compatible with the truncate operation
+      auto transformed_result = transform_term->transform_func()->Transform(literal);
```
zhjwpku, line 307: I'm wondering whether we can simply count the code points in the literal and compare that with `TruncateTransform.width()`. If they match, we can rewrite it as a `StartsWith`. We don't need to transform the extended_literal then.

zhjwpku: Also, if the number of code points in the literal is greater than `TruncateTransform.width()`, we can rewrite it as `AlwaysFalse`. I created PR [1] for the Java implementation to do this.

[1] https://github.com/apache/iceberg/pull/16383

PR #290, opened 2025-11-04
refactor: simplify test structure for rest catalog models

wgtmac:
- simplified rest catalog model definitions.
- added common functions to operate json ser/de.
- refactored cases to be able to be shared.


liurenjie1024 (approved):
Thanks @wgtmac for this fix!

PR #292, opened 2025-11-04
feat: add config, error and validation for rest types

HeartLinked:


src/iceberg/catalog/rest/types.h
```diff
@@ -34,6 +33,26 @@
 
 namespace iceberg::rest {
 
+/// \brief Server-provided configuration for the catalog.
+struct ICEBERG_REST_EXPORT CatalogConfig {
+  std::unordered_map<std::string, std::string> overrides;  // required
+  std::unordered_map<std::string, std::string> defaults;   // required
```
```suggestion
  std::unordered_map<std::string, std::string> defaults;   // required
  std::unordered_map<std::string, std::string> overrides;  // required
```

wgtmac: Let's switch the order to be more natural.

src/iceberg/meson.build
```diff
@@ -121,6 +127,7 @@ iceberg_lib = library(
     dependencies: iceberg_deps,
     include_directories: iceberg_include_dir,
     install: true,
+    cpp_args: iceberg_common_cpp_args,
     gnu_symbol_visibility: 'inlineshidden',
```
wgtmac:
```suggestion
    gnu_symbol_visibility: 'inlineshidden',
    cpp_args: iceberg_common_cpp_args,
```

Let's put similar flags closer

src/iceberg/meson.build
```diff
@@ -15,6 +15,12 @@
 # specific language governing permissions and limitations
 # under the License.
 
+cpp = meson.get_compiler('cpp')
+iceberg_common_cpp_args = []
+if cpp.get_id() == 'msvc'
+    iceberg_common_cpp_args += cpp.get_supported_arguments(['/bigobj'])
+endif
```
wgtmac: Is it better to move them to above line 124?

src/iceberg/meson.build
```diff
@@ -15,6 +15,12 @@
 # specific language governing permissions and limitations
 # under the License.
 
+cpp = meson.get_compiler('cpp')
+iceberg_common_cpp_args = []
+if cpp.get_id() == 'msvc'
+    iceberg_common_cpp_args += cpp.get_supported_arguments(['/bigobj'])
```
wgtmac: Please add some comments on why we need this. BTW, does the issue happen when compiling the test executable? If yes, perhaps only add this flag to the test target?

cc @WillAyd

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -59,9 +60,77 @@ constexpr std::string_view kDestination = "destination";
...
 
 }  // namespace
 
+nlohmann::json ToJson(const CatalogConfig& config) {
+  nlohmann::json json;
+  json[kOverrides] = config.overrides;
+  json[kDefaults] = config.defaults;
+  if (!config.endpoints.empty()) {
+    json[kEndpoints] = config.endpoints;
+  }
```
wgtmac:
```suggestion
  SetContainerField(json, kEndpoints, config.endpoints);
```

src/iceberg/catalog/rest/types.h
```diff
@@ -34,6 +33,26 @@
...
+  std::vector<std::string> endpoints;
+};
+
+/// \brief JSON error payload returned in a response with further details on the error.
+struct ICEBERG_REST_EXPORT ErrorModel {
+  std::string message;  // required
+  std::string type;     // required
+  uint16_t code;        // required
```
wgtmac: Why `uint16_t` but not `int32_t`? The spec says `type: integer`.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -59,9 +60,77 @@ constexpr std::string_view kDestination = "destination";
...
+}
+
+nlohmann::json ToJson(const ErrorModel& error) {
+  nlohmann::json json;
+  json[kMessage] = error.message;
+  json[kType] = error.type;
+  json[kCode] = error.code;
+  if (!error.stack.empty()) {
+    json[kStack] = error.stack;
+  }
```
wgtmac:
```suggestion
  SetContainerField(json, kStack, error.stack);
```

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -59,9 +60,77 @@ constexpr std::string_view kDestination = "destination";
...
+  return json;
+}
+
+Result<ErrorModel> ErrorModelFromJson(const nlohmann::json& json) {
```
wgtmac, line 104: It seems to be inconsistent with Java where all fields are nullable.

```java
  public static ErrorResponse fromJson(JsonNode jsonNode) {
    Preconditions.checkArgument(
        jsonNode != null && jsonNode.isObject(),
        "Cannot parse error response from non-object value: %s",
        jsonNode);
    Preconditions.checkArgument(jsonNode.has(ERROR), "Cannot parse missing field: error");
    JsonNode error = JsonUtil.get(ERROR, jsonNode);
    String message = JsonUtil.getStringOrNull(MESSAGE, error);
    String type = JsonUtil.getStringOrNull(TYPE, error);
    Integer code = JsonUtil.getIntOrNull(CODE, error);
    List<String> stack = JsonUtil.getStringListOrNull(STACK, error);
    return ErrorResponse.builder()
        .withMessage(message)
        .withType(type)
        .responseCode(code)
        .withStackTrace(stack)
        .build();
  }
```

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+  return {};
+}
+
+Status Validator::Validate(const ErrorModel& error) {
+  if (error.message.empty() || error.type.empty()) [[unlikely]] {
+    return Invalid("Invalid error model: missing required fields");
+  }
+
+  if (error.code < 400 || error.code > 600) [[unlikely]] {
+    return Invalid("Invalid error model: code must be between 400 and 600");
```
wgtmac: nit: print the code in the message.

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // stack is optional, no validation needed
+  return {};
+}
+
+Status Validator::Validate(const ErrorResponse& response) { return {}; }
```
wgtmac, line 57: It is worth adding a comment that `We don't validate the error field because ErrorModel::Validate has been called in the FromJson`.

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+  if (!common.empty()) {
+    std::string keys;
+    bool first = true;
+    for (const std::string& s : common) {
+      if (!std::exchange(first, false)) keys += ", ";
+      keys += s;
+    }
+
+    return Invalid(
+        "Invalid namespace properties update: cannot simultaneously set and remove keys: "
```
wgtmac:
```suggestion
        "Invalid namespace update: cannot simultaneously set and remove keys: "
```

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+
+Status Validator::Validate(const CreateNamespaceRequest& request) { return {}; }
+
+Status Validator::Validate(const CreateNamespaceResponse& response) { return {}; }
+
+Status Validator::Validate(const GetNamespaceResponse& response) { return {}; }
+
+Status Validator::Validate(const UpdateNamespacePropertiesRequest& request) {
+  // keys in updates and removals must not overlap
+  if (request.removals.empty() || request.updates.empty()) [[unlikely]] {
```
wgtmac:
```suggestion
  if (request.removals.empty() || request.updates.empty()) {
```

The unlikely attribute should not be used like this.

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+    return {};
+  }
+
+  std::unordered_set<std::string> remove_set(request.removals.begin(),
```
wgtmac: Can we use `std::string_view` instead of copying every key here?

wgtmac: Perhaps it much cleaner to use `std::set_intersection`?

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+Status Validator::Validate(const LoadTableResult& result) {
+  if (!result.metadata) [[unlikely]] {
+    return Invalid("Invalid metadata: null");
+  }
+  return {};
+}
+
+Status Validator::Validate(const RegisterTableRequest& request) {
+  if (request.name.empty()) [[unlikely]] {
+    return Invalid("Invalid table name: empty");
```
wgtmac:
```suggestion
    return Invalid("Missing table name");
```

It is misleading since users may think the table name is called `empty`.

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  if (request.metadata_location.empty()) [[unlikely]] {
+    return Invalid("Invalid metadata location: empty");
```
wgtmac:
```suggestion
    return Invalid("Empty metadata location");
```

src/iceberg/catalog/rest/validator.cc
wgtmac: Although the Java impl is doing the same thing. However, I think calling validate on fields recursively is a more robust approach. This is not a strong opinion. If we finally choose this approach, then we should remove this `Validator` class and let each object to add a `Validate` function.

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  return {};
+}
+
+Status Validator::Validate(const RenameTableRequest& request) {
+  if (request.source.ns.levels.empty() || request.source.name.empty()) [[unlikely]] {
```
wgtmac: How about adding a `Status Validator::Validate(const TableIdentifier& identifier)`? BTW, I don't think we need to check `ns.levels.empty()`, do we? It is totally fine to create a table without any namespace.

src/iceberg/catalog/rest/json_internal.cc
```diff
@@ -59,9 +60,76 @@ constexpr std::string_view kDestination = "destination";
...
+  json[kMessage] = error.message;
+  json[kType] = error.type;
+  json[kCode] = error.code;
+  SetContainerField(json, kStack, error.stack);
+  return json;
+}
+
+Result<ErrorModel> ErrorModelFromJson(const nlohmann::json& json) {
+  ErrorModel error;
+  // NOTE: Iceberg's Java implementation allows missing required fields (message, type,
```
wgtmac, line 106: @Fokko I don't think this is something serious but there is a slight inconsistency between Java impl and the spec.

src/iceberg/catalog/rest/validator.cc
```diff
@@ -0,0 +1,146 @@
...
+    std::string keys;
+    for (size_t i = 0; i < common.size(); ++i) {
+      if (i) keys += ", ";
+      keys += common[i];
+    }
+
+    return Invalid(
+        "Invalid namespace update: cannot simultaneously set and remove keys: "
+        "[{}]",
+        keys);
```
wgtmac: nit: you can include `formatter_internal.h` to directly print `common`

src/iceberg/test/meson.build
```diff
@@ -96,6 +96,12 @@ if get_option('rest').enabled()
     }
 endif
 
+cpp = meson.get_compiler('cpp')
+cpp_bigobj_args = []
+if cpp.get_id() == 'msvc'
+    iceberg_common_cpp_args += cpp.get_supported_arguments(['/bigobj'])
```
wgtmac: @WillAyd Do we need to apply the flag to a single source file only?

src/iceberg/test/meson.build
```diff
@@ -96,6 +96,12 @@ if get_option('rest').enabled()
     }
 endif
 
+cpp = meson.get_compiler('cpp')
+cpp_bigobj_args = []
+if cpp.get_id() == 'msvc'
```
WillAyd: Just some extra consideration points for youf:

1. The `if cpp.get_id() == 'msvc'` and `cpp.get_supported_arguments(...)` calls are redundant; the latter is all you need. I don't think the redundancy here is bad, but if you have a lot of arguments its usually easier to let Meson figure out what is applicable, rather than branching on each compiler
2. Rather than applying this at the target level, you _might_ want to apply /bigobj to all targets in the project. So in the root configuration something like:

```python
args = cpp.get_supported_arguments(['/bigobj'])
add_project_arguments(args, language: 'cpp')
```

Would lessen redundancy as more targets require this

WillAyd (approved):

src/iceberg/test/meson.build
wgtmac: Nice design! Thanks @WillAyd!

PR #294, opened 2025-11-04
feat: implement validation for table update requirements

shangxinli:
Implement Validate() methods for remaining table requirement classes:
- AssertDoesNotExist: validates table doesn't exist
- AssertRefSnapshotID: validates snapshot references (branches/tags)
- AssertLastAssignedFieldId: validates last assigned field ID
- AssertLastAssignedPartitionId: validates last assigned partition ID
- AssertDefaultSpecID: validates default partition spec ID
- AssertDefaultSortOrderID: validates default sort order ID

All implementations follow the pattern established in AssertCurrentSchemaID and provide descriptive error messages for validation failures.

### DIFF
--- a/src/iceberg/test/table_metadata_builder_test.cc
+++ b/src/iceberg/test/table_metadata_builder_test.cc
@@ -354,9 +354,7 @@ TEST_F(TableMetadataBuilderTest, TableRequirementAssertLastAssignedFieldIdMismat
 TEST_F(TableMetadataBuilderTest, TableRequirementAssertLastAssignedFieldIdNullBase) {
   table::AssertLastAssignedFieldId requirement(10);
 
-  auto status = requirement.Validate(nullptr);
-  EXPECT_THAT(status, IsError(ErrorKind::kCommitFailed));
-  EXPECT_THAT(status, HasErrorMessage("metadata is missing"));
+  EXPECT_THAT(requirement.Validate(nullptr), IsOk());
 }
 
 TEST_F(TableMetadataBuilderTest, TableRequirementAssertLastAssignedPartitionIdSuccess) {


zhjwpku (approved):
LGTM

src/iceberg/table_requirement.cc
```diff
@@ -45,12 +52,55 @@ Status AssertUUID::Validate(const TableMetadata* base) const {
...
 Status AssertLastAssignedFieldId::Validate(const TableMetadata* base) const {
-  return NotImplemented(
-      "AssertCurrentTableLastAssignedFieldId::Validate not implemented");
+  // Validate that the last assigned field ID matches the expected value
+
+  if (base == nullptr) {
+    return CommitFailed("Requirement failed: current table metadata is missing");
+  }
+
+  if (base->last_column_id != last_assigned_field_id_) {
```
wgtmac:
```suggestion
  if (base && base->last_column_id != last_assigned_field_id_) {
```

The Java impl allows base metadata is missing. I think it makes sense for a new table.

src/iceberg/table_requirement.cc
```diff
@@ -75,16 +125,53 @@ Status AssertCurrentSchemaID::Validate(const TableMetadata* base) const {
...
 Status AssertLastAssignedPartitionId::Validate(const TableMetadata* base) const {
-  return NotImplemented(
-      "AssertCurrentTableLastAssignedPartitionId::Validate not implemented");
+  // Validate that the last assigned partition ID matches the expected value
+
+  if (base == nullptr) {
+    return CommitFailed("Requirement failed: current table metadata is missing");
+  }
+
+  if (base->last_partition_id != last_assigned_partition_id_) {
```
wgtmac:
```suggestion
  if (base && base->last_partition_id != last_assigned_partition_id_) {
```

src/iceberg/table_requirement.cc
```diff
@@ -45,12 +52,55 @@ Status AssertUUID::Validate(const TableMetadata* base) const {
 }
 
 Status AssertRefSnapshotID::Validate(const TableMetadata* base) const {
-  return NotImplemented("AssertTableRefSnapshotID::Validate not implemented");
+  // Validate that a reference (branch or tag) points to the expected snapshot ID
+
+  if (base == nullptr) {
```
wgtmac, line 52: This is inconsistent with the Java impl. Could you double check?

```java
    public void validate(TableMetadata base) {
      SnapshotRef ref = base.ref(name);
      if (ref != null) {
        String type = ref.isBranch() ? "branch" : "tag";
        if (snapshotId == null) {
          // a null snapshot ID means the ref should not exist already
          throw new CommitFailedException(
              "Requirement failed: %s %s was created concurrently", type, name);
        } else if (snapshotId != ref.snapshotId()) {
          throw new CommitFailedException(
              "Requirement failed: %s %s has changed: expected id %s != %s",
              type, name, snapshotId, ref.snapshotId());
        }
      } else if (snapshotId != null) {
        throw new CommitFailedException(
            "Requirement failed: branch or tag %s is missing, expected %s", name, snapshotId);
      }
    }
```

src/iceberg/table_requirement.cc
shangxinli: Good catch

src/iceberg/table_requirement.cc
shangxinli: changed

src/iceberg/table_requirement.cc
shangxinli: Changed

wgtmac (approved):
Thanks a lot @shangxinli!

I just made a push to add some missing checks and removed strange comments generated by AI. :)

src/iceberg/table_requirement.cc
```diff
@@ -19,43 +19,77 @@
...
+  return {};
 }
 
 Status AssertLastAssignedFieldId::Validate(const TableMetadata* base) const {
-  return NotImplemented(
-      "AssertCurrentTableLastAssignedFieldId::Validate not implemented");
+  if (base == nullptr) {
+    return CommitFailed("Requirement failed: current table metadata is missing");
+  }
+
```
wgtmac:
```suggestion
```

PR #300, opened 2025-11-07
feat: add SortOrder::Make and SortOrder::Validate

zhjwpku:


src/iceberg/transform.h
```diff
@@ -150,6 +150,10 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
   Result<std::shared_ptr<TransformFunction>> Bind(
       const std::shared_ptr<Type>& source_type) const;
 
+  /// \brief Returns the Type produced by this transform given a source type.
+  Result<std::shared_ptr<Type>> ResultType(
+      const std::shared_ptr<Type>& source_type) const;
```
wgtmac: nit: `const Type&` might be a better parameter type.

src/iceberg/sort_order.cc
```diff
@@ -80,4 +86,66 @@ bool SortOrder::Equals(const SortOrder& other) const {
...
+      return InvalidArgument("Cannot find source column for sort field: {}", field);
+    }
+
+    const auto& source_type = schema_field.value().get().type();
+
+    if (!source_type->is_primitive()) {
+      return InvalidArgument("Cannot sort by non-primitive source field: {}",
+                             *source_type);
+    }
+
```
wgtmac:
```suggestion
    const auto& source_type = schema_field.value().get().type();
```

I think the check is covered by line below?

src/iceberg/sort_order.cc
```diff
@@ -80,4 +86,66 @@ bool SortOrder::Equals(const SortOrder& other) const {
...
+    ICEBERG_RETURN_UNEXPECTED(field.transform()->ResultType(source_type));
```
wgtmac: Perhaps we should add `bool Transform::CanTransform(const Type&) const` and use it here? It is a little bit odd that the result type is not used here.

src/iceberg/sort_order.cc
```diff
@@ -80,4 +86,66 @@ bool SortOrder::Equals(const SortOrder& other) const {
   return order_id_ == other.order_id_ && fields_ == other.fields_;
 }
 
+bool SortOrder::IsBoundToSchema(const Schema& schema) const {
```
wgtmac: It seems that we don't need this function? Or should it be in the test file? Perhaps you can change it to `Status SortOrder::Validate(const Schema& schema)`?

src/iceberg/sort_order.cc
zhjwpku: Yeah, that's redundant, deleted.

src/iceberg/sort_order.cc
zhjwpku: I've changed the Transform::ResultType to Transform::CanTransform, which seems more natural.

src/iceberg/sort_order.cc
zhjwpku: Changed to SortOrder::Validate.

src/iceberg/sort_order.cc
```diff
@@ -80,4 +86,60 @@ bool SortOrder::Equals(const SortOrder& other) const {
...
+
+  if (fields.empty() && sort_id != kUnsortedOrderId) [[unlikely]] {
+    return InvalidArgument("Sort order must have at least one sort field");
+  }
+
+  for (const auto& field : fields) {
+    ICEBERG_ASSIGN_OR_RAISE(auto schema_field, schema.FindFieldById(field.source_id()));
+    if (schema_field == std::nullopt) {
+      return InvalidArgument("Cannot find schema field for sort field: {}", field);
+    }
```
wgtmac, line 120:
```suggestion
    if (schema_field == std::nullopt) {
      continue;
    }
```

I think we should not error out if the field is missing. The Java impl is as below:

```java
  public SortOrder bind(Schema schema) {
    SortOrder.Builder builder = SortOrder.builderFor(schema).withOrderId(orderId);

    for (UnboundSortField field : fields) {
      Type sourceType = schema.findType(field.sourceId);
      Transform<?, ?> transform;
      if (sourceType != null) {
        transform = Transforms.fromString(sourceType, field.transform.toString());
      } else {
        transform = field.transform;
      }
      builder.addSortField(transform, field.sourceId, field.direction, field.nullOrder);
    }

    return builder.build();
  }
```

src/iceberg/sort_order.h
```diff
@@ -69,6 +72,28 @@ class ICEBERG_EXPORT SortOrder : public util::Formattable {
     return lhs.Equals(rhs);
   }
 
+  /// \brief Validates the sort order against a schema.
+  /// \param schema The schema to validate against.
+  /// \return Error status if the sort order is not bound to the schema.
```
wgtmac:
```suggestion
  /// \return Error status if the sort order has any invalid transform.
```

src/iceberg/sort_order.cc
zhjwpku: Yeah, UnboundSortOrder doesn't check, but checkCompatibility should do the check, see [1].

[1] https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/SortOrder.java#L300-L302

src/iceberg/transform.h
```diff
@@ -150,6 +150,11 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
   Result<std::shared_ptr<TransformFunction>> Bind(
       const std::shared_ptr<Type>& source_type) const;
 
+  /// \brief Checks whether this function can be applied to the given Type.
+  /// \param source_type The source type to check.
+  /// \return true if this transform can be applied to the type, false otherwise
+  bool CanTransform(const Type& source_type) const;
```
lishuxu, line 156: Consider changing the parameter to const std::shared_ptr& for consistency?

zhjwpku: I think a const ref of Type here should be fine, the Bind function takes a const shared_ptr& because the returned TransformFunction holds its own shared_ptr.

I'm open to using const shared_ptr& if more people think it's the better approach.

src/iceberg/sort_order.cc
wgtmac: Good catch!

PR #304, opened 2025-11-10
refactor: remove Validator and add Validate methods directly to rest models

HeartLinked:


src/iceberg/catalog/rest/meson.build
```diff
@@ -51,6 +47,6 @@ meson.override_dependency('iceberg-rest', iceberg_rest_dep)
 pkg.generate(iceberg_rest_lib)
 
 install_headers(
-    ['rest_catalog.h', 'types.h', 'json_internal.h', 'validator.h'],
+    ['rest_catalog.h', 'types.h', 'json_internal.h'],
```
wgtmac:
```suggestion
    ['rest_catalog.h', 'types.h'],
```

src/iceberg/table_identifier.cc
```diff
@@ -0,0 +1,33 @@
...
+ * under the License.
+ */
+
+#include "iceberg/table_identifier.h"
+
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+Status TableIdentifier::Validate() const {
```
wgtmac: Why not directly implement this in the header file?

src/iceberg/catalog/rest/types.h
```diff
@@ -19,14 +19,19 @@
...
 #include <memory>
 #include <string>
 #include <unordered_map>
 #include <vector>
 
 #include "iceberg/catalog/rest/iceberg_rest_export.h"
+#include "iceberg/result.h"
 #include "iceberg/table_identifier.h"
 #include "iceberg/type_fwd.h"
+#include "iceberg/util/formatter_internal.h"
```
wgtmac: We cannot add internal header to an installed header. If this is not used, please remove it. Otherwise, we need to add back `types.cc` to move all function definitions to it.

src/iceberg/catalog/rest/types.h
```diff
@@ -46,36 +60,112 @@ struct ICEBERG_REST_EXPORT ErrorModel {
...
+    std::vector<std::string_view> common;
+    std::ranges::set_intersection(sorted_removals, sorted_update_keys,
+                                  std::back_inserter(common));
+
+    if (!common.empty()) {
+      return Invalid(
+          "Invalid namespace update: cannot simultaneously set and remove keys: {}",
+          common);
+    }
+    return {};
```
```suggestion
    for (const auto& key : removals) {
      if (updates.contains(key)) {
        return Invalid("Duplicate key to update and remove: {}", key);
      }
    }
    return {};
```

I just found that `updates` is a map. Then we don't need to be that lengthy.

src/iceberg/catalog/rest/types.h
HeartLinked: I've removed this.

wgtmac (approved):

PR #305, opened 2025-11-10
refactor: make SortOrder constructor private

zhjwpku:
Per offline discussion with @wgtmac

### DIFF
--- a/src/iceberg/test/sort_order_test.cc
+++ b/src/iceberg/test/sort_order_test.cc
@@ -259,10 +259,10 @@ TEST_F(SortOrderTest, MakeUnboundSortOrder) {
   SortField sort_field_invalid(999, Transform::Identity(), SortDirection::kAscending,
                                NullOrder::kFirst);
 
-  auto sort_order =
-      SortOrder::Make(1, std::vector<SortField>{*sort_field1_, sort_field_invalid});
-  ASSERT_THAT(sort_order, IsOk());
-  auto validate_status = sort_order.value()->Validate(*schema_);
+  ICEBERG_UNWRAP_OR_FAIL(
+      auto sort_order,
+      SortOrder::Make(1, std::vector<SortField>{*sort_field1_, sort_field_invalid}));
+  auto validate_status = sort_order->Validate(*schema_);
   EXPECT_THAT(validate_status, IsError(ErrorKind::kInvalidArgument));
   EXPECT_THAT(validate_status,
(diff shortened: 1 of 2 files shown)


src/iceberg/json_internal.h
```diff
@@ -69,8 +69,15 @@ ICEBERG_EXPORT nlohmann::json ToJson(const SortOrder& sort_order);
...
 /// \param json The JSON object representing a `SortOrder`.
+/// \param current_schema The current schema associated with the sort order.
 /// \return An `expected` value containing either a `SortOrder` object or an error. If the
 /// JSON is malformed or missing expected fields, an error will be returned.
+ICEBERG_EXPORT Result<std::unique_ptr<SortOrder>> SortOrderFromJson(
+    const nlohmann::json& json, const std::shared_ptr<Schema>& current_schema);
+
+/// \brief Deserializes a JSON object into a `SortOrder` object.
+///
+/// Same as above but the returned SortOrder will not be validated against any schema.
```
wgtmac: Should we remove this function? I assume schema is always required to deserialize a sort order? I'm fine to keep it just in case we need it in the future.

src/iceberg/test/sort_order_test.cc
```diff
@@ -70,9 +70,9 @@ TEST(SortOrderTest, Basics) {
                         NullOrder::kFirst);
     SortField st_field2(7, identity_transform, SortDirection::kDescending,
                         NullOrder::kFirst);
-    SortOrder sort_order(100, {st_field1, st_field2});
-    ASSERT_EQ(sort_order, sort_order);
-    std::span<const SortField> fields = sort_order.fields();
+    ICEBERG_UNWRAP_OR_FAIL(auto sort_order, SortOrder::Make(100, {st_field1, st_field2}));
```
wgtmac: Should we add some invalid cases like invalid sort_id or transform?

wgtmac (approved):
Generally LGTM. I have left some comments for test cases. Thanks @zhjwpku!

src/iceberg/json_internal.h
zhjwpku: I'm also inclined to remove it, but I've kept it here because FromJsonHelper in json_internal_test.cc uses it. Let me update the test and then remove it.

src/iceberg/test/sort_order_test.cc
zhjwpku: done.

Issue #72, opened 2025-04-14
Add `BoundExpression`


Fokko:
Where https://github.com/apache/iceberg-cpp/pull/58 adds the unbound expressions, I think it would be valuable also to add the BoundExpressions. When an expression is passed in from the Query Engine, we need to bind it to the schema(s) of the Iceberg metadata. This converts an `UnboundExpression` to a `BoundExpression` where the unbound one references a field using a name, and the bound one references a field using its ID.

Issue #71, opened 2025-04-14
Add `RewriteNot` to rewrite `Not` from expressions


Fokko:
Once https://github.com/apache/iceberg-cpp/issues/70 has been resolved, we can add a `RewriteNot` expression to remove `Not` operations from expressions.

See Java: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/RewriteNot.java

PR #311, opened 2025-11-11
feat: add expression visitors

### LINKED ISSUE #70: Add `ExpressionVisitor` to evaluate expressions
Once https://github.com/apache/iceberg-cpp/pull/58 is in, we can add an ExpressionVisitor to evaluate expressions: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/ExpressionVisitors.java

wgtmac:
- Added support to And/Or/Not to create folded expressions.
- Refactored `UnboundPredicate` to use virtual inheritance of `Expression` to facilitate visitor pattern.
- Added template class `ExpressionVisitor` and `BoundVisitor`.
- Implemented `Binder`, `IsBoundVisitor` and `RewriteNot`.

Closes #70 #71 #72

### DIFF
--- a/src/iceberg/expression/expression.cc
+++ b/src/iceberg/expression/expression.cc
@@ -135,7 +135,7 @@ Result<std::shared_ptr<Expression>> Not::MakeFolded(std::shared_ptr<Expression> 
     return not_expr.child();
   }
 
-  return ::iceberg::Not::Make(std::move(child));
+  return Not::Make(std::move(child));
 }
 
 Not::Not(std::shared_ptr<Expression> child) : child_(std::move(child)) {


src/iceberg/expression/expression_visitor.h
```diff
@@ -0,0 +1,314 @@
...
+/// \brief Base visitor for traversing expression trees.
+///
+/// This visitor traverses expression trees in postfix order (children before parents)
+/// and calls appropriate visitor methods for each node. Subclasses can override
+/// specific methods to implement custom behavior.
+///
+/// \tparam R The return type produced by visitor methods
+template <typename R>
+class ICEBERG_EXPORT ExpressionVisitor {
+  using ParamType = typename std::conditional_t<std::is_fundamental_v<R>, R, const R&>;
```
HuaHuaY:
```suggestion
  using ParamType = std::conditional_t<std::is_fundamental_v<R>, R, const R&>;
```

src/iceberg/expression/expression_visitor.h
```diff
@@ -0,0 +1,314 @@
...
+/// from child nodes are passed to parent nodes.
+///
+/// \tparam R The return type produced by the visitor
+/// \tparam V The visitor type (must derive from ExpressionVisitor<R>)
+/// \param expr The expression to traverse
+/// \param visitor The visitor to use for traversal
+/// \return The result produced by the visitor for the root expression node
+template <typename R, typename V>
+  requires std::derived_from<V, ExpressionVisitor<R>>
+Result<R> Visit(const std::shared_ptr<Expression>& expr, V& visitor);
```
HuaHuaY: Is it a duplicate function declaration?

src/iceberg/expression/expression_visitor.h
```diff
@@ -0,0 +1,314 @@
...
+    ICEBERG_DCHECK(pred != nullptr, "UnboundPredicate cannot be null");
+    return NotSupported("Not a bound predicate: {}", pred->ToString());
+  }
+};
+
+/// \brief Traverse an expression tree with a visitor.
+///
+/// This function traverses the given expression tree in postfix order (children
+/// before parents) and calls appropriate visitor methods for each node. Results
+/// from child nodes are passed to parent nodes.
```
HuaHuaY:
```suggestion
/// This function traverses the given expression tree in postorder traversal and calls
/// appropriate visitor methods for each node. Results from child nodes are passed to
/// parent nodes.
```

src/iceberg/expression/expression_visitor.h
```diff
@@ -0,0 +1,314 @@
...
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+/// \brief Base visitor for traversing expression trees.
+///
+/// This visitor traverses expression trees in postfix order (children before parents)
+/// and calls appropriate visitor methods for each node. Subclasses can override
+/// specific methods to implement custom behavior.
```
HuaHuaY:
```suggestion
/// This visitor traverses expression trees in postorder traversal and calls appropriate
/// visitor methods for each node. Subclasses can override specific methods to implement
/// custom behavior.
```

src/iceberg/expression/expression_visitor.h
```diff
@@ -0,0 +1,314 @@
...
+/// \tparam R The return type produced by the visitor
+/// \tparam V The visitor type (must derive from ExpressionVisitor<R>)
+/// \param expr The expression to traverse
+/// \param visitor The visitor to use for traversal
+/// \return The result produced by the visitor for the root expression node
+template <typename R, typename V>
+  requires std::derived_from<V, ExpressionVisitor<R>>
+Result<R> Visit(const std::shared_ptr<Expression>& expr, V& visitor);
+
+template <typename R, typename V>
```
HuaHuaY:
```suggestion
template <typename R, typename V>
```

Is it a duplicate function declaration?

src/iceberg/expression/expression.cc
```diff
@@ -120,6 +120,24 @@ Result<std::unique_ptr<Not>> Not::Make(std::shared_ptr<Expression> child) {
...
+    return True::Instance();
+  }
+
+  // not(not(x)) = x
+  if (child->op() == Expression::Operation::kNot) {
+    const auto& not_expr = internal::checked_cast<const ::iceberg::Not&>(*child);
+    return not_expr.child();
+  }
+
+  return ::iceberg::Not::Make(std::move(child));
```
HuaHuaY:
```suggestion
  return Not::Make(std::move(child));
```

src/iceberg/expression/expression.h
```diff
@@ -137,6 +143,45 @@ class ICEBERG_EXPORT And : public Expression {
...
+  /// \brief Creates a folded And expression from two sub-expressions.
+  ///
+  /// \param left The left operand of the AND expression
+  /// \param right The right operand of the AND expression
+  /// \param args Additional operands of the AND expression
+  /// \return A Result containing a shared pointer to the folded And expression, or an
+  /// error if left or right is nullptr
+  /// \note A folded And expression is an expression that is equivalent to the original
+  /// that is equivalent to the original expression, but with the And operation removed.
+  /// For example, (true and x) = x.
```
HuaHuaY:
```suggestion
  /// \note A folded And expression is an expression that is equivalent to the original
  /// expression, but with the And operation removed. For example, (true and x) = x.
```

src/iceberg/expression/expression.h
```diff
@@ -175,6 +220,44 @@ class ICEBERG_EXPORT Or : public Expression {
...
+  /// \brief Creates a folded Or expression from two sub-expressions.
+  ///
+  /// \param left The left operand of the OR expression
+  /// \param right The right operand of the OR expression
+  /// \param args Additional operands of the OR expression
+  /// \return A Result containing a shared pointer to the folded Or expression, or an
+  /// error if left or right is nullptr
+  /// \note A folded Or expression is an expression that is equivalent to the original
+  /// that is equivalent to the original expression, but with the Or operation removed.
+  /// For example, (false or x) = x.
```
HuaHuaY:
```suggestion
  /// \note A folded Or expression is an expression that is equivalent to the original
  /// expression, but with the Or operation removed. For example, (false or x) = x.
```

src/iceberg/expression/expression.h
```diff
@@ -211,6 +294,16 @@ class ICEBERG_EXPORT Not : public Expression {
   /// \return A Result containing a unique pointer to Not, or an error if child is nullptr
   static Result<std::unique_ptr<Not>> Make(std::shared_ptr<Expression> child);
 
+  /// \brief Creates a folded Not expression from a child expression.
+  ///
+  /// \param child The expression to negate
+  /// \return A Result containing a shared pointer to the folded Not expression, or an
+  /// error if child is nullptr \note A folded Not expression is an expression that is
+  /// equivalent to the original expression, but with the Not operation removed. For
+  /// example, not(not(x)) = x.
```
HuaHuaY:
```suggestion
  /// error if child is nullptr
  /// \note A folded Not expression is an expression that is equivalent to the original
  /// expression, but with the Not operation removed. For example, not(not(x)) = x.
```

HuaHuaY (changes requested):

zhjwpku (approved):
+1, thanks for working on this.

src/iceberg/expression/binder.cc
```diff
@@ -0,0 +1,95 @@
...
+Result<bool> IsBoundVisitor::AlwaysFalse() { return true; }
+
+Result<bool> IsBoundVisitor::Not(bool child_result) { return child_result; }
+
+Result<bool> IsBoundVisitor::And(bool left_result, bool right_result) {
+  return left_result && right_result;
+}
+
+Result<bool> IsBoundVisitor::Or(bool left_result, bool right_result) {
+  return left_result && right_result;
```
gty404, line 84:
```suggestion
  return left_result || right_result;
```

src/iceberg/expression/binder.cc
```diff
@@ -0,0 +1,95 @@
...
+Result<std::shared_ptr<Expression>> Binder::Predicate(
+    const std::shared_ptr<UnboundPredicate>& pred) {
+  ICEBERG_DCHECK(pred != nullptr, "Predicate cannot be null");
+  return pred->Bind(schema_, case_sensitive_);
+}
+
+Result<std::shared_ptr<Expression>> Binder::Predicate(
+    const std::shared_ptr<BoundPredicate>& pred) {
+  ICEBERG_DCHECK(pred != nullptr, "Predicate cannot be null");
+  return InvalidExpression("Found already bound predicate: {}", pred->ToString());
```
gty404, line 64: Does additional commentary need to be added here? If an already bound expression is encountered, choose to report an error rather than directly using it.

src/iceberg/expression/binder.cc
wgtmac: I don't think this is correct because this is not a logical operation. We can only return true if both sides are bound expressions. See java impl for reference: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/Binder.java#L208-L245

src/iceberg/expression/binder.cc
wgtmac: Did you mean that we need to add a comment on why an error is returned here?

PR #316, opened 2025-11-13
refactor: make PartitionSpec ctor private and add factory methods with validation

HeartLinked:


src/iceberg/util/type_util.h
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+#pragma once
+
+#include <functional>
+#include <stack>
+
+#include "iceberg/type.h"
+
+/// \file iceberg/util/type_util.h
```
wgtmac, line 27: We need to install this file in the `meson.build`.

src/iceberg/util/type_util.h
```diff
@@ -0,0 +1,101 @@
...
+/// Utility functions for Iceberg types.
+
+namespace iceberg {
+
+/// \brief Index parent field IDs for all fields in a struct hierarchy.
+/// \param root_struct The root struct type to analyze
+/// \return A map from field ID to its parent struct field ID
+std::unordered_map<int32_t, int32_t> indexParents(const StructType& root_struct) {
```
wgtmac:
```suggestion
ICEBERG_EXPORT static std::unordered_map<int32_t, int32_t> IndexParents(const StructType& root_struct) {
```

src/iceberg/util/type_util.h
```diff
@@ -0,0 +1,101 @@
...
+
+        parent_id_stack.push(key_field.field_id());
+        visit(*key_field.type());
+        parent_id_stack.pop();
+
+        parent_id_stack.push(value_field.field_id());
+        visit(*value_field.type());
+        parent_id_stack.pop();
+        break;
+      }
```
wgtmac:
```suggestion
      case TypeId::kStruct:
      case TypeId::kList:
      case TypeId::kMap: {
        const auto& nested_type = static_cast<const NestedType&>(type);
        for (const auto& field : nested_type.fields()) {
          if (!parent_id_stack.empty()) {
            id_to_parent[field.field_id()] = parent_id_stack.top();
          }

          parent_id_stack.push(field.field_id());
          visit(*field.type());
          parent_id_stack.pop();
        }
        break;
      }
```

src/iceberg/util/type_util.h
```diff
@@ -0,0 +1,101 @@
...
+/// \brief Index parent field IDs for all fields in a struct hierarchy.
+/// \param root_struct The root struct type to analyze
+/// \return A map from field ID to its parent struct field ID
+std::unordered_map<int32_t, int32_t> indexParents(const StructType& root_struct) {
+  std::unordered_map<int32_t, int32_t> id_to_parent;
+  std::stack<int32_t> parent_id_stack;
+
+  // Recursive function to visit and build parent relationships
+  std::function<void(const Type&)> visit;
+  visit = [&](const Type& type) {
```
wgtmac:
```suggestion
  auto visit = [&](const Type& type) -> void {
```

src/iceberg/util/type_util.h
```diff
@@ -0,0 +1,101 @@
...
+    switch (type.type_id()) {
+      case TypeId::kStruct: {
+        const auto& struct_type = static_cast<const StructType&>(type);
+        for (const auto& field : struct_type.fields()) {
+          if (!parent_id_stack.empty()) {
+            id_to_parent[field.field_id()] = parent_id_stack.top();
```
wgtmac, line 53: `field.field_id()` might be invalid or duplicates with other fields if we don't validate the input `root_struct`. If you don't want to change the return type to be `Result<...>`, perhaps add a comment to let readers know what is the expected input. Otherwise we need to validate `field.field_id()` is non-negative and `id_to_parent.emplace(field.field_id(), parent_id_stack.top())` is successful (i.e. not exists already).

src/iceberg/util/type_util.h
wgtmac: We must add `static` or `inline` to avoid duplicate definition. It can happen when two files include this.

src/iceberg/partition_spec.cc
```diff
@@ -104,4 +108,68 @@ bool PartitionSpec::Equals(const PartitionSpec& other) const {
...
+Status PartitionSpec::Validate(const Schema& schema, bool allowMissingFields) const {
+  std::unordered_map<int32_t, int32_t> parents = indexParents(schema);
+  for (const auto& partition_field : fields_) {
+    ICEBERG_ASSIGN_OR_RAISE(auto source_field,
+                            schema.FindFieldById(partition_field.source_id()));
+    // In the case the underlying field is dropped, we cannot check if they are compatible
+    if (allowMissingFields && !source_field.has_value()) {
+      continue;
+    }
+    auto field_transform = partition_field.transform();
```
```suggestion
    const auto& field_transform = partition_field.transform();
```

src/iceberg/partition_spec.cc
```diff
@@ -104,4 +108,68 @@ bool PartitionSpec::Equals(const PartitionSpec& other) const {
...
+    // void transform, see: https://iceberg.apache.org/spec/#partition-transforms. We
+    // don't care about the source type since a VoidTransform is always compatible and
+    // skip the checks
+    if (field_transform->transform_type() != TransformType::kVoid) {
+      if (!source_field.has_value()) {
+        return InvalidArgument("Cannot find source column for partition field: {}",
+                               partition_field);
+      }
+      const auto& source_type = source_field.value().get().type();
+      if (!partition_field.transform()->CanTransform(*source_type)) {
```
```suggestion
      if (!field_transform->CanTransform(*source_type)) {
```

src/iceberg/partition_spec.cc
```diff
@@ -104,4 +108,68 @@ bool PartitionSpec::Equals(const PartitionSpec& other) const {
...
+        return InvalidArgument("Invalid source type {} for transform {}",
+                               source_type->ToString(),
+                               partition_field.transform()->ToString());
```
```suggestion
                               field_transform->ToString());
```

src/iceberg/partition_spec.cc
```diff
@@ -104,4 +108,68 @@ bool PartitionSpec::Equals(const PartitionSpec& other) const {
...
+    }
+  }
+  return {};
+}
+
+Result<std::unique_ptr<PartitionSpec>> PartitionSpec::Make(
+    const Schema& schema, int32_t spec_id, std::vector<PartitionField> fields,
+    std::optional<int32_t> last_assigned_field_id, bool allowMissingFields) {
+  auto partition_spec =
+      std::make_unique<PartitionSpec>(spec_id, std::move(fields), last_assigned_field_id);
```
wgtmac: Why not making ctors private and change downstreams all at once?

src/iceberg/test/partition_spec_test.cc
```diff
@@ -152,4 +152,174 @@ TEST(PartitionSpecTest, PartitionTypeTest) {
   EXPECT_EQ(pt_field3, partition_type.value()->fields()[2]);
 }
 
+TEST(PartitionSpecTest, MissingSourceColumn) {
+  SchemaField field1(1, "id", int64(), false);
```
wgtmac: nit: use SchemaField::MakeRequired and use a one-liner to create the Schema.

src/iceberg/partition_spec.h
```diff
@@ -80,6 +80,38 @@ class ICEBERG_EXPORT PartitionSpec : public util::Formattable {
...
+  /// \param fields The partition fields.
+  /// \param last_assigned_field_id The last assigned field ID assigned to ensure new
+  /// fields get unique IDs.
+  /// \param allowMissingFields Whether to skip validation for partition fields whose
+  /// source columns have been dropped from the schema.
+  /// \return A Result containing the partition spec or an error.
+  static Result<std::unique_ptr<PartitionSpec>> Make(
+      const Schema& schema, int32_t spec_id, std::vector<PartitionField> fields,
+      std::optional<int32_t> last_assigned_field_id = std::nullopt,
+      bool allowMissingFields = false);
```
```suggestion
      bool allow_missing_fields = false);
```

src/iceberg/test/partition_spec_test.cc
```diff
@@ -152,4 +152,174 @@ TEST(PartitionSpecTest, PartitionTypeTest) {
...
+  SchemaField field2(2, "ts", timestamp(), false);
+  SchemaField field3(6, "s", string(), false);
+  Schema schema({field1, field2, field3}, Schema::kInitialSchemaId);
+
+  // Try to create partition field with non-existent source column ID 999
+  PartitionField pt_field_invalid(999, 1000, "missing_partition", Transform::Identity());
+
+  auto result = PartitionSpec::Make(schema, 1, {pt_field_invalid});
```
wgtmac: Please also check `check_missing_fields` work

src/iceberg/test/partition_spec_test.cc
```diff
@@ -152,4 +152,174 @@ TEST(PartitionSpecTest, PartitionTypeTest) {
...
+  EXPECT_THAT(result.error().message,
+              ::testing::HasSubstr("Cannot find source column for partition field"));
+}
+
+TEST(PartitionSpecTest, InvalidTransformForType) {
+  // Test Day transform on string type (should fail)
+  SchemaField field_string(6, "s", string(), false);
+  Schema schema_string({field_string}, Schema::kInitialSchemaId);
+
+  PartitionField pt_field_invalid(6, 1005, "s_day", Transform::Day());
```
wgtmac, line 163: Can we also check that void transform does not error out?

src/iceberg/test/partition_spec_test.cc
```diff
@@ -152,4 +152,174 @@ TEST(PartitionSpecTest, PartitionTypeTest) {
...
+
+  auto result_string = PartitionSpec::Make(schema_string, 1, {pt_field_invalid});
+  EXPECT_FALSE(result_string.has_value());
+  EXPECT_THAT(result_string.error().message, ::testing::HasSubstr("Invalid source type"));
+}
+
+TEST(PartitionSpecTest, SourceIdNotFound) {
```
wgtmac, line 175: Isn't it duplicate with `TEST(PartitionSpecTest, MissingSourceColumn)`?

src/iceberg/test/partition_spec_test.cc
```diff
@@ -152,4 +152,174 @@ TEST(PartitionSpecTest, PartitionTypeTest) {
...
+  SchemaField field1(1, "id", int64(), false);
+  SchemaField field2(2, "ts", timestamp(), false);
+  SchemaField field3(6, "s", string(), false);
+  Schema schema({field1, field2, field3}, Schema::kInitialSchemaId);
+
+  // Try to create partition field with non-existent source column ID 999
+  PartitionField pt_field_invalid(999, 1000, "missing_partition", Transform::Identity());
+
+  auto result = PartitionSpec::Make(schema, 1, {pt_field_invalid});
+  EXPECT_FALSE(result.has_value());
```
wgtmac: Let's verify the error kind. Same for all below.

src/iceberg/test/partition_spec_test.cc
```diff
@@ -152,4 +152,174 @@ TEST(PartitionSpecTest, PartitionTypeTest) {
...
+      std::make_shared<StructType>(std::vector<SchemaField>{field1, field2});
+  SchemaField outer_struct(11, "MyStruct", struct_type, false);
+
+  Schema schema({outer_struct}, Schema::kInitialSchemaId);
+
+  // Partition on a field within the struct (id field with ID 1)
+  PartitionField pt_field(1, 1000, "id_partition", Transform::Identity());
+
+  auto result = PartitionSpec::Make(schema, 1, {pt_field});
+  EXPECT_TRUE(result.has_value()) << result.error().message;
```
wgtmac:
```suggestion
  EXPECT_THAT(result, IsOK());
```

wgtmac: Same for below.

src/iceberg/test/partition_spec_test.cc
wgtmac: Same for below. The point is to make test shorter.

src/iceberg/util/type_util.h
wgtmac: Please add a test for this. Perhaps in the `type_test.cc` is fine.

wgtmac (changes requested):

src/iceberg/json_internal.cc
```diff
@@ -542,9 +543,17 @@ Result<std::unique_ptr<PartitionSpec>> PartitionSpecFromJson(
...
   }
-  // TODO(Li Feiyang):use a new PartitionSpec::Make to find the source field of each
-  // partition field from schema and then verify it
-  return std::make_unique<PartitionSpec>(spec_id, std::move(partition_fields));
+
+  std::unique_ptr<PartitionSpec> spec;
+  if (default_spec_id == spec_id) {
+    ICEBERG_ASSIGN_OR_RAISE(
+        spec, PartitionSpec::Make(*schema, spec_id, std::move(partition_fields),
+                                  true /*allow_missing_fields*/));
```
wgtmac:
```suggestion
        spec, PartitionSpec::Make(*schema, spec_id, std::move(partition_fields),
                                  /*allow_missing_fields=*/false));
```

It is illegal for the current spec to have a missing source field. You can verify it by checking the Java code below that `bind(schema)` is called.

```java
  public PartitionSpec bind(Schema schema) {
    return copyToBuilder(schema).build();
  }

  public PartitionSpec bind(Schema schema, boolean ignoreMissingFields) {
    return copyToBuilder(schema).build(ignoreMissingFields);
  }
```

src/iceberg/json_internal.cc
```diff
@@ -542,9 +543,17 @@ Result<std::unique_ptr<PartitionSpec>> PartitionSpecFromJson(
...
+  } else {
+    ICEBERG_ASSIGN_OR_RAISE(spec,
+                            PartitionSpec::Make(spec_id, std::move(partition_fields)));
```
wgtmac:
```suggestion
    ICEBERG_ASSIGN_OR_RAISE(spec, PartitionSpec::Make(*schema, spec_id,
      std::move(partition_fields), /*allow_missing_fields=*/true));
```

The Java code does not call `checkCompatibility` in this case but `PartitionSpec.Builder.copyToBuilder(Schema schema)` is still called as below where non-missing field has to validate its transform:

```java
  private PartitionSpec.Builder copyToBuilder(Schema schema) {
    PartitionSpec.Builder builder = PartitionSpec.builderFor(schema).withSpecId(specId);

    for (UnboundPartitionField field : fields) {
      Type fieldType = schema.findType(field.sourceId);
      Transform<?, ?> transform;
      if (fieldType != null) {
        transform = Transforms.fromString(fieldType, field.transform.toString());
      } else {
        transform = Transforms.fromString(field.transform.toString());
      }
      if (field.partitionId != null) {
        builder.add(field.sourceId, field.partitionId, field.name, transform);
      } else {
        builder.add(field.sourceId, field.name, transform);
      }
    }

    return builder;
  }
```

src/iceberg/json_internal.cc
```diff
@@ -917,10 +926,10 @@ Status ParsePartitionSpecs(const nlohmann::json& json, int8_t format_version,
...
     }
 
-    // TODO(Li Feiyang):use a new PartitionSpec::Make to find the source field of each
-    // partition field from schema and then verify it
-    auto spec =
-        std::make_unique<PartitionSpec>(PartitionSpec::kInitialSpecId, std::move(fields));
+    // Create partition spec with schema validation
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto spec, PartitionSpec::Make(*current_schema, PartitionSpec::kInitialSpecId,
+                                       std::move(fields), true /*allow_missing_fields*/));
```
wgtmac:
```suggestion
                                       std::move(fields), /*allow_missing_fields=*/false));
```

ditto

PR #317, opened 2025-11-14
feat: add partition summary and write added/existing/deleted entries to manifest writer

zhjwpku:
including AddEntry, AddExistingEntry, AddDeleteEntry

### DIFF
--- a/src/iceberg/manifest_adapter.cc
+++ b/src/iceberg/manifest_adapter.cc
@@ -27,7 +27,6 @@
 #include "iceberg/arrow/nanoarrow_status_internal.h"
 #include "iceberg/manifest_entry.h"
 #include "iceberg/manifest_list.h"
-#include "iceberg/partition_summary_internal.h"
 #include "iceberg/result.h"
 #include "iceberg/schema.h"
 #include "iceberg/util/checked_cast.h"
@@ -321,16 +320,13 @@ Status ManifestEntryAdapter::AppendDataFile(
           ICEBERG_NANOARROW_RETURN_UNEXPECTED(ArrowArrayAppendNull(child_array, 1));
         }
         break;
-      case 142: {
-        // first_row_id (optional int64)
-        ICEBERG_ASSIGN_OR_RAISE(auto first_row_id, GetFirstRowId(file));
-        if (first_row_id.has_value()) {
-          ICEBERG_RETURN_UNEXPECTED(AppendField(child_array, first_row_id.value()));
+      case 142:  // first_row_id (optional int64)
+        if (file.first_row_id.has_value()) {
+          ICEBERG_RETURN_UNEXPECTED(AppendField(child_array, file.first_row_id.value()));
         } else {
           ICEBERG_NANOARROW_RETURN_UNEXPECTED(ArrowArrayAppendNull(child_array, 1));
         }
         break;
-      }
       case 143: {
         // referenced_data_file (optional string)
         ICEBERG_ASSIGN_OR_RAISE(auto referenced_data_file, GetReferenceDataFile(file));
@@ -395,8 +391,6 @@ Status ManifestEntryAdapter::AppendInternal(const ManifestEntry& entry) {
   if (entry.data_file == nullptr) [[unlikely]] {
     return InvalidManifest("Missing required data_file field from manifest entry.");
   }
-
-  ICEBERG_RETURN_UNEXPECTED(partition_summary_->Update(entry.data_file->partition));
 
   const auto& fields = manifest_schema_->fields();
   for (size_t i = 0; i < fields.size(); i++) {
--- a/src/iceberg/manifest_adapter.h
+++ b/src/iceberg/manifest_adapter.h
@@ -76,13 +76,7 @@ class ICEBERG_EXPORT ManifestEntryAdapter : public ManifestAdapter {
 
   const std::shared_ptr<PartitionSpec>& partition_spec() const { return partition_spec_; }
 
-  const std::unique_ptr<PartitionSummary>& partition_summary() const {
-    return partition_summary_;
-  }
-
-  /// \brief Create a ManifestFile object without setting file metadata, such as
-  /// location, file size, key metadata, etc.
-  Result<ManifestFile> ToManifestFile() const;
+  const std::shared_ptr<StructType>& partition_type() const { return partition_type_; }
 
  protected:
   Status AppendInternal(const ManifestEntry& entry);
@@ -106,9 +100,9 @@ class ICEBERG_EXPORT ManifestEntryAdapter : public ManifestAdapter {
   std::optional<int64_t> snapshot_id_;
   std::shared_ptr<PartitionSpec> partition_spec_;
   std::shared_ptr<Schema> current_schema_;
+  std::shared_ptr<StructType> partition_type_;
   std::shared_ptr<Schema> manifest_schema_;
   const ManifestContent content_;
-  std::unique_ptr<PartitionSummary> partition_summary_;
 };
 
 /// \brief Adapter for appending a list of `ManifestFile`s to an `ArrowArray`.
--- a/src/iceberg/manifest_entry.h
+++ b/src/iceberg/manifest_entry.h
@@ -268,8 +268,7 @@ struct ICEBERG_EXPORT DataFile {
 
   bool operator==(const DataFile& other) const = default;
 
-  std::shared_ptr<DataFile> Clone() const;
-
+  /// \brief Get the schema of the data file with the given partition type.
   static std::shared_ptr<StructType> Type(std::shared_ptr<StructType> partition_type);
 };
 
@@ -316,6 +315,10 @@ struct ICEBERG_EXPORT ManifestEntry {
   ManifestEntry AsAdded() const {
     ManifestEntry copy = *this;
     copy.status = ManifestStatus::kAdded;
+    if (copy.data_file->first_row_id.has_value()) {
+      copy.data_file = std::make_unique<DataFile>(*copy.data_file);
+      copy.data_file->first_row_id = std::nullopt;
+    }
     return copy;
   }
 
--- a/src/iceberg/manifest_reader_internal.cc
+++ b/src/iceberg/manifest_reader_internal.cc
@@ -328,7 +328,7 @@ Status ParseLiteral(ArrowArrayView* view_of_partition, int64_t row_idx,
 }
 
 Status ParseDataFile(const std::shared_ptr<StructType>& data_file_schema,
-                     ArrowArrayView* view_of_column,
+                     ArrowArrayView* view_of_column, std::optional<int64_t>& first_row_id,
                      std::vector<ManifestEntry>& manifest_entries) {
   if (view_of_column->storage_type != ArrowType::NANOARROW_TYPE_STRUCT) {
     return InvalidManifest("DataFile field should be a struct.");
@@ -432,10 +432,25 @@ Status ParseDataFile(const std::shared_ptr<StructType>& data_file_schema,
         PARSE_PRIMITIVE_FIELD(manifest_entries[row_idx].data_file->sort_order_id,
                               view_of_file_field, int32_t);
         break;
-      case 16:
+      case 16: {
         PARSE_PRIMITIVE_FIELD(manifest_entries[row_idx].data_file->first_row_id,
                               view_of_file_field, int64_t);
+        if (first_row_id.has_value()) {
+          std::ranges::for_each(manifest_entries, [&first_row_id](ManifestEntry& entry) {
+            if (entry.status != ManifestStatus::kDeleted &&
+                !entry.data_file->first_row_id.has_value()) {
+              entry.data_file->first_row_id = first_row_id.value();
+              first_row_id = first_row_id.value() + entry.data_file->record_count;
+            }
+          });
+        } else {
+          // data file's first_row_id is null when the manifest's first_row_id is null
+          std::ranges::for_each(manifest_entries, [](ManifestEntry& entry) {
+            entry.data_file->first_row_id = std::nullopt;
+          });
+        }
         break;
+      }
       case 17:
         PARSE_STRING_FIELD(manifest_entries[row_idx].data_file->referenced_data_file,
                            view_of_file_field);
@@ -455,9 +470,9 @@ Status ParseDataFile(const std::shared_ptr<StructType>& data_file_schema,
   return {};
 }
 
-Result<std::vector<ManifestEntry>> ParseManifestEntry(ArrowSchema* schema,
-                                                      ArrowArray* array_in,
-                                                      const Schema& iceberg_schema) {
+Result<std::vector<ManifestEntry>> ParseManifestEntry(
+    ArrowSchema* schema, ArrowArray* array_in, const Schema& iceberg_schema,
+    std::optional<int64_t>& first_row_id) {
   if (schema->n_children != array_in->n_children) {
     return InvalidManifest("Columns size not match between schema:{} and array:{}",
                            schema->n_children, array_in->n_children);
@@ -512,8 +527,8 @@ Result<std::vector<ManifestEntry>> ParseManifestEntry(ArrowSchema* schema,
       case 4: {
         auto data_file_schema =
             internal::checked_pointer_cast<StructType>(field.value()->get().type());
-        ICEBERG_RETURN_UNEXPECTED(
-            ParseDataFile(data_file_schema, view_of_column, manifest_entries));
+        ICEBERG_RETURN_UNEXPECTED(ParseDataFile(data_file_schema, view_of_column,
+                                                first_row_id, manifest_entries));
         break;
       }
       default:
@@ -533,7 +548,7 @@ Result<std::vector<ManifestEntry>> ManifestReaderImpl::Entries() const {
       internal::ArrowArrayGuard array_guard(&result.value());
       ICEBERG_ASSIGN_OR_RAISE(
           auto parse_result,
-          ParseManifestEntry(&arrow_schema, &result.value(), *schema_));
+          ParseManifestEntry(&arrow_schema, &result.value(), *schema_, first_row_id_));
       manifest_entries.insert(manifest_entries.end(),
                               std::make_move_iterator(parse_result.begin()),
                               std::make_move_iterator(parse_result.end()));
--- a/src/iceberg/manifest_writer.cc
+++ b/src/iceberg/manifest_writer.cc
@@ -34,20 +34,38 @@
 
 namespace iceberg {
 
+ManifestWriter::ManifestWriter(std::unique_ptr<Writer> writer,
+                               std::unique_ptr<ManifestEntryAdapter> adapter,
+                               std::string_view manifest_location,
+                               std::optional<int64_t> first_row_id)
+    : writer_(std::move(writer)),
+      adapter_(std::move(adapter)),
+      manifest_location_(manifest_location),
+      first_row_id_(first_row_id),
+      partition_summary_(
+          std::make_unique<PartitionSummary>(*adapter_->partition_type())) {}
+
+ManifestWriter::~ManifestWriter() = default;
+
 Status ManifestWriter::WriteAddedEntry(std::shared_ptr<DataFile> file,
                                        std::optional<int64_t> data_sequence_number) {
-  ManifestEntry added;
-  added.status = ManifestStatus::kAdded;
-  added.snapshot_id = adapter_->snapshot_id();
-  if (file->first_row_id.has_value()) {
-    // suppress first_row_id if set
-    added.data_file = file->Clone();
-    added.data_file->first_row_id = std::nullopt;
-  } else {
-    added.data_file = std::move(file);
+  if (!file) [[unlikely]] {
+    return InvalidArgument("Data file cannot be null");
   }
-  added.sequence_number = data_sequence_number;
-  added.file_sequence_number = std::nullopt;
+
+  ManifestEntry added{
+      .status = ManifestStatus::kAdded,
+      .snapshot_id = adapter_->snapshot_id(),
+      .sequence_number = data_sequence_number,
+      .file_sequence_number = std::nullopt,
+      .data_file = std::move(file),
+  };
+
+  // Suppress first_row_id for added entries
+  if (added.data_file->first_row_id.has_value()) {
+    added.data_file = std::make_unique<DataFile>(*added.data_file);
+    added.data_file->first_row_id = std::nullopt;
+  }
 
   return WriteEntry(added);
 }
@@ -55,19 +73,12 @@ Status ManifestWriter::WriteAddedEntry(std::shared_ptr<DataFile> file,
 Status ManifestWriter::WriteAddedEntry(const ManifestEntry& entry) {
   // Update the entry status to `Added`
   auto added = entry.AsAdded();
-  // Set the snapshot id to the current snapshot id
   added.snapshot_id = adapter_->snapshot_id();
-  if (added.data_file->first_row_id.has_value()) {
-    // suppress first_row_id if set
-    added.data_file = added.data_file->Clone();
-    added.data_file->first_row_id = std::nullopt;
-  }
-  // Set the sequence number to nullopt if it is invalid(smaller than 0)
+  // Set the sequence number to nullopt if it is invalid (smaller than 0)
   if (added.sequence_number.has_value() &&
       added.sequence_number.value() < TableMetadata::kInitialSequenceNumber) {
     added.sequence_number = std::nullopt;
   }
-  // Set the file sequence number to nullopt
   added.file_sequence_number = std::nullopt;
 
   return WriteEntry(added);
@@ -77,6 +88,10 @@ Status ManifestWriter::WriteExistingEntry(std::shared_ptr<DataFile> file,
                                           int64_t file_snapshot_id,
                                           int64_t data_sequence_number,
                                           std::optional<int64_t> file_sequence_number) {
+  if (!file) [[unlikely]] {
+    return InvalidArgument("Data file cannot be null");
+  }
+
   ManifestEntry existing;
   existing.status = ManifestStatus::kExisting;
   existing.snapshot_id = file_snapshot_id;
@@ -89,13 +104,16 @@ Status ManifestWriter::WriteExistingEntry(std::shared_ptr<DataFile> file,
 
 Status ManifestWriter::WriteExistingEntry(const ManifestEntry& entry) {
   // Update the entry status to `Existing`
-  auto existing = entry.AsExisting();
-  return WriteEntry(existing);
+  return WriteEntry(entry.AsExisting());
 }
 
 Status ManifestWriter::WriteDeletedEntry(std::shared_ptr<DataFile> file,
                                          int64_t data_sequence_number,
                                          std::optional<int64_t> file_sequence_number) {
+  if (!file) [[unlikely]] {
+    return InvalidArgument("Data file cannot be null");
+  }
+
   ManifestEntry deleted;
   deleted.status = ManifestStatus::kDeleted;
   deleted.snapshot_id = adapter_->snapshot_id();
@@ -116,12 +134,18 @@ Status ManifestWriter::WriteDeletedEntry(const ManifestEntry& entry) {
 }
 
 Status ManifestWriter::WriteEntry(const ManifestEntry& entry) {
+  if (!entry.data_file) [[unlikely]] {
+    return InvalidArgument("Data file cannot be null");
+  }
+
   ICEBERG_RETURN_UNEXPECTED(CheckDataFile(*entry.data_file));
   if (adapter_->size() >= kBatchSize) {
     ICEBERG_ASSIGN_OR_RAISE(auto array, adapter_->FinishAppending());
     ICEBERG_RETURN_UNEXPECTED(writer_->Write(array));
     ICEBERG_RETURN_UNEXPECTED(adapter_->StartAppending());
   }
+
+  ICEBERG_RETURN_UNEXPECTED(partition_summary_->Update(entry.data_file->partition));
 
   // update statistics
   switch (entry.status) {
@@ -198,29 +222,29 @@ Result<ManifestFile> ManifestWriter::ToManifestFile() const {
     return Invalid("Cannot get ManifestFile before closing the writer.");
   }
 
-  ManifestFile manifest_file;
-  manifest_file.partition_spec_id = adapter_->partition_spec()->spec_id();
-  manifest_file.content = adapter_->content();
-  // sequence_number and min_sequence_number with kInvalidSequenceNumber will be
-  // replace with real sequence number in `ManifestListWriter`.
-  manifest_file.sequence_number = TableMetadata::kInvalidSequenceNumber;
-  manifest_file.min_sequence_number =
-      min_sequence_number_.value_or(TableMetadata::kInvalidSequenceNumber);
+  ICEBERG_ASSIGN_OR_RAISE(auto partitions, partition_summary_->Summaries());
+  ICEBERG_ASSIGN_OR_RAISE(auto manifest_length, writer_->length());
 
-  manifest_file.existing_files_count = existing_files_count_;
-  manifest_file.added_files_count = add_files_count_;
-  manifest_file.existing_files_count = existing_files_count_;
-  manifest_file.deleted_files_count = delete_files_count_;
-  manifest_file.added_rows_count = add_rows_count_;
-  manifest_file.existing_rows_count = existing_rows_count_;
-  manifest_file.deleted_rows_count = delete_rows_count_;
-  ICEBERG_ASSIGN_OR_RAISE(manifest_file.partitions,
-                          adapter_->partition_summary()->Summaries());
-  manifest_file.manifest_path = manifest_location_;
-  ICEBERG_ASSIGN_OR_RAISE(manifest_file.manifest_length, writer_->length());
-  manifest_file.added_snapshot_id =
-      adapter_->snapshot_id().value_or(Snapshot::kInvalidSnapshotId);
-  return manifest_file;
+  return ManifestFile{
+      .manifest_path = manifest_location_,
+      .manifest_length = manifest_length,
+      .partition_spec_id = adapter_->partition_spec()->spec_id(),
+      .content = adapter_->content(),
+      // sequence_number and min_sequence_number with kInvalidSequenceNumber will be
+      // replace with real sequence number in `ManifestListWriter`.
+      .sequence_number = TableMetadata::kInvalidSequenceNumber,
+      .min_sequence_number =
+          min_sequence_number_.value_or(TableMetadata::kInvalidSequenceNumber),
+      .added_snapshot_id = adapter_->snapshot_id().value_or(Snapshot::kInvalidSnapshotId),
+      .added_files_count = add_files_count_,
+      .existing_files_count = existing_files_count_,
+      .deleted_files_count = delete_files_count_,
+      .added_rows_count = add_rows_count_,
+      .existing_rows_count = existing_rows_count_,
+      .deleted_rows_count = delete_rows_count_,
+      .partitions = std::move(partitions),
+      .first_row_id = first_row_id_,
+  };
 }
 
 Result<std::unique_ptr<Writer>> OpenFileWriter(
@@ -271,7 +295,7 @@ Result<std::unique_ptr<ManifestWriter>> ManifestWriter::MakeV1Writer(
       OpenFileWriter(manifest_location, std::move(schema), std::move(file_io),
                      adapter->metadata(), "manifest_entry"));
   return std::make_unique<ManifestWriter>(std::move(writer), std::move(adapter),
-                                          manifest_location);
+                                          manifest_location, std::nullopt);
 }
 
 Result<std::unique_ptr<ManifestWriter>> ManifestWriter::MakeV2Writer(
@@ -301,7 +325,7 @@ Result<std::unique_ptr<ManifestWriter>> ManifestWriter::MakeV2Writer(
       OpenFileWriter(manifest_location, std::move(schema), std::move(file_io),
                      adapter->metadata(), "manifest_entry"));
   return std::make_unique<ManifestWriter>(std::move(writer), std::move(adapter),
-                                          manifest_location);
+                                          manifest_location, std::nullopt);
 }
 
 Result<std::unique_ptr<ManifestWriter>> ManifestWriter::MakeV3Writer(
@@ -333,7 +357,7 @@ Result<std::unique_ptr<ManifestWriter>> ManifestWriter::MakeV3Writer(
       OpenFileWriter(manifest_location, std::move(schema), std::move(file_io),
                      adapter->metadata(), "manifest_entry"));
   return std::make_unique<ManifestWriter>(std::move(writer), std::move(adapter),
-                                          manifest_location);
+                                          manifest_location, first_row_id);
 }
 
 Status ManifestListWriter::Add(const ManifestFile& file) {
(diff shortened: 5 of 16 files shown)


src/iceberg/partition_summary_internal.h
```diff
@@ -0,0 +1,65 @@
...
+  bool contains_null_{false};
+  bool contains_nan_{false};
+  std::optional<Literal> lower_bound_;
+  std::optional<Literal> upper_bound_;
+};
+
+class PartitionSummary {
+ public:
+  /// \brief Update the partition summary with partition.
+  Status Update(const std::vector<Literal>& partition);
```
wgtmac:
```suggestion
  /// \brief Update the partition summary with partition values.
  Status Update(const std::vector<Literal>& partition_values);
```

`partition` was a little bit vague in its meaning.

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/partition_summary_internal.h"
```
wgtmac: Why adding `_internal` suffix to it? Isn't it a public api?

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+#include <memory>
+
+#include "iceberg/expression/literal.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+Status PartitionFieldStats::Update(const Literal& value) {
+  if (type_->type_id() != value.type()->type_id()) {
```
wgtmac: I think we need to support the allowed schema evolution rule here. Otherwise we end up with losing partition stats if field change like `int`->`long` or `float`->`double` happens.

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+  }
+  return {};
+}
+
+PartitionFieldSummary PartitionFieldStats::Finish() const {
+  PartitionFieldSummary summary;
+  summary.contains_null = contains_null_;
+  summary.contains_nan = contains_nan_;
+  if (lower_bound_) {
+    summary.lower_bound = lower_bound_->Serialize().value();
```
wgtmac: We need to handle any possible error returned by `lower_bound_->Serialize()` and `upper_bound_->Serialize()` below. Perhaps changing the function to return `Result< PartitionFieldSummary>` as well?

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+
+std::vector<PartitionFieldSummary> PartitionSummary::Summaries() const {
+  std::vector<PartitionFieldSummary> summaries;
+  for (const auto& field_stat : field_stats_) {
+    summaries.push_back(field_stat.Finish());
+  }
+  return summaries;
+}
+
+Result<std::unique_ptr<PartitionSummary>> PartitionSummary::Make(
```
wgtmac: If there is no validation in the `Make` function, we don't need to add it. It is more convenient to directly use the constructor.

This is not a strong opinion. Perhaps we can validate the partition type in the future.

src/iceberg/partition_summary.cc
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  for (size_t i = 0; i < partition.size(); i++) {
+    ICEBERG_RETURN_UNEXPECTED(field_stats_[i].Update(partition[i]));
+  }
+  return {};
+}
+
+std::vector<PartitionFieldSummary> PartitionSummary::Summaries() const {
+  std::vector<PartitionFieldSummary> summaries;
```
wgtmac, line 94: nit: call reserve on `summaries`.

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+  }
+  if (upper_bound_) {
+    summary.upper_bound = upper_bound_->Serialize().value();
+  }
+  return summary;
+}
+
+Status PartitionSummary::Update(const std::vector<Literal>& partition) {
+  if (partition.size() != field_stats_.size()) {
+    return InvalidArgument("partition size does not match field stats size");
```
wgtmac: Let's also include the value in the error message.

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  for (size_t i = 0; i < partition.size(); i++) {
+    ICEBERG_RETURN_UNEXPECTED(field_stats_[i].Update(partition[i]));
```
wgtmac: Perhaps we can call `partition[i].CastTo()` to handle the schema evolution here so `PartitionFieldStats::Update` can enforce strong type equality? This looks simpler.

src/iceberg/manifest_entry.h
```diff
@@ -315,6 +306,17 @@ struct ICEBERG_EXPORT ManifestEntry {
   inline static const SchemaField kFileSequenceNumber =
       SchemaField::MakeOptional(4, "file_sequence_number", iceberg::int64());
 
+  /// \brief Check if this manifest entry is deleted.
+  constexpr bool IsAlive() const {
+    return status == ManifestStatus::kAdded || status == ManifestStatus::kExisting;
+  }
+
+  /// \brief Create a copy of this manifest entry.
+  ManifestEntry Copy() const {
```
wgtmac: Do we actually need this function? Can we just use `operator=`?

src/iceberg/manifest_list.h
```diff
@@ -198,6 +198,12 @@ struct ICEBERG_EXPORT ManifestFile {
   bool operator==(const ManifestFile& other) const = default;
 
   static const std::shared_ptr<Schema>& Type();
+
+  /// \brief Create a copy of this manifest file.
+  ManifestFile Copy() const {
```
wgtmac: Same question for `operator=`

src/iceberg/manifest_reader_internal.cc
```diff
@@ -367,13 +367,16 @@ Status ParseDataFile(const std::shared_ptr<StructType>& data_file_schema,
           return InvalidManifest("Field:{} should be a struct.", field_name);
         }
         if (view_of_file_field->n_children > 0) {
-          auto view_of_partition = view_of_file_field->children[0];
-          for (int64_t row_idx = 0; row_idx < view_of_partition->length; row_idx++) {
-            if (ArrowArrayViewIsNull(view_of_partition, row_idx)) {
-              break;
+          for (int64_t partition_idx = 0; partition_idx < view_of_file_field->n_children;
+               partition_idx++) {
+            auto view_of_partition = view_of_file_field->children[partition_idx];
```
wgtmac, line 372: This is the fix to a bug that multiple partition fields are not correctly handled?

src/iceberg/manifest_writer.h
```diff
@@ -54,9 +69,15 @@ class ICEBERG_EXPORT ManifestWriter {
   /// \brief Close writer and flush to storage.
   Status Close();
 
+  /// \brief Get the metrics of written manifest file.
```
wgtmac:
```suggestion
  /// \brief Get the metrics of written manifest file.
  /// \note Only valid after the file is closed.
```

src/iceberg/manifest_writer.h
```diff
@@ -54,9 +69,15 @@ class ICEBERG_EXPORT ManifestWriter {
...
+  std::optional<Metrics> metrics() const;
+
   /// \brief Get the content of the manifest.
   ManifestContent content() const;
 
+  /// \brief Get the ManifestFile object.
+  Result<ManifestFile> ToManifestFile() const;
```
wgtmac, line 118: I think we should be consistent for `metrics()` and `ToManifestFile()`. What about using the following:

```cpp
Result<Metrics> metrics() const;
Result<ManifestFile> ToManifestFile() const;  // unchanged
```

We just return invalid state error when file is not closed.

src/iceberg/metrics.h
```diff
@@ -32,12 +32,12 @@ namespace iceberg {
 /// \brief Iceberg file format metrics
 struct ICEBERG_EXPORT Metrics {
   int64_t row_count = 0;
```
wgtmac:
```suggestion
  std::optional<int64_t> row_count = 0;
```

Should we make it optional?

src/iceberg/manifest_writer.h
wgtmac: Perhaps we should also change `FileWriter::metrics` to return `Result<Metrics>` as well and return not supported error for some writers.

src/iceberg/manifest_writer.h
```diff
@@ -103,6 +124,10 @@ class ICEBERG_EXPORT ManifestWriter {
   static constexpr int64_t kBatchSize = 1024;
   std::unique_ptr<Writer> writer_;
   std::unique_ptr<ManifestEntryAdapter> adapter_;
+  bool closed_{false};
+  std::string manifest_location_;
+  // TODO(zhjwpku): consider passing key metadata when makeing ManifestWriter.
+  std::vector<uint8_t> key_metadata_;
```
wgtmac: We don't need this in the near future.

src/iceberg/manifest_writer.cc
```diff
@@ -171,7 +211,8 @@ Status ManifestListWriter::Add(const ManifestFile& file) {
     ICEBERG_RETURN_UNEXPECTED(writer_->Write(array));
     ICEBERG_RETURN_UNEXPECTED(adapter_->StartAppending());
   }
-  return adapter_->Append(file);
+  auto copy = file.Copy();
+  return adapter_->Append(copy);
```
wgtmac:
```suggestion
  return adapter_->Append(file);
```

Please revert change for `ManifestListWriter`.

src/iceberg/manifest_adapter.h
```diff
@@ -61,18 +60,45 @@ class ICEBERG_EXPORT ManifestAdapter {
...
 class ICEBERG_EXPORT ManifestEntryAdapter : public ManifestAdapter {
  public:
-  ManifestEntryAdapter(std::shared_ptr<PartitionSpec> partition_spec,
+  ManifestEntryAdapter(std::optional<int64_t> snapshot_id_,
+                       std::shared_ptr<PartitionSpec> partition_spec,
                        std::shared_ptr<Schema> current_schema, ManifestContent content);
 
   ~ManifestEntryAdapter() override;
 
-  virtual Status Append(const ManifestEntry& entry) = 0;
```
wgtmac, line 69: The main purpose of the adapter is:
1. Convert this `ManifestEntry` and then append to arrow array. 
2. Deal with differences among table format versions.

I'd suggest keeping the signature unchanged. Perhaps all changes in this file should be moved to manifest_writer.h/cc.

src/iceberg/manifest_writer.h
```diff
@@ -36,15 +38,28 @@ namespace iceberg {
...
+        manifest_location_(manifest_location) {}
 
   ~ManifestWriter() = default;
 
-  /// \brief Write manifest entry to file.
+  /// \brief Add a new Manifest entry.
   /// \param entry Manifest entry to write.
   /// \return Status::OK() if entry was written successfully
-  Status Add(const ManifestEntry& entry);
+  Status AddEntry(const ManifestEntry& entry);
```
wgtmac: What about adding following functions to deal with different use cases:

```cpp
  // Write the entry that all its fields are populated correctly.
  Status WriteEntry(const ManifestEntry& entry);

  // Write the input as a new entry to add.
  Status WriteAddedEntry(std::shared_ptr<DataFile> file,
                         std::optional<int64_t> data_sequence_number);
  Status WriteAddedEntry(const ManifestEntry& entry);

  // Write the input as an existing entry.
  Status WriteExistingEntry(std::shared_ptr<DataFile> file, int64_t file_snapshot_id,
                            int64_t data_sequence_number,
                            std::optional<int64_t> file_sequence_number);
  Status WriteExistingEntry(const ManifestEntry& entry);

  // Write the input as a deleted entry.
  Status WriteDeletedEntry(std::shared_ptr<DataFile> file, int64_t data_sequence_number,
                           std::optional<int64_t> file_sequence_number);
  Status WriteDeletedEntry(const ManifestEntry& entry);

```

wgtmac: `WriteEntry` can be the delegate function to accept writes from other functions. Inside each `WriteXXXEntry` function, we can just create a new entry and the copy should be cheap.

src/iceberg/partition_summary_internal.cc
zhjwpku: I think PartitionSummary is an internal data structure temporarily used by manifest_adapter to generate PartitionFieldSummary, it's not a public api I guess.

src/iceberg/manifest_writer.h
zhjwpku: Done.

src/iceberg/manifest_writer.h
zhjwpku: removed.

src/iceberg/manifest_reader_internal.cc
zhjwpku: Yes, I was thinking about moving this to a seperate PR, but I can't compose a reasonable test, so just leave it here, the cases added in this PR can hit this logic.

src/iceberg/partition_summary_internal.cc
zhjwpku: I do the `CastTo` way you mentioned below.

src/iceberg/partition_summary_internal.cc
zhjwpku: done.

src/iceberg/partition_summary_internal.cc
zhjwpku: Removed.

src/iceberg/partition_summary_internal.cc
zhjwpku: done.

src/iceberg/manifest_writer.h
zhjwpku: Good idea, I tried this, please take a look again.

src/iceberg/metrics.h
zhjwpku: DataFile.record_count is a pure int64_t, I guess it's ok here?

src/iceberg/manifest_adapter.cc
```diff
@@ -161,6 +164,110 @@ ManifestEntryAdapter::~ManifestEntryAdapter() {
   }
 }
 
+Status ManifestEntryAdapter::AddEntry(ManifestEntry& entry) {
+  ICEBERG_RETURN_UNEXPECTED(CheckDataFile(*entry.data_file));
+  entry.status = ManifestStatus::kAdded;
+  entry.snapshot_id = snapshot_id_;
+  if (entry.sequence_number.has_value() &&
+      entry.sequence_number.value() < TableMetadata::kInitialSequenceNumber) {
+    entry.sequence_number = std::nullopt;
```
Jinchul81: Please add a comment why `sequence_number` should be initialized here.

src/iceberg/manifest_adapter.cc
```diff
@@ -161,6 +164,110 @@ ManifestEntryAdapter::~ManifestEntryAdapter() {
...
+  }
+  entry.file_sequence_number = std::nullopt;
```
Jinchul81: Why is `file_sequence_number` initialized here?

src/iceberg/manifest_adapter.cc
```diff
@@ -161,6 +164,110 @@ ManifestEntryAdapter::~ManifestEntryAdapter() {
...
+      min_sequence_number_.value_or(TableMetadata::kInvalidSequenceNumber);
+  manifest_file.existing_files_count = existing_files_count_;
+  manifest_file.added_snapshot_id = snapshot_id_.value_or(Snapshot::kInvalidSnapshotId);
+  manifest_file.added_files_count = add_files_count_;
+  manifest_file.existing_files_count = existing_files_count_;
+  manifest_file.deleted_files_count = delete_files_count_;
+  manifest_file.added_rows_count = add_rows_count_;
+  manifest_file.existing_rows_count = existing_rows_count_;
+  manifest_file.deleted_rows_count = delete_rows_count_;
+  manifest_file.partitions = std::move(partition_summary_->Summaries());
```
Jinchul81: Please clarify the ownership movement. Please add a comment why it is required.

src/iceberg/manifest_list.h
```diff
@@ -72,17 +72,17 @@ struct ICEBERG_EXPORT PartitionFieldSummary {
   static const StructType& Type();
 };
 
+/// \brief The type of files tracked by the manifest, either data or delete files; 0 for
+/// all v1 manifests
+enum class ManifestContent {
```
Jinchul81, line 77: I would prefer `ManifestFileContent` rather than `ManifestContent`.

src/iceberg/manifest_list.h
```diff
@@ -218,21 +224,21 @@ struct ICEBERG_EXPORT ManifestList {
...
-    case ManifestFile::Content::kDeletes:
+    case ManifestContent::kDeletes:
       return "deletes";
   }
   std::unreachable();
 }
 
 /// \brief Get the relative manifest content type from name
-ICEBERG_EXPORT constexpr Result<ManifestFile::Content> ManifestFileContentFromString(
+ICEBERG_EXPORT constexpr Result<ManifestContent> ManifestContentFromString(
```
Jinchul81: Please make it a static function.

src/iceberg/manifest_list.h
```diff
@@ -218,21 +224,21 @@ struct ICEBERG_EXPORT ManifestList {
 };
 
 /// \brief Get the relative manifest content type name
-ICEBERG_EXPORT constexpr std::string_view ToString(ManifestFile::Content type) noexcept {
+ICEBERG_EXPORT constexpr std::string_view ToString(ManifestContent type) noexcept {
```
Jinchul81: Please make it a static function.

src/iceberg/manifest_reader_internal.cc
```diff
@@ -367,13 +367,16 @@ Status ParseDataFile(const std::shared_ptr<StructType>& data_file_schema,
...
+            for (int64_t row_idx = 0; row_idx < view_of_partition->length; row_idx++) {
+              if (ArrowArrayViewIsNull(view_of_partition, row_idx)) {
```
Jinchul81, line 374: Can you please add a comment why we stop here if finding null?

src/iceberg/manifest_writer.cc
```diff
@@ -29,18 +29,39 @@
...
+    ICEBERG_ASSIGN_OR_RAISE(auto array, adapter_->FinishAppending());
+    ICEBERG_RETURN_UNEXPECTED(writer_->Write(array));
+    ICEBERG_RETURN_UNEXPECTED(adapter_->StartAppending());
+  }
+  auto copy = entry.Copy();
+  return adapter_->AddDeleteEntry(copy);
+}
+
+Status ManifestWriter::AddExisting(const ManifestEntry& entry) {
+  if (adapter_->size() >= kBatchSize) {
```
Jinchul81, line 142: Please refactor code. The following code snippets are redundant in `AddEntry`, `AddDelete`, and `AddExisting`.
```
  if (adapter_->size() >= kBatchSize) {
    ICEBERG_ASSIGN_OR_RAISE(auto array, adapter_->FinishAppending());
    ICEBERG_RETURN_UNEXPECTED(writer_->Write(array));
    ICEBERG_RETURN_UNEXPECTED(adapter_->StartAppending());
  }
  auto copy = entry.Copy();
  ```

File: src/iceberg/manifest_writer.h [hunk 27]
```diff
@@ -103,6 +124,10 @@ class ICEBERG_EXPORT ManifestWriter {
   static constexpr int64_t kBatchSize = 1024;
   std::unique_ptr<Writer> writer_;
   std::unique_ptr<ManifestEntryAdapter> adapter_;
+  bool closed_{false};
+  std::string manifest_location_;
+  // TODO(zhjwpku): consider passing key metadata when makeing ManifestWriter.
```
Jinchul81: Please file an issue and use the tracking id instead of your id. I know other code already used personal id, but it's not helpful to follow what to do.

src/iceberg/metrics.h
```diff
@@ -32,12 +32,12 @@ namespace iceberg {
...
-  std::unordered_map<int64_t, int64_t> column_sizes;
-  std::unordered_map<int64_t, int64_t> value_counts;
-  std::unordered_map<int64_t, int64_t> null_value_counts;
-  std::unordered_map<int64_t, int64_t> nan_value_counts;
-  std::unordered_map<int64_t, Literal> lower_bounds;
-  std::unordered_map<int64_t, Literal> upper_bounds;
+  std::unordered_map<int32_t, int64_t> column_sizes;
```
Jinchul81, line 36: What's the reason to shrink the size?

src/iceberg/partition_summary_internal.cc
Jinchul81: Missing `[[unlikely]]`.

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+
+#include "iceberg/expression/literal.h"
+#include "iceberg/manifest_list.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+Status PartitionFieldStats::Update(const Literal& value) {
+  if (type_->type_id() != value.type()->type_id()) {
+    return InvalidArgument("value is not compatible with type");
```
Jinchul81: Please put more details such as actual and expected type ids.

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,95 @@
...
+    summary.lower_bound = lower_bound_->Serialize().value();
+  }
+  if (upper_bound_) {
+    summary.upper_bound = upper_bound_->Serialize().value();
+  }
+  return summary;
+}
+
+Status PartitionSummary::Update(const std::vector<Literal>& partition) {
+  if (partition.size() != field_stats_.size()) {
```
Jinchul81: Missing `[[unlikely]]`.

src/iceberg/partition_summary_internal.h
```diff
@@ -0,0 +1,65 @@
...
+
+#pragma once
+
+#include "iceberg/expression/literal.h"
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+class PartitionFieldStats {
```
Jinchul81, line 29: Missing the class description.

src/iceberg/partition_summary_internal.h
```diff
@@ -0,0 +1,65 @@
...
+
+ private:
+  std::shared_ptr<Type> type_{nullptr};
+  bool contains_null_{false};
+  bool contains_nan_{false};
+  std::optional<Literal> lower_bound_;
+  std::optional<Literal> upper_bound_;
+};
+
+class PartitionSummary {
```
Jinchul81, line 51: Missing the class description.

src/iceberg/partition_summary_internal.h
```diff
@@ -0,0 +1,65 @@
...
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+class PartitionFieldStats {
+ public:
+  explicit PartitionFieldStats(const std::shared_ptr<Type>& type) : type_(type) {}
+
+  Status Update(const Literal& value);
```
Jinchul81, line 34: Please add a comment.

src/iceberg/partition_summary_internal.h
```diff
@@ -0,0 +1,65 @@
...
+
+  PartitionFieldSummary Finish() const;
```
Jinchul81: Please add a comment.

src/iceberg/partition_summary_internal.h
```diff
@@ -0,0 +1,65 @@
...
+  std::vector<PartitionFieldSummary> Summaries() const;
+
+  /// \brief Create a PartitionSummary from the partition type.
+  /// \param partition_type The partition type.
+  /// \return A Result containing a unique pointer to the PartitionSummary.
+  static Result<std::unique_ptr<PartitionSummary>> Make(const StructType& partition_type);
+
+ private:
+  /// \brief Create a PartitionSummary with the given field stats.
+  explicit PartitionSummary(std::vector<PartitionFieldStats> field_stats)
```
Jinchul81, line 54: Why should we hide the ctor? Please add a comment for reasoning.

src/iceberg/test/manifest_reader_writer_test.cc
```diff
@@ -66,7 +68,20 @@ class ManifestReaderWriterTestBase : public TempFileTestBase {
...
+    // Make a copy of expected_entries but set snapshot_id if provided
+    if (snapshot_id.has_value()) {
+      std::vector<ManifestEntry> adjusted_expected_entries;
+      for (const auto& entry : expected_entries) {
+        ManifestEntry copy = entry;
+        copy.snapshot_id = snapshot_id;
+        adjusted_expected_entries.push_back(copy);
+      }
+      ASSERT_EQ(read_result.value(), adjusted_expected_entries);
+      return;
```
Jinchul81: Why should we stop here?

src/iceberg/v2_metadata.cc
```diff
@@ -154,7 +151,9 @@ Status ManifestFileAdapterV2::Init() {
   return ToArrowSchema(*manifest_list_schema_, &schema_);
 }
 
-Status ManifestFileAdapterV2::Append(const ManifestFile& file) {
+Status ManifestFileAdapterV2::Append(ManifestFile& file) {
+  // TODO(zhjwpku): Should we set sequence_number here?
```
Jinchul81: Please file an issue and use the tracking id instead of your id. I know other code already used personal id, but it's not helpful to follow what to do.

src/iceberg/v3_metadata.cc
```diff
@@ -189,7 +188,12 @@ Status ManifestFileAdapterV3::Init() {
   return ToArrowSchema(*manifest_list_schema_, &schema_);
 }
 
-Status ManifestFileAdapterV3::Append(const ManifestFile& file) {
+Status ManifestFileAdapterV3::Append(ManifestFile& file) {
+  // TODO(zhjwpku): Should we set sequence_number/first_row_id here?
```
Jinchul81: Please file an issue and use the tracking id instead of your id. I know other code already used personal id, but it's not helpful to follow what to do.

src/iceberg/manifest_list.h
wgtmac: I think `ManifestContent` is fine which is used by the Java impl.

src/iceberg/manifest_writer.cc
```diff
@@ -50,11 +147,28 @@ Status ManifestWriter::Close() {
...
+Result<Metrics> ManifestWriter::metrics() const { return writer_->metrics(); }
+
+Result<ManifestFile> ManifestWriter::ToManifestFile() const {
+  if (!closed_) [[unlikely]] {
+    return Invalid("Cannot get ManifestFile before closing the writer.");
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto manifest_file, adapter_->ToManifestFile());
+  manifest_file.manifest_path = manifest_location_;
+  manifest_file.manifest_length = writer_->length().value_or(0);
```
wgtmac: Shouldn't we return error if `writer_->length()` returns error?

src/iceberg/manifest_writer.cc
```diff
@@ -38,9 +111,33 @@ Status ManifestWriter::Add(const ManifestEntry& entry) {
   return adapter_->Append(entry);
 }
 
+Status ManifestWriter::CheckDataFile(const DataFile& file) const {
+  switch (adapter_->content()) {
+    case ManifestContent::kData:
+      if (file.content != DataFile::Content::kData) {
+        return InvalidArgument(
+            "Manifest content type: data, data file content should be: data, but got: {}",
+            ToString(file.content));
```
wgtmac:
```suggestion
        return InvalidArgument(
            "Cannot write {} file to data manifest file", ToString(file.content));
```

src/iceberg/manifest_writer.cc
```diff
@@ -38,9 +111,33 @@ Status ManifestWriter::Add(const ManifestEntry& entry) {
...
+      }
+      break;
+    case ManifestContent::kDeletes:
+      if (file.content != DataFile::Content::kPositionDeletes &&
+          file.content != DataFile::Content::kEqualityDeletes) {
+        return InvalidArgument(
+            "Manifest content type: deletes, data file content should be: "
+            "position_deletes or equality_deletes, but got: {}",
+            ToString(file.content));
```
wgtmac:
```suggestion
        return InvalidArgument(
            "Cannot write {} file to delete manifest file", ToString(file.content));
```

src/iceberg/manifest_writer.h
```diff
@@ -100,9 +156,20 @@ class ICEBERG_EXPORT ManifestWriter {
       std::shared_ptr<Schema> current_schema, ManifestContent content);
 
  private:
+  /// \brief Write the entry that all its fields are populated correctly.
+  /// \param entry Manifest entry to write.
+  /// \return Status::OK() if entry was written successfully
+  /// \note All other write entry variants delegate to this method after populating
+  /// the necessary fields.
+  Status WriteEntry(const ManifestEntry& entry);
```
wgtmac: I think this can be public because `AddAll(const std::vector<ManifestEntry>& entries)` is also public.

src/iceberg/partition_summary_internal.h
```diff
@@ -0,0 +1,65 @@
...
+  explicit PartitionFieldStats(const std::shared_ptr<Type>& type) : type_(type) {}
+
+  Status Update(const Literal& value);
+
+  Result<PartitionFieldSummary> Finish() const;
+
+  const std::shared_ptr<Type>& type() const { return type_; }
+
+ private:
+  std::shared_ptr<Type> type_{nullptr};
```
wgtmac: Perhaps `const Type& type_` is better to avoid copy.

src/iceberg/partition_summary_internal.h
wgtmac, line 29: Add ICEBERG_EXPORT if used in the test.

src/iceberg/partition_summary_internal.cc
wgtmac: Fair enough. Just note that only header file is required to add the suffix. We can remove the suffix from .cc file.

src/iceberg/partition_summary_internal.cc
```diff
@@ -0,0 +1,105 @@
...
+  summary.contains_null = contains_null_;
+  summary.contains_nan = contains_nan_;
+  if (lower_bound_) {
+    ICEBERG_ASSIGN_OR_RAISE(auto serialized_lower, lower_bound_->Serialize())
+    summary.lower_bound = std::move(serialized_lower);
+  }
+  if (upper_bound_) {
+    ICEBERG_ASSIGN_OR_RAISE(auto serialized_upper, upper_bound_->Serialize())
+    summary.upper_bound = std::move(serialized_upper);
+  }
```
wgtmac, line 66:
```suggestion
  if (lower_bound_) {
    ICEBERG_ASSIGN_OR_RAISE(summary.lower_bound, lower_bound_->Serialize())
  }
  if (upper_bound_) {
    ICEBERG_ASSIGN_OR_RAISE(summary.upper_bound, upper_bound_->Serialize())
  }
```

src/iceberg/partition_summary.cc
```diff
@@ -0,0 +1,105 @@
...
+  }
+  if (upper_bound_) {
+    ICEBERG_ASSIGN_OR_RAISE(auto serialized_upper, upper_bound_->Serialize())
+    summary.upper_bound = std::move(serialized_upper);
+  }
+  return summary;
+}
+
+PartitionSummary::PartitionSummary(const StructType& partition_type) {
+  std::vector<PartitionFieldStats> field_stats;
```
wgtmac: nit: call reserve on field_stats

wgtmac: BTW, why not directly use `field_stats_`?

src/iceberg/manifest_adapter.cc
```diff
@@ -161,6 +164,27 @@ ManifestEntryAdapter::~ManifestEntryAdapter() {
...
+      min_sequence_number_.value_or(TableMetadata::kInvalidSequenceNumber);
+  manifest_file.existing_files_count = existing_files_count_;
+  manifest_file.added_files_count = add_files_count_;
+  manifest_file.existing_files_count = existing_files_count_;
+  manifest_file.deleted_files_count = delete_files_count_;
+  manifest_file.added_rows_count = add_rows_count_;
+  manifest_file.existing_rows_count = existing_rows_count_;
+  manifest_file.deleted_rows_count = delete_rows_count_;
+  ICEBERG_ASSIGN_OR_RAISE(auto partition_summary, partition_summary_->Summaries());
+  manifest_file.partitions = std::move(partition_summary);
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(manifest_file.partitions, partition_summary_->Summaries());
```

src/iceberg/manifest_adapter.cc
```diff
@@ -161,6 +164,27 @@ ManifestEntryAdapter::~ManifestEntryAdapter() {
   }
 }
 
+Result<ManifestFile> ManifestEntryAdapter::ToManifestFile() const {
+  ManifestFile manifest_file;
```
wgtmac: nit: use aggregate initialization for struct

src/iceberg/manifest_writer.cc
```diff
@@ -21,15 +21,88 @@
...
 
 namespace iceberg {
 
-Status ManifestWriter::Add(const ManifestEntry& entry) {
+Status ManifestWriter::WriteAddedEntry(std::shared_ptr<DataFile> file,
+                                       std::optional<int64_t> data_sequence_number) {
+  ManifestEntry added;
+  added.status = ManifestStatus::kAdded;
+  added.snapshot_id = adapter_->snapshot_id();
+  added.data_file = std::move(file);
```
wgtmac: Just reminder that first_row_id should be suppressed if set: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/GenericManifestEntry.java#L79

Same for the `WriteAddedEntry` function below

src/iceberg/v2_metadata.cc
```diff
@@ -83,11 +83,18 @@ Status ManifestEntryAdapterV2::Init() {
...
   ICEBERG_ASSIGN_OR_RAISE(auto partition_type,
                           partition_spec_->PartitionType(*current_schema_));
+  partition_summary_ = std::make_unique<PartitionSummary>(*partition_type);
   manifest_schema_ = EntrySchema(std::move(partition_type));
   return ToArrowSchema(*manifest_schema_, &schema_);
 }
 
 Status ManifestEntryAdapterV2::Append(const ManifestEntry& entry) {
+  if (entry.IsAlive() && entry.sequence_number.has_value()) {
+    if (!min_sequence_number_.has_value() ||
```
wgtmac: Why the maintenance of `min_sequence_number_` is duplicated in v2 and v3 but not v1? I think the best place is in the ManifestWriter::WriteEntry together with stats update.

src/iceberg/manifest_adapter.cc
```diff
@@ -161,6 +164,27 @@ ManifestEntryAdapter::~ManifestEntryAdapter() {
   }
 }
 
+Result<ManifestFile> ManifestEntryAdapter::ToManifestFile() const {
```
wgtmac: It looks strange that both `ManifestEntryAdapter` and `ManifestWriter` support `ToManifestFile()` but the one returned by `ManifestEntryAdapter` is partial. Should we remove the function here and maintain all entry stats in the ManifestWriter to make it look less strange?

src/iceberg/metrics.h
wgtmac: The reason is that now we return `Result<Metrics>` and we need to make it explicit that row_count is unavailable. In the Java impl it is also a `Long` type.

src/iceberg/manifest_adapter.cc
zhjwpku: This logic has been removed since GetSequenceNumber handle this, the reason for it been null is because for newly added data, its sequence number is inherited from manifest metadata, see [1].

[1] https://iceberg.apache.org/spec/#sequence-number-inheritance

src/iceberg/manifest_adapter.cc
zhjwpku: ditto

src/iceberg/manifest_adapter.cc
zhjwpku: This has been changed to:

```
ICEBERG_ASSIGN_OR_RAISE(auto partition_summary, partition_summary_->Summaries());
manifest_file.partitions = std::move(partition_summary);
```

I agree at the version of you comment, the `move` is not necessary, I guess meson build would warn under warning_level=2. But after changing, the `move` can avoid an extra copy.

src/iceberg/manifest_list.h
zhjwpku: I'm a bit confused here, are you worried that this function will be included in multiple translation units?
If so, `constexpr` implies `inline`, so this should be fine. We already use this pattern in several other places, correct me if I'm wrong.

src/iceberg/manifest_writer.h
zhjwpku: key metadata will not be supported for now, I've removed this comment. Your suggestion is reasonable, thanks.

src/iceberg/manifest_reader_internal.cc
zhjwpku: Good question. The fix here was intended to handle multiple partition values, but I'm not sure why we didn't follow the case 2 pattern. @dongxiao1198 can you share some insight?

src/iceberg/metrics.h
zhjwpku: Because these fields are serialized as <int32_t, int64_t> pair. per spec [1]

[1] https://iceberg.apache.org/spec/#data-file-fields

src/iceberg/partition_summary_internal.h
zhjwpku: The reason was there was a Make method, but now I remove the Make and bring the ctor to public.

src/iceberg/v3_metadata.cc
zhjwpku: These TODOs were meant to attract reviewers' closer attention. I've removed them now.

src/iceberg/manifest_adapter.cc
zhjwpku: After seeing Gang's comment, I change it to the following:

```
ICEBERG_ASSIGN_OR_RAISE(manifest_file.partitions, partition_summary_->Summaries());
```

src/iceberg/manifest_writer.cc
zhjwpku: Since DataFile is a struct, I can not think of a better way than adding a Clone method to DataFile, any suggestion?

src/iceberg/v2_metadata.cc
zhjwpku: I do this mostly because the test case check, v1 doesn't maintain sequence number, so its min_sequence_number_ should be kInvalidSequenceNumber. I think we can do what you suggested and change the test case accordingly.

PR #319, opened 2025-11-14
chore: add support to customize download url of vendored dependencies

### LINKED ISSUE #239: build deps mirror for China
now the cmake deps are all github.com.

developers in China are easy to build failed occurs to network.

Subham-KRLX:
Fixes #239

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,25 @@ C++ implementation of [Apache Iceberg™](https://iceberg.apache.org/).
 
 ## Customizing Dependency URLs
 
-If you experience network issues when downloading dependencies, you can customize the download URLs using environment variables. See the available options documented in [`cmake_modules/IcebergThirdpartyToolchain.cmake`](cmake_modules/IcebergThirdpartyToolchain.cmake#L24-L40).
+If you experience network issues when downloading dependencies, you can customize the download URLs using environment variables.
+
+The following environment variables can be set to customize dependency URLs:
+
+- `ICEBERG_ARROW_URL`: Apache Arrow tarball URL
+- `ICEBERG_AVRO_URL`: Apache Avro tarball URL
+- `ICEBERG_AVRO_GIT_URL`: Apache Avro git repository URL
+- `ICEBERG_NANOARROW_URL`: Nanoarrow tarball URL
+- `ICEBERG_CROARING_URL`: CRoaring tarball URL
+- `ICEBERG_NLOHMANN_JSON_URL`: nlohmann-json tarball URL
+- `ICEBERG_SPDLOG_URL`: spdlog tarball URL
+- `ICEBERG_CPR_URL`: cpr tarball URL
+
+Example usage:
+
+```bash
+export ICEBERG_ARROW_URL="https://your-mirror.com/apache-arrow-22.0.0.tar.gz"
+cmake -S . -B build
+```
 
 ## Build
 


README.md
```diff
@@ -31,6 +31,8 @@ C++ implementation of [Apache Iceberg™](https://iceberg.apache.org/).
 - CMake 3.25 or higher
 - C++23 compliant compiler
 
+> **Note**: For developers in China experiencing network issues when downloading dependencies, see [BUILD_CHINA.md](BUILD_CHINA.md) for mirror configuration.
```
zhjwpku: -1 from me. sorry about that :(

One quick question: do other Apache projects handle this the same way for China?

iceberg-cpp should build against system libraries dependencies, and I think that's probably the right direction.

HuaHuaY: Agree with you. We have CMake options such as `ICEBERG_ARROW_URL` which allow users to customize urls. If other projects use specific urls for China, I would also support following this approach.

Subham-KRLX: Thanks for the review!
To clarify: this PR doesn’t introduce region specific behavior or change defaults. It only provides optional mirror URLs just like `ICEBERG_ARROW_URL` to make the build workable for developers in China who frequently hit GitHub timeouts.

If preferred, I can simplify the PR so it only adds optional CMake variables + documentation, matching how other Apache projects handle this.
Please let me know and I’ll update it accordingly.

Xuanwo: Here is my -1 too. I don't think it's good idea to add other mirrors in `ARROW_SOURCE_URL` too.

Subham-KRLX: After researching Apache best practices I removed all hardcoded mirror URLs and kept only optional environment variables for custom mirrors just like ICEBERG_ARROW_URL. I added documentation showing how users can set these variables if needed. No defaults changed—everything still defaults to the original URLs. This keeps the build flexible respects Apache guidelines and helps developers facing network issues. Please review and let me know if you need any changes.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -221,9 +246,17 @@ endfunction()
 function(resolve_nanoarrow_dependency)
   prepare_fetchcontent()
 
+  if(DEFINED ENV{ICEBERG_NANOARROW_URL})
```
wgtmac, line 265: Could you help define these variables to the top of this file: https://github.com/apache/iceberg-cpp/blob/main/cmake_modules/IcebergThirdpartyToolchain.cmake#L24

BUILD_CHINA.md
```diff
@@ -0,0 +1,118 @@
...
+
+1. **Use system packages**: Install dependencies via your system package manager
+2. **Use a proxy**: Set `https_proxy` environment variable
+3. **Pre-download**: Manually download tarballs to `~/.cmake/Downloads/`
+
+## Getting Help
+
+If you continue experiencing build issues, please open an issue at https://github.com/apache/iceberg-cpp/issues with details about which dependency failed.
+
+# Building in China
```
wgtmac: This document seems not well organized. I'm not sure if it is a good practice to mention China explicitly. In order not to maintain this file, can we remove it and only modify README to let users know that we can use env variables to customize thirdparty dependencies?

README.md
```diff
@@ -31,6 +31,10 @@ C++ implementation of [Apache Iceberg™](https://iceberg.apache.org/).
 - CMake 3.25 or higher
 - C++23 compliant compiler
 
+## Customizing Dependency URLs
+
+If you experience network issues when downloading dependencies, you can customize the download URLs using environment variables. See the available options documented in [`cmake_modules/IcebergThirdpartyToolchain.cmake`](cmake_modules/IcebergThirdpartyToolchain.cmake#L24-L40).
```
zhjwpku: IcebergThirdpartyToolchain.cmake can be changed, the link you put here is not stable. Why not directly put the quoted lines here?

wgtmac (approved):

Issue #314, opened 2025-11-12
Support UUID type


wgtmac:
The UUID type is not fully supported. The list is as below but it may not be exhaustive:

- [ ] Arrow UUID canonical extension type conversion, including schema and array.
- [ ] File reader and writer: Avro, Parquet.
- [ ] Literal value.
- [ ] StructLike scalar value.

PR #320, opened 2025-11-14
feat: add eval support to bound term

wgtmac:
- add struct like accessor
- support schema to find accessor by field id
- bound term can evaluate struct-like

### DIFF
--- a/src/iceberg/row/struct_like.cc
+++ b/src/iceberg/row/struct_like.cc
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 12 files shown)


src/iceberg/row/struct_like.cc
```diff
@@ -0,0 +1,129 @@
...
+    case TypeId::kTimestamp:
+      return Literal::Timestamp(std::get<int64_t>(scalar));
+    case TypeId::kTimestampTz:
+      return Literal::TimestampTz(std::get<int64_t>(scalar));
+    case TypeId::kFixed: {
+      const auto& fixed_data = std::get<std::string_view>(scalar);
+      return Literal::Fixed(std::vector<uint8_t>(fixed_data.cbegin(), fixed_data.cend()));
+    }
+    case TypeId::kUuid:
+      // TODO(gangwu): Implement UUID type
```
zhjwpku, line 120: We seems to have a Uuid class under `util` dir, can we use that?

wgtmac: I have created https://github.com/apache/iceberg-cpp/issues/314 to address all aspects of the uuid data type. I think the missing parts are more than the scalar part.

PR #321, opened 2025-11-15
feat: transactional UpdateProperties method support

lishuxu:
feat: transactional UpdateProperties method support

Table/Catalog produce a Transaction (e.g., NewTransaction or staged create/replace).
Clients build PendingUpdate actions (e.g., UpdateProperties, AppendFiles), each validating via Apply() (conflict checks, reserved keys, format-version parsing, schema/metrics validation).
Transaction::CommitTransaction() is expected to gather pending updates and call the catalog’s UpdateTable (with derived TableUpdate objects and requirements) in one atomic commit; on success, metadata refreshes; on failure, errors propagate

### DIFF
--- a/src/iceberg/type_fwd.h
+++ b/src/iceberg/type_fwd.h
@@ -183,8 +183,6 @@ class TableUpdateContext;
 class PendingUpdate;
 class UpdateProperties;
 
-struct PropertiesUpdateChanges;
-
 /// ----------------------------------------------------------------------------
 /// TODO: Forward declarations below are not added yet.
 /// ----------------------------------------------------------------------------


zhjwpku (approved):
This is consistent with the rest catalog open api, so +1, thanks.

src/iceberg/transaction.h
```diff
@@ -39,6 +39,76 @@ class ICEBERG_EXPORT Transaction {
   /// \return this transaction's table
   virtual const std::shared_ptr<Table>& table() const = 0;
 
+  /// \brief Create a new schema addition operation
+  ///
+  /// \return a new AddSchema
+  virtual std::shared_ptr<AddSchema> AddSchema() = 0;
```
wgtmac:
```suggestion
  virtual std::unique_ptr<AddSchema> AddSchema() = 0;
```

I'd suggest returning std::unique_ptr for all actions here and below.

src/iceberg/transaction.h
```diff
@@ -39,6 +39,76 @@ class ICEBERG_EXPORT Transaction {
...
+  /// \param schema_id the schema id to set as current
+  /// \return a new SetCurrentSchema
+  virtual std::shared_ptr<SetCurrentSchema> SetCurrentSchema(int32_t schema_id) = 0;
+
+  /// \brief Create a new remove schemas operation
+  ///
+  /// \param schema_ids the schema ids to remove
+  /// \return a new RemoveSchemas
+  virtual std::shared_ptr<RemoveSchemas> RemoveSchemas(
+      const std::vector<int32_t>& schema_ids) = 0;
```
wgtmac:
```suggestion
      std::span<const int32_t> schema_ids) = 0;
```

Should we use span to accept broader range of input types for here and below?

src/iceberg/type_fwd.h
```diff
@@ -149,6 +149,25 @@ class ArrayLike;
 class MapLike;
 
 class TableUpdate;
+class AssignUUID;
```
Jinchul81: Can you please reorder the class names? Please follow the Google CPP guideline.

HeartLinked (approved):

src/iceberg/catalog/memory/in_memory_catalog.h
```diff
@@ -22,6 +22,8 @@
 #include <mutex>
 
 #include "iceberg/catalog.h"
+#include "iceberg/table_requirement.h"
```
HuaHuaY: I suggest to use forward declarations instead of including header file.

lishuxu: Because TransactionCatalog needs to gather and store all TableRequirement/TableUpdate objects for a given Transaction, I changed the UpdateTable interface parameter from a reference to a non-reference type. Therefore, the header files need be included here instead of relying on forward declarations.

HuaHuaY: `std::vector<std::unique_ptr<>>` doesn't need the complete type.

lishuxu: Yes, I see. I have remove the "includes" from catalog header files

src/iceberg/catalog.h
```diff
@@ -123,8 +123,8 @@ class ICEBERG_EXPORT Catalog {
   /// \return a Table instance or ErrorKind::kAlreadyExists if the table already exists
   virtual Result<std::unique_ptr<Table>> UpdateTable(
       const TableIdentifier& identifier,
-      const std::vector<std::unique_ptr<TableRequirement>>& requirements,
-      const std::vector<std::unique_ptr<TableUpdate>>& updates) = 0;
+      std::vector<std::unique_ptr<TableRequirement>> requirements,
```
wgtmac: Can we revert this change? If later we support retry on failed commits, these `requirements` and `updates` cannot be reused since they are moved away.

src/iceberg/catalog.h
```diff
@@ -184,6 +184,11 @@ class ICEBERG_EXPORT Catalog {
   /// \return a Table instance or ErrorKind::kAlreadyExists if the table already exists
   virtual Result<std::shared_ptr<Table>> RegisterTable(
       const TableIdentifier& identifier, const std::string& metadata_file_location) = 0;
+
+  /// \brief Set whether the last operation in a transaction has been committed
+  ///
+  /// \param committed true if the last operation has been committed, false otherwise
+  virtual void SetLastOperationCommitted(bool committed) = 0;
```
wgtmac, line 191: This function is only used by transaction catalog so it should not appear here.

src/iceberg/table.cc
```diff
@@ -114,7 +115,7 @@ std::unique_ptr<UpdateProperties> Table::UpdateProperties() const {
 }
 
 std::unique_ptr<Transaction> Table::NewTransaction() const {
-  throw NotImplemented("Table::NewTransaction is not implemented");
+  return std::make_unique<BaseTransaction>(shared_from_this(), catalog_);
```
wgtmac: I'd recommend directly pass `Table*` to it. Similarly, for `catalog_` we can use `catalog_.get()` to just use pointer. The created transaction object should not outlive the table and catalog objects. It is unnecessary to increase the reference counter of them.

I'm open to keep the current design. WDYT @HuaHuaY?

src/iceberg/base_transaction.h
```diff
@@ -0,0 +1,134 @@
...
+class ICEBERG_EXPORT BaseTransaction : public Transaction {
+ public:
+  BaseTransaction(std::shared_ptr<const Table> table, std::shared_ptr<Catalog> catalog);
+  ~BaseTransaction() override = default;
+
+  const std::shared_ptr<const Table>& table() const override;
+
+  std::unique_ptr<::iceberg::UpdateProperties> UpdateProperties() override;
+
+  std::unique_ptr<AppendFiles> NewAppend() override;
```
wgtmac:
```suggestion
  std::shared_ptr<UpdateProperties> NewUpdateProperties() override;

  std::shared_ptr<AppendFiles> NewAppend() override;
```

It seems that returning shared_ptr will make it easier to collect and transform the update internally. Adding `New` prefix also can avoid `iceberg::` prefix in the return type.

src/iceberg/base_transaction.h
```diff
@@ -0,0 +1,134 @@
...
+#include "iceberg/table_identifier.h"
+#include "iceberg/table_requirement.h"
+#include "iceberg/table_update.h"
+#include "iceberg/transaction.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Base class for transaction implementations
+class ICEBERG_EXPORT BaseTransaction : public Transaction {
```
wgtmac: If `BaseTransaction` is the only subclass, we can just implement all features to `Transaction` just like `Table`.

src/iceberg/base_transaction.cc
```diff
@@ -0,0 +1,146 @@
...
+#include "iceberg/transaction_catalog.h"
+#include "iceberg/update/update_properties.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+BaseTransaction::BaseTransaction(std::shared_ptr<const Table> table,
+                                 std::shared_ptr<Catalog> catalog)
+    : table_(std::move(table)) {
+  ICEBERG_DCHECK(table_ != nullptr, "table must not be null");
```
wgtmac: We need to add a `Transaction::Make` function to return `Result<Transaction>` and check these arguments explicitly.

src/iceberg/base_transaction.h
```diff
@@ -0,0 +1,134 @@
...
+  /// \param updates the list of table updates to apply
+  /// \return Status::OK if the updates were applied successfully, or an error status
+  Status ApplyUpdates(const std::vector<std::unique_ptr<TableUpdate>>& updates);
+
+  /// \brief Create a new table update if the last operation has been committed
+  ///
+  /// \param args the arguments to forward to the update constructor
+  /// \return a new UpdateType instance or an error status
+  template <typename UpdateType, typename... Args>
+  Result<std::unique_ptr<UpdateType>> CheckAndCreateUpdate(Args&&... args) {
```
wgtmac: This is over-complicated. Let's switch to `std::shared_ptr` for these update types.

src/iceberg/base_transaction.h
wgtmac: BTW, do we want to introduce `TransactionType` from the Java impl as well? If would be good if we can add this to make it clear.

src/iceberg/base_transaction.cc
```diff
@@ -0,0 +1,146 @@
...
+  catalog_ = std::make_shared<TransactionCatalog>(std::move(catalog), this);
+}
+
+const std::shared_ptr<const Table>& BaseTransaction::table() const { return table_; }
+
+std::unique_ptr<UpdateProperties> BaseTransaction::UpdateProperties() {
+  auto update = CheckAndCreateUpdate<::iceberg::UpdateProperties>(
+      table_->name(), catalog_, CurrentMetadata());
+  if (!update.has_value()) {
+    ERROR_TO_EXCEPTION(update.error());
```
wgtmac: We cannot throw. Perhaps we should also use `Result` wrapper for these update return types.

src/iceberg/transaction_catalog.h
```diff
@@ -0,0 +1,135 @@
...
+namespace iceberg {
+
+/**
+ * @brief Lightweight catalog wrapper for BaseTransaction.
+ *
+ * For read-only operations, TransactionCatalog simply forwards to the wrapped catalog.
+ * For mutating calls such as UpdateTable or HasLastOperationCommitted, it delegates back
+ * to the owning BaseTransaction so staged updates remain private until commit.
+ */
+class ICEBERG_EXPORT TransactionCatalog : public Catalog {
```
wgtmac, line 38: Can we move this class to `base_transaction.cc`?

src/iceberg/base_transaction.cc
```diff
@@ -0,0 +1,146 @@
...
+  auto pending_requirements = ConsumePendingRequirements();
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto updated_table,
+      catalog_->catalog_impl()->UpdateTable(
+          table_->name(), std::move(pending_requirements), std::move(pending_updates)));
+
+  // update table to the new version
+  if (updated_table) {
+    table_ = std::shared_ptr<Table>(std::move(updated_table));
```
wgtmac: Why do we need this `table_` after commit has been done? It seems useless?

src/iceberg/base_transaction.cc
```diff
@@ -0,0 +1,146 @@
...
+  ICEBERG_RETURN_UNEXPECTED(ApplyUpdates(updates));
+
+  for (auto& requirement : requirements) {
+    context_.pending_requirements.emplace_back(std::move(requirement));
+  }
+  for (auto& update : updates) {
+    context_.pending_updates.emplace_back(std::move(update));
+  }
+
+  return std::make_unique<Table>(
```
wgtmac, line 121: Why do we need to return a table?

src/iceberg/base_transaction.cc
```diff
@@ -0,0 +1,146 @@
...
+  if (!HasLastOperationCommitted()) {
+    return InvalidState("Cannot commit transaction: last operation has not committed");
+  }
+
+  auto pending_updates = ConsumePendingUpdates();
+  if (pending_updates.empty()) {
+    return {};
+  }
+
+  auto pending_requirements = ConsumePendingRequirements();
```
wgtmac: What if we need to support retry on failed commits? In the current approach, these two lists are gone.

src/iceberg/table.cc
HuaHuaY: When the table‘s content changes, or we drop a table, what should the executing transaction hold?

src/iceberg/catalog.h
lishuxu: We have to collect requirements/updates in transaction, so I use shared_ptr to replace unique_ptr for arguments requirements/updates

src/iceberg/base_transaction.h
lishuxu: Thanks for the suggestion.

In the current transaction implementation, we don’t need to collect or retain PendingUpdate instances across the transaction lifecycle, so returning std::unique_ptr keeps the ownership semantics explicit and avoids introducing shared ownership where it isn’t required.

I agree that returning shared_ptr could make internal aggregation or transformation of updates easier if the transaction logic evolves in that direction.

src/iceberg/base_transaction.h
lishuxu: I’d prefer to keep the current BaseTransaction design for clearer interfaces.

As for TransactionType, we can introduce it later, following the Java implementation.

src/iceberg/base_transaction.cc
lishuxu: Yes，done

lishuxu:
This PR duplicates the changes in https://github.com/apache/iceberg-cpp/pull/418
Closing this one to avoid duplication.

PR #327, opened 2025-11-19
feat: implement expression evaluator

wgtmac:


wgtmac:
@zhjwpku @HuaHuaY @HeartLinked PTAL

src/iceberg/expression/evaluator.cc
```diff
@@ -0,0 +1,163 @@
...
+#include "iceberg/expression/binder.h"
+#include "iceberg/expression/expression_visitor.h"
+#include "iceberg/schema.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+class Evaluator::EvalVisitor : public BoundVisitor<bool> {
+ public:
+  void UpdateRow(const StructLike* row) { row_ = row; }
```
HuaHuaY: I think we can create a new object in `Evaluator::Eval` like java's implementation and change `UpdateRow` to constructor. So we don't need `ICEBERG_DCHECK(row_, "Row is not set")` in every methods.

src/iceberg/expression/evaluator.cc
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  Result<bool> IsNull(const std::shared_ptr<BoundTerm>& term) override {
+    ICEBERG_DCHECK(row_, "Row is not set");
+    ICEBERG_ASSIGN_OR_RAISE(auto value, term->Evaluate(*row_));
+    return value.IsNull();
+  }
+
+  Result<bool> NotNull(const std::shared_ptr<BoundTerm>& term) override {
+    ICEBERG_DCHECK(row_, "Row is not set");
```
zhjwpku: Not a strong opinion, we could adopt the same pattern used for Eq/NotEq, where NotXXX calls XXX and returns !result. Ditto for Lt/GtEq and Gt/LtEq.

src/iceberg/expression/evaluator.cc
zhjwpku: nit: SetRow? since this method doesn't update anything.

wgtmac:
Thanks for the review @zhjwpku @HuaHuaY! I have addressed all comments.

HuaHuaY (approved):

liurenjie1024 (approved):

PR #328, opened 2025-11-19
refactor: suppress warnings

Jinchul81:
Warnings have been suppressed according to the following patterns:
* Applied `[[maybe_unused]]`
* Explicitly initialize the member variables in struct.
* Unaligned type comparison between signed and unsigned. Converted unsigned to singed.
* Hiding virtual functions in the base class. Declare `using base::func`.

### DIFF
--- a/src/iceberg/catalog/memory/in_memory_catalog.cc
+++ b/src/iceberg/catalog/memory/in_memory_catalog.cc
@@ -384,23 +384,25 @@ Result<std::vector<TableIdentifier>> InMemoryCatalog::ListTables(
 }
 
 Result<std::unique_ptr<Table>> InMemoryCatalog::CreateTable(
-    const TableIdentifier& identifier, const Schema& schema, const PartitionSpec& spec,
-    const std::string& location,
-    const std::unordered_map<std::string, std::string>& properties) {
+    [[maybe_unused]] const TableIdentifier& identifier,
+    [[maybe_unused]] const Schema& schema, [[maybe_unused]] const PartitionSpec& spec,
+    [[maybe_unused]] const std::string& location,
+    [[maybe_unused]] const std::unordered_map<std::string, std::string>& properties) {
   return NotImplemented("create table");
 }
 
 Result<std::unique_ptr<Table>> InMemoryCatalog::UpdateTable(
-    const TableIdentifier& identifier,
-    const std::vector<std::unique_ptr<TableRequirement>>& requirements,
-    const std::vector<std::unique_ptr<TableUpdate>>& updates) {
+    [[maybe_unused]] const TableIdentifier& identifier,
+    [[maybe_unused]] const std::vector<std::unique_ptr<TableRequirement>>& requirements,
+    [[maybe_unused]] const std::vector<std::unique_ptr<TableUpdate>>& updates) {
   return NotImplemented("update table");
 }
 
 Result<std::shared_ptr<Transaction>> InMemoryCatalog::StageCreateTable(
-    const TableIdentifier& identifier, const Schema& schema, const PartitionSpec& spec,
-    const std::string& location,
-    const std::unordered_map<std::string, std::string>& properties) {
+    [[maybe_unused]] const TableIdentifier& identifier,
+    [[maybe_unused]] const Schema& schema, [[maybe_unused]] const PartitionSpec& spec,
+    [[maybe_unused]] const std::string& location,
+    [[maybe_unused]] const std::unordered_map<std::string, std::string>& properties) {
   return NotImplemented("stage create table");
 }
 
@@ -409,14 +411,15 @@ Result<bool> InMemoryCatalog::TableExists(const TableIdentifier& identifier) con
   return root_namespace_->TableExists(identifier);
 }
 
-Status InMemoryCatalog::DropTable(const TableIdentifier& identifier, bool purge) {
+Status InMemoryCatalog::DropTable(const TableIdentifier& identifier,
+                                  [[maybe_unused]] bool purge) {
   std::unique_lock lock(mutex_);
   // TODO(Guotao): Delete all metadata files if purge is true.
   return root_namespace_->UnregisterTable(identifier);
 }
 
-Status InMemoryCatalog::RenameTable(const TableIdentifier& from,
-                                    const TableIdentifier& to) {
+Status InMemoryCatalog::RenameTable([[maybe_unused]] const TableIdentifier& from,
+                                    [[maybe_unused]] const TableIdentifier& to) {
   std::unique_lock lock(mutex_);
   return NotImplemented("rename table");
 }
@@ -455,7 +458,8 @@ Result<std::shared_ptr<Table>> InMemoryCatalog::RegisterTable(
 }
 
 std::unique_ptr<Catalog::TableBuilder> InMemoryCatalog::BuildTable(
-    const TableIdentifier& identifier, const Schema& schema) const {
+    [[maybe_unused]] const TableIdentifier& identifier,
+    [[maybe_unused]] const Schema& schema) const {
   throw IcebergError("not implemented");
 }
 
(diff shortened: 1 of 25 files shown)


src/iceberg/catalog/rest/types.h
```diff
@@ -39,9 +39,9 @@ namespace iceberg::rest {
 
 /// \brief Server-provided configuration for the catalog.
 struct ICEBERG_REST_EXPORT CatalogConfig {
-  std::unordered_map<std::string, std::string> defaults;   // required
-  std::unordered_map<std::string, std::string> overrides;  // required
-  std::vector<std::string> endpoints;
+  std::unordered_map<std::string, std::string> defaults{};   // required
```
wgtmac, line 42: We don't need to add `{}` for everything, at least not for STL containers and std::string. Please try to minimize lines of change like this.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -384,23 +384,25 @@ Result<std::vector<TableIdentifier>> InMemoryCatalog::ListTables(
 }
 
 Result<std::unique_ptr<Table>> InMemoryCatalog::CreateTable(
-    const TableIdentifier& identifier, const Schema& schema, const PartitionSpec& spec,
-    const std::string& location,
-    const std::unordered_map<std::string, std::string>& properties) {
+    [[maybe_unused]] const TableIdentifier& identifier,
```
HuaHuaY, line 387: I think we can remove the parameters' names instead of adding `[[maybe_unused]]` to keep the code clear.

src/iceberg/expression/literal.cc
```diff
@@ -245,7 +245,7 @@ Result<Literal> LiteralCaster::CastFromBinary(
   switch (target_type->type_id()) {
     case TypeId::kFixed: {
       auto target_fixed_type = internal::checked_pointer_cast<FixedType>(target_type);
-      if (binary_val.size() == target_fixed_type->length()) {
+      if (static_cast<int32_t>(binary_val.size()) == target_fixed_type->length()) {
```
HuaHuaY: In general, we should convert small types to large types. There are some similar codes in this PR that need to be modified.

src/iceberg/test/schema_util_test.cc
```diff
@@ -82,14 +82,14 @@ std::shared_ptr<Type> CreateNestedStruct() {
   });
 }
 
-std::shared_ptr<Type> CreateListOfList() {
+[[maybe_unused]] std::shared_ptr<Type> CreateListOfList() {
```
HuaHuaY, line 85: @wgtmac Does this method not being used mean we are missing some tests?

src/iceberg/util/conversions.cc
```diff
@@ -200,7 +200,7 @@ Result<Literal::Value> Conversions::FromBytes(const PrimitiveType& type,
       return Literal::Value{std::vector<uint8_t>(data.begin(), data.end())};
     case TypeId::kFixed: {
       const auto& fixed_type = static_cast<const FixedType&>(type);
-      if (data.size() != fixed_type.length()) {
+      if (static_cast<int32_t>(data.size()) != fixed_type.length()) {
```
HuaHuaY: ditto

src/iceberg/schema.h
```diff
@@ -103,6 +103,7 @@ class ICEBERG_EXPORT Schema : public StructType {
   friend bool operator==(const Schema& lhs, const Schema& rhs) { return lhs.Equals(rhs); }
 
  private:
+  using Type::Equals;
```
HuaHuaY, line 106: @wgtmac It seems that we should override the `Equals` methods in `Schema` instead of adding `Schema::Equals(const Schema& other) const` at line 108?

src/iceberg/catalog/rest/types.h
Jinchul81: It looks like they are necessary. The following warnings occurred when removing the initialization. Do you have any better idea?
```
../src/iceberg/test/rest_json_internal_test.cc:774:55: warning: missing initializer for member ‘iceberg::rest::CatalogConfig::endpoints’
../src/iceberg/test/rest_json_internal_test.cc:780:76: warning: missing initializer for member ‘iceberg::rest::CatalogConfig::overrides’
```

src/iceberg/schema.h
Jinchul81: @lidavidm Did you intend to introduce non-overriding function `Equals` in this commit ebacc54dd55e74144e84c0a02e8445c590e353af?

src/iceberg/expression/literal.cc
Jinchul81: Let me rewrite the change like this:
```
      if (target_fixed_type->length() >= 0 &&
          binary_val.size() == static_cast<size_t>(target_fixed_type->length())) {
        return Literal::Fixed(std::move(binary_val));
      }
```

src/iceberg/catalog/memory/in_memory_catalog.cc
Jinchul81: I am not sure your suggestion seems to be aligned with the original design. These functions overrode the pure virtual functions in `Catalog`.

HuaHuaY: I don't get your point. Removing the names in child class don't change the function signature even though it overrides virtual function.

```cpp
Result<std::unique_ptr<Table>> InMemoryCatalog::CreateTable(
    const TableIdentifier&, const Schema&, const PartitionSpec&,
    const std::string&,
    const std::unordered_map<std::string, std::string>&) {
  return NotImplemented("create table");
}
```

src/iceberg/expression/literal.cc
HuaHuaY: I don't think we need to check the length is not smaller than 0.

src/iceberg/catalog/memory/in_memory_catalog.cc
Jinchul81: @HuaHuaY Let's compare the two options. I prefer to use `[[maybe_unused]]` rather than omitting the parameter name. Adding the attribute can bring better readability and maintenance when enabling parameters.

src/iceberg/schema.h
Jinchul81: Let me keep this change because this commit is to refactor the code.

src/iceberg/expression/literal.cc
Jinchul81: Okay, let me assume the rhs will never have a negative value.

src/iceberg/catalog/memory/in_memory_catalog.cc
HuaHuaY: According to C++ core guidelines https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-unused, `[[maybe_unused]]` is used when parameters are conditionally unused. 
I think it's the responsibility of who implements this feature in the future to add the name back. If we don't need the variable, we should not give a name to it.

src/iceberg/schema.h
HuaHuaY: I didn't make it clear. This and the previous comments are just to remind wgtmac. You don't need to modify these.

src/iceberg/catalog/rest/types.h
wgtmac: I'd suggest adding `-Wno-missing-field-initializers` to suppress such kind of warnings.

src/iceberg/catalog/memory/in_memory_catalog.cc
wgtmac: I've seen several patterns regarding to this:

```
void func_a([[may_unused]] int a);
void func_b(int /*a*/);
void func_c(int);
```

I think `void func_b(int /*a*/)` strikes a good balance.

PR #334, opened 2025-11-20
feat: add PendingUpdate interface for table changes

shangxinli:
Add PendingUpdate<T> template class that provides a builder pattern API for constructing and committing table metadata changes. This interface follows the Java PendingUpdate design with C++ idioms:

- Apply() returns Result<T> with uncommitted changes for validation
- Commit() returns Status and atomically commits changes to the table
- Template parameter T allows type-safe results (Snapshot, Schema, etc.)

The interface is designed for future implementations like AppendFiles, UpdateSchema, UpdateProperties, and other table update operations.

### DIFF
--- a/src/iceberg/result.h
+++ b/src/iceberg/result.h
@@ -35,12 +35,11 @@ enum class ErrorKind {
   kDecompressError,
   kInvalid,  // For general invalid errors
   kInvalidArgument,
-  kValidationFailed,
   kInvalidArrowData,
   kInvalidExpression,
-  kInvalidSchema,
   kInvalidManifest,
   kInvalidManifestList,
+  kInvalidSchema,
   kIOError,
   kJsonParseError,
   kNoSuchNamespace,
@@ -50,6 +49,7 @@ enum class ErrorKind {
   kNotImplemented,
   kNotSupported,
   kUnknownError,
+  kValidationFailed,
 };
 
 /// \brief Error with a kind and a message.
@@ -85,12 +85,11 @@ DEFINE_ERROR_FUNCTION(CommitStateUnknown)
 DEFINE_ERROR_FUNCTION(DecompressError)
 DEFINE_ERROR_FUNCTION(Invalid)
 DEFINE_ERROR_FUNCTION(InvalidArgument)
-DEFINE_ERROR_FUNCTION(ValidationFailed)
 DEFINE_ERROR_FUNCTION(InvalidArrowData)
 DEFINE_ERROR_FUNCTION(InvalidExpression)
-DEFINE_ERROR_FUNCTION(InvalidSchema)
 DEFINE_ERROR_FUNCTION(InvalidManifest)
 DEFINE_ERROR_FUNCTION(InvalidManifestList)
+DEFINE_ERROR_FUNCTION(InvalidSchema)
 DEFINE_ERROR_FUNCTION(IOError)
 DEFINE_ERROR_FUNCTION(JsonParseError)
 DEFINE_ERROR_FUNCTION(NoSuchNamespace)
@@ -100,6 +99,7 @@ DEFINE_ERROR_FUNCTION(NotFound)
 DEFINE_ERROR_FUNCTION(NotImplemented)
 DEFINE_ERROR_FUNCTION(NotSupported)
 DEFINE_ERROR_FUNCTION(UnknownError)
+DEFINE_ERROR_FUNCTION(ValidationFailed)
 
 #undef DEFINE_ERROR_FUNCTION
 
--- a/src/iceberg/test/CMakeLists.txt
+++ b/src/iceberg/test/CMakeLists.txt
@@ -79,12 +79,12 @@ add_iceberg_test(schema_test
 
 add_iceberg_test(table_test
                  SOURCES
-                 test_common.cc
                  json_internal_test.cc
-                 table_test.cc
+                 pending_update_test.cc
                  schema_json_test.cc
                  table_metadata_builder_test.cc
-                 pending_update_test.cc)
+                 table_test.cc
+                 test_common.cc)
 
 add_iceberg_test(expression_test
                  SOURCES
--- a/src/iceberg/test/pending_update_test.cc
+++ b/src/iceberg/test/pending_update_test.cc
@@ -37,6 +37,7 @@ class MockPendingUpdate : public PendingUpdateTyped<MockSnapshot> {
     if (should_fail_) {
       return ValidationFailed("Mock validation failed");
     }
+    apply_called_ = true;
     return MockSnapshot{};
   }
 
@@ -44,7 +45,6 @@ class MockPendingUpdate : public PendingUpdateTyped<MockSnapshot> {
     if (should_fail_commit_) {
       return CommitFailed("Mock commit failed");
     }
-    apply_called_ = true;
     commit_called_ = true;
     return {};
   }


src/iceberg/pending_update.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /// \brief Apply and commit the pending changes to the table
+  ///
+  /// Changes are committed by calling the underlying table's commit operation.
+  ///
+  /// Once the commit is successful, the updated table will be refreshed.
+  ///
+  /// \return Status::OK if the commit was successful, or an error status if
+  ///         validation failed or the commit encountered conflicts
+  virtual Status Commit() = 0;
```
gty404, line 53: Define explicit Status codes for commit failure categories (validation failed, conflicts, commit-state-unknown) so callers can distinguish the same cases Java documents ?

src/iceberg/pending_update.h
```diff
@@ -0,0 +1,75 @@
...
+/// Iceberg table. Subclasses implement specific types of table updates such as
+/// schema changes, property updates, or snapshot-producing operations like
+/// appends and deletes.
+///
+/// Apply() can be used to validate and inspect the uncommitted changes before
+/// committing. Commit() applies the changes and commits them to the table.
+///
+/// \tparam T The type of result returned by Apply()
+template <typename T>
+class ICEBERG_EXPORT PendingUpdate {
```
gty404, line 39: Both PendingUpdateBase/PendingUpdateTyped\<T\>  and PendingUpdate\<T\> have advantages and disadvantages for the caller. Does iceberg-java have scenarios where PendingUpdate is used directly?

src/iceberg/pending_update.h
wgtmac: I think we have already added some. If not, we can extend it: https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/result.h#L31. Is this what you expected?

src/iceberg/pending_update.h
wgtmac: image

wgtmac: If we want to use the same approach in the `Transaction`, we need to make `PendingUpdate` non-templated.

src/iceberg/pending_update.h
gty404: My meaning is that the status code returned by the Commit interface is in what scenario, and if the status code is clearly named, it can follow the definition in result.h.

src/iceberg/pending_update.h
gty404: Yes, if we need to collect a list of PendingUpdate, we need PendingUpdate without template parameters. We can obtain the specific implementation class through type promotion.

wgtmac: Agreed.

src/iceberg/pending_update.h
shangxinli: make sense

src/iceberg/pending_update.h
shangxinli: Yeah, make sense

src/iceberg/pending_update.h
```diff
@@ -0,0 +1,75 @@
...
+  ///
+  /// \return Status::OK if the commit was successful, or an error status if
+  ///         validation failed or the commit encountered conflicts
+  virtual Status Commit() = 0;
+
+ protected:
+  PendingUpdate() = default;
+
+  // Non-copyable, movable
+  PendingUpdate(const PendingUpdate&) = delete;
```
Jinchul81, line 56: Please move it to the public space.

src/iceberg/pending_update.h
```diff
@@ -0,0 +1,75 @@
...
+  PendingUpdate& operator=(const PendingUpdate&) = delete;
```
Jinchul81, line 57: Please move it to the public space.

src/iceberg/pending_update.h
```diff
@@ -28,48 +28,67 @@
 
 namespace iceberg {
 
-/// \brief Base class for table metadata changes using builder pattern
```
Jinchul81: I don't prefer you to delete the context that you're using builder pattern. This is the reason why the ctors of this class are in the protected region, right?

Jinchul81 (commented):
Please don't forget to add unit tests for `PendingUpdate` in the follow up PRs.

src/iceberg/pending_update.h
shangxinli: ack

src/iceberg/pending_update.h
```diff
@@ -0,0 +1,94 @@
...
+
+/// \brief Non-template base class for table metadata changes using builder pattern
+///
+/// This base class allows storing different types of PendingUpdate operations
+/// in the same collection (e.g., in Transaction). It provides the common Commit()
+/// interface that all updates share.
+///
+/// This matches the Java Iceberg pattern where BaseTransaction stores a
+/// List<PendingUpdate> without type parameters.
+class ICEBERG_EXPORT PendingUpdateBase {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT PendingUpdate {
```

Should we use `PendingUpdate` as the base name? `std::vector<std::unique_ptr<PendingUpdate>>` looks nicer than `std::vector<std::unique_ptr<PendingUpdateBase>>` IMHO.

src/iceberg/pending_update.h
```diff
@@ -0,0 +1,94 @@
...
+/// returns the specific result type for each operation. Subclasses implement
+/// specific types of table updates such as schema changes, property updates, or
+/// snapshot-producing operations like appends and deletes.
+///
+/// Apply() can be used to validate and inspect the uncommitted changes before
+/// committing. Commit() applies the changes and commits them to the table.
+///
+/// \tparam T The type of result returned by Apply()
+template <typename T>
+class ICEBERG_EXPORT PendingUpdate : public PendingUpdateBase {
```
wgtmac: We can rename it to `BasePendingUpdate`.

src/iceberg/pending_update.h
shangxinli: sure

src/iceberg/pending_update.h
shangxinli: Rename it ti PendingUpdateTyped

src/iceberg/result.h
```diff
@@ -35,6 +35,7 @@ enum class ErrorKind {
   kDecompressError,
   kInvalid,  // For general invalid errors
   kInvalidArgument,
+  kValidationFailed,
```
wgtmac: Please sort it alphabetically.

src/iceberg/result.h
```diff
@@ -84,6 +85,7 @@ DEFINE_ERROR_FUNCTION(CommitStateUnknown)
 DEFINE_ERROR_FUNCTION(DecompressError)
 DEFINE_ERROR_FUNCTION(Invalid)
 DEFINE_ERROR_FUNCTION(InvalidArgument)
+DEFINE_ERROR_FUNCTION(ValidationFailed)
```
wgtmac: ditto

src/iceberg/test/CMakeLists.txt
```diff
@@ -83,7 +83,8 @@ add_iceberg_test(table_test
                  json_internal_test.cc
                  table_test.cc
                  schema_json_test.cc
-                 table_metadata_builder_test.cc)
+                 table_metadata_builder_test.cc
+                 pending_update_test.cc)
```
wgtmac: These files are unfortunately not sorted alphabetically. Could you help fix it as well?

wgtmac (approved):
Generally LGTM. I've left some minor comments. Thanks @shangxinli!

@gty404 @Jinchul81 Do you have more comments?

src/iceberg/test/pending_update_test.cc
```diff
@@ -0,0 +1,99 @@
...
+      return ValidationFailed("Mock validation failed");
+    }
+    return MockSnapshot{};
+  }
+
+  Status Commit() override {
+    if (should_fail_commit_) {
+      return CommitFailed("Mock commit failed");
+    }
+    apply_called_ = true;
```
HeartLinked: Should we remove `apply_called_ = true` here? I don't think we should assume that `Apply` has already been called by the user when `Commit` is invoked.I haven't seen any content in the comments that requires users to guarantee this.

HeartLinked (approved):
LGTM. I only have a minor question and comment.

src/iceberg/result.h
shangxinli: sure

src/iceberg/result.h
shangxinli: sure

src/iceberg/test/CMakeLists.txt
shangxinli: sure

src/iceberg/test/pending_update_test.cc
shangxinli: Good catch

Issue #330, opened 2025-11-20
Add aggregate expression


wgtmac:
Add `Aggregate` to the `Expression` family. The current expression implementation looks pretty much similar to the Java implementation. The work items would include:

- [ ] `Aggregate`
- [ ] `UnboundAggregate`
- [ ] `BoundAggregate`
- [ ] `CountAggregate`: include `CountNull`,`CountNotNull`,`CountStar`
- [ ] `ValueAggregate`: include `MaxAggregate` and `MinAggregate`
- [ ] `AggregateEvaluator`

SuKi2cn:
Hi, I’ve opened PR #335 implementing aggregate expressions and evaluator for this issue.
When you have time, I’d really appreciate any feedback on the design and implementation.

PR #335, opened 2025-11-21
feat: add aggregate expressions and evaluator

SuKi2cn:
This PR addresses issue #330 by introducing aggregate expressions & execution support.

* Add aggregate expression family (count / count_null / count_star / max / min)
  with bound/unbound types, visitor and binder support.

* Add `AggregateEvaluator` for count/max/min execution over `StructLike` rows.

* Expose aggregate factories in `Expressions` and wire into CMake/Meson builds
  with new aggregate tests.

### DIFF
--- a/src/iceberg/expression/aggregate.cc
+++ b/src/iceberg/expression/aggregate.cc
@@ -23,8 +23,7 @@
 #include <optional>
 #include <vector>
 
-#include "iceberg/exception.h"
-#include "iceberg/expression/binder.h"
+#include "iceberg/expression/literal.h"
 #include "iceberg/row/struct_like.h"
 #include "iceberg/type.h"
 #include "iceberg/util/checked_cast.h"
@@ -39,154 +38,122 @@ std::shared_ptr<PrimitiveType> GetPrimitiveType(const BoundTerm& term) {
   return internal::checked_pointer_cast<PrimitiveType>(term.type());
 }
 
-class CountNonNullAggregator : public BoundAggregate::Aggregator {
+class CountAggregator : public BoundAggregate::Aggregator {
  public:
-  explicit CountNonNullAggregator(std::shared_ptr<BoundTerm> term)
-      : term_(std::move(term)) {}
+  explicit CountAggregator(const CountAggregate& aggregate) : aggregate_(aggregate) {}
 
   Status Update(const StructLike& row) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto literal, term_->Evaluate(row));
-    if (!literal.IsNull()) {
-      ++count_;
-    }
+    ICEBERG_ASSIGN_OR_RAISE(auto count, aggregate_.CountFor(row));
+    count_ += count;
     return {};
   }
 
   Literal GetResult() const override { return Literal::Long(count_); }
 
  private:
-  std::shared_ptr<BoundTerm> term_;
-  int64_t count_ = 0;
-};
-
-class CountNullAggregator : public BoundAggregate::Aggregator {
- public:
-  explicit CountNullAggregator(std::shared_ptr<BoundTerm> term)
-      : term_(std::move(term)) {}
-
-  Status Update(const StructLike& row) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto literal, term_->Evaluate(row));
-    if (literal.IsNull()) {
-      ++count_;
-    }
-    return {};
-  }
-
-  Literal GetResult() const override { return Literal::Long(count_); }
-
- private:
-  std::shared_ptr<BoundTerm> term_;
-  int64_t count_ = 0;
-};
-
-class CountStarAggregator : public BoundAggregate::Aggregator {
- public:
-  Status Update(const StructLike& /*row*/) override {
-    ++count_;
-    return {};
-  }
-
-  Literal GetResult() const override { return Literal::Long(count_); }
-
- private:
+  const CountAggregate& aggregate_;
   int64_t count_ = 0;
 };
 
 class MaxAggregator : public BoundAggregate::Aggregator {
  public:
-  explicit MaxAggregator(std::shared_ptr<BoundTerm> term) : term_(std::move(term)) {}
+  explicit MaxAggregator(const MaxAggregate& aggregate)
+      : aggregate_(aggregate),
+        current_(Literal::Null(GetPrimitiveType(*aggregate_.term()))) {}
 
-  Status Update(const StructLike& row) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto val_literal, term_->Evaluate(row));
-    if (val_literal.IsNull()) {
+  Status Update(const StructLike& data) override {
+    ICEBERG_ASSIGN_OR_RAISE(auto value, aggregate_.Evaluate(data));
+    if (value.IsNull()) {
       return {};
     }
-    if (!current_) {
-      current_ = std::move(val_literal);
+    if (current_.IsNull()) {
+      current_ = std::move(value);
       return {};
     }
 
-    auto ordering = val_literal <=> *current_;
-    if (ordering == std::partial_ordering::unordered) {
-      return InvalidExpression("Cannot compare literals of type {}",
-                               val_literal.type()->ToString());
+    if (auto ordering = value <=> current_;
+        ordering == std::partial_ordering::unordered) {
+      return InvalidArgument("Cannot compare literal {} with current value {}",
+                             value.ToString(), current_.ToString());
+    } else if (ordering == std::partial_ordering::greater) {
+      current_ = std::move(value);
     }
 
-    if (ordering == std::partial_ordering::greater) {
-      current_ = std::move(val_literal);
-    }
     return {};
   }
 
-  Literal GetResult() const override {
-    return current_.value_or(Literal::Null(GetPrimitiveType(*term_)));
-  }
+  Literal GetResult() const override { return current_; }
 
  private:
-  std::shared_ptr<BoundTerm> term_;
-  std::optional<Literal> current_;
+  const MaxAggregate& aggregate_;
+  Literal current_;
 };
 
 class MinAggregator : public BoundAggregate::Aggregator {
  public:
-  explicit MinAggregator(std::shared_ptr<BoundTerm> term) : term_(std::move(term)) {}
+  explicit MinAggregator(const MinAggregate& aggregate)
+      : aggregate_(aggregate),
+        current_(Literal::Null(GetPrimitiveType(*aggregate_.term()))) {}
 
-  Status Update(const StructLike& row) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto val_literal, term_->Evaluate(row));
-    if (val_literal.IsNull()) {
+  Status Update(const StructLike& data) override {
+    ICEBERG_ASSIGN_OR_RAISE(auto value, aggregate_.Evaluate(data));
+    if (value.IsNull()) {
       return {};
     }
-    if (!current_) {
-      current_ = std::move(val_literal);
+    if (current_.IsNull()) {
+      current_ = std::move(value);
       return {};
     }
 
-    auto ordering = val_literal <=> *current_;
-    if (ordering == std::partial_ordering::unordered) {
-      return InvalidExpression("Cannot compare literals of type {}",
-                               val_literal.type()->ToString());
-    }
-
-    if (ordering == std::partial_ordering::less) {
-      current_ = std::move(val_literal);
+    if (auto ordering = value <=> current_;
+        ordering == std::partial_ordering::unordered) {
+      return InvalidArgument("Cannot compare literal {} with current value {}",
+                             value.ToString(), current_.ToString());
+    } else if (ordering == std::partial_ordering::less) {
+      current_ = std::move(value);
     }
     return {};
   }
 
-  Literal GetResult() const override {
-    return current_.value_or(Literal::Null(GetPrimitiveType(*term_)));
-  }
+  Literal GetResult() const override { return current_; }
 
  private:
-  std::shared_ptr<BoundTerm> term_;
-  std::optional<Literal> current_;
+  const MinAggregate& aggregate_;
+  Literal current_;
 };
 
 }  // namespace
 
-// -------------------- Bound aggregates --------------------
-
-std::string BoundAggregate::ToString() const {
-  std::string term_name;
-  if (op() != Expression::Operation::kCountStar) {
-    ICEBERG_DCHECK(term() != nullptr, "Bound aggregate should have term unless COUNT(*)");
-    term_name = term()->reference()->name();
-  }
+template <TermType T>
+std::string Aggregate<T>::ToString() const {
+  ICEBERG_DCHECK(IsSupportedOp(op()), "Unexpected aggregate operation");
+  ICEBERG_DCHECK(op() == Expression::Operation::kCountStar || term() != nullptr,
+                 "Aggregate term should not be null except for COUNT(*)");
 
   switch (op()) {
     case Expression::Operation::kCount:
-      return std::format("count({})", term_name);
+      return std::format("count({})", term()->ToString());
     case Expression::Operation::kCountNull:
-      return std::format("count_null({})", term_name);
+      return std::format("count_if({} is null)", term()->ToString());
     case Expression::Operation::kCountStar:
       return "count(*)";
     case Expression::Operation::kMax:
-      return std::format("max({})", term_name);
+      return std::format("max({})", term()->ToString());
     case Expression::Operation::kMin:
-      return std::format("min({})", term_name);
+      return std::format("min({})", term()->ToString());
     default:
-      return "Aggregate";
+      return std::format("Invalid aggregate: {}", ::iceberg::ToString(op()));
   }
+}
+
+// -------------------- CountAggregate --------------------
+
+Result<Literal> CountAggregate::Evaluate(const StructLike& data) const {
+  return CountFor(data).transform([](int64_t count) { return Literal::Long(count); });
+}
+
+std::unique_ptr<BoundAggregate::Aggregator> CountAggregate::NewAggregator() const {
+  return std::unique_ptr<BoundAggregate::Aggregator>(new CountAggregator(*this));
 }
 
 CountNonNullAggregate::CountNonNullAggregate(std::shared_ptr<BoundTerm> term)
@@ -201,13 +168,9 @@ Result<std::unique_ptr<CountNonNullAggregate>> CountNonNullAggregate::Make(
       new CountNonNullAggregate(std::move(term)));
 }
 
-Result<Literal> CountNonNullAggregate::Evaluate(const StructLike& data) const {
-  ICEBERG_ASSIGN_OR_RAISE(auto literal, term()->Evaluate(data));
-  return Literal::Long(literal.IsNull() ? 0 : 1);
-}
-
-std::unique_ptr<BoundAggregate::Aggregator> CountNonNullAggregate::NewAggregator() const {
-  return std::unique_ptr<BoundAggregate::Aggregator>(new CountNonNullAggregator(term()));
+Result<int64_t> CountNonNullAggregate::CountFor(const StructLike& data) const {
+  return term()->Evaluate(data).transform(
+      [](const auto& val) { return val.IsNull() ? 0 : 1; });
 }
 
 CountNullAggregate::CountNullAggregate(std::shared_ptr<BoundTerm> term)
@@ -221,13 +184,9 @@ Result<std::unique_ptr<CountNullAggregate>> CountNullAggregate::Make(
   return std::unique_ptr<CountNullAggregate>(new CountNullAggregate(std::move(term)));
 }
 
-Result<Literal> CountNullAggregate::Evaluate(const StructLike& data) const {
-  ICEBERG_ASSIGN_OR_RAISE(auto literal, term()->Evaluate(data));
-  return Literal::Long(literal.IsNull() ? 1 : 0);
-}
-
-std::unique_ptr<BoundAggregate::Aggregator> CountNullAggregate::NewAggregator() const {
-  return std::unique_ptr<BoundAggregate::Aggregator>(new CountNullAggregator(term()));
+Result<int64_t> CountNullAggregate::CountFor(const StructLike& data) const {
+  return term()->Evaluate(data).transform(
+      [](const auto& val) { return val.IsNull() ? 1 : 0; });
 }
 
 CountStarAggregate::CountStarAggregate()
@@ -237,12 +196,8 @@ Result<std::unique_ptr<CountStarAggregate>> CountStarAggregate::Make() {
   return std::unique_ptr<CountStarAggregate>(new CountStarAggregate());
 }
 
-Result<Literal> CountStarAggregate::Evaluate(const StructLike& data) const {
-  return Literal::Long(1);
-}
-
-std::unique_ptr<BoundAggregate::Aggregator> CountStarAggregate::NewAggregator() const {
-  return std::unique_ptr<BoundAggregate::Aggregator>(new CountStarAggregator());
+Result<int64_t> CountStarAggregate::CountFor(const StructLike& /*data*/) const {
+  return 1;
 }
 
 MaxAggregate::MaxAggregate(std::shared_ptr<BoundTerm> term)
@@ -253,12 +208,11 @@ std::shared_ptr<MaxAggregate> MaxAggregate::Make(std::shared_ptr<BoundTerm> term
 }
 
 Result<Literal> MaxAggregate::Evaluate(const StructLike& data) const {
-  ICEBERG_ASSIGN_OR_RAISE(auto literal, term()->Evaluate(data));
-  return literal;
+  return term()->Evaluate(data);
 }
 
 std::unique_ptr<BoundAggregate::Aggregator> MaxAggregate::NewAggregator() const {
-  return std::unique_ptr<BoundAggregate::Aggregator>(new MaxAggregator(term()));
+  return std::unique_ptr<BoundAggregate::Aggregator>(new MaxAggregator(*this));
 }
 
 MinAggregate::MinAggregate(std::shared_ptr<BoundTerm> term)
@@ -269,12 +223,11 @@ std::shared_ptr<MinAggregate> MinAggregate::Make(std::shared_ptr<BoundTerm> term
 }
 
 Result<Literal> MinAggregate::Evaluate(const StructLike& data) const {
-  ICEBERG_ASSIGN_OR_RAISE(auto literal, term()->Evaluate(data));
-  return literal;
+  return term()->Evaluate(data);
 }
 
 std::unique_ptr<BoundAggregate::Aggregator> MinAggregate::NewAggregator() const {
-  return std::unique_ptr<BoundAggregate::Aggregator>(new MinAggregator(term()));
+  return std::unique_ptr<BoundAggregate::Aggregator>(new MinAggregator(*this));
 }
 
 // -------------------- Unbound binding --------------------
@@ -292,42 +245,25 @@ Result<std::shared_ptr<Expression>> UnboundAggregateImpl<B>::Bind(
 
   switch (this->op()) {
     case Expression::Operation::kCountStar:
-      return CountStarAggregate::Make().transform([](auto aggregate) {
-        return std::shared_ptr<CountStarAggregate>(std::move(aggregate));
-      });
+      return CountStarAggregate::Make();
     case Expression::Operation::kCount:
-      return CountNonNullAggregate::Make(std::move(bound_term))
-          .transform([](auto aggregate) {
-            return std::shared_ptr<CountNonNullAggregate>(std::move(aggregate));
-          });
+      return CountNonNullAggregate::Make(std::move(bound_term));
     case Expression::Operation::kCountNull:
-      return CountNullAggregate::Make(std::move(bound_term))
-          .transform([](auto aggregate) {
-            return std::shared_ptr<CountNullAggregate>(std::move(aggregate));
-          });
+      return CountNullAggregate::Make(std::move(bound_term));
     case Expression::Operation::kMax:
-    case Expression::Operation::kMin: {
-      if (!bound_term) {
-        return InvalidExpression("Aggregate requires a term");
-      }
-      if (!bound_term->type()->is_primitive()) {
-        return InvalidExpression("Aggregate requires primitive type, got {}",
-                                 bound_term->type()->ToString());
-      }
-      if (this->op() == Expression::Operation::kMax) {
-        return MaxAggregate::Make(std::move(bound_term));
-      }
+      return MaxAggregate::Make(std::move(bound_term));
+    case Expression::Operation::kMin:
       return MinAggregate::Make(std::move(bound_term));
-    }
     default:
-      return NotSupported("Unsupported aggregate operation");
+      return NotSupported("Unsupported aggregate operation: {}",
+                          ::iceberg::ToString(this->op()));
   }
 }
 
 template <typename B>
 Result<std::shared_ptr<UnboundAggregateImpl<B>>> UnboundAggregateImpl<B>::Make(
     Expression::Operation op, std::shared_ptr<UnboundTerm<B>> term) {
-  if (!IsSupportedOp(op)) {
+  if (!Aggregate<UnboundTerm<B>>::IsSupportedOp(op)) {
     return NotSupported("Unsupported aggregate operation: {}", ::iceberg::ToString(op));
   }
   if (op != Expression::Operation::kCountStar && !term) {
@@ -338,31 +274,8 @@ Result<std::shared_ptr<UnboundAggregateImpl<B>>> UnboundAggregateImpl<B>::Make(
       new UnboundAggregateImpl<B>(op, std::move(term)));
 }
 
-template <typename B>
-std::string UnboundAggregateImpl<B>::ToString() const {
-  ICEBERG_DCHECK(UnboundAggregateImpl<B>::IsSupportedOp(this->op()),
-                 "Unexpected aggregate operation");
-  ICEBERG_DCHECK(
-      this->op() == Expression::Operation::kCountStar || this->term() != nullptr,
-      "Aggregate term should not be null unless COUNT(*)");
-
-  auto term_str = this->term() ? this->term()->ToString() : std::string{};
-  switch (this->op()) {
-    case Expression::Operation::kCount:
-      return std::format("count({})", term_str);
-    case Expression::Operation::kCountNull:
-      return std::format("count_if({} is null)", term_str);
-    case Expression::Operation::kCountStar:
-      return "count(*)";
-    case Expression::Operation::kMax:
-      return std::format("max({})", term_str);
-    case Expression::Operation::kMin:
-      return std::format("min({})", term_str);
-    default:
-      return "Aggregate";
-  }
-}
-
+template class Aggregate<UnboundTerm<BoundReference>>;
+template class Aggregate<BoundTerm>;
 template class UnboundAggregateImpl<BoundReference>;
 
 // -------------------- AggregateEvaluator --------------------
@@ -376,9 +289,9 @@ class AggregateEvaluatorImpl : public AggregateEvaluator {
       std::vector<std::unique_ptr<BoundAggregate::Aggregator>> aggregators)
       : aggregates_(std::move(aggregates)), aggregators_(std::move(aggregators)) {}
 
-  Status Update(const StructLike& row) override {
+  Status Update(const StructLike& data) override {
     for (auto& aggregator : aggregators_) {
-      ICEBERG_RETURN_UNEXPECTED(aggregator->Update(row));
+      ICEBERG_RETURN_UNEXPECTED(aggregator->Update(data));
     }
     return {};
   }
--- a/src/iceberg/expression/aggregate.h
+++ b/src/iceberg/expression/aggregate.h
@@ -23,7 +23,6 @@
 /// Aggregate expression definitions.
 
 #include <memory>
-#include <optional>
 #include <span>
 #include <string>
 #include <vector>
@@ -45,9 +44,18 @@ class ICEBERG_EXPORT Aggregate : public virtual Expression {
 
   const std::shared_ptr<T>& term() const { return term_; }
 
+  std::string ToString() const override;
+
  protected:
   Aggregate(Expression::Operation op, std::shared_ptr<T> term)
       : operation_(op), term_(std::move(term)) {}
+
+  static constexpr bool IsSupportedOp(Expression::Operation op) {
+    return op == Expression::Operation::kCount ||
+           op == Expression::Operation::kCountNull ||
+           op == Expression::Operation::kCountStar || op == Expression::Operation::kMax ||
+           op == Expression::Operation::kMin;
+  }
 
   Expression::Operation operation_;
   std::shared_ptr<T> term_;
@@ -79,19 +87,10 @@ class ICEBERG_EXPORT UnboundAggregateImpl : public UnboundAggregate,
   Result<std::shared_ptr<Expression>> Bind(const Schema& schema,
                                            bool case_sensitive) const override;
 
-  std::string ToString() const override;
-
  private:
-  static constexpr bool IsSupportedOp(Expression::Operation op) {
-    return op == Expression::Operation::kCount ||
-           op == Expression::Operation::kCountNull ||
-           op == Expression::Operation::kCountStar || op == Expression::Operation::kMax ||
-           op == Expression::Operation::kMin;
-  }
-
   UnboundAggregateImpl(Expression::Operation op, std::shared_ptr<UnboundTerm<B>> term)
       : BASE(op, std::move(term)) {
-    ICEBERG_DCHECK(IsSupportedOp(op), "Unexpected aggregate operation");
+    ICEBERG_DCHECK(BASE::IsSupportedOp(op), "Unexpected aggregate operation");
     ICEBERG_DCHECK(op == Expression::Operation::kCountStar || BASE::term() != nullptr,
                    "Aggregate term cannot be null except for COUNT(*)");
   }
@@ -103,34 +102,36 @@ class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound 
   using Aggregate<BoundTerm>::op;
   using Aggregate<BoundTerm>::term;
 
+  /// \brief Base class for aggregators.
   class Aggregator {
    public:
     virtual ~Aggregator() = default;
 
-    virtual Status Update(const StructLike& row) = 0;
+    virtual Status Update(const StructLike& data) = 0;
+
     virtual Status Update(const DataFile& file) {
-      return NotSupported("Aggregating DataFile not supported");
+      return NotImplemented("Update(DataFile) not implemented");
     }
+
+    /// \brief Get the result of the aggregation.
+    /// \return The result of the aggregation.
+    /// \note It is an undefined behavior to call this method if any previous Update call
+    /// has returned an error.
     virtual Literal GetResult() const = 0;
   };
 
   std::shared_ptr<BoundReference> reference() override {
-    ICEBERG_DCHECK(term_ != nullptr || op() == Expression::Operation::kCountStar,
+    ICEBERG_DCHECK(term() != nullptr || op() == Expression::Operation::kCountStar,
                    "Bound aggregate term should not be null except for COUNT(*)");
-    return term_ ? term_->reference() : nullptr;
+    return term() ? term()->reference() : nullptr;
   }
-  std::string ToString() const override;
 
   Result<Literal> Evaluate(const StructLike& data) const override = 0;
 
   bool is_bound_aggregate() const override { return true; }
 
-  enum class Kind : int8_t {
-    kCount = 0,
-    kValue,
-  };
-
-  virtual Kind kind() const = 0;
+  /// \brief Create a new aggregator for this aggregate.
+  /// \note The returned aggregator cannot outlive the BoundAggregate that creates it.
   virtual std::unique_ptr<Aggregator> NewAggregator() const = 0;
 
  protected:
@@ -141,7 +142,12 @@ class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound 
 /// \brief Base class for COUNT aggregates.
 class ICEBERG_EXPORT CountAggregate : public BoundAggregate {
  public:
-  Kind kind() const override { return Kind::kCount; }
+  Result<Literal> Evaluate(const StructLike& data) const final;
+
+  std::unique_ptr<Aggregator> NewAggregator() const override;
+
+  /// \brief Count for a single row. Subclasses implement this.
+  virtual Result<int64_t> CountFor(const StructLike& data) const = 0;
 
  protected:
   CountAggregate(Expression::Operation op, std::shared_ptr<BoundTerm> term)
@@ -154,8 +160,7 @@ class ICEBERG_EXPORT CountNonNullAggregate : public CountAggregate {
   static Result<std::unique_ptr<CountNonNullAggregate>> Make(
       std::shared_ptr<BoundTerm> term);
 
-  Result<Literal> Evaluate(const StructLike& data) const override;
-  std::unique_ptr<Aggregator> NewAggregator() const override;
+  Result<int64_t> CountFor(const StructLike& data) const override;
 
  private:
   explicit CountNonNullAggregate(std::shared_ptr<BoundTerm> term);
@@ -167,8 +172,7 @@ class ICEBERG_EXPORT CountNullAggregate : public CountAggregate {
   static Result<std::unique_ptr<CountNullAggregate>> Make(
       std::shared_ptr<BoundTerm> term);
 
-  Result<Literal> Evaluate(const StructLike& data) const override;
-  std::unique_ptr<Aggregator> NewAggregator() const override;
+  Result<int64_t> CountFor(const StructLike& data) const override;
 
  private:
   explicit CountNullAggregate(std::shared_ptr<BoundTerm> term);
@@ -179,8 +183,7 @@ class ICEBERG_EXPORT CountStarAggregate : public CountAggregate {
  public:
   static Result<std::unique_ptr<CountStarAggregate>> Make();
 
-  Result<Literal> Evaluate(const StructLike& data) const override;
-  std::unique_ptr<Aggregator> NewAggregator() const override;
+  Result<int64_t> CountFor(const StructLike& data) const override;
 
  private:
   CountStarAggregate();
@@ -191,9 +194,8 @@ class ICEBERG_EXPORT MaxAggregate : public BoundAggregate {
  public:
   static std::shared_ptr<MaxAggregate> Make(std::shared_ptr<BoundTerm> term);
 
-  Kind kind() const override { return Kind::kValue; }
-
   Result<Literal> Evaluate(const StructLike& data) const override;
+
   std::unique_ptr<Aggregator> NewAggregator() const override;
 
  private:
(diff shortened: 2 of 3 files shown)


src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,186 @@
...
+/// \brief Base class for aggregate expressions.
+class ICEBERG_EXPORT Aggregate : public Expression {
+ public:
+  ~Aggregate() override = default;
+
+  Expression::Operation op() const override { return operation_; }
+
+  bool is_unbound_aggregate() const override { return false; }
+  bool is_bound_aggregate() const override { return false; }
+
```
wgtmac:
```suggestion
```

We don't need to override them here.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,186 @@
...
+ protected:
+  BoundAggregate(Expression::Operation op, std::shared_ptr<BoundTerm> term)
+      : Aggregate(op), term_(std::move(term)) {}
+
+ private:
+  std::shared_ptr<BoundTerm> term_;
+};
+
+/// \brief COUNT aggregate variants.
+class ICEBERG_EXPORT CountAggregate : public UnboundAggregate {
```
wgtmac: We don't need to define subclass for unbound aggregates. Please check the current design of `Predicate` in both C++ and Java implementations.

SuKi2cn: Gotcha! In progress.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+#include <vector>
+
+#include "iceberg/expression/expression.h"
+#include "iceberg/expression/term.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+template <typename T>
+concept AggregateTermType = std::derived_from<T, Term>;
```
wgtmac: nit: `predicate.h` has defined a duplicate `concept TermType`. Perhaps we can move it to `term.h` and reuse it here?

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+};
+
+/// \brief Template for unbound aggregates that carry a term and operation.
+template <typename B>
+class ICEBERG_EXPORT UnboundAggregateImpl : public UnboundAggregate,
+                                            public Aggregate<UnboundTerm<B>> {
+  using BASE = Aggregate<UnboundTerm<B>>;
+
+ public:
+  enum class CountMode { kNonNull, kNull, kStar, kNone };
```
wgtmac: What is `kNone` for? We don't even need `CountMode`, do we? `UnboundAggregateImpl` should also support value mode.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+/// \brief Base aggregate holding an operation and a term.
+template <AggregateTermType T>
+class ICEBERG_EXPORT Aggregate : public virtual Expression {
+ public:
+  ~Aggregate() override = default;
+
+  Expression::Operation op() const override { return operation_; }
+
+  const std::shared_ptr<T>& term() const { return term_; }
+
```
wgtmac, line 46: Please also add a `std::string ToString() const` similar to the Java impl.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+#include "iceberg/row/struct_like.h"
+#include "iceberg/type.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
+
+std::string OperationToPrefix(Expression::Operation op) {
```
Jinchul81: Why don't we use `std::string_view` instead of `std::string` if the return values should be constant string literals?

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
   return unbound_transform;
 }
 
+// Aggregates
+
+std::shared_ptr<CountAggregate> Expressions::Count(std::string name) {
```
Jinchul81: You can change the parameter to `std::string&&` to avoid an unnecessary copy and clearly express that the function intends to take ownership of the string. Using an rvalue reference makes the API’s intent clearer and can reduce overhead in performance-critical paths.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+  return Count(Ref(std::move(name)));
+}
+
+std::shared_ptr<CountAggregate> Expressions::Count(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, CountAggregate::Count(std::move(expr)));
+  return std::shared_ptr<CountAggregate>(std::move(agg));
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountNull(std::string name) {
```
Jinchul81: Same above.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+  return CountNull(Ref(std::move(name)));
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountNull(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, CountAggregate::CountNull(std::move(expr)));
+  return std::shared_ptr<CountAggregate>(std::move(agg));
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountNotNull(std::string name) {
```
Jinchul81: Same above.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+  ICEBERG_ASSIGN_OR_THROW(auto agg, CountAggregate::Count(std::move(expr)));
+  return std::shared_ptr<CountAggregate>(std::move(agg));
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountStar() {
+  auto agg = CountAggregate::CountStar();
+  return std::shared_ptr<CountAggregate>(std::move(agg));
+}
+
+std::shared_ptr<ValueAggregate> Expressions::Max(std::string name) {
```
Jinchul81: Same above.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+  return Max(Ref(std::move(name)));
+}
+
+std::shared_ptr<ValueAggregate> Expressions::Max(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, ValueAggregate::Max(std::move(expr)));
+  return std::shared_ptr<ValueAggregate>(std::move(agg));
+}
+
+std::shared_ptr<ValueAggregate> Expressions::Min(std::string name) {
```
Jinchul81: Same above.

src/iceberg/expression/expression_visitor.h
```diff
@@ -77,6 +78,22 @@ class ICEBERG_EXPORT ExpressionVisitor {
   /// \brief Visit an unbound predicate.
   /// \param pred The unbound predicate to visit
   virtual Result<R> Predicate(const std::shared_ptr<UnboundPredicate>& pred) = 0;
+
+  /// \brief Visit a bound aggregate.
+  /// \param aggregate The bound aggregate to visit.
```
Jinchul81, line 84: Please add comment for the return.

src/iceberg/expression/expression_visitor.h
```diff
@@ -77,6 +78,22 @@ class ICEBERG_EXPORT ExpressionVisitor {
...
+  virtual Result<R> Aggregate(const std::shared_ptr<BoundAggregate>& aggregate) {
+    ICEBERG_DCHECK(aggregate != nullptr, "Bound aggregate cannot be null");
+    return NotSupported("Bound aggregate is not supported by this visitor: {}",
+                        aggregate->ToString());
+  }
+
+  /// \brief Visit an unbound aggregate.
+  /// \param aggregate The unbound aggregate to visit.
+  virtual Result<R> Aggregate(const std::shared_ptr<UnboundAggregate>& aggregate) {
```
Jinchul81, line 93: Please add comment for the return.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+      return {};
+    }
+
+    if (!current_) {
+      current_ = std::move(literal);
+      return {};
+    }
+
+    auto ordering = literal <=> *current_;
+    if (ordering == std::partial_ordering::unordered) {
```
Jinchul81: Please pull up this condition. We can exit early if the conditions are not satisfied.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+      return InvalidExpression("Cannot compare literals of type {}",
+                               literal.type()->ToString());
+    }
+
+    if (op_ == Expression::Operation::kMax) {
```
Jinchul81: Please add a comment why `current_` should be overwritten here.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+  return Min(Ref(std::move(name)));
+}
+
+std::shared_ptr<ValueAggregate> Expressions::Min(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, ValueAggregate::Min(std::move(expr)));
+  return std::shared_ptr<ValueAggregate>(std::move(agg));
```
Jinchul81: It looks like `return std::move(agg));` is enough.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+}
+
+std::shared_ptr<ValueAggregate> Expressions::Max(std::string name) {
+  return Max(Ref(std::move(name)));
+}
+
+std::shared_ptr<ValueAggregate> Expressions::Max(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, ValueAggregate::Max(std::move(expr)));
+  return std::shared_ptr<ValueAggregate>(std::move(agg));
```
Jinchul81: It looks like return `std::move(agg));` is enough.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+
+std::shared_ptr<CountAggregate> Expressions::CountNotNull(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, CountAggregate::Count(std::move(expr)));
+  return std::shared_ptr<CountAggregate>(std::move(agg));
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountStar() {
+  auto agg = CountAggregate::CountStar();
+  return std::shared_ptr<CountAggregate>(std::move(agg));
```
Jinchul81: It looks like return `std::move(agg));` is enough.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountNotNull(std::string name) {
+  return CountNotNull(Ref(std::move(name)));
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountNotNull(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, CountAggregate::Count(std::move(expr)));
+  return std::shared_ptr<CountAggregate>(std::move(agg));
```
Jinchul81: It looks like return `std::move(agg));` is enough.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountNull(std::string name) {
+  return CountNull(Ref(std::move(name)));
+}
+
+std::shared_ptr<CountAggregate> Expressions::CountNull(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, CountAggregate::CountNull(std::move(expr)));
+  return std::shared_ptr<CountAggregate>(std::move(agg));
```
Jinchul81: It looks like return `std::move(agg));` is enough.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,63 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+// Aggregates
+
+std::shared_ptr<CountAggregate> Expressions::Count(std::string name) {
+  return Count(Ref(std::move(name)));
+}
+
+std::shared_ptr<CountAggregate> Expressions::Count(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  ICEBERG_ASSIGN_OR_THROW(auto agg, CountAggregate::Count(std::move(expr)));
+  return std::shared_ptr<CountAggregate>(std::move(agg));
```
Jinchul81: It looks like return `std::move(agg));` is enough.

src/iceberg/test/aggregate_test.cc
```diff
@@ -0,0 +1,132 @@
...
+
+namespace {
+
+class VectorStructLike : public StructLike {
+ public:
+  explicit VectorStructLike(std::vector<Scalar> fields) : fields_(std::move(fields)) {}
+
+  Result<Scalar> GetField(size_t pos) const override {
+    if (pos >= fields_.size()) {
+      return InvalidArgument("Position {} out of range", pos);
```
Jinchul81, line 42: Please also show the number of fields to find out the cause easily.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,186 @@
...
+  bool is_unbound_aggregate() const override { return true; }
+
+  /// \brief Returns the unbound reference if the aggregate has a term.
+  virtual std::shared_ptr<NamedReference> reference() override = 0;
+
+ protected:
+  explicit UnboundAggregate(Expression::Operation op) : Aggregate(op) {}
+};
+
+/// \brief Bound aggregate with an optional term.
```
Jinchul81: What's the definition of term here? Why does aggregate optionally require term? Please put this into the comment.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,186 @@
...
+      : Aggregate(op), term_(std::move(term)) {}
+
+ private:
+  std::shared_ptr<BoundTerm> term_;
+};
+
+/// \brief COUNT aggregate variants.
+class ICEBERG_EXPORT CountAggregate : public UnboundAggregate {
+ public:
+  enum class Mode { kNonNull, kNull, kStar };
```
Jinchul81: Please add a comment for the description for them.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+      return "max";
+    case Expression::Operation::kMin:
+      return "min";
+    case Expression::Operation::kCount:
+    case Expression::Operation::kCountStar:
+      return "count";
+    default:
+      break;
+  }
+  return "aggregate";
```
Jinchul81: I don't think this is the right return value. Why do we have to map the others to `aggregate`?

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+}
+
+Result<std::unique_ptr<CountAggregate>> CountAggregate::CountNull(
+    std::shared_ptr<UnboundTerm<BoundReference>> term) {
+  auto ref = term->reference();
+  return std::unique_ptr<CountAggregate>(new CountAggregate(
+      Expression::Operation::kCount, Mode::kNull, std::move(term), std::move(ref)));
+}
+
+std::unique_ptr<CountAggregate> CountAggregate::CountStar() {
```
Jinchul81: Why does it return `std::unique_ptr<CountAggregate>` instead of `Result<std::unique_ptr<CountAggregate>>`?

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+  return std::unique_ptr<CountAggregate>(new CountAggregate(
+      Expression::Operation::kCountStar, Mode::kStar, nullptr, nullptr));
+}
+
+std::string CountAggregate::ToString() const {
+  if (mode_ == Mode::kStar) {
```
Jinchul81: Why did you special handling for `kStar`?

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+  return aggregate;
+}
+
+BoundCountAggregate::BoundCountAggregate(Expression::Operation op,
+                                         CountAggregate::Mode mode,
+                                         std::shared_ptr<BoundTerm> term)
+    : BoundAggregate(op, std::move(term)), mode_(mode) {}
+
+std::string BoundCountAggregate::ToString() const {
+  if (mode_ == CountAggregate::Mode::kStar) {
```
Jinchul81: Why did you special handling for `kStar`?

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,186 @@
...
+
+  std::shared_ptr<NamedReference> reference() override { return reference_; }
+
+  std::string ToString() const override;
+
+  Result<std::shared_ptr<Expression>> Bind(const Schema& schema,
+                                           bool case_sensitive) const override;
+
+ private:
+  CountAggregate(Expression::Operation op, Mode mode,
```
Jinchul81: Please add a comment why public ctor is not allowed.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Why did you special handling for `kStar`?

The special handling for `kStar` is intentional and mirrors the Java implementation.

In the Java version, `Aggregate.toString()` has a dedicated branch for `Operation.COUNT_STAR` that renders: `count(*)`

In the C++ design, the different COUNT variants are represented via `BoundCountAggregate::Mode`
(`kNonNull`, `kNull`, `kStar`). Therefore, this `kStar` special case is the equivalent of the Java
`COUNT_STAR` branch, ensuring we correctly print `count(*)` for `COUNT(*)` semantics.

In other words:
- `Mode::kStar`   → `COUNT(*)`
- `Mode::kNonNull` → `COUNT(column)`
- `Mode::kNull` → `COUNT_NULL(column)`

This preserves consistency with the Java behavior while adapting to the C++ expression model.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Why don't we use `std::string_view` instead of `std::string` if the return values should be constant string literals?

Good point — these return values are all string literals with static lifetime, so `std::string_view` is more appropriate here. I'll update the signature accordingly.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Why did you special handling for `kStar`?

Good catch — mapping all remaining operations to `"aggregate"` is not really correct.

In the Java implementation, `Aggregate.toString()` handles each aggregate operation explicitly and the `default` branch throws an `UnsupportedOperationException("Invalid aggregate: " + op())`, so there is no generic `"aggregate"` fallback.

I’ll update the C++ helper to mirror that behavior: only the supported aggregate operations will be handled explicitly, and for anything else we’ll treat it as invalid/unreachable instead of returning `"aggregate"`. This way we don’t silently hide unsupported operations and stay consistent with the Java implementation.

src/iceberg/expression/aggregate.cc
Jinchul81: Please make the return type `constexpr std::string_view`.

src/iceberg/expression/aggregate.cc
Jinchul81: I am still curious the special handling should be required in not only C++ but also Java. Can you please try to apply `kStar` in the generic manner? I don't accept we just follow the implementation of the Java version.

src/iceberg/expression/aggregate.cc
Jinchul81: Let's set aside the idea of copying the Java implementation. Can you please clarify why we cannot make it as the generic approach?

SuKi2cn: > Let's set aside the idea of copying the Java implementation. Can you please clarify why we cannot make it as the generic approach?

Got it, thanks for clarifying.

I understand your point. The goal here is not to blindly mirror the Java implementation, but to design a more generic and idiomatic C++ solution.

I'll take some time to revisit the current design and see how `kStar` can be modeled in a more generic way (e.g. reducing special-casing in `ToString()` and improving consistency with the Predicate style), while still preserving the same external semantics.

I’ll follow up with an updated implementation once I’ve reworked this part.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+    case Mode::kStar:
+      break;
+  }
+  std::unreachable();
+}
+
+Result<std::shared_ptr<Expression>> CountAggregate::Bind(const Schema& schema,
+                                                         bool case_sensitive) const {
+  std::shared_ptr<BoundTerm> bound_term;
+  if (term_ != nullptr) {
```
Jinchul81: Please write the code like `if (term_ != nullptr) [[unlikely]] {` to help branch prediction.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,289 @@
...
+};
+
+}  // namespace
+
+Result<std::unique_ptr<AggregateEvaluator>> AggregateEvaluator::Make(
+    std::shared_ptr<BoundAggregate> aggregate) {
+  ICEBERG_DCHECK(aggregate != nullptr, "Aggregate cannot be null");
+
+  if (auto count = std::dynamic_pointer_cast<BoundCountAggregate>(aggregate)) {
+    if (count->mode() != CountAggregate::Mode::kStar && !count->term()) {
```
Jinchul81: Please write the code like `if (count->mode() != CountAggregate::Mode::kStar && !count->term()) [[unlikely]] {` to help branch prediction.

src/iceberg/test/aggregate_test.cc
```diff
@@ -0,0 +1,132 @@
...
+namespace iceberg {
+
+namespace {
+
+class VectorStructLike : public StructLike {
+ public:
+  explicit VectorStructLike(std::vector<Scalar> fields) : fields_(std::move(fields)) {}
+
+  Result<Scalar> GetField(size_t pos) const override {
+    if (pos >= fields_.size()) {
```
Jinchul81, line 41: Please write the code like `if (pos >= fields_.size()) [[unlikely]] {` to help branch prediction.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+ public:
+  ~Aggregate() override = default;
+
+  Expression::Operation op() const override { return operation_; }
+
+  const std::shared_ptr<T>& term() const { return term_; }
+
+ protected:
+  Aggregate(Expression::Operation op, std::shared_ptr<T> term)
+      : operation_(op), term_(std::move(term)) {}
```
wgtmac, line 51: nit: add `ICEBERG_DCHECK(term != nullptr)` when `op` is not count star.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+
+ public:
+  enum class CountMode { kNonNull, kNull, kStar, kNone };
+
+  UnboundAggregateImpl(Expression::Operation op, std::shared_ptr<UnboundTerm<B>> term,
+                       CountMode count_mode = CountMode::kNone)
+      : BASE(op, std::move(term)), count_mode_(count_mode) {}
+
+  std::shared_ptr<NamedReference> reference() override {
+    return BASE::term() ? BASE::term()->reference() : nullptr;
```
wgtmac, line 84: nit: add a `ICEBERG_DCHECK` for expected `op`

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+class ICEBERG_EXPORT UnboundAggregateImpl : public UnboundAggregate,
+                                            public Aggregate<UnboundTerm<B>> {
+  using BASE = Aggregate<UnboundTerm<B>>;
+
+ public:
+  enum class CountMode { kNonNull, kNull, kStar, kNone };
+
+  UnboundAggregateImpl(Expression::Operation op, std::shared_ptr<UnboundTerm<B>> term,
+                       CountMode count_mode = CountMode::kNone)
+      : BASE(op, std::move(term)), count_mode_(count_mode) {}
```
wgtmac: Please add a static `Make` function and use private constructor so we can make sure that created expression does not have invalid fields.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+};
+
+/// \brief Base class for bound aggregates.
+class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound {
+ public:
+  using Aggregate<BoundTerm>::op;
+  using Aggregate<BoundTerm>::term;
+
+  std::shared_ptr<BoundReference> reference() override {
+    return term_ ? term_->reference() : nullptr;
```
wgtmac: ditto for ICEBERG_DCHECK

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+
+  virtual Kind kind() const = 0;
+
+ protected:
+  BoundAggregate(Expression::Operation op, std::shared_ptr<BoundTerm> term)
+      : Aggregate<BoundTerm>(op, std::move(term)) {}
+};
+
+/// \brief Bound COUNT aggregate.
+class ICEBERG_EXPORT BoundCountAggregate : public BoundAggregate {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT CountAggregate : public BoundAggregate {
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+ public:
+  enum class Mode { kNonNull, kNull, kStar };
+
+  BoundCountAggregate(Expression::Operation op, Mode mode,
```
wgtmac: Same here. Make the ctor private and add a static `Make` function.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+                      std::shared_ptr<BoundTerm> term);
+
+  Mode mode() const { return mode_; }
+
+  Kind kind() const override { return Kind::kCount; }
+
+  std::string ToString() const override;
+  Result<Literal> Evaluate(const StructLike& data) const override;
```
wgtmac:
```suggestion

  Result<Literal> Evaluate(const StructLike& data) const override;
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+
+ private:
+  Mode mode_;
+};
+
+/// \brief Bound MAX/MIN aggregate.
+class ICEBERG_EXPORT BoundValueAggregate : public BoundAggregate {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT ValueAggregate : public BoundAggregate {
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+ public:
+  BoundValueAggregate(Expression::Operation op, std::shared_ptr<BoundTerm> term);
```
wgtmac: ditto

src/iceberg/expression/expressions.h
```diff
@@ -101,6 +102,48 @@ class ICEBERG_EXPORT Expressions {
   static std::shared_ptr<UnboundTransform> Transform(
       std::string name, std::shared_ptr<Transform> transform);
 
+  // Aggregates
+
+  /// \brief Create COUNT(col) aggregate.
```
wgtmac:
```suggestion
  /// \brief Create a COUNT aggregate for a field name.
```

Let's be consistent with the above comment style. Same for all below.

src/iceberg/expression/expressions.cc
```diff
@@ -81,6 +82,76 @@ std::shared_ptr<UnboundTransform> Expressions::Transform(
...
+// Aggregates
+
+std::shared_ptr<UnboundAggregateImpl<BoundReference>> Expressions::Count(
+    std::string name) {
+  return Count(Ref(std::move(name)));
+}
+
+std::shared_ptr<UnboundAggregateImpl<BoundReference>> Expressions::Count(
+    std::shared_ptr<UnboundTerm<BoundReference>> expr) {
+  auto agg = std::make_shared<UnboundAggregateImpl<BoundReference>>(
```
wgtmac: See my other comments. Let's add `Make` functions to these types and throw when there is any error. Same for other functions in this file.

src/iceberg/expression/expression_visitor.h
```diff
@@ -77,6 +78,22 @@ class ICEBERG_EXPORT ExpressionVisitor {
   /// \brief Visit an unbound predicate.
   /// \param pred The unbound predicate to visit
   virtual Result<R> Predicate(const std::shared_ptr<UnboundPredicate>& pred) = 0;
+
+  /// \brief Visit a bound aggregate.
+  /// \param aggregate The bound aggregate to visit.
+  virtual Result<R> Aggregate(const std::shared_ptr<BoundAggregate>& aggregate) {
+    ICEBERG_DCHECK(aggregate != nullptr, "Bound aggregate cannot be null");
+    return NotSupported("Bound aggregate is not supported by this visitor: {}",
```
wgtmac: The error message is misleading since we are printing the aggregate as the visitor name.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,256 @@
...
+
+namespace {
+
+Result<std::shared_ptr<PrimitiveType>> GetPrimitiveType(const BoundTerm& term) {
+  auto primitive = std::dynamic_pointer_cast<PrimitiveType>(term.type());
+  if (primitive == nullptr) {
+    return InvalidExpression("Aggregate requires primitive type, got {}",
+                             term.type()->ToString());
+  }
+  return primitive;
```
wgtmac:
```suggestion
  if (!term.type().is_primitive()) {
    return InvalidExpression("Aggregate requires primitive type, got {}",
                             term.type()->ToString());
  }
  return internal::checked_pointer_cast<PrimitiveType>(term.type());
```

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,256 @@
...
+      return std::format("count_null({})", term()->reference()->name());
+    case Mode::kNonNull:
+      return std::format("count({})", term()->reference()->name());
+    case Mode::kStar:
+      break;
+  }
+  std::unreachable();
+}
+
+Result<Literal> BoundCountAggregate::Evaluate(const StructLike& data) const {
```
wgtmac: What's the tradeoff between these two options:
- Current approach that uses a single `CountAggregate` for all kinds.
- Add dedicated subclass for CountNull, CountNotNull and CountStar.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,256 @@
...
+  std::unreachable();
+}
+
+template class UnboundAggregateImpl<BoundReference>;
+
+// -------------------- AggregateEvaluator --------------------
+
+namespace {
+
+class AggregateEvaluatorImpl : public AggregateEvaluator {
```
wgtmac, line 285: Why not following the Java impl of aggregator class hierarchy? If we need to add support for `DataFile`, these lines will be doubled with a lot of duplicate code.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,175 @@
...
+      std::shared_ptr<BoundAggregate> aggregate);
+
+  /// \brief Create an evaluator for multiple bound aggregates.
+  /// \param aggregates Aggregates to evaluate in one pass; order is preserved in
+  /// Results().
+  static Result<std::unique_ptr<AggregateEvaluator>> MakeList(
+      std::vector<std::shared_ptr<BoundAggregate>> aggregates);
+
+  /// \brief Add a row to the aggregate.
+  virtual Status Add(const StructLike& row) = 0;
```
wgtmac: Update?

src/iceberg/expression/aggregate.cc
wgtmac: Personally I prefer to mimic the Java class design if there is no compelling reason to deviate.

wgtmac (changes requested):

SuKi2cn:
@wgtmac @Jinchul81 Thanks for the review, I'll take some time to address these changes and follow up shortly.

src/iceberg/expression/aggregate.cc
SuKi2cn: Thanks for bringing this up.

Between the two options:
- a single CountAggregate keeps the class count smaller but relies more heavily on branching based on `op()`, which can make the semantics harder to follow;
- dedicated subclasses more closely match the Java design, improve semantic clarity, and reduce internal conditional logic.

Given that there isn’t a compelling reason to deviate, I’ll align with the Java approach and refactor this into dedicated subclasses for CountNull, CountNotNull and CountStar.

src/iceberg/expression/aggregate.cc
SuKi2cn: Thanks, that’s a good point.

src/iceberg/expression/aggregate.cc
Jinchul81: Please file an issue for the follow up item and add a comment including the reason why you've applied special handling and `TODO(#XYZ)`.

src/iceberg/expression/aggregate.h
wgtmac: This is not resolved.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+           op == Expression::Operation::kCountNull ||
+           op == Expression::Operation::kCountStar || op == Expression::Operation::kMax ||
+           op == Expression::Operation::kMin;
+  }
+
+  UnboundAggregateImpl(Expression::Operation op, std::shared_ptr<UnboundTerm<B>> term)
+      : BASE(op, std::move(term)) {
+    ICEBERG_DCHECK(IsSupportedOp(op), "Unexpected aggregate operation");
+    ICEBERG_DCHECK(op == Expression::Operation::kCountStar || BASE::term() != nullptr,
+                   "Aggregate term cannot be null unless COUNT(*)");
```
wgtmac:
```suggestion
                   "Aggregate term cannot be null except for COUNT(*)");
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+    virtual Status Update(const StructLike& row) = 0;
+    virtual Status Update(const DataFile& file) {
+      return NotSupported("Aggregating DataFile not supported");
+    }
+    virtual Result<Literal> ResultLiteral() const = 0;
+  };
+
+  std::shared_ptr<BoundReference> reference() override {
+    ICEBERG_DCHECK(term_ != nullptr || op() == Expression::Operation::kCountStar,
+                   "Bound aggregate term should not be null unless COUNT(*)");
```
wgtmac:
```suggestion
                   "Bound aggregate term should not be null except for COUNT(*)");
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+  std::string ToString() const override;
+  Result<Literal> Evaluate(const StructLike& data) const override;
+  Result<std::unique_ptr<Aggregator>> NewAggregator() const override;
+
+ private:
+  CountStarAggregate();
+};
+
+/// \brief Bound MAX/MIN aggregate.
+class ICEBERG_EXPORT ValueAggregate : public BoundAggregate {
```
wgtmac: Why not adding subclasses for MIN and MAX for ValueAggregate and ValueAggregator?

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+      std::vector<std::shared_ptr<BoundAggregate>> aggregates);
+
+  /// \brief Update aggregates with a row.
+  virtual Status Update(const StructLike& row) = 0;
+
+  /// \brief Final aggregated value.
+  virtual Result<std::vector<Literal>> Results() const = 0;
+
+  /// \brief Convenience accessor when only one aggregate is evaluated.
+  virtual Result<Literal> ResultLiteral() const = 0;
```
wgtmac:
```suggestion
  virtual Result<std::vector<Literal>> GetResults() const = 0;

  /// \brief Convenience accessor when only one aggregate is evaluated.
  virtual Result<Literal> GetResult() const = 0;
```

Perhaps this looks more consistent.

src/iceberg/expression/expression_visitor.h
```diff
@@ -77,6 +79,22 @@ class ICEBERG_EXPORT ExpressionVisitor {
   /// \brief Visit an unbound predicate.
   /// \param pred The unbound predicate to visit
   virtual Result<R> Predicate(const std::shared_ptr<UnboundPredicate>& pred) = 0;
+
+  /// \brief Visit a bound aggregate.
+  /// \param aggregate The bound aggregate to visit.
+  virtual Result<R> Aggregate(const std::shared_ptr<BoundAggregate>& aggregate) {
+    ICEBERG_DCHECK(aggregate != nullptr, "Bound aggregate cannot be null");
+    return NotSupported("Visitor {} does not support bound aggregate: {}",
+                        typeid(*this).name(), aggregate->ToString());
```
wgtmac:
```suggestion
    return NotSupported("Visitor {} does not support bound aggregate",
                        typeid(*this).name());
```

src/iceberg/expression/expression_visitor.h
```diff
@@ -77,6 +79,22 @@ class ICEBERG_EXPORT ExpressionVisitor {
...
+  }
+
+  /// \brief Visit an unbound aggregate.
+  /// \param aggregate The unbound aggregate to visit.
+  virtual Result<R> Aggregate(const std::shared_ptr<UnboundAggregate>& aggregate) {
+    ICEBERG_DCHECK(aggregate != nullptr, "Unbound aggregate cannot be null");
+    return NotSupported("Visitor {} does not support unbound aggregate: {}",
+                        typeid(*this).name(), aggregate->ToString());
```
wgtmac:
```suggestion
    ICEBERG_DCHECK(aggregate != nullptr, "Unbound aggregate cannot be null");
    return NotSupported("Visitor {} does not support unbound aggregate",
                        typeid(*this).name());
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+ public:
+  Kind kind() const override { return Kind::kCount; }
+
+ protected:
+  CountAggregate(Expression::Operation op, std::shared_ptr<BoundTerm> term)
+      : BoundAggregate(op, std::move(term)) {}
+};
+
+/// \brief COUNT(term) aggregate.
+class ICEBERG_EXPORT CountNonNullAggregate : public CountAggregate {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT CountNonNull : public CountAggregate {
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+
+  class ICEBERG_EXPORT Aggregator {
+   public:
+    virtual ~Aggregator() = default;
+
+    virtual Status Update(const StructLike& row) = 0;
+    virtual Status Update(const DataFile& file) {
+      return NotSupported("Aggregating DataFile not supported");
+    }
+    virtual Result<Literal> ResultLiteral() const = 0;
```
wgtmac:
```suggestion
    virtual Result<Literal> GetResult() const = 0;
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+  }
+};
+
+/// \brief Base class for bound aggregates.
+class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound {
+ public:
+  using Aggregate<BoundTerm>::op;
+  using Aggregate<BoundTerm>::term;
+
+  class ICEBERG_EXPORT Aggregator {
```
wgtmac:
```suggestion
  class Aggregator {
```

I think we don't need to add export for inner class.

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+
+  /// \brief Create an evaluator for a single bound aggregate.
+  /// \param aggregate The bound aggregate to evaluate across rows.
+  static Result<std::unique_ptr<AggregateEvaluator>> Make(
+      std::shared_ptr<BoundAggregate> aggregate);
+
+  /// \brief Create an evaluator for multiple bound aggregates.
+  /// \param aggregates Aggregates to evaluate in one pass; order is preserved in
+  /// Results().
+  static Result<std::unique_ptr<AggregateEvaluator>> MakeList(
```
wgtmac:
```suggestion
  static Result<std::unique_ptr<AggregateEvaluator>> Make(
```

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+      std::vector<std::shared_ptr<BoundAggregate>> aggregates);
+
+  /// \brief Update aggregates with a row.
+  virtual Status Update(const StructLike& row) = 0;
+
+  /// \brief Final aggregated value.
+  virtual Result<std::vector<Literal>> Results() const = 0;
```
wgtmac: Should we return `Result<std::span<const Literal>>`?

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,398 @@
...
+
+  Status Update(const StructLike& row) override {
+    ICEBERG_ASSIGN_OR_RAISE(auto literal, term_->Evaluate(row));
+    if (!literal.IsNull()) {
+      ++count_;
+    }
+    return {};
+  }
+
+  Result<Literal> ResultLiteral() const override { return Literal::Long(count_); }
```
wgtmac: Do we actually need to use `Result` to wrap the return type?

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,398 @@
...
+};
+
+}  // namespace
+
+// -------------------- Bound aggregates --------------------
+
+CountNonNullAggregate::CountNonNullAggregate(std::shared_ptr<BoundTerm> term)
+    : CountAggregate(Expression::Operation::kCount, std::move(term)) {}
+
+Result<std::shared_ptr<CountNonNullAggregate>> CountNonNullAggregate::Make(
```
wgtmac:
```suggestion
Result<std::unique_ptr<CountNonNullAggregate>> CountNonNullAggregate::Make(
```

wgtmac: Same for other Make functions.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,398 @@
...
+  if (op != Expression::Operation::kCountStar && !term) {
+    return InvalidExpression("Aggregate term cannot be null unless COUNT(*)");
+  }
+
+  return std::shared_ptr<UnboundAggregateImpl<B>>(
+      new UnboundAggregateImpl<B>(op, std::move(term)));
+}
+
+template <typename B>
+std::string UnboundAggregateImpl<B>::ToString() const {
```
wgtmac: Can we move this to `Aggregate::ToString()` so all other overrides can be removed.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,398 @@
...
+
+template <typename B>
+Result<std::shared_ptr<Expression>> UnboundAggregateImpl<B>::Bind(
+    const Schema& schema, bool case_sensitive) const {
+  ICEBERG_DCHECK(UnboundAggregateImpl<B>::IsSupportedOp(this->op()),
+                 "Unexpected aggregate operation");
+
+  std::shared_ptr<B> bound_term;
+  if (this->term()) {
+    ICEBERG_ASSIGN_OR_THROW(bound_term, this->term()->Bind(schema, case_sensitive));
```
wgtmac: We cannot throw in this function. Please use `ICEBERG_ASSIGN_OR_RAISE` for this. Same for below.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  switch (this->op()) {
+    case Expression::Operation::kCountStar: {
+      ICEBERG_ASSIGN_OR_THROW(auto aggregate, CountStarAggregate::Make());
+      return aggregate;
+    }
```
wgtmac:
```suggestion
    case Expression::Operation::kCountStar:
      return CountStarAggregate::Make();
```

Same for below.

src/iceberg/expression/aggregate.cc
```diff
@@ -0,0 +1,398 @@
...
+    case Expression::Operation::kCount: {
+      if (!bound_term) {
```
wgtmac: Shouldn't this be checked by CountNonNullAggregate::Make?

src/iceberg/expression/aggregate.h
```diff
@@ -0,0 +1,230 @@
...
+
+  enum class Kind : int8_t {
+    // Count aggregates (COUNT, COUNT_STAR, COUNT_NULL)
+    kCount = 0,
+    // Value aggregates (MIN, MAX)
+    kValue,
+  };
+
+  virtual Kind kind() const = 0;
+  virtual Result<std::unique_ptr<Aggregator>> NewAggregator() const = 0;
```
wgtmac: If it will not return any error, we can use `std::unique_ptr<Aggregator>` as its return type.

src/iceberg/test/aggregate_test.cc
```diff
@@ -0,0 +1,160 @@
...
+#include "iceberg/expression/expressions.h"
+#include "iceberg/row/struct_like.h"
+#include "iceberg/schema.h"
+#include "iceberg/test/matchers.h"
+#include "iceberg/type.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
```
wgtmac, line 33: These test cases look insufficient to me. I think we need to covert at least the following:

- Verify unbound aggregates of different operations are created successfully and invalid inputs are rejected.
- Verify binding unbound to bound aggregates work as expected for different operations.
- Verify ToString and other member functions work.
- Verify bound aggregates evaluation work as expected.
- Verify different evaluators work fine for all operations and single/multiple aggregate inputs.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Do we actually need to use `Result` to wrap the return type?

You are right. I’ll update the interface and implementations to return `Literal` directly instead.

src/iceberg/expression/aggregate.cc
SuKi2cn: > 



> Same for other Make functions.

Applied this suggestion and changed all `Make` functions to return `std::unique_ptr`, converting to `std::shared_ptr` at the call sites where shared ownership is actually required.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Same for below.

Since the `Make` now return `Result<std::unique_ptr<...>>`, I used `transform` here to convert the result to `std::shared_ptr<Expression>`, keeping ownership single until binding.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Shouldn't this be checked by CountNonNullAggregate::Make?

I can't agree more. This check belongs in `CountNonNullAggregate::Make`, since a non-null term is part of the aggregate’s own invariant. Moving it there centralizes validation, avoids duplicating the same logic in `Bind`, and ensures any future callers of `Make()` also get the correct error handling automatically.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Can we move this to `Aggregate::ToString()` so all other overrides can be removed.

Thanks, that makes sense. I’ve moved the common `ToString()` implementation into Aggregate so all aggregates share the same formatting logic, and removed the redundant overrides from the concrete classes.

src/iceberg/expression/aggregate.h
SuKi2cn: > If it will not return any error, we can use `std::unique_ptr<Aggregator>` as its return type.

Agreed. Since `NewAggregator` cannot fail, I’ve switched its return type to `std::unique_ptr<Aggregator>` and removed the `Result` wrapper.

src/iceberg/expression/aggregate.h
SuKi2cn: > Why not adding subclasses for MIN and MAX for ValueAggregate and ValueAggregator?

That makes sense. I’ll split the current `ValueAggregate` & `ValueAggregator` into separate `Min` & `Max` subclasses so that each operation has its own concrete type and we don’t need the `Kind` enum to distinguish them. This will also better match the Java implementation.

src/iceberg/expression/aggregate.h
SuKi2cn: > Should we return `Result<std::span<const Literal>>`?

Good suggestion. I’ve updated `Results()` to return `Result<std::span<const Literal>>` so callers get a lightweight read-only view without extra allocations or copies.

src/iceberg/test/aggregate_test.cc
SuKi2cn: > These test cases look insufficient to me. I think we need to covert at least the following:
> 
> * Verify unbound aggregates of different operations are created successfully and invalid inputs are rejected.
> * Verify binding unbound to bound aggregates work as expected for different operations.
> * Verify ToString and other member functions work.
> * Verify bound aggregates evaluation work as expected.
> * Verify different evaluators work fine for all operations and single/multiple aggregate inputs.

Thank you for your review. I agree the current test coverage is not sufficient.
I will expand the test cases to cover the scenarios you listed, including creation and validation of unbound aggregates, binding behavior across different operations, ToString and other member functions, as well as evaluation correctness for both single and multiple aggregates.

src/iceberg/test/aggregate_test.cc
```diff
@@ -0,0 +1,240 @@
...
+
+class VectorStructLike : public StructLike {
+ public:
+  explicit VectorStructLike(std::vector<Scalar> fields) : fields_(std::move(fields)) {}
```
wgtmac: Please note that `Scalar` uses std::string_view for string/binary types so it may have memory lifetime issue.

wgtmac (approved):
Thanks for working on this, @SuKi2cn! This looks good to me.

SuKi2cn:
> Thanks for working on this, @SuKi2cn! This looks good to me.

Thanks a lot for the review and guidance, @wgtmac ! I really appreciate it. Happy Thanksgiving!

PR #337, opened 2025-11-21
fix: build unit tests with shared library

HuaHuaY:


CMakeLists.txt
```diff
@@ -45,6 +45,21 @@ option(ICEBERG_BUILD_REST "Build rest catalog client" ON)
...
 
+if(ICEBERG_BUILD_SHARED)
+  set(ICEBERG_TEST_LINKAGE_DEFAULT "shared")
+else()
+  set(ICEBERG_TEST_LINKAGE_DEFAULT "static")
+endif()
+set(ICEBERG_TEST_LINKAGE ${ICEBERG_TEST_LINKAGE_DEFAULT} CACHE STRING "Linkage of Iceberg libraries with unit tests executables")
+if(ICEBERG_BUILD_TESTS)
+  if(ICEBERG_TEST_LINKAGE STREQUAL "shared" AND NOT ICEBERG_BUILD_SHARED)
+    message(FATAL_ERROR "If using ICEBERG_TEST_LINKAGE=shared, must also pass ICEBERG_BUILD_SHARED=on")
```
wgtmac:
```suggestion
    message(FATAL_ERROR "If using ICEBERG_TEST_LINKAGE=shared, must also pass ICEBERG_BUILD_SHARED=ON")
```

CMakeLists.txt
```diff
@@ -45,6 +45,21 @@ option(ICEBERG_BUILD_REST "Build rest catalog client" ON)
...
+  endif()
+  if(ICEBERG_TEST_LINKAGE STREQUAL "static" AND NOT ICEBERG_BUILD_STATIC)
+    message(FATAL_ERROR "If using ICEBERG_TEST_LINKAGE=static, must also pass ICEBERG_BUILD_STATIC=on")
```
wgtmac:
```suggestion
    message(FATAL_ERROR "If using ICEBERG_TEST_LINKAGE=static, must also pass ICEBERG_BUILD_STATIC=ON")
```

wgtmac (approved):

src/iceberg/test/CMakeLists.txt
```diff
@@ -52,12 +52,20 @@ function(add_iceberg_test test_name)
 
   target_sources(${test_name} PRIVATE ${ARG_SOURCES})
 
-  if(ARG_USE_BUNDLE)
-    target_link_libraries(${test_name} PRIVATE iceberg_bundle_static GTest::gtest_main
-                                               GTest::gmock)
+  target_link_libraries(${test_name} PRIVATE GTest::gtest_main GTest::gmock)
+
+  if("${ICEBERG_TEST_LINKAGE}" STREQUAL "shared")
```
zhjwpku, line 48: Could we just use `ICEBERG_BUILD_SHARED` and `ICEBERG_BUILD_STATIC` to control the linkage behavior? For example:
- If only `ICEBERG_BUILD_SHARED=ON`, link against the shared libraries.
- If only `ICEBERG_BUILD_STATIC=ON`, link against the static libraries.
- If both are enabled, default to linking against the shared libraries (as linking time will be shorter).

HuaHuaY: I just follow the cmakelist file in arrow https://github.com/apache/arrow/blob/main/cpp/cmake_modules/DefineOptions.cmake#L249

Both are fine for me.

PR #338, opened 2025-11-21
fix: shared library uses hidden visibility by default

HuaHuaY:


cmake_modules/IcebergBuildUtils.cmake
```diff
@@ -152,6 +152,10 @@ function(add_iceberg_lib LIB_NAME)
 
     string(TOUPPER ${LIB_NAME} VISIBILITY_NAME)
     target_compile_definitions(${LIB_NAME}_shared PRIVATE ${VISIBILITY_NAME}_EXPORTING)
+    set_target_properties(${LIB_NAME}_shared
```
zhjwpku, line 155: Just curious, can we use the following:

```
set(CMAKE_C_VISIBILITY_PRESET hidden)
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
```

I searched github, quite a few projects adopt the above, it seems more nature to me, but I might miss some details.

HuaHuaY: I follow the cmakelist of parquet.
https://github.com/apache/arrow/blob/main/cpp/CMakeLists.txt#L569

I'm not sure whether it's useful to import such an option.

wgtmac (approved):

PR #344, opened 2025-11-26
refactor: unify lazy init for table

wgtmac:
Introduced a `TableMetadataCache` to support lazy initialization of mappings for schemas, partition specs, sort orders and snapshots.

### DIFF
--- a/src/iceberg/table_metadata.cc
+++ b/src/iceberg/table_metadata.cc
@@ -22,7 +22,6 @@
 #include <algorithm>
 #include <chrono>
 #include <format>
-#include <ranges>
 #include <string>
 
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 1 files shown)


src/iceberg/table_metadata.cc
```diff
@@ -162,6 +162,70 @@ bool operator==(const TableMetadata& lhs, const TableMetadata& rhs) {
...
+}
+
+Result<TableMetadataCache::SnapshotsMapRef> TableMetadataCache::GetSnapshotsById() const {
+  return snapshot_map_.Get(metadata_);
+}
+
+Result<TableMetadataCache::SchemasMap> TableMetadataCache::InitSchemasMap(
+    const TableMetadata* metadata) {
+  SchemasMap schemas_map;
+  schemas_map.reserve(metadata->schemas.size());
```
HuaHuaY: Just FYI. We can use `std::ranges::to` in `#include <ranges>` and don't need to call `reserve`. For example:
```cpp
Result<TableMetadataCache::SchemasMap> TableMetadataCache::InitSchemasMap(
    const TableMetadata* metadata) {
  return metadata->schemas | std::views::transform([](auto& schema) {
           return std::pair(schema->schema_id().value(), schema);
         }) |
         std::ranges::to<SchemasMap>();
}
```

wgtmac: ../src/iceberg/table_metadata.cc: In static member function ‘static iceberg::Result<std::unordered_map<int, std::shared_ptr<iceberg::Schema>, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<const int, std::shared_ptr<iceberg::Schema> > > > > iceberg::TableMetadataCache::InitSchemasMap(const iceberg::TableMetadata*)’:
../src/iceberg/table_metadata.cc:194:23: error: ‘to’ is not a member of ‘std::ranges’
  194 |          std::ranges::to<SchemasMap>();
      |                       ^~
../src/iceberg/table_metadata.cc:194:36: error: expected primary-expression before ‘>’ token
  194 |          std::ranges::to<SchemasMap>();
      |                                    ^
../src/iceberg/table_metadata.cc:194:38: error: expected primary-expression before ‘)’ token
  194 |          std::ranges::to<SchemasMap>();
      |

HuaHuaY: Sad. `std::ranges::to` is supported by gcc 14. But we use gcc 13 in ci.
I test in my local environment. `vendoredarrow-src` can be compiled by gcc 14 but can not be compiled by clang base on gcc 14. The difference is cmake will generate different compile flags. Clang will report that `atomic_load` is deprecated.
Maybe we can upgrade gcc's version in ci from 13 to 14.

zhjwpku (approved):

wgtmac:
Thanks @HuaHuaY @HeartLinked @Fokko @zhjwpku for the review!

PR #354, opened 2025-11-27
feat: implement struct like for partition values

wgtmac:
- Added PartitionValues to extend StructLike
- Implemented PartitionMap
- Implemented PartitionSet

### DIFF
--- a/src/iceberg/partition_summary.cc
+++ b/src/iceberg/partition_summary.cc
@@ -82,11 +82,11 @@ Status PartitionSummary::Update(const PartitionValues& partition_values) {
   }
 
   for (size_t i = 0; i < partition_values.num_fields(); i++) {
+    ICEBERG_ASSIGN_OR_RAISE(auto val, partition_values.ValueAt(i));
     ICEBERG_ASSIGN_OR_RAISE(
-        auto literal,
-        partition_values.ValueAt(i)->get().CastTo(
-            internal::checked_pointer_cast<PrimitiveType>(field_stats_[i].type())));
-    ICEBERG_RETURN_UNEXPECTED(field_stats_[i].Update(literal));
+        auto lit, val.get().CastTo(internal::checked_pointer_cast<PrimitiveType>(
+                      field_stats_[i].type())));
+    ICEBERG_RETURN_UNEXPECTED(field_stats_[i].Update(lit));
   }
   return {};
 }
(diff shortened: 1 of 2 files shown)


src/iceberg/partition_summary.cc
```diff
@@ -74,16 +75,16 @@ PartitionSummary::PartitionSummary(const StructType& partition_type) {
...
-                           partition_values.size(), field_stats_.size());
+                           partition_values.num_fields(), field_stats_.size());
   }
 
-  for (size_t i = 0; i < partition_values.size(); i++) {
+  for (size_t i = 0; i < partition_values.num_fields(); i++) {
     ICEBERG_ASSIGN_OR_RAISE(
         auto literal,
-        partition_values[i].CastTo(
+        partition_values.ValueAt(i)->get().CastTo(
```
shangxinli: ValueAt() returns Result<std::reference_wrapper<const Literal>>, but  we're using ->get() directly. Since ValueAt() can return an error (e.g., if the index is out of bounds), we should check the Result before accessing the value. Otherwise, we might hit undefined behavior at runtime. We can something like: 
 ICEBERG_ASSIGN_OR_RAISE(auto partition_value_ref, partition_values.ValueAt(i));
  const auto& partition_value = partition_value_ref.get();

src/iceberg/util/partition_value_util.h
```diff
@@ -0,0 +1,218 @@
...
+
+  /// \brief Clear all entries from the map.
+  void clear() { map_.clear(); }
+
+  /// \brief Check if the map contains a key.
+  /// \param spec_id The partition spec ID.
+  /// \param values The partition values.
+  /// \return true if the key exists, false otherwise.
+  bool contains(int32_t spec_id, const PartitionValues& values) const {
+    return map_.contains(PartitionKey{spec_id, values});
```
shangxinli: In this case, every lookup will create a temporary PartitionKey{spec_id, values} which copies the entire PartitionValues vector? Consider heterogeneous lookup?

wgtmac: Good suggestion! Added transparent hash for heterogeneous lookup.

src/iceberg/partition_summary.cc
wgtmac: My bad.

shangxinli:
LGTM

Fokko:
Thanks @shangxinli @dongxiao1198, and @zhjwpku for the reviews 🙌

PR #357, opened 2025-11-27
feat: add inclusive metrics evaluator

dongxiao1198:
feat: add inclusive metrics evaluator

### DIFF
--- a/src/iceberg/expression/inclusive_metrics_evaluator.cc
+++ b/src/iceberg/expression/inclusive_metrics_evaluator.cc
@@ -86,7 +86,7 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
     if (ContainsNullsOnly(id)) {
       return kRowCannotMatch;
     }
-    if (std::dynamic_pointer_cast<BoundReference>(expr) == nullptr) {
+    if (dynamic_cast<const BoundReference*>(expr.get()) == nullptr) {
       return kRowsMightMatch;
     }
     auto it = data_file_.nan_value_counts.find(id);
@@ -97,13 +97,12 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
   }
 
   Result<bool> NotNaN(const std::shared_ptr<Bound>& expr) override {
-    if (std::dynamic_pointer_cast<BoundReference>(expr) == nullptr) {
+    if (dynamic_cast<const BoundReference*>(expr.get()) == nullptr) {
       // identity transforms are already removed by this time
       return kRowsMightMatch;
     }
 
     int32_t id = expr->reference()->field().field_id();
-
     if (ContainsNaNsOnly(id)) {
       return kRowCannotMatch;
     }
@@ -117,20 +116,18 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
     if (ContainsNullsOnly(id) || ContainsNaNsOnly(id)) {
       return kRowCannotMatch;
     }
-    ICEBERG_ASSIGN_OR_RAISE(auto lower_result, LowerBound(expr));
-    if (!lower_result.has_value() || lower_result.value().IsNull() ||
-        lower_result.value().IsNaN()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto lower, LowerBound(expr));
+    if (!lower.has_value() || lower->IsNull() || lower->IsNaN()) {
       // NaN indicates unreliable bounds. See the InclusiveMetricsEvaluator docs for more.
       return kRowsMightMatch;
     }
-    const auto& lower = lower_result.value();
 
     // this also works for transforms that are order preserving:
     // if a transform f is order preserving, a < b means that f(a) <= f(b).
     // because lower <= a for all values of a in the file, f(lower) <= f(a).
     // when f(lower) >= X then f(a) >= f(lower) >= X, so there is no a such that f(a) < X
     // f(lower) >= X means rows cannot match
-    if (lower >= lit) {
+    if (lower.value() >= lit) {
       return kRowCannotMatch;
     }
 
@@ -144,20 +141,18 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
       return kRowCannotMatch;
     }
 
-    ICEBERG_ASSIGN_OR_RAISE(auto lower_result, LowerBound(expr));
-    if (!lower_result.has_value() || lower_result.value().IsNull() ||
-        lower_result.value().IsNaN()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto lower, LowerBound(expr));
+    if (!lower.has_value() || lower->IsNull() || lower->IsNaN()) {
       // NaN indicates unreliable bounds. See the InclusiveMetricsEvaluator docs for more.
       return kRowsMightMatch;
     }
-    const auto& lower = lower_result.value();
 
     // this also works for transforms that are order preserving:
     // if a transform f is order preserving, a < b means that f(a) <= f(b).
     // because lower <= a for all values of a in the file, f(lower) <= f(a).
     // when f(lower) > X then f(a) >= f(lower) > X, so there is no a such that f(a) <= X
     // f(lower) > X means rows cannot match
-    if (lower > lit) {
+    if (lower.value() > lit) {
       return kRowCannotMatch;
     }
 
@@ -171,12 +166,11 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
       return kRowCannotMatch;
     }
 
-    ICEBERG_ASSIGN_OR_RAISE(auto upper_result, UpperBound(expr));
-    if (!upper_result.has_value() || upper_result.value().IsNull()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto upper, UpperBound(expr));
+    if (!upper.has_value() || upper->IsNull()) {
       return kRowsMightMatch;
     }
-    const auto& upper = upper_result.value();
-    if (upper <= lit) {
+    if (upper.value() <= lit) {
       return kRowCannotMatch;
     }
 
@@ -190,12 +184,11 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
       return kRowCannotMatch;
     }
 
-    ICEBERG_ASSIGN_OR_RAISE(auto upper_result, UpperBound(expr));
-    if (!upper_result.has_value() || upper_result.value().IsNull()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto upper, UpperBound(expr));
+    if (!upper.has_value() || upper->IsNull()) {
       return kRowsMightMatch;
     }
-    const auto& upper = upper_result.value();
-    if (upper < lit) {
+    if (upper.value() < lit) {
       return kRowCannotMatch;
     }
 
@@ -209,21 +202,18 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
       return kRowCannotMatch;
     }
 
-    ICEBERG_ASSIGN_OR_RAISE(auto lower_result, LowerBound(expr));
-    if (lower_result.has_value() && !lower_result.value().IsNull() &&
-        !lower_result.value().IsNaN()) {
-      const auto& lower = lower_result.value();
-      if (!lower.IsNaN() && lower > lit) {
+    ICEBERG_ASSIGN_OR_RAISE(auto lower, LowerBound(expr));
+    if (lower.has_value() && !lower->IsNull() && !lower->IsNaN()) {
+      if (lower.value() > lit) {
         return kRowCannotMatch;
       }
     }
 
-    ICEBERG_ASSIGN_OR_RAISE(auto upper_result, UpperBound(expr));
-    if (!upper_result.has_value() || upper_result.value().IsNull()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto upper, UpperBound(expr));
+    if (!upper.has_value() || upper->IsNull()) {
       return kRowsMightMatch;
     }
-    const auto& upper = upper_result.value();
-    if (upper < lit) {
+    if (upper.value() < lit) {
       return kRowCannotMatch;
     }
 
@@ -249,28 +239,25 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
       return kRowsMightMatch;
     }
 
-    ICEBERG_ASSIGN_OR_RAISE(auto lower_result, LowerBound(expr));
-    if (!lower_result.has_value() || lower_result.value().IsNull() ||
-        lower_result.value().IsNaN()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto lower, LowerBound(expr));
+    if (!lower.has_value() || lower->IsNull() || lower->IsNaN()) {
       // NaN indicates unreliable bounds. See the InclusiveMetricsEvaluator docs for more.
       return kRowsMightMatch;
     }
-    const auto& lower = lower_result.value();
     auto literals_view = literal_set | std::views::filter([&](const Literal& lit) {
-                           return lower <= lit;
+                           return lower.value() <= lit;
                          });
     // if all values are less than lower bound, rows cannot match
     if (literals_view.empty()) {
       return kRowCannotMatch;
     }
 
-    ICEBERG_ASSIGN_OR_RAISE(auto upper_result, UpperBound(expr));
-    if (!upper_result.has_value() || upper_result.value().IsNull()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto upper, UpperBound(expr));
+    if (!upper.has_value() || upper->IsNull()) {
       return kRowsMightMatch;
     }
-    const auto& upper = upper_result.value();
     auto filtered_view = literals_view | std::views::filter([&](const Literal& lit) {
-                           return upper >= lit;
+                           return upper.value() >= lit;
                          });
     // if remaining values are greater than upper bound, rows cannot match
     if (filtered_view.empty()) {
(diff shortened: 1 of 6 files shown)


src/iceberg/test/inclusive_metrics_evaluator_test.cc
```diff
@@ -0,0 +1,385 @@
...
+          bound_values[3] = Literal::Int(std::get<int32_t>(value)).Serialize().value();
+        } else if (key == "salary") {
+          bound_values[4] = Literal::Double(std::get<double>(value)).Serialize().value();
+        } else if (key == "active") {
+          bound_values[5] = Literal::Boolean(std::get<bool>(value)).Serialize().value();
+        }
+      }
+    };
+
+    auto data_file = std::make_shared<DataFile>();
```
HeartLinked, line 89: You can use designated initializers here.

dongxiao1198:
> Modify meson files too?

done

src/iceberg/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(ICEBERG_SOURCES
     expression/literal.cc
     expression/predicate.cc
     expression/rewrite_not.cc
+    expression/inclusive_metrics_evaluator.cc
```
wgtmac: sort alphabetically.

src/iceberg/expression/inclusive_metrics_evaluator.h
```diff
@@ -0,0 +1,75 @@
...
+
+class ICEBERG_EXPORT InclusiveMetricsEvaluator {
+ public:
+  /// \brief Make a inclusive metrics evaluator
+  ///
+  /// \param expr The expression to evaluate
+  /// \param schema The schema of the table
+  /// \param case_sensitive Whether field name matching is case-sensitive
+  static Result<std::unique_ptr<InclusiveMetricsEvaluator>> Make(
+      std::shared_ptr<Expression> expr, const std::shared_ptr<Schema>& schema,
```
wgtmac:
```suggestion
      std::shared_ptr<Expression> expr, const Schema& schema,
```

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+#include "iceberg/schema.h"
+#include "iceberg/transform.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace {
+static const bool ROWS_MIGHT_MATCH = true;
+static const bool ROWS_CANNOT_MATCH = false;
+static const int32_t IN_PREDICATE_LIMIT = 200;
```
wgtmac:
```suggestion
constexpr bool kRowsMightMatch = true;
constexpr bool kRowCannotMatch = false;
constexpr int32_t kInPredicateLimit = 200;
```

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    //  and unknown bound terms are not non-null preserving
+    return false;
+  }
+
+ private:
+  const DataFile& data_file_;
+};
+
+InclusiveMetricsEvaluator::InclusiveMetricsEvaluator(
+    const std::shared_ptr<Expression>& expr)
```
wgtmac:
```suggestion
    std::shared_ptr<Expression> expr)
```

We shouldn't move a const ref.

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto bound_expr,
+                          Binder::Bind(*schema, rewrite_expr, case_sensitive));
+  return std::unique_ptr<InclusiveMetricsEvaluator>(
+      new InclusiveMetricsEvaluator(std::move(bound_expr)));
+}
+
+Result<bool> InclusiveMetricsEvaluator::Eval(const DataFile& data_file) const {
+  if (data_file.record_count <= 0) {
+    return ROWS_CANNOT_MATCH;
+  }
```
wgtmac:
```suggestion
  if (data_file.record_count == 0) {
    return kRowsCannotMatch;
  }
  if (data_file.record_count < 0) {
    return kRowsMightMatch;
  }
```

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+  Result<bool> And(bool left_result, bool right_result) override {
+    return left_result && right_result;
+  }
+
+  Result<bool> Or(bool left_result, bool right_result) override {
+    return left_result || right_result;
+  }
+
+  Result<bool> IsNull(const std::shared_ptr<BoundTerm>& term) override {
+    if (IsNonNullPreserving(term)) {
```
wgtmac:
```suggestion
    // if the column has no null values, the expression cannot match
    if (IsNonNullPreserving(term)) {
```

Let's preserve the original comment

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+      }
+    }
+    return ROWS_MIGHT_MATCH;
+  }
+
+  Result<bool> NotNull(const std::shared_ptr<BoundTerm>& term) override {
+    // no need to check whether the field is required because binding evaluates that case
+    // if the column has no non-null values, the expression cannot match
+
+    // all terms are null preserving. see #isNullPreserving(Bound)
```
wgtmac, line 73:
```suggestion
    // if the column has no non-null values, the expression cannot match.
    // all terms are null preserving.
```

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+  }
+
+  Result<bool> Or(bool left_result, bool right_result) override {
+    return left_result || right_result;
+  }
+
+  Result<bool> IsNull(const std::shared_ptr<BoundTerm>& term) override {
+    if (IsNonNullPreserving(term)) {
+      // number of non-nulls is the same as for the ref
+      int id = term->reference()->field().field_id();
```
wgtmac:
```suggestion
      int32_t id = term->reference()->field().field_id();
```

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+  }
+
+  Result<bool> IsNaN(const std::shared_ptr<BoundTerm>& term) override {
+    // when there's no nanCounts information, but we already know the column only contains
+    // null, it's guaranteed that there's no NaN value
+    int id = term->reference()->field().field_id();
+    if (ContainsNullsOnly(id)) {
+      return ROWS_CANNOT_MATCH;
+    }
+    auto ptr = std::dynamic_pointer_cast<BoundReference>(term);
```
wgtmac: This may be slow. At least we should use `dynamic_cast<BoundReference*>(term.get())`. A better solution is to add some virtual functions to `Bound` so we can know exactly the type.

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    if (ptr == nullptr) {
+      return ROWS_MIGHT_MATCH;
+    }
+    if (!data_file_.nan_value_counts.empty() &&
+        data_file_.nan_value_counts.contains(id) &&
```
wgtmac: nit: use `find` to combine `contains` and `at`

src/iceberg/expression/inclusive_metrics_evaluator.cc
wgtmac: Same for below

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    return ROWS_MIGHT_MATCH;
+  }
+
+  Result<bool> Lt(const std::shared_ptr<BoundTerm>& term, const Literal& lit) override {
+    // all terms are null preserving. see #isNullPreserving(Bound)
+    int id = term->reference()->field().field_id();
+    if (ContainsNullsOnly(id) || ContainsNaNsOnly(id)) {
+      return ROWS_CANNOT_MATCH;
+    }
+    auto lower_result = LowerBound(term);
```
wgtmac: Isn't it an error if `!lower_result.has_value()`? Should we directly return it?

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    if (!lower_result.has_value() || lower_result.value().IsNaN()) {
```
wgtmac: We need to check `lower_result.value().IsNull()` here.

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+
+  Result<bool> LtEq(const std::shared_ptr<BoundTerm>& term, const Literal& lit) override {
+    // all terms are null preserving. see #isNullPreserving(Bound)
+    int id = term->reference()->field().field_id();
+    if (ContainsNullsOnly(id) || ContainsNaNsOnly(id)) {
+      return ROWS_CANNOT_MATCH;
+    }
+
+    auto lower_result = LowerBound(term);
+    if (!lower_result.has_value() || lower_result.value().IsNaN()) {
```
wgtmac: ditto

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+      // NaN indicates unreliable bounds. See the InclusiveMetricsEvaluator docs for more.
+      return ROWS_MIGHT_MATCH;
+    }
+    const auto& lower = lower_result.value();
+    std::vector<Literal> literals;
+    for (const auto& lit : literal_set) {
```
wgtmac: nit: use ranges to only execute these on views.

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    // because the bounds are not necessarily a min or max value, this cannot be answered
+    // using them. notIn(col, {X, ...}) with (X, Y) doesn't guarantee that X is a value in
+    // col.
+    return ROWS_MIGHT_MATCH;
+  }
+
+  Result<bool> StartsWith(const std::shared_ptr<BoundTerm>& term,
+                          const Literal& lit) override {
+    if (term->kind() == BoundTerm::Kind::kTransform &&
+        std::dynamic_pointer_cast<BoundTransform>(term)->transform()->transform_type() !=
```
wgtmac: use internal::checked_cast

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    int id = term->reference()->field().field_id();
+    if (ContainsNullsOnly(id)) {
+      return ROWS_CANNOT_MATCH;
+    }
+    if (lit.type()->type_id() != TypeId::kString) {
+      return ROWS_CANNOT_MATCH;
+    }
+    std::string prefix = get<std::string>(lit.value());
```
wgtmac:
```suggestion
    const std::string& prefix = get<std::string>(lit.value());
```

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+
+  bool ContainsNaNsOnly(int32_t id) {
+    return !data_file_.nan_value_counts.empty() &&
+           data_file_.nan_value_counts.contains(id) && !data_file_.value_counts.empty() &&
+           data_file_.value_counts.at(id) == data_file_.nan_value_counts.at(id);
+  }
+
+  Result<Literal> LowerBound(const std::shared_ptr<BoundTerm>& term) {
+    if (term->kind() == BoundTerm::Kind::kReference) {
+      return ParseLowerBound(*std::dynamic_pointer_cast<BoundReference>(term));
```
wgtmac: use internal::checked_cast

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+      ICEBERG_ASSIGN_OR_RAISE(auto upper, ParseLowerBound(*boundTransform.reference()));
+      ICEBERG_ASSIGN_OR_RAISE(auto transform_func,
+                              transform->Bind(boundTransform.reference()->type()));
+      return transform_func->Transform(upper);
+    }
+
+    return NotFound("Transform upper bound not found.");
+  }
+
+  // TODO(xiao.dong) handle extract lower and upper bounds
```
wgtmac: Remove these for now?

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    } else {
+      return NotFound("Upper bound not found.");
+    }
+  }
+
+  Result<Literal> ParseLowerBound(const BoundReference& ref) {
+    int id = ref.field().field_id();
+    auto type = ref.type();
+    if (!type->is_primitive()) {
+      return Invalid("Lower bound of non-primitive type is not supported.");
```
wgtmac: Return `InvalidArgument` or `NotSupported`?

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+    }
+    auto primitive_type = std::dynamic_pointer_cast<PrimitiveType>(type);
+    if (!data_file_.lower_bounds.empty() && data_file_.lower_bounds.contains(id)) {
+      ICEBERG_ASSIGN_OR_RAISE(
+          auto lower,
+          Literal::Deserialize(data_file_.lower_bounds.at(id), primitive_type));
+      return lower;
+    }
+
+    return NotFound("Lower bound not found.");
```
wgtmac: The `Error` status in the `Result` usually means an error has been found and cannot proceed. We can either change the return type to `Result<std::optional<Literal>>` or return `Literal::Null()` for a missing value (which is a valid case).

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+  Result<Literal> TransformLowerBound(BoundTransform& boundTransform) {
+    auto transform = boundTransform.transform();
+    if (transform->PreservesOrder()) {
+      ICEBERG_ASSIGN_OR_RAISE(auto lower, ParseLowerBound(*boundTransform.reference()));
+      ICEBERG_ASSIGN_OR_RAISE(auto transform_func,
+                              transform->Bind(boundTransform.reference()->type()));
+      return transform_func->Transform(lower);
+    }
+
+    return NotFound("Transform lower bound not found.");
```
wgtmac: Same here

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,544 @@
...
+
+  }
+*/
+
+  /** Returns true if the expression term produces a non-null value for non-null input. */
+  bool IsNonNullPreserving(const std::shared_ptr<BoundTerm>& term) {
+    if (term->kind() == BoundTerm::Kind::kReference) {
+      return true;
+    } else if (term->kind() == BoundTerm::Kind::kTransform) {
+      return std::dynamic_pointer_cast<BoundTransform>(term)
```
wgtmac: ditto

src/iceberg/test/inclusive_metrics_evaluator_test.cc
```diff
@@ -0,0 +1,385 @@
...
+#include "iceberg/expression/expressions.h"
+#include "iceberg/manifest_entry.h"
+#include "iceberg/schema.h"
+#include "iceberg/test/matchers.h"
+#include "iceberg/type.h"
+
+namespace iceberg {
+
+namespace {
+static const bool ROWS_MIGHT_MATCH = true;
```
wgtmac: ditto, use `k` prefix for constants.

src/iceberg/test/inclusive_metrics_evaluator_test.cc
```diff
@@ -0,0 +1,385 @@
...
+static const bool ROWS_CANNOT_MATCH = false;
+}  // namespace
+using TestVariant = std::variant<bool, int32_t, int64_t, double, std::string>;
+
+class InclusiveMetricsEvaluatorTest : public ::testing::Test {
```
wgtmac, line 44: Is it possible to port all cases from the Java `TestInclusiveMetricsEvaluator`? We did same thing for manifest reader/writer and evaluator to have better coverage.

dongxiao1198: I migrated all java case into this ut, but the cases I write before were kept anyway, the ut file is a little bit large, but the coverage should be better than java cases.

gty404 (approved):

src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -0,0 +1,546 @@
...
+      }
+    }
+
+    return std::nullopt;
+  }
+
+  Result<std::optional<Literal>> TransformUpperBound(BoundTransform& boundTransform) {
+    auto transform = boundTransform.transform();
+    if (transform->PreservesOrder()) {
+      ICEBERG_ASSIGN_OR_RAISE(auto upper, ParseLowerBound(*boundTransform.reference()));
```
wgtmac:
```suggestion
      ICEBERG_ASSIGN_OR_RAISE(auto upper, ParseUpperBound(*boundTransform.reference()));
```

We might need to port the TestInclusiveMetricsEvaluatorWithTransforms as well.

PR #358, opened 2025-11-28
feat: Add error collection pattern to PendingUpdate

shangxinli:
Implement error vector collection pattern in PendingUpdateTyped to allow builder methods to accumulate validation errors instead of failing fast. This enables users to see all validation issues at once rather than fixing them one by one.

Changes:
- Add protected error collection methods (AddError, CheckErrors, HasErrors, ClearErrors) to PendingUpdateTyped base class
- Add std::vector<Error> errors_ member to collect validation errors
- Update PendingUpdate documentation with usage examples
- Add 6 comprehensive unit tests demonstrating error collection

Pattern follows TableMetadataBuilder implementation.

All tests pass (11/11).

### DIFF
--- a/src/iceberg/pending_update.h
+++ b/src/iceberg/pending_update.h
@@ -73,33 +73,9 @@ class ICEBERG_EXPORT PendingUpdate {
 /// Apply() can be used to validate and inspect the uncommitted changes before
 /// committing. Commit() applies the changes and commits them to the table.
 ///
-/// Error Collection Pattern:
-/// Builder methods in subclasses should use AddError() to collect validation
-/// errors instead of returning immediately. The accumulated errors will be
-/// returned by Apply() or Commit() when they are called. This allows users
-/// to see all validation errors at once rather than fixing them one by one.
-///
-/// Example usage in a subclass:
-/// \code
-///   MyUpdate& SetProperty(std::string_view key, std::string_view value) {
-///     if (key.empty()) {
-///       AddError(ErrorKind::kInvalidArgument, "Property key cannot be empty");
-///       return *this;
-///     }
-///     // ... continue with normal logic
-///     return *this;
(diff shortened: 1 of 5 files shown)


src/iceberg/pending_update.h
```diff
@@ -89,6 +116,51 @@ class ICEBERG_EXPORT PendingUpdateTyped : public PendingUpdate {
...
+  ///
+  /// Errors are accumulated and will be returned by Apply() or Commit().
+  /// This allows builder methods to continue and collect all errors rather
+  /// than failing fast on the first error.
+  ///
+  /// \param kind The kind of error
+  /// \param message The error message
+  void AddError(ErrorKind kind, std::string message) {
+    errors_.emplace_back(kind, std::move(message));
+  }
```
wgtmac: nit: we can add `void AddError(Error err)` to directly add an existing error without change.

src/iceberg/pending_update.h
```diff
@@ -72,6 +75,30 @@ class ICEBERG_EXPORT PendingUpdate {
 /// Apply() can be used to validate and inspect the uncommitted changes before
 /// committing. Commit() applies the changes and commits them to the table.
 ///
+/// Error Collection Pattern:
```
wgtmac: Perhaps we can make it a separate class so that other builders can leverage this as well? For example, TableMetadataBuilder uses the same pattern: https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/table_metadata.cc#L279

src/iceberg/pending_update.h
shangxinli: sure

src/iceberg/pending_update.h
shangxinli: make sense

PR #363, opened 2025-11-28
feat: add update table properties support

WZhuo:


src/iceberg/metrics_config.cc
```diff
@@ -0,0 +1,49 @@
...
+    const std::unordered_map<std::string, std::string>& updates, const Schema& schema) {
+  for (const auto& [key, value] : updates) {
+    if (!key.starts_with(TableProperties::kMetricModeColumnConfPrefix)) {
+      continue;
+    }
+    auto field_name =
+        std::string_view(key).substr(TableProperties::kMetricModeColumnConfPrefix.size());
+    auto field = schema.FindFieldByName(field_name);
+    if (!field.has_value() || !field.value().has_value()) {
+      return InvalidArgument(
```
wgtmac:
```suggestion
      return ValidationFailed(
```

src/iceberg/test/CMakeLists.txt
```diff
@@ -160,6 +161,7 @@ if(ICEBERG_BUILD_BUNDLE)
 
   add_iceberg_test(scan_test USE_BUNDLE SOURCES file_scan_task_test.cc)
 
+  add_iceberg_test(metrics_test USE_BUNDLE SOURCES metrics_config_test.cc)
```
wgtmac: TBH I don't think it is worth adding a new test executable. Perhaps we can add it to the `table_test`.

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
```
wgtmac, line 20: We need to add `src/iceberg/update/CMakeLists.txt` to install headers. Its content should be like:

```
iceberg_install_all_headers(iceberg/update)
```

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+#include "iceberg/file_format.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/pending_update.h"
+#include "iceberg/table_identifier.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Updates table properties.
+class ICEBERG_EXPORT UpdateProperties : public PendingUpdateTyped<void> {
```
wgtmac:
```suggestion
class ICEBERG_EXPORT UpdateProperties : public PendingUpdateTyped<std::unordered_map<std::string, std::string>> {
```

Should `Apply` return a map instead of void?

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \return Reference to this UpdateProperties for chaining
+  UpdateProperties& Set(std::string key, std::string value);
+
+  /// \brief Marks a property for removal.
+  ///
+  /// If the key was previously set for update, this operation cancels the update.
+  ///
+  /// \param key The property key to remove
+  /// \return Reference to this UpdateProperties for chaining
+  UpdateProperties& Remove(std::string key);
```
wgtmac:
```suggestion
  UpdateProperties& Remove(const std::string& key);
```

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \return Reference to this UpdateProperties for chaining
+  UpdateProperties& DefaultFormat(FileFormatType format);
+
+  /// \brief Applies the property changes without committing them.
+  ///
+  /// Validates the pending property changes but does not commit them to the table.
+  /// This method can be used to validate changes before actually committing them.
+  ///
+  /// \return Status::OK if the changes are valid, or an error if validation fails
+  Result<void> Apply() override;
```
wgtmac: Should it return a map instead of void? It is fine to keep void if not consumed by callers.

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \return OK if the changes are valid and committed successfully, or an error
+  Status Commit() override;
+
+ private:
+  TableIdentifier identifier_;
+  std::shared_ptr<Catalog> catalog_;
+  std::shared_ptr<TableMetadata> metadata_;
+
+  std::unordered_map<std::string, std::string> updates_;
+  std::vector<std::string> removals_;
```
wgtmac: Should it be a unordered_set?

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+/// \brief Updates table properties.
+class ICEBERG_EXPORT UpdateProperties : public PendingUpdateTyped<void> {
+ public:
+  /// \brief Constructs a UpdateProperties for the specified table.
+  ///
+  /// \param identifier The table identifier
+  /// \param catalog The catalog containing the table
+  /// \param metadata The current table metadata
+  UpdateProperties(TableIdentifier identifier, std::shared_ptr<Catalog> catalog,
+                   std::shared_ptr<TableMetadata> metadata);
```
wgtmac:
```suggestion
  UpdateProperties(TableIdentifier identifier, Catalog* catalog,
                   TableMetadata* base);
```

Let's avoid unnecessary shared_ptr copies and make it explicit it is a base metadata. If catalog and base will never be nullptr, perhaps we can use `Catalog& catalog, TableMetadata& base`?

src/iceberg/update/update_properties.cc
```diff
@@ -0,0 +1,130 @@
...
+UpdateProperties::UpdateProperties(TableIdentifier identifier,
+                                   std::shared_ptr<Catalog> catalog,
+                                   std::shared_ptr<TableMetadata> metadata)
+    : identifier_(std::move(identifier)),
+      catalog_(std::move(catalog)),
+      metadata_(std::move(metadata)) {}
+
+UpdateProperties& UpdateProperties::Set(std::string key, std::string value) {
+  if (std::ranges::find(removals_, key) != std::ranges::end(removals_)) {
+    return *this;
```
wgtmac: We can call `return AddError(ErrorKind::kInvalidArgument, std::format("Cannot remove and update the same key: {}", key))`.

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \return Reference to this UpdateProperties for chaining
+  UpdateProperties& Remove(std::string key);
+
+  /// \brief Sets the default file format for the table.
+  ///
+  /// This is a convenience method for setting the "write.format.default" property.
+  ///
+  /// \param format The file format type to use as default
+  /// \return Reference to this UpdateProperties for chaining
+  UpdateProperties& DefaultFormat(FileFormatType format);
```
wgtmac: TBH, I don't think we need this special function.

src/iceberg/update/update_properties.cc
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  if (!TableProperties::reserved_properties().contains(key) ||
```
wgtmac, line 53: Why do we need this check?

src/iceberg/update/update_properties.cc
```diff
@@ -0,0 +1,130 @@
...
+      key == TableProperties::kFormatVersion.key()) {
+    updates_.emplace(std::move(key), std::move(value));
+  }
+
+  return *this;
+}
+
+UpdateProperties& UpdateProperties::Remove(std::string key) {
+  if (updates_.contains(key)) {
+    return *this;
```
wgtmac: Use AddError and return.

src/iceberg/update/update_properties.h
wgtmac: I think you are right. Usually `Apply()` is called internally in the `Commit()` where all internal states are accessible. Perhaps we can simply change the return type to `Status` for the base type and totally remove the `PendingUpdateTyped` subclass.

src/iceberg/update/update_properties.cc
```diff
@@ -0,0 +1,130 @@
...
+  if (format_version_.has_value()) {
+    updates.emplace_back(
+        std::make_unique<table::UpgradeFormatVersion>(format_version_.value()));
+  };
+
+  if (!updates.empty()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto requirements,
+                            TableRequirements::ForUpdateTable(*metadata_, updates));
+    ICEBERG_RETURN_UNEXPECTED(
+        catalog_->UpdateTable(identifier_, requirements, std::move(updates)));
```
wgtmac:
```suggestion
        catalog_->UpdateTable(identifier_, requirements, updates));
```

src/iceberg/metrics_config.cc
```diff
@@ -0,0 +1,49 @@
...
+
+Status MetricsConfig::VerifyReferencedColumns(
+    const std::unordered_map<std::string, std::string>& updates, const Schema& schema) {
+  for (const auto& [key, value] : updates) {
+    if (!key.starts_with(TableProperties::kMetricModeColumnConfPrefix)) {
+      continue;
+    }
+    auto field_name =
+        std::string_view(key).substr(TableProperties::kMetricModeColumnConfPrefix.size());
+    auto field = schema.FindFieldByName(field_name);
```
wgtmac, line 39:
```suggestion
    ICEBERG_ASSIGN_OR_RAISE(auto field, schema.FindFieldByName(field_name));
```

src/iceberg/metrics_config.cc
```diff
@@ -0,0 +1,49 @@
...
+    if (!field.has_value() || !field.value().has_value()) {
```
wgtmac, line 40:
```suggestion
    if (!field.has_value()) {
```

src/iceberg/table_properties.h
```diff
@@ -231,6 +231,8 @@ class ICEBERG_EXPORT TableProperties : public ConfigBase<TableProperties> {
       "write.metadata.delete-after-commit.enabled", false};
   inline static Entry<int32_t> kMetricsMaxInferredColumnDefaults{
       "write.metadata.metrics.max-inferred-column-defaults", 100};
+  static constexpr std::string_view kMetricModeColumnConfPrefix =
```
HeartLinked: We need inline here to avoid One Definition Rule (ODR) violations.

src/iceberg/table.h
```diff
@@ -104,6 +104,12 @@ class ICEBERG_EXPORT Table {
   /// \return a vector of history entries
   const std::vector<SnapshotLogEntry>& history() const;
 
+  /// \brief Create a new UpdateProperties to update table properties and commit the
+  /// changes
+  ///
+  /// \return a new UpdateProperties instance
+  virtual std::unique_ptr<iceberg::UpdateProperties> UpdateProperties() const;
```
HeartLinked, line 111:
```suggestion
  virtual std::unique_ptr<UpdateProperties> UpdateProperties() const;
```

src/iceberg/table.cc
```diff
@@ -110,6 +109,10 @@ const std::vector<SnapshotLogEntry>& Table::history() const {
   return metadata_->snapshot_log;
 }
 
+std::unique_ptr<UpdateProperties> Table::UpdateProperties() const {
+  return std::make_unique<iceberg::UpdateProperties>(identifier_, catalog_, metadata_);
```
HeartLinked, line 113:
```suggestion
  return std::make_unique<UpdateProperties>(identifier_, catalog_, metadata_);
```

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \param identifier The table identifier
+  /// \param catalog The catalog containing the table
+  /// \param metadata The current table metadata
+  UpdateProperties(TableIdentifier identifier, std::shared_ptr<Catalog> catalog,
+                   std::shared_ptr<TableMetadata> metadata);
+
+  /// \brief Sets a property key-value pair.
+  ///
+  /// If the key was previously marked for removal, this operation cancels the removal.
+  /// Reserved properties are handled according to TableProperties rules.
```
HeartLinked: What is "TableProperties rules"? 🤔

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+  /// \param key The property key to remove
+  /// \return Reference to this UpdateProperties for chaining
+  UpdateProperties& Remove(std::string key);
+
+  /// \brief Sets the default file format for the table.
+  ///
+  /// This is a convenience method for setting the "write.format.default" property.
+  ///
+  /// \param format The file format type to use as default
+  /// \return Reference to this UpdateProperties for chaining
```
HeartLinked:
```suggestion
  /// \brief Sets the default file format for the table.
  /// \note This method will setting the "write.format.default" property.
  /// \param format The file format type to use as default
  /// \return Reference to this UpdateProperties for chaining
```

src/iceberg/update/update_properties.h
```diff
@@ -0,0 +1,97 @@
...
+  ///
+  /// Validates the pending property changes but does not commit them to the table.
+  /// This method can be used to validate changes before actually committing them.
+  ///
+  /// \return Status::OK if the changes are valid, or an error if validation fails
+  Result<void> Apply() override;
+
+  /// \brief Commits the property changes to the table.
+  ///
+  /// Validates the changes and applies them to the table through the catalog.
```
HeartLinked, line 70: I think use `\note` would be better too.

src/iceberg/update/update_properties.cc
WZhuo: Follow java impl, reserved properties can not be set explicitly. The reserved properties are set according to the table state.

src/iceberg/metrics_config.cc
WZhuo: Need to transform the error returned, but not return the error directly

src/iceberg/update/update_properties.h
WZhuo: Follow the repo's comments style, there's no '\note' anywhere else?

src/iceberg/update/update_properties.h
WZhuo: I'm not sure how reserved properties are set, but it is certain that they cannot be explicitly set by users.

src/iceberg/update/update_properties.h
WZhuo: It's a convenience method.

src/iceberg/table_properties.h
WZhuo: fix it

src/iceberg/update/update_properties.h
WZhuo: I think so. Apply() return Status, and remove PendingUpdateTyped subclass. Fix it.

src/iceberg/test/CMakeLists.txt
WZhuo: ok

src/iceberg/table.cc
WZhuo: Duplicate name with the method, specify the namespace

src/iceberg/table.h
WZhuo: Duplicate class name and method name, specify the namespace

src/iceberg/update/update_properties.cc
WZhuo: fix it

src/iceberg/update/update_properties.cc
WZhuo: fix it

src/iceberg/update/update_properties.cc
WZhuo: fix it

src/iceberg/update/update_properties.h
WZhuo: fix it

src/iceberg/update/update_properties.h
WZhuo: I think it's better shared_ptr, so that UpdateProperties::Commit can be execute in different context.

src/iceberg/update/update_properties.h
WZhuo: It can be unordered_set. fix it

src/iceberg/metrics_config.cc
WZhuo: Both return ValidationFailed for !field.has_value(), !field.value().has_value()

src/iceberg/metrics_config.cc
wgtmac: Isn't the original error swallowed in this case?

src/iceberg/test/metrics_config_test.cc
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  {
+    // Invalid column reference
+    std::unordered_map<std::string, std::string> updates;
+    updates[std::string(TableProperties::kMetricModeColumnConfPrefix) + "nonexistent"] =
+        "counts";
+
+    auto result = MetricsConfig::VerifyReferencedColumns(updates, *schema_);
+    EXPECT_FALSE(result.has_value())
```
wgtmac: nit: you can `#include "iceberg/test/matchers.h"` and then write

```
EXPECT_THAT(result, IsError(ErrorKind::kValidationFailed));
EXPECT_THAT(result, HasErrorMessage("..."));
```

src/iceberg/test/update_properties_test.cc
```diff
@@ -0,0 +1,195 @@
...
+ */
+
+#include "iceberg/update/update_properties.h"
+
+#include <cstddef>
+#include <memory>
+
+#include <gtest/gtest.h>
+
+#include "gmock/gmock.h"
```
wgtmac:
```suggestion
#include <gmock/gmock.h>

```

src/iceberg/test/update_properties_test.cc
```diff
@@ -0,0 +1,195 @@
...
+}
+
+TEST_F(UpdatePropertiesTest, SetRemoveConflict) {
+  {
+    // Set a property that is already marked for removal
+    UpdateProperties update(identifier_, catalog_, metadata_);
+    update.Set("key1", "value1").Remove("key1");
+
+    auto result = update.Apply();
+    EXPECT_THAT(result, IsError(ErrorKind::kValidationFailed));
```
wgtmac, line 93: It would be good if we can check the error message.

src/iceberg/update/update_properties.h
wgtmac: Sounds good. I still suggest renaming `metadata` to `base` to make it explicit.

Do we need to check if catalog or base is null? We cannot throw in the ctor, perhaps we should check them in Apply and Commit.

src/iceberg/update/update_properties.h
wgtmac: Yes I understand. I still think we don't need special case the file format config to make the API simple. The Java impl seems have some historical reasons and we don't need to follow everything.

src/iceberg/update/update_properties.cc
```diff
@@ -0,0 +1,136 @@
...
+namespace iceberg {
+
+UpdateProperties::UpdateProperties(TableIdentifier identifier,
+                                   std::shared_ptr<Catalog> catalog,
+                                   std::shared_ptr<TableMetadata> metadata)
+    : identifier_(std::move(identifier)),
+      catalog_(std::move(catalog)),
+      metadata_(std::move(metadata)) {}
+
+UpdateProperties& UpdateProperties::Set(std::string key, std::string value) {
```
wgtmac: What about using `(const std::string& key, const std::string& value)` as the parameters?

src/iceberg/update/update_properties.cc
```diff
@@ -0,0 +1,136 @@
...
+}
+
+UpdateProperties& UpdateProperties::DefaultFormat(FileFormatType format) {
+  updates_.emplace(TableProperties::kDefaultFileFormat.key(), ToString(format));
+  return *this;
+}
+
+Status UpdateProperties::Apply() {
+  if (!metadata_) {
+    return InvalidArgument("Cannot commit an empty table");
```
wgtmac:
```suggestion
    return InvalidArgument("Base metadata is required to apply property updates");
```

src/iceberg/update/update_properties.cc
```diff
@@ -0,0 +1,136 @@
...
+      return InvalidArgument("Format version '{}' is out of range", iter->second);
+    }
+
+    updates_.erase(iter);
+  }
+
+  if (metadata_->Schema().has_value()) {
+    ICEBERG_RETURN_UNEXPECTED(
+        MetricsConfig::VerifyReferencedColumns(updates_, *metadata_->Schema().value()));
+  }
```
wgtmac:
```suggestion
  if (auto schema = metadata_->Schema(); schema.has_value()) {
    ICEBERG_RETURN_UNEXPECTED(
        MetricsConfig::VerifyReferencedColumns(updates_, *schema.value()));
  }
```

wgtmac (commented):
This looks good now. I just left some minor comments.

src/iceberg/update/update_properties.cc
WZhuo: const std::string& is better

src/iceberg/test/update_properties_test.cc
WZhuo: Fix it

src/iceberg/test/metrics_config_test.cc
WZhuo: Replace all EXPECT_xxx(result.has_value()) with EXPECT_THAT(result, IsError/IsOk);

src/iceberg/test/update_properties_test.cc
```diff
@@ -0,0 +1,207 @@
...
+#include "iceberg/update/update_properties.h"
+
+#include <cstddef>
+#include <memory>
+#include <string>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "gmock/gmock.h"
```
wgtmac:
```suggestion
```

PR #366, opened 2025-11-28
refactor: clean TableBuilder in catalog

lishuxu:


HeartLinked (approved):
LGTM. Would it be more appropriate for this PR title to use refactor instead of feat?

PR #20, opened 2024-12-27
Add GitHub cpp-linter-action

wgtmac:
image

PR #367, opened 2025-11-28
fix: clang-tidy in cpplint should compile header files in C++23

HuaHuaY:


.github/workflows/cpp-linter.yml
```diff
@@ -57,8 +60,9 @@ jobs:
           thread-comments: true
           ignore: 'build|cmake_modules|ci'
           database: build
+          verbosity: 'debug'
```
wgtmac, line 63: Do we need to remove this? Does it affect the run time?

.github/workflows/cpp-linter.yml
```diff
@@ -29,6 +29,9 @@ on:
 jobs:
   cpp-linter:
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      pull-requests: write
```
wgtmac, line 34: I've tried this but it does not work. See discussion in https://github.com/apache/iceberg-cpp/pull/20#issuecomment-2566081248

HuaHuaY: Or should we remove `thread-comments: true`?

.github/workflows/cpp-linter.yml
HuaHuaY: It will display detailed output when clang-tidy fails. I think it doesn't affect the run time.

PR #368, opened 2025-11-28
refactor: make ConfigBase subclass constructors private

HeartLinked:


HuaHuaY (approved):

zhjwpku (approved):

src/iceberg/file_reader.cc
```diff
@@ -60,12 +60,12 @@ Result<std::unique_ptr<Reader>> ReaderFactoryRegistry::Open(
 }
 
 std::unique_ptr<ReaderProperties> ReaderProperties::default_properties() {
-  return std::make_unique<ReaderProperties>();
+  return std::unique_ptr<ReaderProperties>(new ReaderProperties());
```
shangxinli, line 63: For this specific case is actually safe because it is single expression, immediately returned and no other allocations happening. But I thought we generally prefer to use make_unique for it's safety.

HuaHuaY: `std::make_unique` needs a public constructor, so it can't be used when constructor is marked to private.

PR #370, opened 2025-11-28
feat: add ExpireSnapshots as concrete class

shangxinli:
Add ExpireSnapshots as a concrete implementation extending PendingUpdateTyped, following Gang's suggestion to avoid pure interfaces when there's only one implementation.

Changes:
- ExpireSnapshots is now a concrete class (not pure interface)
- Builder methods (ExpireSnapshotId, ExpireOlderThan, etc.) are non-virtual for performance
- Apply() and Commit() remain virtual (from PendingUpdateTyped)
- Added placeholder implementations in expire_snapshots.cc
- Updated tests to use concrete class directly
- Added source file to CMakeLists.txt and meson.build

This design provides:
- Simpler one-class implementation (vs interface + impl)
- No virtual function overhead for builder methods
- Still polymorphic through PendingUpdate base (for Transaction)
- Maintains fluent API and type safety

Full implementation of Apply() and Commit() will be added in follow-up PRs.

### DIFF
--- a/src/iceberg/pending_update.h
+++ b/src/iceberg/pending_update.h
@@ -69,4 +69,44 @@ class ICEBERG_EXPORT PendingUpdate : public ErrorCollector {
   PendingUpdate() = default;
 };
 
+/// \brief Template class for type-safe table metadata changes using builder pattern
+///
+/// PendingUpdateTyped extends PendingUpdate with a type-safe Apply() method that
+/// returns the specific result type for each operation. Subclasses implement
+/// specific types of table updates such as schema changes, property updates, or
+/// snapshot-producing operations like appends and deletes.
+///
+/// Apply() can be used to validate and inspect the uncommitted changes before
+/// committing. Commit() applies the changes and commits them to the table.
+///
+/// \tparam T The type of result returned by Apply()
+template <typename T>
+class ICEBERG_EXPORT PendingUpdateTyped : public PendingUpdate {
+ public:
+  ~PendingUpdateTyped() override = default;
+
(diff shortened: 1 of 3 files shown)


src/iceberg/test/CMakeLists.txt
```diff
@@ -97,6 +97,8 @@ add_iceberg_test(expression_test
                  literal_test.cc
                  predicate_test.cc)
 
+add_iceberg_test(update_test SOURCES ../update/test/expire_snapshots_test.cc)
```
zhjwpku: nit: I think this test file should be placed under src/iceberg/test/update/

shangxinli: Fixed

src/iceberg/table.h
```diff
@@ -115,6 +115,11 @@ class ICEBERG_EXPORT Table {
   /// \return a pointer to the new Transaction
   virtual std::unique_ptr<Transaction> NewTransaction() const;
 
+  /// \brief Create a new expire snapshots operation for this table
+  ///
+  /// \return a shared pointer to the new ExpireSnapshots operation
+  virtual std::shared_ptr<ExpireSnapshots> NewExpireSnapshots();
```
zhjwpku: nit: return unique_ptr?

zhjwpku (approved):

src/iceberg/table.h
shangxinli: Changed earlier but didn't reply this comment. Sorry.

PR #371, opened 2025-11-29
feat: add transform project

zhjwpku:
Add Transform::Project for inclusive predicate projection

This PR implements Transform::Project, which transforms a BoundPredicate to an inclusive predicate on partition values. StrictProject will be added in a separate PR to keep the review easier.

Move template implementations of Expressions into header file, or the linker will shout out the following:

Undefined symbols for architecture x86_64:
  "std::__1::shared_ptr<iceberg::UnboundPredicateImpl<iceberg::BoundReference>> iceberg::Expressions::In<iceberg::BoundReference>(std::__1::shared_ptr<iceberg::UnboundTerm<iceberg::BoundReference>>, std::initializer_list<iceberg::Literal>)", referenced from:
      iceberg::ProjectionUtil::FixInclusiveTimeProjection(std::__1::shared_ptr<iceberg::UnboundPredicateImpl<iceberg::BoundReference>> const&) in transform.cc.o

### DIFF
--- a/src/iceberg/test/transform_test.cc
+++ b/src/iceberg/test/transform_test.cc
@@ -1078,8 +1078,6 @@ TEST_F(TransformProjectTest, BucketProjectWithMatchingTransformedChild) {
   // The predicate's term should be a transform
   EXPECT_EQ(bound_pred->term()->kind(), Term::Kind::kTransform);
 
-  auto dummy = Expressions::NotEqual<BoundTransform>(bucket_term, Literal::Int(5));
-
   // When the transform matches, Project should use RemoveTransform and return the
   // predicate
   ICEBERG_ASSIGN_OR_THROW(auto projected,


src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,476 @@
...
+    bool is_source_date_type = func->source_type()->type_id() == TypeId::kDate;
+
+    switch (predicate->op()) {
+      case Expression::Operation::kLt: {
+        // adjust closed and then transform ltEq
+        ICEBERG_ASSIGN_OR_RAISE(
+            auto transformed,
+            func->Transform(is_source_date_type
+                                ? Literal::Date(std::get<T>(literal.value()) - 1)
+                                : Literal::Int(std::get<T>(literal.value()) - 1)));
```
shangxinli: What if kLt operation with long type?

zhjwpku: Good point, I failed to consider the Literal type in TruncateInteger, changed.

src/iceberg/type_fwd.h
```diff
@@ -161,6 +161,9 @@ class PendingUpdate;
 template <typename T>
 class PendingUpdateTyped;
 
+class BoundPredicate;
```
wgtmac: Move them closer to the expression family?

src/iceberg/transform.h
```diff
@@ -164,11 +165,23 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
...
+  /// \brief Transforms a BoundPredicate to an inclusive predicate on the partition values
+  /// produced by the transform.
+  ///
+  /// This inclusive transform guarantees that if predicate->Test(value) is true, then
+  /// Projected(transform(value)) is true.
+  /// \param name The name of the partition column.
+  /// \param predicate The predicate to project.
+  /// \return A Result containing either a shared pointer to the projected predicate or an
+  /// Error if the projection fails.
+  Result<std::shared_ptr<UnboundPredicate>> Project(
```
wgtmac: Can we return std::unique_ptr instead?

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,499 @@
...
+  static Result<std::shared_ptr<UnboundPredicate>> TransformSet(
+      std::string_view name, const std::shared_ptr<BoundSetPredicate>& predicate,
+      const std::shared_ptr<TransformFunction>& func) {
+    std::vector<Literal> transformed;
+    transformed.reserve(predicate->literal_set().size());
+    for (const auto& lit : predicate->literal_set()) {
+      ICEBERG_ASSIGN_OR_RAISE(auto transformed_lit, func->Transform(lit));
+      transformed.push_back(std::move(transformed_lit));
+    }
+    return Expressions::Predicate(predicate->op(), std::string(name),
```
wgtmac: We cannot use convenience functions in the `Expressions` because they may throw and they are only supposed to be used by users. We need to stick with Make functions to deal with any error.

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,499 @@
...
+      const std::shared_ptr<TransformFunction>& func) {
+    const Literal& boundary = predicate->literal();
+
+    ICEBERG_ASSIGN_OR_RAISE(auto transformed, func->Transform(boundary));
+
+    switch (predicate->op()) {
+      case Expression::Operation::kLt:
+      case Expression::Operation::kLtEq:
+        return Expressions::Predicate(Expression::Operation::kLtEq, std::string(name),
+                                      transformed);
```
wgtmac:
```suggestion
                                      std::move(transformed));
```

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,499 @@
...
+    transformed.reserve(predicate->literal_set().size());
+    for (const auto& lit : predicate->literal_set()) {
+      ICEBERG_ASSIGN_OR_RAISE(auto transformed_lit, func->Transform(lit));
+      transformed.push_back(std::move(transformed_lit));
+    }
+    return Expressions::Predicate(predicate->op(), std::string(name),
+                                  std::move(transformed));
+  }
+
+  static Result<std::shared_ptr<UnboundPredicate>> TruncateArray(
```
wgtmac: I know Java impl uses the same name. However, I am still confused that it is generic enough to be used by transform other than truncate and it has nothing to do with array.

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,499 @@
...
+
+    switch (predicate->op()) {
+      case Expression::Operation::kLt: {
+        // adjust closed and then transform ltEq
+        if constexpr (std::is_same_v<T, int32_t>) {
+          ICEBERG_ASSIGN_OR_RAISE(
+              auto transformed,
+              func->Transform(literal.type()->type_id() == TypeId::kDate
+                                  ? Literal::Date(std::get<T>(literal.value()) - 1)
+                                  : Literal::Int(std::get<T>(literal.value()) - 1)));
```
wgtmac: This does not look safe since `predicate` may be of any type but `TruncateInteger` indicates that only integer variants are accepted.

src/iceberg/transform.h
zhjwpku: Done.

src/iceberg/util/projection_util_internal.h
zhjwpku: I rename it with TruncateByteArray, it's used for string and binary truncation. I also add a GenericTransform which is used as a fallback for any special cases.

src/iceberg/util/projection_util_internal.h
zhjwpku: I split the temporal usage into another function TransformTemporal.

src/iceberg/transform.h
```diff
@@ -164,11 +165,23 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
...
 
+  /// \brief Transforms a BoundPredicate to an inclusive predicate on the partition values
+  /// produced by the transform.
+  ///
+  /// This inclusive transform guarantees that if predicate->Test(value) is true, then
+  /// Projected(transform(value)) is true.
+  /// \param name The name of the partition column.
+  /// \param predicate The predicate to project.
+  /// \return A Result containing either a unique pointer to the projected predicate or an
+  /// Error if the projection fails.
```
wgtmac: It can also return `nullptr` which should be documented as well.

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,640 @@
...
+    transformed.reserve(predicate->literal_set().size());
+    for (const auto& lit : predicate->literal_set()) {
+      ICEBERG_ASSIGN_OR_RAISE(auto transformed_lit, func->Transform(lit));
+      transformed.push_back(std::move(transformed_lit));
+    }
+    ICEBERG_ASSIGN_OR_RAISE(auto ref, NamedReference::Make(std::string(name)));
+    ICEBERG_ASSIGN_OR_RAISE(auto pred,
+                            UnboundPredicateImpl<BoundReference>::Make(
+                                predicate->op(), std::move(ref), std::move(transformed)));
+    return pred;
```
wgtmac:
```suggestion
    return UnboundPredicateImpl<BoundReference>::Make(
      predicate->op(), std::move(ref), std::move(transformed)));
```

wgtmac: Same for below

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,640 @@
...
+            ICEBERG_ASSIGN_OR_RAISE(auto pred,
+                                    UnboundPredicateImpl<BoundReference>::Make(
+                                        Expression::Operation::kGtEq, std::move(ref),
+                                        std::move(transformed)));
+            return pred;
+          }
+          default:
+            return GenericTransform(std::move(ref), predicate, func);
+        }
+        std::unreachable();
```
wgtmac: Why there is a std::unreachable() but other branches not?

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,640 @@
...
+                                             std::move(ref), std::move(transformed)));
+      return pred;
+    }
+
+    return nullptr;
+  }
+
+  // Fixes an inclusive projection to account for incorrectly transformed values.
+  // align with Java implementation:
+  // https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/transforms/ProjectionUtil.java#L275
```
wgtmac: The link might be invalid over time because it points to the main branch.

src/iceberg/transform.cc
```diff
@@ -240,6 +246,66 @@ bool Transform::SatisfiesOrderOf(const Transform& other) const {
...
+      if (predicate->term()->kind() == Term::Kind::kTransform) {
+        const auto boundTransform =
+            internal::checked_pointer_cast<BoundTransform>(predicate->term());
+        if (*this == *boundTransform->transform()) {
+          return ProjectionUtil::RemoveTransform(name, predicate);
+        } else {
+          return nullptr;
+        }
+      }
+      ICEBERG_ASSIGN_OR_RAISE(auto func, Bind(predicate->term()->type()));
```
wgtmac: What if the term kind is kExtract here?

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,640 @@
...
+            internal::checked_pointer_cast<BoundSetPredicate>(predicate);
+        ICEBERG_ASSIGN_OR_RAISE(
+            auto pred, UnboundPredicateImpl<BoundReference>::Make(
+                           predicate->op(), std::move(ref),
+                           std::vector<Literal>(setPredicate->literal_set().begin(),
+                                                setPredicate->literal_set().end())));
+        return pred;
+      }
+    }
+    std::unreachable();
```
wgtmac, line 555: Should it be moved out of the switch block?

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,640 @@
...
+        ICEBERG_ASSIGN_OR_RAISE(
+            auto pred, UnboundPredicateImpl<BoundReference>::Make(
+                           predicate->op(), std::move(ref),
+                           std::vector<Literal>(setPredicate->literal_set().begin(),
+                                                setPredicate->literal_set().end())));
+        return pred;
+      }
+    }
+
+    std::unreachable();
```
wgtmac, line 417: Should we return nullptr for consistency?

src/iceberg/util/projection_util_internal.h
```diff
@@ -0,0 +1,640 @@
...
+      return TruncateByteArray(name, predicate, func);
+    }
+
+    const auto& truncate_transform =
+        internal::checked_pointer_cast<TruncateTransform>(func);
+    const auto& str_value = std::get<std::string>(predicate->literal().value());
+    const auto width = truncate_transform->width();
+    ICEBERG_ASSIGN_OR_RAISE(auto ref, NamedReference::Make(std::string(name)));
+
+    if (str_value.length() < width) {
```
wgtmac: Is this correct? `str_value.length()` is the number of bytes not the number of UTF-8 chars but `width` is the number of UTF-8 chars.

zhjwpku: Good catch, we should consider UTF-8, will fix.

src/iceberg/transform.h
zhjwpku: Done

src/iceberg/util/projection_util_internal.h
zhjwpku: Fixed by changing to a permanent link.

src/iceberg/util/projection_util_internal.h
zhjwpku: We can't reach here, so the unreachable should be fine.

src/iceberg/util/projection_util_internal.h
zhjwpku: It's already out of the switch block.

src/iceberg/util/projection_util_internal.h
zhjwpku: Fixed by adding a CodePointCount function to do the count.

src/iceberg/test/transform_test.cc
```diff
@@ -914,4 +919,767 @@ TEST(TransformCanTransformTest, CanTransform) {
...
+  auto unbound = Expressions::Equal<BoundTransform>(bucket_term, Literal::Int(5));
+  ICEBERG_ASSIGN_OR_THROW(auto bound,
+                          unbound->Bind(*int_schema_, /*case_sensitive=*/true));
+  auto bound_pred = std::dynamic_pointer_cast<BoundPredicate>(bound);
+  ASSERT_NE(bound_pred, nullptr);
+
+  // The predicate's term should be a transform
+  EXPECT_EQ(bound_pred->term()->kind(), Term::Kind::kTransform);
+
+  auto dummy = Expressions::NotEqual<BoundTransform>(bucket_term, Literal::Int(5));
```
wgtmac: Unused?

zhjwpku: removed, sorry about the oversight :)

Issue #360, opened 2025-11-28
Aggregate evaluator supports evaluating DataFile


wgtmac:
As https://github.com/apache/iceberg-cpp/pull/335 is merged, we need to support evaluating `DataFile` on Aggregate evaluators.

PR #373, opened 2025-11-30
feat:  add DataFile aggregate evaluation

SuKi2cn:
This PR implements the `DataFil` evaluation for aggregates addressing issue #360.
- `MAX`/`MIN` now support evaluation from `DataFile` metrics.
- Count-family aggregates now follow Java semantics:
  - `CountNonNull`: requires value + null counts
  - `CountNull`: requires null counts
  - `CountStar`: requires non-negative record_count
  Missing metrics mark the aggregator invalid and produce null results.
- Added `AggregateEvaluator::AllAggregatorsValid()`.
  Mirrors Java’s allAggregatorsValid() to indicate when aggregates can be reliably computed from `DataFile` metrics.
- Added `MAX`/`MIN` overloads for `UnboundTerm<BoundTransform>`.
- Updated and extended tests:

### DIFF
--- a/src/iceberg/expression/aggregate.cc
+++ b/src/iceberg/expression/aggregate.cc
@@ -282,11 +282,6 @@ Result<Literal> CountAggregate::Evaluate(const DataFile& file) const {
   return CountFor(file).transform(Literal::Long);
 }
 
-Result<Literal> CountAggregate::Evaluate(const DataFile& file) const {
-  ICEBERG_ASSIGN_OR_RAISE(auto count, CountFor(file));
-  return Literal::Long(count);
-}
-
 std::unique_ptr<BoundAggregate::Aggregator> CountAggregate::NewAggregator() const {
   return std::unique_ptr<BoundAggregate::Aggregator>(new CountAggregator(*this));
 }
@@ -531,16 +526,7 @@ class AggregateEvaluatorImpl : public AggregateEvaluator {
   }
 
   bool AllAggregatorsValid() const override {
-<<<<<<< ours
     return std::ranges::all_of(aggregators_, &BoundAggregate::Aggregator::IsValid);
-=======
-    for (const auto& aggregator : aggregators_) {
-      if (!aggregator->IsValid()) {
-        return false;
-      }
-    }
-    return true;
->>>>>>> theirs
   }
 
  private:


src/iceberg/expression/aggregate.cc
```diff
@@ -38,6 +40,58 @@ std::shared_ptr<PrimitiveType> GetPrimitiveType(const BoundTerm& term) {
   return internal::checked_pointer_cast<PrimitiveType>(term.type());
 }
 
+Result<Scalar> LiteralToScalar(const Literal& literal) {
```
wgtmac: Move this to struct_like.h/.cc?

src/iceberg/expression/aggregate.h
```diff
@@ -143,11 +148,16 @@ class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound
...
  public:
   Result<Literal> Evaluate(const StructLike& data) const final;
+  Result<Literal> Evaluate(const DataFile& file) const override;
 
   std::unique_ptr<Aggregator> NewAggregator() const override;
 
   /// \brief Count for a single row. Subclasses implement this.
   virtual Result<int64_t> CountFor(const StructLike& data) const = 0;
+  /// \brief Count using metrics from a data file. Nullopt if not available.
+  virtual Result<std::optional<int64_t>> CountFor(const DataFile& file) const;
```
wgtmac: Is it better to declare it as `Result<int64_t> CountFor(const DataFile& file) const`? We will call `HasValue` before `CountFor`, so it is fine to return an error if it does not have value.

src/iceberg/expression/aggregate.cc
```diff
@@ -38,6 +40,58 @@ std::shared_ptr<PrimitiveType> GetPrimitiveType(const BoundTerm& term) {
...
+      return NotSupported("Cannot convert literal of type {} to Scalar",
+                          literal.type()->ToString());
+  }
+}
+
+class SingleValueStructLike : public StructLike {
+ public:
+  explicit SingleValueStructLike(Literal literal) : literal_(std::move(literal)) {}
+
+  Result<Scalar> GetField(size_t /*pos*/) const override {
```
wgtmac: Should we return error if pos is not 0?

src/iceberg/expression/aggregate.h
wgtmac: BTW, why not making it a pure virtual function?

src/iceberg/expression/aggregate.h
```diff
@@ -127,6 +128,10 @@ class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound
   }
 
   Result<Literal> Evaluate(const StructLike& data) const override = 0;
+  virtual Result<Literal> Evaluate(const DataFile& file) const;
```
wgtmac: Why not pure virtual?

src/iceberg/expression/aggregate.h
```diff
@@ -127,6 +128,10 @@ class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound
...
+
+  /// \brief Whether metrics in the data file are sufficient to evaluate.
+  virtual bool HasValue(const DataFile& file) const { return false; }
```
wgtmac: Same here

src/iceberg/expression/aggregate.cc
```diff
@@ -173,6 +366,23 @@ Result<int64_t> CountNonNullAggregate::CountFor(const StructLike& data) const {
       [](const auto& val) { return val.IsNull() ? 0 : 1; });
 }
 
+Result<std::optional<int64_t>> CountNonNullAggregate::CountFor(
+    const DataFile& file) const {
+  auto field_id = GetFieldId(term());
+  auto value_count = GetMapValue(file.value_counts, field_id);
+  auto null_count = GetMapValue(file.null_value_counts, field_id).value_or(0);
```
wgtmac: Is this correct? `HasValue` has guaranteed that `field_id` exists in the `file.null_value_counts`, so we don't need to call `value_or(0)`, right? We can simply return `NotFound` error if any of value_count and null_count is missing.

wgtmac: Perhaps it is simpler if we directly call HasValue in CountFor and return error if false?

src/iceberg/expression/aggregate.cc
```diff
@@ -211,10 +445,34 @@ Result<Literal> MaxAggregate::Evaluate(const StructLike& data) const {
...
+  auto field_id = GetFieldId(term());
+  auto upper = GetMapValue(file.upper_bounds, field_id);
+  auto ptype = GetPrimitiveType(*term());
+  if (!upper.has_value()) {
+    SingleValueStructLike data(Literal::Null(ptype));
+    return term()->Evaluate(data);
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto literal, Literal::Deserialize(*upper, ptype));
+  SingleValueStructLike data(std::move(literal));
```
wgtmac: Should we consider reusing SingleValueStructLike? The Java impl does this in the ValueAggregate (which we don't have yet)

src/iceberg/expression/aggregate.cc
SuKi2cn: > Move this to struct_like.h/.cc?

Make sense. It will be reusable.

src/iceberg/expression/aggregate.h
SuKi2cn: > Is it better to declare it as `Result<int64_t> CountFor(const DataFile& file) const`? We will call `HasValue` before `CountFor`, so it is fine to return an error if it does not have value.

Got it. it matches the way `HasValue` is intended to be used.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Should we consider reusing SingleValueStructLike? The Java impl does this in the ValueAggregate (which we don't have yet)

Totally agree.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Is this correct? `HasValue` has guaranteed that `field_id` exists in the `file.null_value_counts`, so we don't need to call `value_or(0)`, right? We can simply return `NotFound` error if any of value_count and null_count is missing.

> Perhaps it is simpler if we directly call HasValue in CountFor and return error if false?

Good catch. I didn’t notice that earlier. I’ll update the implementation.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Should we return error if pos is not 0?

I moved `LiteralToScalar`/`SingleValueStructLike` into `row/struct_like` as the small adapter we use when evaluating aggregates from file metrics. I originally tried to return an error when `pos != 0`, but that breaks the metrics aggregation path: bound terms carry the original field position (often 1,2,…) so `Count`/`Max`/`Min` on file metrics all fail (ctest reproduces this). The Java equivalent `ValueAggregate` ([ValueAggregate](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/ValueAggregate.java#L60-L68)) also ignores the index for the same reason. It isn’t a reusable general StructLike; it’s a narrow, internal adapter, so I think we need to ignore the incoming index here for correctness. If anything in my understanding is off, please let me know.

wgtmac: SGTM. Thanks for the explanation!

src/iceberg/test/aggregate_test.cc
```diff
@@ -236,4 +237,239 @@ TEST(AggregateTest, MultipleAggregatesInEvaluator) {
...
+
+  ICEBERG_UNWRAP_OR_FAIL(auto results, evaluator->GetResults());
+  ASSERT_EQ(results.size(), aggregates.size());
+  // Truncate width 10: max(truncate(23)) -> 20, min(truncate(5)) -> 0
+  EXPECT_EQ(std::get<int32_t>(results[0].value()), 20);
+  EXPECT_EQ(std::get<int32_t>(results[1].value()), 0);
+  EXPECT_TRUE(evaluator->AllAggregatorsValid());
+}
+
+TEST(AggregateTest, DataFileAggregatorParityWithJava) {
```
HeartLinked:
```suggestion
TEST(AggregateTest, DataFileAggregatorParity) {
```
modify this?

src/iceberg/test/aggregate_test.cc
```diff
@@ -236,4 +237,239 @@ TEST(AggregateTest, MultipleAggregatesInEvaluator) {
...
+    return std::pair{std::move(lower_bounds), std::move(upper_bounds)};
+  };
+
+  auto [b1_lower, b1_upper] = make_bounds(1, 33, 2345);
+  DataFile file;
+  file.file_path = "file.avro";
+  file.record_count = 50;
+  file.value_counts = {{1, 50}, {3, 50}, {4, 50}};
+  file.null_value_counts = {{1, 10}, {3, 50}, {4, 10}};
+  file.lower_bounds = b1_lower;
```
HeartLinked: use designated initializers to simplify code？

src/iceberg/test/aggregate_test.cc
SuKi2cn: > modify this?

Yep. It's way better.

src/iceberg/test/aggregate_test.cc
SuKi2cn: > use designated initializers to simplify code？

Gotcha.

src/iceberg/row/struct_like.h
```diff
@@ -55,6 +55,9 @@ using Scalar = std::variant<std::monostate,  // for null
                             std::shared_ptr<ArrayLike>,   // for list
                             std::shared_ptr<MapLike>>;    // for map
 
+/// \brief Convert a Literal to a Scalar
+Result<Scalar> LiteralToScalar(const Literal& literal);
```
wgtmac:
```suggestion
ICEBERG_EXPORT Result<Scalar> LiteralToScalar(const Literal& literal);
```

src/iceberg/expression/aggregate.cc
```diff
@@ -115,13 +188,73 @@ class MinAggregator : public BoundAggregate::Aggregator {
...
+
+  bool IsValid() const override { return valid_; }
 
  private:
   const MinAggregate& aggregate_;
   Literal current_;
+  bool valid_ = true;
 };
 
+bool HasMapKey(const std::map<int32_t, int64_t>& map, int32_t key) {
```
HuaHuaY: Why do you wrap `contains`?

src/iceberg/expression/aggregate.cc
```diff
@@ -152,6 +285,11 @@ Result<Literal> CountAggregate::Evaluate(const StructLike& data) const {
   return CountFor(data).transform([](int64_t count) { return Literal::Long(count); });
 }
 
+Result<Literal> CountAggregate::Evaluate(const DataFile& file) const {
+  ICEBERG_ASSIGN_OR_RAISE(auto count, CountFor(file));
+  return Literal::Long(count);
```
HuaHuaY:
```suggestion
  return CountFor(file).transform(Literal::Long);
```
Please also help simplify `CountAggregate::Evaluate(const StructLike& data) const` above.

src/iceberg/expression/aggregate.cc
```diff
@@ -315,6 +533,15 @@ class AggregateEvaluatorImpl : public AggregateEvaluator {
     return all.front();
   }
 
+  bool AllAggregatorsValid() const override {
+    for (const auto& aggregator : aggregators_) {
+      if (!aggregator->IsValid()) {
+        return false;
+      }
+    }
+    return true;
```
HuaHuaY:
```suggestion
    return std::ranges::all_of(aggregators_, &BoundAggregate::Aggregator::IsValid);
```
Need to add `#include <algorithm>`.

src/iceberg/expression/aggregate.h
```diff
@@ -143,11 +147,16 @@ class ICEBERG_EXPORT BoundAggregate : public Aggregate<BoundTerm>, public Bound
 class ICEBERG_EXPORT CountAggregate : public BoundAggregate {
  public:
   Result<Literal> Evaluate(const StructLike& data) const final;
+  Result<Literal> Evaluate(const DataFile& file) const override;
```
HuaHuaY, line 150:
```suggestion
  Result<Literal> Evaluate(const DataFile& file) const final;
```

src/iceberg/expression/aggregate.cc
SuKi2cn: > Why do you wrap `contains`?

Got it. It's redundant.

src/iceberg/expression/aggregate.cc
SuKi2cn: > Need to add `#include <algorithm>`.

Thank you!

SuKi2cn:
> This PR carries a lot of unrelated changes. Could you please fix that?

If convenient, please review pr #400. Appreciate for your time and help.

Issue #332, opened 2025-11-20
Improve Avro implementation


wgtmac:
The current Avro writer and reader implementations are not that efficient because we need to use `avro::GenericDatum` as the intermediate layer:

- [x] [Avro writer](https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/avro/avro_writer.cc#L104): `ArrowArray` -> `avro::GenericDatum` -> `avro::Encoder`

- [x] [Avro reader](https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/avro/avro_reader.cc#L134): `avro::Decoder` -> `avro::GenericDatum` -> `ArrowArray` (https://github.com/apache/iceberg-cpp/pull/374)

It would be more efficient if we can get rid of conversion from/to `avro::GenericDatum` by directly using the encoder/decoder in the Avro writer/reader. It can greatly improve the performance of manifest file I/O.

shangxinli:
+1

PR #374, opened 2025-11-30
feat: eliminate GenericDatum in Avro reader for performance

shangxinli:
Replace GenericDatum intermediate layer with direct Avro decoder access to improve manifest I/O performance.

Changes:
- Add avro_direct_decoder_internal.h with DecodeAvroToBuilder API
- Add avro_direct_decoder.cc implementing direct Avro→Arrow decoding
  - Primitive types: bool, int, long, float, double, string, binary, fixed
  - Temporal types: date, time, timestamp
  - Logical types: uuid, decimal
  - Nested types: struct, list, map
  - Union type handling for nullable fields
- Modify avro_reader.cc to use DataFileReaderBase with direct decoder
  - Replace DataFileReader<GenericDatum> with DataFileReaderBase
  - Use decoder.decodeInt(), decodeLong(), etc. directly
  - Remove GenericDatum allocation and extraction overhead
- Update CMakeLists.txt to include new decoder source

Performance improvement:
- Before: Avro binary → GenericDatum → Extract → Arrow (3 steps)
- After: Avro binary → decoder.decodeInt() → Arrow (2 steps)

This matches Java implementation which uses Decoder directly via ValueReader interface, avoiding intermediate object allocation.

All avro_test cases pass.

Issue: #332

### DIFF
--- a/src/iceberg/avro/avro_direct_decoder.cc
+++ b/src/iceberg/avro/avro_direct_decoder.cc
@@ -486,13 +486,6 @@ Status DecodePrimitiveValueToBuilder(const ::avro::NodePtr& avro_node,
             ToString(avro_node));
       }
 
-      const auto& decimal_type =
-          internal::checked_cast<const DecimalType&>(projected_type);
-
-      // Note: Avro C++ LogicalType doesn't expose precision/scale getters,
-      // so we rely on schema projection validation
-
-      // Use Avro schema's fixed size (not calculated)
       size_t byte_width = avro_node->fixedSize();
       auto* builder = internal::checked_cast<::arrow::Decimal128Builder*>(array_builder);
 
--- a/src/iceberg/avro/avro_reader.cc
+++ b/src/iceberg/avro/avro_reader.cc
@@ -75,10 +75,7 @@ struct ReadContext {
   DecodeContext decode_context_;
 };
 
-// TODO(gang.wu): there are a lot to do to make this reader work.
-// 1. prune the reader schema based on the projection
-// 2. read key-value metadata from the avro file
-// 3. collect basic reader metrics
+// TODO(gang.wu): collect basic reader metrics
 class AvroReader::Impl {
  public:
   Status Open(const ReaderOptions& options) {
@@ -217,7 +214,7 @@ class AvroReader::Impl {
       return Invalid("Reader is not opened");
     }
 
-    const ::avro::Metadata metadata = GetReaderMetadata();
+    const auto& metadata = GetReaderMetadata();
     std::unordered_map<std::string, std::string> metadata_map;
     metadata_map.reserve(metadata.size());
 
@@ -280,19 +277,18 @@ class AvroReader::Impl {
     return arrow_array;
   }
 
-  // Helper: Check if past sync point
   bool IsPastSync() const {
-    if (!split_end_) return false;
+    if (!split_end_) {
+      return false;
+    }
     return use_direct_decoder_ ? base_reader_->pastSync(split_end_.value())
                                : datum_reader_->pastSync(split_end_.value());
   }
 
-  // Helper: Get metadata from appropriate reader
-  ::avro::Metadata GetReaderMetadata() const {
+  const ::avro::Metadata& GetReaderMetadata() const {
     return use_direct_decoder_ ? base_reader_->metadata() : datum_reader_->metadata();
   }
 
-  // Helper: Close the appropriate reader
   void CloseReader() {
     if (use_direct_decoder_) {
       if (base_reader_) {
@@ -307,7 +303,6 @@ class AvroReader::Impl {
     }
   }
 
-  // Helper: Get reader schema
   const ::avro::ValidSchema& GetReaderSchema() const {
     return use_direct_decoder_ ? base_reader_->readerSchema()
                                : datum_reader_->readerSchema();


src/iceberg/avro/avro_reader.cc
```diff
@@ -140,12 +141,15 @@ class AvroReader::Impl {
...
+      if (!reader_->hasMore()) {
         break;
       }
+      reader_->decr();
+
+      // Use direct decoder instead of GenericDatum
       ICEBERG_RETURN_UNEXPECTED(
-          AppendDatumToBuilder(reader_->readerSchema().root(), *context_->datum_,
-                               projection_, *read_schema_, context_->builder_.get()));
+          DecodeAvroToBuilder(reader_->readerSchema().root(), reader_->decoder(),
```
wgtmac: I think it would be better to use a feature flag to enable users to use the old reader just in case there is any bug.

shangxinli: Good suggestion!

wgtmac:
Could you help add some test cases? The original avro reader is simply a wrapper on top of avro-cpp so existing test cases are very minimal. We need better test coverage to be confident. Perhaps we can use the avro writer to create in-memory avro files (using `MakeMockFileIO` and then use the new reader to read them. It is also nice to add some benchmarks to prove that this indeed improves the reading performance compared to the original implementation.

src/iceberg/avro/avro_direct_decoder.cc
```diff
@@ -0,0 +1,597 @@
...
+Status DecodeFieldToBuilder(const ::avro::NodePtr& avro_node, ::avro::Decoder& decoder,
+                            const FieldProjection& projection,
+                            const SchemaField& projected_field,
+                            ::arrow::ArrayBuilder* array_builder);
+
+/// \brief Skip an Avro value based on its schema without decoding
+Status SkipAvroValue(const ::avro::NodePtr& avro_node, ::avro::Decoder& decoder) {
+  switch (avro_node->type()) {
+    case ::avro::AVRO_NULL:
+      // Nothing to skip
```
Smith-Cruise: Better to call `decoder.decodeNull()`? Even if nothing todo.

shangxinli: Good point

src/iceberg/avro/avro_reader.cc
```diff
@@ -238,17 +277,54 @@ class AvroReader::Impl {
...
+  bool use_direct_decoder_{true};
   // The end of the split to read and used to terminate the reading.
   std::optional<int64_t> split_end_;
   // The schema to read.
   std::shared_ptr<::iceberg::Schema> read_schema_;
   // The projection result to apply to the read schema.
   SchemaProjection projection_;
-  // The avro reader to read the data into a datum.
-  std::unique_ptr<::avro::DataFileReader<::avro::GenericDatum>> reader_;
+  // The avro reader base - provides direct access to decoder (new path).
```
wgtmac: I would rather not call them new or legacy path...

src/iceberg/avro/avro_direct_decoder.cc
```diff
@@ -0,0 +1,597 @@
...
+      return {};
+    }
+
+    case TypeId::kBinary: {
+      if (avro_node->type() != ::avro::AVRO_BYTES) {
+        return InvalidArgument("Expected Avro bytes for binary field, got: {}",
+                               ToString(avro_node));
+      }
+      auto* builder = internal::checked_cast<::arrow::BinaryBuilder*>(array_builder);
+      std::vector<uint8_t> bytes;
```
wgtmac: Do we want to avoid frequent small object allocation like this? Perhaps we can reuse it by adding a `class DecodeContext` where a `std::vector<uint8_t> scratch` object is supposed to be used from it.

wgtmac: Same for other temp vector and string variables.

src/iceberg/avro/avro_direct_decoder.cc
```diff
@@ -0,0 +1,597 @@
...
+      ICEBERG_ARROW_RETURN_NOT_OK(builder->Append(uuid_data.data()));
+      return {};
+    }
+
+    case TypeId::kDecimal: {
+      if (avro_node->type() != ::avro::AVRO_FIXED) {
+        return InvalidArgument("Expected Avro fixed for decimal field, got: {}",
+                               ToString(avro_node));
+      }
+      if (avro_node->logicalType().type() != ::avro::LogicalType::DECIMAL) {
```
wgtmac: Can we combine the two `if`s above just like other branches?

src/iceberg/avro/avro_direct_decoder.cc
```diff
@@ -0,0 +1,597 @@
...
+
+      // Validate byte width is sufficient for precision
+      // Max value with P digits: 10^P - 1, needs ceil(P * log(10) / log(256)) bytes
+      size_t min_bytes = (decimal_type.precision() * 415) / 1000 + 1;  // ceil(P * 0.415)
+      if (byte_width < min_bytes) {
+        return InvalidArgument(
+            "Decimal byte width {} insufficient for precision {}, need at least {} bytes",
+            byte_width, decimal_type.precision(), min_bytes);
+      }
+
```
wgtmac:
```suggestion
```

I don't think we need this check.

src/iceberg/avro/avro_direct_decoder.cc
```diff
@@ -0,0 +1,597 @@
...
+      ICEBERG_RETURN_UNEXPECTED(DecodeFieldToBuilder(
+          avro_field_node, decoder, field_projection, expected_field, field_builder));
+    }
+  }
+
+  // Handle null fields (fields in projection but not in Avro)
+  for (size_t proj_idx = 0; proj_idx < projections.size(); ++proj_idx) {
+    const auto& field_projection = projections[proj_idx];
+    if (field_projection.kind == FieldProjection::Kind::kNull) {
```
wgtmac, line 205: Return error for other unsupported `kind`?

src/iceberg/avro/avro_direct_decoder.cc
```diff
@@ -0,0 +1,597 @@
...
+  if (avro_node->type() != ::avro::AVRO_RECORD) {
+    return InvalidArgument("Expected Avro record, got type: {}", ToString(avro_node));
+  }
+
+  auto* struct_builder = internal::checked_cast<::arrow::StructBuilder*>(array_builder);
+  ICEBERG_ARROW_RETURN_NOT_OK(struct_builder->Append());
+
+  // Build a map from Avro field index to projection index
+  // -1 means the field should be skipped
+  std::vector<int> avro_to_projection(avro_node->leaves(), -1);
```
wgtmac: I'm not sure if we need to cache `avro_to_projection` somewhere, perhaps in the `DecodeContext` that I propose in the below comment. We can use `const FieldProjection*` as the key to find the cached value.

src/iceberg/avro/avro_reader.cc
```diff
@@ -62,12 +63,12 @@ Result<std::unique_ptr<AvroInputStream>> CreateInputStream(const ReaderOptions&
 
 // A stateful context to keep track of the reading progress.
 struct ReadContext {
-  // The datum to reuse for reading the data.
-  std::unique_ptr<::avro::GenericDatum> datum_;
   // The arrow schema to build the record batch.
   std::shared_ptr<::arrow::Schema> arrow_schema_;
   // The builder to build the record batch.
   std::shared_ptr<::arrow::ArrayBuilder> builder_;
+  // GenericDatum for legacy path (only used if direct decoder is disabled)
```
```suggestion
  // GenericDatum for intermediate data (only used if direct decoder is disabled)
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -91,10 +94,21 @@ class AvroReader::Impl {
...
+    ::avro::ValidSchema file_schema;
+
+    if (use_direct_decoder_) {
+      // New path: Create base reader for direct decoder access
+      auto base_reader =
+          std::make_unique<::avro::DataFileReaderBase>(std::move(input_stream));
+      file_schema = base_reader->dataSchema();
+      base_reader_ = std::move(base_reader);
+    } else {
+      // Legacy path: Create DataFileReader<GenericDatum>
```
```suggestion
      // Create a reader to read into generic datum
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -91,10 +94,21 @@ class AvroReader::Impl {
...
                             CreateInputStream(options, kDefaultBufferSize));
 
-    // Create a base reader without setting reader schema to enable projection.
-    auto base_reader =
-        std::make_unique<::avro::DataFileReaderBase>(std::move(input_stream));
-    ::avro::ValidSchema file_schema = base_reader->dataSchema();
+    ::avro::ValidSchema file_schema;
+
+    if (use_direct_decoder_) {
+      // New path: Create base reader for direct decoder access
```
```suggestion
      // Create a base reader to directly use its decoder
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -137,25 +160,37 @@ class AvroReader::Impl {
...
     while (context_->builder_->length() < batch_size_) {
-      if (split_end_ && reader_->pastSync(split_end_.value())) {
+      if (IsPastSync()) {
         break;
       }
-      if (!reader_->read(*context_->datum_)) {
-        break;
+
+      if (use_direct_decoder_) {
+        // New path: Use direct decoder
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -137,25 +160,37 @@ class AvroReader::Impl {
...
+        if (!base_reader_->hasMore()) {
+          break;
+        }
+        base_reader_->decr();
+
+        ICEBERG_RETURN_UNEXPECTED(
+            DecodeAvroToBuilder(GetReaderSchema().root(), base_reader_->decoder(),
+                                projection_, *read_schema_, context_->builder_.get()));
+      } else {
+        // Legacy path: Use GenericDatum
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -214,6 +248,11 @@ class AvroReader::Impl {
     }
     context_->builder_ = builder_result.MoveValueUnsafe();
 
+    // Initialize GenericDatum for legacy path
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -238,17 +277,54 @@ class AvroReader::Impl {
...
+      }
+    } else {
+      if (datum_reader_) {
+        datum_reader_->close();
+        datum_reader_.reset();
+      }
+    }
+  }
+
+  // Helper: Get reader schema
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -238,17 +277,54 @@ class AvroReader::Impl {
...
+    return use_direct_decoder_ ? base_reader_->pastSync(split_end_.value())
+                               : datum_reader_->pastSync(split_end_.value());
+  }
+
+  // Helper: Get metadata from appropriate reader
+  ::avro::Metadata GetReaderMetadata() const {
+    return use_direct_decoder_ ? base_reader_->metadata() : datum_reader_->metadata();
+  }
+
+  // Helper: Close the appropriate reader
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -238,17 +277,54 @@ class AvroReader::Impl {
     return arrow_array;
   }
 
+  // Helper: Check if past sync point
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -238,17 +277,54 @@ class AvroReader::Impl {
...
+  bool IsPastSync() const {
+    if (!split_end_) return false;
+    return use_direct_decoder_ ? base_reader_->pastSync(split_end_.value())
+                               : datum_reader_->pastSync(split_end_.value());
+  }
+
+  // Helper: Get metadata from appropriate reader
```
```suggestion
```

src/iceberg/avro/avro_reader.cc
```diff
@@ -238,17 +277,54 @@ class AvroReader::Impl {
     return arrow_array;
   }
 
+  // Helper: Check if past sync point
+  bool IsPastSync() const {
+    if (!split_end_) return false;
```
```suggestion
    if (!split_end_) {
      return false;
    }
```

src/iceberg/file_reader.h
```diff
@@ -76,6 +76,10 @@ class ReaderProperties : public ConfigBase<ReaderProperties> {
   /// \brief The batch size to read.
   inline static Entry<int64_t> kBatchSize{"read.batch-size", 4096};
 
+  /// \brief Use direct Avro decoder (true) or GenericDatum-based decoder (false).
+  /// Default: true (use direct decoder for better performance).
+  inline static Entry<bool> kAvroUseDirectDecoder{"avro.use-direct-decoder", true};
```
wgtmac: I would call it `read.avro.skip-datum`, WDYT?

src/iceberg/test/avro_test.cc
```diff
@@ -244,4 +246,234 @@ TEST_F(AvroReaderTest, AvroWriterNestedType) {
...
+
+  std::string expected_string = R"([
+    [1, []],
+    [2, [10, 20, 30]]
+  ])";
+
+  WriteAndVerify(schema, expected_string);
+}
+
+// Skip Fixed and UUID tests for now - they require specific binary encoding
```
wgtmac:
```suggestion
```

src/iceberg/test/avro_test.cc
```diff
@@ -244,4 +246,234 @@ TEST_F(AvroReaderTest, AvroWriterNestedType) {
   WriteAndVerify(schema, expected_string);
 }
 
+// Comprehensive tests using in-memory MockFileIO
```
wgtmac:
```suggestion
```

src/iceberg/test/avro_test.cc
```diff
@@ -244,4 +246,234 @@ TEST_F(AvroReaderTest, AvroWriterNestedType) {
...
+  std::string expected_string = R"([
+    [1, ["tag1", "tag2", "tag3"]],
+    [2, ["foo", "bar"]],
+    [3, []]
+  ])";
+
+  WriteAndVerify(schema, expected_string);
+}
+
+TEST_F(AvroReaderTest, MapType) {
```
wgtmac: Can we test a map with a non-string-typed key?

src/iceberg/test/avro_test.cc
```diff
@@ -244,4 +246,234 @@ TEST_F(AvroReaderTest, AvroWriterNestedType) {
...
+    [2, null, 25],
+    [3, "Charlie", null],
+    [4, null, null]
+  ])";
+
+  WriteAndVerify(schema, expected_string);
+}
+
+// Test both direct decoder and GenericDatum paths
+TEST_F(AvroReaderTest, DirectDecoderVsGenericDatum) {
```
wgtmac: I would recommend changing it to parameterized test to enable/disable direct decoder for all cases in this file.

src/iceberg/test/avro_test.cc
```diff
@@ -244,4 +246,234 @@ TEST_F(AvroReaderTest, AvroWriterNestedType) {
...
+                                .projection = schema,
+                                .properties = std::move(reader_properties)});
+    ASSERT_THAT(reader_result, IsOk());
+    auto reader = std::move(reader_result.value());
+    ASSERT_NO_FATAL_FAILURE(VerifyNextBatch(*reader, expected_string));
+    ASSERT_NO_FATAL_FAILURE(VerifyExhausted(*reader));
+  }
+}
+
+TEST_F(AvroReaderTest, LargeDataset) {
```
wgtmac: It seems that we don't have test for projection. For example, only some columns are selected and they are reordered.

wgtmac:
BTW, it is possible to add a executable under `iceberg/avro` folder named `avro_scan`? We can use it to benchmark on a local avro file and be able to config whether to skip datum or not.

src/iceberg/avro/avro_reader.cc
shangxinli: Yeah, let's use direct/GenericDatum

src/iceberg/avro/avro_direct_decoder.cc
shangxinli: Make sense

src/iceberg/avro/avro_direct_decoder.cc
shangxinli: yes

src/iceberg/avro/avro_direct_decoder.cc
shangxinli: sure

src/iceberg/avro/avro_direct_decoder.cc
shangxinli: Good catch! I've added caching for the avro_to_projection mapping in DecodeContext.

src/iceberg/file_reader.h
shangxinli: Good suggestion! I've renamed it to follow the naming convention and be more descriptive.

  Changes:
  - kAvroUseDirectDecoder → kAvroSkipDatum
  - "avro.use-direct-decoder" → "read.avro.skip-datum"

src/iceberg/test/avro_test.cc
shangxinli: Good catch! I've added a test for maps with non-string keys.

src/iceberg/test/avro_test.cc
shangxinli: Good catch! I've added a test for column projection with subset selection and reordering.

src/iceberg/test/avro_test.cc
shangxinli: Converted the Avro tests to use parameterized tests as recommended. Created AvroReaderParameterizedTest fixture and Converted 12 tests to TEST_P.  Added INSTANTIATE_TEST_SUITE_P and Removed DirectDecoderVsGenericDatum test.

PR #383, opened 2025-12-03
feat: add strict metrics evaluator

dongxiao1198:
- add StrictMetricsEvaluator
- add TestCase & migrate all java case to test cc


wgtmac:
Thanks for working on this! I just found some inconsistency with the Java impl and made some slight changes.

PR #386, opened 2025-12-03
feat: implement InMemoryCatalog's UpdateTable

WZhuo:


src/iceberg/util/location_util.h
```diff
@@ -0,0 +1,43 @@
...
+
+class ICEBERG_EXPORT LocationUtil {
+ public:
+  static std::string StripTrailingSlash(const std::string& path) {
+    if (path.empty()) {
+      return "";
+    }
+
+    std::string_view result = path;
+    while (!result.ends_with("://") && result.ends_with("/")) {
```
zhjwpku:
```suggestion
    while (result.ends_with("/") && !result.ends_with("://")) {
```
This might short circuit more quickly, not a strong opinion though.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -314,6 +324,13 @@ Result<std::string> InMemoryNamespace::GetTableMetadataLocation(
   return it->second;
 }
 
+Status InMemoryNamespace::UpdateTableMetadataLocation(
+    TableIdentifier const& table_ident, std::string const& metadata_location) {
```
HuaHuaY: I think it's better to keep the same style `const T&` as function declaration.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -394,7 +411,32 @@ Result<std::unique_ptr<Table>> InMemoryCatalog::UpdateTable(
     const TableIdentifier& identifier,
     const std::vector<std::unique_ptr<TableRequirement>>& requirements,
     const std::vector<std::unique_ptr<TableUpdate>>& updates) {
-  return NotImplemented("update table");
+  std::unique_lock lock(mutex_);
```
HuaHuaY: Use `std::lock_guard` instead of `std::unique_lock` if we don't need to move the guard. Anyone interested can help modify the other `std::unique_lock` in this file.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -394,7 +411,32 @@ Result<std::unique_ptr<Table>> InMemoryCatalog::UpdateTable(
...
+  ICEBERG_ASSIGN_OR_RAISE(auto metadata_location,
+                          root_namespace_->GetTableMetadataLocation(identifier));
+
+  ICEBERG_ASSIGN_OR_RAISE(auto base,
+                          TableMetadataUtil::Read(*file_io_, metadata_location));
+  base->metadata_file_location = metadata_location;
```
HuaHuaY:
```suggestion
  base->metadata_file_location = std::move(metadata_location);
```

src/iceberg/catalog/memory/in_memory_catalog.cc
HuaHuaY, line 414:
```suggestion
  std::lock_guard guard(mutex_);
```
Use `std::lock_guard` instead of `std::unique_lock` if we don't need to move the guard.

src/iceberg/catalog/memory/in_memory_catalog.cc
HuaHuaY:
```suggestion
    const TableIdentifier& table_ident, const std::string& metadata_location) {
```
I think it's better to keep the same style `const T&` as function declaration.

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +290,95 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
   return TableMetadataFromJson(json);
 }
 
+Status TableMetadataUtil::Write(FileIO& io, const TableMetadata* base,
```
HuaHuaY: How about using reference instead of raw pointer?

src/iceberg/table_metadata.h
```diff
@@ -458,14 +467,41 @@ enum class ICEBERG_EXPORT MetadataFileCodecType {
...
+  /// \param codec The codec name.
+  /// \return The file extension of the codec.
+  static Result<std::string> CodecNameToFileExtension(const std::string_view& codec);
+
+  /// \brief Get the file extension from the codec type.
+  /// \param codec The codec type.
+  /// \return The file extension of the codec.
+  static std::string CodecTypeToFileExtension(MetadataFileCodecType codec);
+
+  inline static constexpr std::string_view kTableMetadataFileSuffix = ".metadata.json";
```
HuaHuaY: `static constexpr` is implicitly `inline` and we can remove `inline` here.

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +290,95 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+  ICEBERG_CHECK(metadata != nullptr, "The metadata is nullptr.");
+
+  int version = -1;
+  if (base != nullptr && !base->metadata_file_location.empty()) {
+    // parse current version from location
+    version = ParseVersionFromLocation(base->metadata_file_location);
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(std::string new_file_location,
+                          NewTableMetadataFilePath(*metadata, version + 1));
```
HuaHuaY: Should we plus one when parsing version from location?

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +290,95 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+                                                   const TableMetadata* metadata) {
+  if (!base) {
+    return;
+  }
+
+  bool delete_after_commit = TableProperties::kMetadataDeleteAfterCommitEnabled.value();
+  if (auto it = metadata->properties.find(
+          TableProperties::kMetadataDeleteAfterCommitEnabled.key());
+      it != metadata->properties.end()) {
+    delete_after_commit = StringUtils::ToLower(it->second) == "true" || it->second == "1";
```
HuaHuaY:
```suggestion
    delete_after_commit = StringUtils::EqualsIgnoreCase(it->second, "true") || it->second == "1";
```
`StringUtils::ToLower` allocates new memory and `StringUtils::EqualsIgnoreCase` doesn't.

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +290,95 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+  if (delete_after_commit) {
+    std::ranges::for_each(
+        base->metadata_log |
+            std::views::filter(
+                [current_files =
+                     metadata->metadata_log |
+                     std::ranges::to<std::unordered_set<MetadataLogEntry,
+                                                        MetadataLogEntry::Hasher>>()](
+                    const auto& entry) { return !current_files.contains(entry); }),
+        [&io](const auto& entry) { auto status = io.DeleteFile(entry.metadata_file); });
```
HuaHuaY:
```suggestion
    auto current_files =
        metadata->metadata_log |
        std::ranges::to<std::unordered_set<MetadataLogEntry, MetadataLogEntry::Hasher>>();
    std::ranges::for_each(
        base->metadata_log | std::views::filter([&current_files](const auto& entry) {
          return !current_files.contains(entry);
        }),
        [&io](const auto& entry) { std::ignore = io.DeleteFile(entry.metadata_file); });
```

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +290,95 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+  size_t version_start = metadata_location.find_last_of('/') + 1;
+  size_t version_end = metadata_location.find('-', version_start);
+
+  if (version_end == std::string::npos) {
+    // found filesystem table's metadata
+    return -1;
+  }
+
+  try {
+    return std::stoi(
```
HuaHuaY: If we ensure there are only numbers here, we can use `std::from_chars` here. It doesn't need to use `substr` to create a new string and is faster than `std::stoi`.
```cpp
#include <charconv>

int result;
if (std::from_chars(metadata_location.data() + version_start,
                    metadata_location.data() + version_end, result)
        .ec == std::errc()) {
  return result;
} else {
  return -1;
}
```

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +290,95 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+        metadata_location.substr(version_start, version_end - version_start));
+  } catch (const std::exception& e) {
+    // Unable to parse version from metadata location
+    return -1;
+  }
+}
+
+Result<std::string> TableMetadataUtil::NewTableMetadataFilePath(const TableMetadata& meta,
+                                                                int new_version) {
+  std::string codec_name = "none";
```
HuaHuaY:
```suggestion
  std::string_view codec_name = "none";
```

src/iceberg/table_metadata.cc
```diff
@@ -664,7 +781,32 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataBuilder::Build() {
...
   }
 
-  // 4. Create and return the TableMetadata
+  // 4. Buildup metadata_log from base metadata
+  int32_t max_metadata_log_size = TableProperties::kMetadataPreviousVersionsMax.value();
+  if (auto iter = impl_->metadata.properties.find(
+          TableProperties::kMetadataPreviousVersionsMax.key());
+      iter != impl_->metadata.properties.end()) {
+    try {
+      max_metadata_log_size = std::stoi(iter->second);
```
HuaHuaY: ditto

src/iceberg/table_metadata.h
WZhuo: It's implicitly inline in c++20+,  It's recommended to explicitly define as inline in my previous PR.

src/iceberg/table_metadata.cc
WZhuo: The base can be nullptr, indicating that this is creating a new table.

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: Fix all function declaration style in this file.

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: I think it can be optimized with ReadWriteLock later.

src/iceberg/util/location_util.h
WZhuo: Ok, I got it

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: Fix it

src/iceberg/table_metadata.cc
WZhuo: Fix it

src/iceberg/table_metadata.h
HuaHuaY: Sorry that I have no time last week so I didn't review your previous PR. This project is under at least C++23, so we don't need inline to avoid ODR violations.

src/iceberg/table_metadata.cc
HuaHuaY: You can turn `metadata` into reference.

src/iceberg/table_metadata.cc
WZhuo: The ParseVersionFromLocation method only gets the version of a specified file, not the next version. I think this makes the function's semantics clearer.

src/iceberg/catalog/memory/in_memory_catalog.cc
HuaHuaY: I agree that we may be able to replace the lock type with `std::shared_lock`. If you don't do that in this PR, I hope that you can replace with `std::lock_guard` first. Because we have merged a PR https://github.com/apache/iceberg-cpp/pull/398 which replaced the other `unique_lock` in this file.

src/iceberg/table_metadata.h
```diff
@@ -458,14 +467,40 @@ enum class ICEBERG_EXPORT MetadataFileCodecType {
...
   /// \brief Get the codec type from the table metadata file name.
   ///
   /// \param file_name The name of the table metadata file.
   /// \return The codec type of the table metadata file.
   static Result<MetadataFileCodecType> CodecFromFileName(std::string_view file_name);
 
+  /// \brief Get the file extension from the codec type.
+  /// \param codec The codec name.
+  /// \return The file extension of the codec.
+  static Result<std::string> CodecNameToFileExtension(const std::string_view& codec);
```
```suggestion
  static Result<std::string> CodecNameToFileExtension(std::string_view codec);
```

src/iceberg/table_metadata.h
```diff
@@ -458,14 +467,40 @@ enum class ICEBERG_EXPORT MetadataFileCodecType {
...
 };
 
-/// \brief Utility class for table metadata
-struct ICEBERG_EXPORT TableMetadataUtil {
+struct ICEBERG_EXPORT CodecTypeUtil {
+  /// \brief Returns the MetadataFileCodecType corresponding to the given string.
+  ///
+  /// \param name The string to parse.
+  /// \return The MetadataFileCodecType corresponding to the given string.
+  static Result<MetadataFileCodecType> CodecFromString(const std::string_view& name);
```
```suggestion
  static Result<MetadataFileCodecType> CodecFromString(std::string_view name);
```

It is recommended to pass string_view by value.

src/iceberg/table_metadata.h
```diff
@@ -76,6 +83,8 @@ struct ICEBERG_EXPORT TableMetadata {
   static constexpr int64_t kInvalidSequenceNumber = -1;
   static constexpr int64_t kInitialRowId = 0;
 
+  /// The location of the table metadata file
+  std::string metadata_file_location;
```
wgtmac: I'm hesitant to introduce this field to the `TableMetadata` because `TableMetadata` aims to carry fields that only exist in the metadata json. Can we revert this?

src/iceberg/table_metadata.h
```diff
@@ -58,6 +59,12 @@ struct ICEBERG_EXPORT MetadataLogEntry {
   friend bool operator==(const MetadataLogEntry& lhs, const MetadataLogEntry& rhs) {
     return lhs.timestamp_ms == rhs.timestamp_ms && lhs.metadata_file == rhs.metadata_file;
   }
+
+  struct Hasher {
```
wgtmac: Why do we need to add this? Why not do this for `SnapshotLogEntry`?

src/iceberg/table_metadata.cc
```diff
@@ -48,6 +52,8 @@ const TimePointMs kInvalidLastUpdatedMs = TimePointMs::min();
 constexpr int32_t kLastAdded = -1;
 }  // namespace
 
+static constexpr std::string_view kMetadataFolderName = "metadata";
```
wgtmac: Can we move it to the anonymous namespace above?

src/iceberg/table_metadata.cc
```diff
@@ -220,30 +226,57 @@ Result<TableMetadataCache::SnapshotsMap> TableMetadataCache::InitSnapshotMap(
          std::ranges::to<SnapshotsMap>();
 }
 
-// TableMetadataUtil implementation
+Result<MetadataFileCodecType> CodecTypeUtil::CodecFromString(
```
wgtmac: Perhaps `TableMetadataUtil` should be renamed to `TableMetadataFileUtil` so that we don't need to split a `CodecTypeUtil`? `CodecTypeUtil` is also confusing because it is tight to codec supported by only metadata file not those general codecs supported by data files (e.g. parquet).

src/iceberg/table_metadata.cc
wgtmac: I still don't think mixing metadata location with `TableMetadata` is a good idea. Can we directly pass the location instead?

src/iceberg/table_metadata.h
```diff
@@ -476,13 +511,56 @@ struct ICEBERG_EXPORT TableMetadataUtil {
...
+  /// \brief Parse the version number from a metadata file location.
+  ///
+  /// Extracts the version number from a metadata file path which follows
+  /// the format: vvvvv-uuid.metadata.json where vvvvv is the zero-padded
+  /// version number.
+  ///
+  /// \param metadata_location The metadata file location string
+  /// \return The parsed version number, or -1 if parsing fails or the
+  ///         location doesn't contain a version
+  static int ParseVersionFromLocation(const std::string_view& metadata_location);
```
```suggestion
  static int ParseVersionFromLocation(std::string_view metadata_location);
```

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +291,88 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+
+  if (delete_after_commit) {
+    auto current_files =
+        metadata.metadata_log |
+        std::ranges::to<std::unordered_set<MetadataLogEntry, MetadataLogEntry::Hasher>>();
+    std::ranges::for_each(
+        base->metadata_log | std::views::filter([&current_files](const auto& entry) {
+          return !current_files.contains(entry);
+        }),
+        [&io](const auto& entry) { auto status = io.DeleteFile(entry.metadata_file); });
```
```suggestion
        [&io](const auto& entry) { std::ignore = io.DeleteFile(entry.metadata_file); });
```

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +291,88 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+  }
+}
+
+int TableMetadataUtil::ParseVersionFromLocation(
+    const std::string_view& metadata_location) {
+  size_t version_start = metadata_location.find_last_of('/') + 1;
```
wgtmac, line 331: Is this correct? What if `/` does not exist?

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +291,88 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
 }
 
+void TableMetadataUtil::DeleteRemovedMetadataFiles(FileIO& io, const TableMetadata* base,
+                                                   const TableMetadata& metadata) {
+  if (!base) {
+    return;
+  }
+
+  bool delete_after_commit = TableProperties::kMetadataDeleteAfterCommitEnabled.value();
+  if (auto it = metadata.properties.find(
```
wgtmac: I was thinking to change `TableMetadata::properties` to directly use `std::shared_ptr<TableProperties>` then downstream can easily check any property. I will put up a PR for this.

src/iceberg/table_metadata.cc
```diff
@@ -258,13 +291,88 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataUtil::Read(
...
+        std::ranges::to<std::unordered_set<MetadataLogEntry, MetadataLogEntry::Hasher>>();
+    std::ranges::for_each(
+        base->metadata_log | std::views::filter([&current_files](const auto& entry) {
+          return !current_files.contains(entry);
+        }),
+        [&io](const auto& entry) { auto status = io.DeleteFile(entry.metadata_file); });
+  }
+}
+
+int TableMetadataUtil::ParseVersionFromLocation(
```
```suggestion
int32_t TableMetadataUtil::ParseVersionFromLocation(
```

src/iceberg/table_metadata.cc
wgtmac: Created https://github.com/apache/iceberg-cpp/pull/407

src/iceberg/table_metadata.cc
WZhuo: It's Ok if '/' does not exist, `version_end` will be std::string::npos, and return -1.

src/iceberg/table_metadata.h
WZhuo: SnapshotLogEntry not used currently. Has leave a TODO in TableMetadataBuilder::Build

src/iceberg/table_metadata.h
WZhuo: If seperate metadata_file_locatin field from TableMetadata, we need to convert more fields in some function
```
Status TableMetadataUtil::Write(FileIO& io, const TableMetadata* base, TableMetadata& metadata)
=>
Status TableMetadataUtil::Write(FileIO& io, const TableMetadata* base, const std::string& base_location, TableMetadata& metadata, std::string& new_location)
```
This filed is highly correlative with TableMetadata.BTW, it's also a filed of TableMetadata in Java impl.

src/iceberg/table_metadata.cc
WZhuo: How about redefine it in the TableMetadataUtil, TableMetadataUtil::Codec

PR #387, opened 2025-12-03
feat: add Transform::ProjectStrict

zhjwpku:


src/iceberg/test/predicate_test.cc
```diff
@@ -607,24 +606,24 @@ std::shared_ptr<BoundPredicate> AssertAndCastToBoundPredicate(
 }  // namespace
 
 TEST_F(PredicateTest, BoundUnaryPredicateTestIsNull) {
-  ICEBERG_ASSIGN_OR_THROW(auto is_null_pred, Expressions::IsNull("name")->Bind(
-                                                 *schema_, /*case_sensitive=*/true));
+  ICEBERG_UNWRAP_OR_FAIL(auto is_null_pred, Expressions::IsNull("name")->Bind(
```
wgtmac, line 609: TBH, I think we should remove `ICEBERG_UNWRAP_OR_FAIL` in favor of `ICEBERG_ASSIGN_OR_THROW` because the former one cannot be used in functions with return type.

src/iceberg/transform.cc
```diff
@@ -306,6 +306,66 @@ Result<std::unique_ptr<UnboundPredicate>> Transform::Project(
   std::unreachable();
 }
 
+Result<std::unique_ptr<UnboundPredicate>> Transform::ProjectStrict(
```
wgtmac, line 309: `Transform::ProjectStrict` looks so similar to `Transform::Project` with the only difference being the calls to `ProjectionUtil`. Perhaps we can use template to dispatch it?

wgtmac: After trying it by myself, I don't think this can help reduce duplicate lines. Ignore the above comment.

src/iceberg/util/projection_util_internal.h
```diff
@@ -40,248 +42,228 @@ namespace iceberg {
...
-        return GenericTransform(std::move(ref), predicate, func);
+        return nullptr;
     }
   }
 
-  template <typename T>
-    requires std::is_same_v<T, int32_t> || std::is_same_v<T, int64_t>
-  static Result<std::unique_ptr<UnboundPredicate>> TruncateInteger(
-      std::string_view name, const std::shared_ptr<BoundLiteralPredicate>& predicate,
+  // Apply to int32, int64, decimal
```
wgtmac:
```suggestion
  // Apply to int32, int64, decimal, and temporal types.
```

PR #398, opened 2025-12-06
refactor: use lock_guard instead of unique_lock when available

HuaHuaY:
Follow the answer in stackoverflow https://stackoverflow.com/a/60172828/11568166
1. `lock_guard` if you need to lock exactly 1 mutex for an entire scope.
2. `scoped_lock` if you need to lock a number of mutexes that is not exactly 1.
3. `unique_lock` if you need to unlock within the scope of the block (which includes use with a condition_variable).

So let clang-tidy only report `modernize-use-scoped-lock` warning when there are multiple `lock_guard`.

### DIFF
--- a/src/iceberg/table_metadata.h
+++ b/src/iceberg/table_metadata.h
@@ -419,7 +419,7 @@ class ICEBERG_EXPORT TableMetadataBuilder : public ErrorCollector {
   Result<std::unique_ptr<TableMetadata>> Build();
 
   /// \brief Destructor
-  ~TableMetadataBuilder();
+  ~TableMetadataBuilder() override;
 
   // Delete copy operations (use BuildFrom to create a new builder)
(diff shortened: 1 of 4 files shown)


src/iceberg/table_metadata.h
```diff
@@ -419,7 +419,7 @@ class ICEBERG_EXPORT TableMetadataBuilder : public ErrorCollector {
   Result<std::unique_ptr<TableMetadata>> Build();
 
   /// \brief Destructor
-  ~TableMetadataBuilder();
+  ~TableMetadataBuilder() override;
```
HuaHuaY, line 422: I find a warning here that it misses `override`.

zhjwpku (approved):
+1

wgtmac (approved):

PR #401, opened 2025-12-07
feat: update partition spec

zhjwpku:


src/iceberg/update/meson.build
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+install_headers(
+    ['update_partition_spec.h', 'update_properties.h'],
```
WZhuo: 'update/update_properties.h' is installed in src/iceberg/meson.build. If we want to unify it into src/iceberg/update/meson.build, could you please help modify it? Thanks.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,599 @@
...
+    : identifier_(std::move(identifier)),
+      catalog_(std::move(catalog)),
+      base_metadata_(std::move(base)) {
+  ICEBERG_DCHECK(catalog_, "Catalog is required to construct UpdatePartitionSpec");
+  ICEBERG_DCHECK(base_metadata_,
+                 "Base table metadata is required to construct UpdatePartitionSpec");
+  format_version_ = base_metadata_->format_version;
+
+  // Get the current/default partition spec
+  ICEBERG_ASSIGN_OR_THROW(spec_, base_metadata_->PartitionSpec());
```
WZhuo: AddError instead of throw exception

src/iceberg/update/meson.build
zhjwpku: Oh, I forgot to check it, will do, thanks

.github/workflows/test.yml
```diff
@@ -137,4 +137,4 @@ jobs:
           meson compile -C builddir
       - name: Test Iceberg
         run: |
-          meson test -C builddir
+          meson test -C builddir --timeout-multiplier=2
```
wgtmac: Why do we need this? Does it work?

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+
+namespace iceberg {
+
+UpdatePartitionSpec::UpdatePartitionSpec(TableIdentifier identifier,
+                                         std::shared_ptr<Catalog> catalog,
+                                         std::shared_ptr<TableMetadata> base)
+    : identifier_(std::move(identifier)),
+      catalog_(std::move(catalog)),
+      base_metadata_(std::move(base)) {
+  ICEBERG_DCHECK(catalog_, "Catalog is required to construct UpdatePartitionSpec");
```
wgtmac: What about calling `AddError` if `catalog` or `base` is null?

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+}
+
+UpdatePartitionSpec& UpdatePartitionSpec::AddNonDefaultSpec() {
+  set_as_default_ = false;
+  return *this;
+}
+
+UpdatePartitionSpec& UpdatePartitionSpec::AddField(const std::string& source_name) {
+  // Find the source field in the schema
+  auto field_result = schema_->FindFieldByName(source_name, case_sensitive_);
```
wgtmac: nit: we can call `BUILDER_RETURN_IF_ERROR` to make it shorter.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+    AddError(ErrorKind::kInvalidArgument,
+             std::format("Cannot find source field: {}", source_name));
+    return *this;
+  }
+
+  auto field_opt = field_result.value();
+  if (!field_opt.has_value()) {
+    AddError(ErrorKind::kInvalidArgument,
+             std::format("Cannot find source field: {}", source_name));
+    return *this;
```
wgtmac:
```suggestion
    return AddError(ErrorKind::kInvalidArgument,
             std::format("Cannot find source field: {}", source_name));
```

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+    if (deletes_.contains(existing.field_id()) && *existing.transform() == *transform) {
+      // If the field was deleted and we're re-adding the same one, just undo the delete
+      return RewriteDeleteAndAddField(existing, name);
+    }
+
+    if (deletes_.find(existing.field_id()) == deletes_.end()) {
+      AddError(
+          ErrorKind::kInvalidArgument,
+          std::format(
+              "Cannot add duplicate partition field for source {} with transform {}, "
```
wgtmac: Can we include the name in the error message? Same for error message below.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+  }
+
+  // Extract parameter from transform string for bucket and truncate
+  // Transform::ToString() returns "bucket[16]" or "truncate[4]" format
+  std::string transform_str = transform->ToString();
+
+  switch (transform->transform_type()) {
+    case TransformType::kIdentity:
+      return source_name;
+    case TransformType::kBucket: {
```
wgtmac: Isn't it an overkill to parse the string if we already have the `transform` instance?

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+                 source_id, transform->ToString(), added_it->second.ToString()));
+    return *this;
+  }
+
+  // Create or recycle the partition field
+  PartitionField new_field = RecycleOrCreatePartitionField(source_id, transform, name);
+
+  // Generate name if not provided
+  std::string field_name;
+  if (name != nullptr) {
```
wgtmac: Should we check `new_field.name().empty()` and then decide whether to generate partition name?

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+  // Handle name conflicts with existing fields
+  auto existing_name_it = name_to_field_.find(field_name);
+  if (existing_name_it != name_to_field_.end()) {
+    const auto& existing_field = existing_name_it->second;
+    if (!deletes_.contains(existing_field.field_id())) {
+      if (IsVoidTransform(existing_field)) {
+        // Rename the old deleted field
+        std::string renamed =
+            std::format("{}_{}", existing_field.name(), existing_field.field_id());
+        renames_[std::string(existing_field.name())] = renamed;
```
wgtmac: The java impl calls `renameField` here to handle some naming conflict cases. Should we follow that as well?

src/iceberg/update/update_partition_spec.h
```diff
@@ -0,0 +1,241 @@
...
+  /// \param term The unbound transform term.
+  /// \return Reference to this for method chaining.
+  UpdatePartitionSpec& AddField(std::shared_ptr<UnboundTerm<BoundTransform>> term);
+
+  /// \brief Add a new partition field with a custom name.
+  ///
+  /// \param name Name for the partition field.
+  /// \param term The unbound term representing the partition transform.
+  /// \return Reference to this for method chaining.
+  UpdatePartitionSpec& AddField(std::optional<std::string> name,
```
wgtmac: Should we directly use `std::string` and use an empty string to denote missing name?

src/iceberg/update/update_partition_spec.h
```diff
@@ -0,0 +1,241 @@
...
+  UpdatePartitionSpec& AddField(const std::string& source_name);
+
+  /// \brief Add a new partition field from an unbound term.
+  ///
+  /// The partition field will use the term's transform or the identity transform if
+  /// the term is a reference.
+  ///
+  /// \param term The unbound term representing the partition transform.
+  /// \return Reference to this for method chaining.
+  UpdatePartitionSpec& AddField(std::shared_ptr<UnboundTerm<BoundReference>> term);
```
wgtmac: Is it possible to use `std::shared_ptr<Term> term` as the input and use `Term::is_unbound()` and `Term::kind()` to check and cast to the right subclass? This may help simplify the API.

src/iceberg/update/update_partition_spec.h
```diff
@@ -0,0 +1,241 @@
...
+
+ private:
+  /// \brief Pair of source ID and transform string for indexing.
+  using TransformKey = std::pair<int32_t, std::string>;
+
+  /// \brief Hash function for TransformKey.
+  struct TransformKeyHash {
+    size_t operator()(const TransformKey& key) const {
+      return std::hash<int32_t>{}(key.first) ^
+             (std::hash<std::string>{}(key.second) << 1);
```
wgtmac:
```suggestion
      return 31 * std::hash<int32_t>{}(key.first) + std::hash<std::string>{}(key.second);
```

src/iceberg/update/update_partition_spec.h
```diff
@@ -0,0 +1,241 @@
...
+  int32_t format_version_;
+  std::shared_ptr<PartitionSpec> spec_;
+  std::shared_ptr<Schema> schema_;
+  bool case_sensitive_{true};
+  bool set_as_default_{true};
+  int32_t last_assigned_partition_id_;
+
+  // Indexes for existing fields
+  std::unordered_map<std::string, PartitionField> name_to_field_;
+  std::unordered_map<TransformKey, PartitionField, TransformKeyHash> transform_to_field_;
```
wgtmac:
```suggestion
  std::unordered_map<std::string, const PartitionField*> name_to_field_;
  std::unordered_map<TransformKey, const PartitionField*, TransformKeyHash> transform_to_field_;
```

It is unnecessary to use a copy here.

src/iceberg/update/update_partition_spec.h
```diff
@@ -0,0 +1,241 @@
...
+
+  // Pending changes
+  std::vector<PartitionField> adds_;
+  std::unordered_map<int32_t, PartitionField> added_time_fields_;
```
wgtmac: ditto, we can directly use pointer to `adds_`.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,608 @@
...
+      return false;
+  }
+}
+
+bool UpdatePartitionSpec::IsVoidTransform(const PartitionField& field) {
+  return field.transform()->transform_type() == TransformType::kVoid;
+}
+
+void UpdatePartitionSpec::CheckForRedundantAddedPartitions(const PartitionField& field) {
+  if (HasErrors()) return;
```
wgtmac: Why this is the only place to call it?

src/iceberg/update/update_partition_spec.cc
zhjwpku: I think checking `new_field.name().empty()` won't work when it find a recycled part field with it's name not empty, am I missing something?

src/iceberg/update/update_partition_spec.h
zhjwpku: Since adds_ can be realloced, using pointer to the array element is not safe. After some digging, I found using PartitionField::ToString as the value should work.

.github/workflows/test.yml
zhjwpku: I came across a timeout of meson build test running, rest_integration_test.
meson test has a default timeout of 30 seconds, maybe we should set `timeout-minutes: 30` for the whole meson job and let alone the meson test timeout.

Anyway, I will delete this in this PR.

src/iceberg/update/update_partition_spec.cc
zhjwpku: Done

src/iceberg/update/update_partition_spec.cc
zhjwpku: I add a Transform::GeneratePartitionName, that way we don't need to expose the param_ out of Transform.

src/iceberg/update/update_partition_spec.cc
zhjwpku: Yeah, I changed to the java impl way.

src/iceberg/update/update_partition_spec.cc
zhjwpku: I think CheckErrors in the Apply should be enough, I tried to remove the HasErrors calls, this one was an oversight.

src/iceberg/update/update_partition_spec.h
zhjwpku: Done.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+}
+
+UpdatePartitionSpec::UpdatePartitionSpec(std::shared_ptr<Transaction> transaction)
+    : PendingUpdate(std::move(transaction)) {
+  const TableMetadata* base_metadata = transaction_->base();
+  if (base_metadata == nullptr) [[unlikely]] {
+    AddError(ErrorKind::kInvalidArgument,
+             "Base table metadata is required to construct UpdatePartitionSpec");
+    return;
+  }
```
wgtmac:
```suggestion
  const TableMetadata& base_metadata = transaction_->current();
```

We should call `current` to get all current accumulated changes.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+    index.emplace(key, &field);
+  }
+  return index;
+}
+
+void UpdatePartitionSpec::BuildHistoricalFieldsIndex() {
+  const TableMetadata* base_metadata = transaction_->base();
+  if (base_metadata == nullptr) {
+    return;
+  }
```
wgtmac:
```suggestion
  const TableMetadata& base_metadata = transaction_->current();
```

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+
+  // Get the current schema
+  auto schema_result = base_metadata->Schema();
+  if (!schema_result.has_value()) {
+    AddError(schema_result.error());
+    return;
+  }
+  schema_ = std::move(schema_result.value());
+
+  last_assigned_partition_id_ = spec_->last_assigned_field_id();
```
wgtmac:
```suggestion
  last_assigned_partition_id_ = base_metadata->last_partition_id;
  if (last_assigned_partition_id_ < PartitionSpec::kLegacyPartitionDataIdStart - 1) {
    last_assigned_partition_id_ = PartitionSpec::kLegacyPartitionDataIdStart - 1;
  }
```

We cannot get `last_assigned_partition_id_` from the current spec since it may be an older one.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+  }
+  historical_fields_.reserve(total_fields);
+
+  // Index all fields from all historical partition specs
+  // Later specs override earlier ones for the same (source_id, transform) key
+  for (const auto& partition_spec : base_metadata->partition_specs) {
+    for (const auto& field : partition_spec->fields()) {
+      TransformKey key{field.source_id(), field.transform()->ToString()};
+      // Use emplace to only insert if key doesn't exist, preserving first occurrence
+      // This ensures we get the earliest field ID for recycling
```
wgtmac: Is this correct? IIUC, `base_metadata->partition_specs` does not guarantee any order of the partition specs.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+      auto field_opt, schema_->FindFieldByName(source_name, case_sensitive_));
+
+  ICEBERG_BUILDER_CHECK(field_opt.has_value(), "Cannot find source field: {}",
+                        source_name);
+  int32_t source_id = field_opt->get().field_id();
+  return AddFieldInternal(nullptr, source_id, Transform::Identity());
+}
+
+UpdatePartitionSpec& UpdatePartitionSpec::AddField(std::shared_ptr<NamedReference> term,
+                                                   std::string part_name) {
```
wgtmac:
```suggestion
UpdatePartitionSpec& UpdatePartitionSpec::AddField(const std::shared_ptr<NamedReference>& term,
                                                   const std::string& part_name) {
```

Both `term` and `part_name` are only accessed internally, we can use const reference to avoid copy.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+  // Bind the term to get the source field
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto bound_ref, term->Bind(*schema_, case_sensitive_));
+  int32_t source_id = bound_ref->field().field_id();
+
+  // Reference terms use identity transform
+  return AddFieldInternal(part_name.empty() ? nullptr : &part_name, source_id,
+                          Transform::Identity());
+}
+
+UpdatePartitionSpec& UpdatePartitionSpec::AddField(std::shared_ptr<UnboundTransform> term,
```
wgtmac: ditto

wgtmac: BTW, should we consolidate this with the above function using the signature below?

```cpp
AddField(const std::shared_ptr<Term>& term, const std::string& part_name)
```

We need to check `term->is_unbound()` and `term->kind()` to cast to the right type.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+
+UpdatePartitionSpec& UpdatePartitionSpec::AddFieldInternal(
+    const std::string* name, int32_t source_id, std::shared_ptr<Transform> transform) {
+  // Cache transform string to avoid repeated ToString() calls
+  const std::string transform_str = transform->ToString();
+  TransformKey validation_key{source_id, transform_str};
+
+  // Check for duplicate name in added fields
+  if (name != nullptr) {
+    ICEBERG_BUILDER_CHECK(!added_field_names_.contains(*name),
```
wgtmac: Move this check to the top?

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+    if (is_deleted && *existing->transform() == *transform) {
+      // If the field was deleted and we're re-adding the same one, just undo the delete
+      return RewriteDeleteAndAddField(*existing, name);
+    }
+
+    ICEBERG_BUILDER_CHECK(
+        is_deleted,
+        "Cannot add duplicate partition field {} for source {} with transform {}, "
+        "conflicts with {}",
+        name ? *name : "unknown", source_id, transform_str, existing->ToString());
```
wgtmac:
```suggestion
        "Cannot add duplicate partition field '{}' for source {} with transform {}, "
        "conflicts with {}",
        name ? *name : "", source_id, transform_str, existing->ToString());
```

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+  }
+
+  // Check if already being added
+  auto added_it = transform_to_added_field_.find(validation_key);
+  ICEBERG_BUILDER_CHECK(
+      added_it == transform_to_added_field_.end(),
+      "Cannot add duplicate partition field {} for source {} with transform {}, "
+      "already added: {}",
+      name ? *name : "unknown", source_id, transform_str, added_it->second);
```
wgtmac:
```suggestion
      "Cannot add duplicate partition field '{}' for source {} with transform {}, "
      "already added: {}",
      name ? *name : "", source_id, transform_str, added_it->second);
```

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+    result.updates.emplace_back(
+        std::make_unique<table::SetDefaultPartitionSpec>(spec_id));
+  }
+
+  return result;
+}
+
+int32_t UpdatePartitionSpec::AssignFieldId() { return ++last_assigned_partition_id_; }
+
+PartitionField UpdatePartitionSpec::RecycleOrCreatePartitionField(
```
wgtmac, line 359: Please add the comment from the java impl of `recycleOrCreatePartitionField`. I think it is helpful for understanding it.

src/iceberg/update/update_partition_spec.cc
wgtmac: Recycled partition field will always carry a non-empty field name. If we skip name comparison in the `RecycleOrCreatePartitionField` when `name` is empty, we can safely generate the partition name here when `new_field`'s name is empty.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+  // Bind the term to get the source field and transform
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto bound_transform,
+                                   term->Bind(*schema_, case_sensitive_));
+  int32_t source_id = bound_transform->reference()->field().field_id();
+  return AddFieldInternal(part_name.empty() ? nullptr : &part_name, source_id,
+                          bound_transform->transform());
+}
+
+UpdatePartitionSpec& UpdatePartitionSpec::AddFieldInternal(
+    const std::string* name, int32_t source_id, std::shared_ptr<Transform> transform) {
```
wgtmac: If my understanding of the partition name is correct, we can use `const std::string& name` here to make it simpler.

src/iceberg/update/update_partition_spec.cc
wgtmac: Same question for `RemoveField`.

src/iceberg/update/update_partition_spec.cc
wgtmac: I think we need to rephrase the comment here to avoid misleading understanding, or simply remove it.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+                                field.transform());
+      } else {
+        new_fields.push_back(field);
+      }
+    } else if (format_version_ < 2) {
+      // In V1, deleted fields are replaced with void transform
+      auto rename_it = renames_.find(std::string(field.name()));
+      std::string field_name =
+          rename_it != renames_.end() ? rename_it->second : std::string(field.name());
+      new_fields.emplace_back(field.source_id(), field.field_id(), field_name,
```
wgtmac:
```suggestion
      new_fields.emplace_back(field.source_id(), field.field_id(), std::move(field_name),
```

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+      // Field is kept, check for rename
+      auto rename_it = renames_.find(std::string(field.name()));
+      if (rename_it != renames_.end()) {
+        new_fields.emplace_back(field.source_id(), field.field_id(), rename_it->second,
+                                field.transform());
+      } else {
+        new_fields.push_back(field);
+      }
+    } else if (format_version_ < 2) {
+      // In V1, deleted fields are replaced with void transform
```
wgtmac: Let's add the comment from java impl?

wgtmac: // field IDs were not required for v1 and were assigned sequentially in each partition spec
        // starting at 1,000.
        // to maintain consistent field ids across partition specs in v1 tables, any partition field
        // that is removed
        // must be replaced with a null transform. null values are always allowed in partition data.

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+                              Transform::Void());
+    }
+    // In V2, deleted fields are simply removed
+  }
+
+  // Add new fields
+  new_fields.insert(new_fields.end(), adds_.begin(), adds_.end());
+
+  // Determine the new spec ID
+  int32_t new_spec_id = spec_ ? spec_->spec_id() + 1 : PartitionSpec::kInitialSpecId;
```
wgtmac: The logic here and below are not necessary. The right approach is to implement `TableMetadataBuilder::SetDefaultPartitionSpec` and `TableMetadataBuilder::AddPartitionSpec` to let it handle reusing or creating a new spec_id and then call it from here.

@HeartLinked is refactoring `PendingUpdate::Apply` and `Transaction` so we need to wait that PR before making such changes.

src/iceberg/transform.cc
```diff
@@ -388,6 +388,28 @@ std::string Transform::ToString() const {
...
+      // Format: sourceName_bucket_N (matching Java: sourceName + "_bucket_" + numBuckets)
+      return std::format("{}_bucket_{}", source_name, std::get<int32_t>(param_));
+    case TransformType::kTruncate:
+      // Format: sourceName_trunc_N (matching Java: sourceName + "_trunc_" + width)
+      return std::format("{}_trunc_{}", source_name, std::get<int32_t>(param_));
+    case TransformType::kYear:
+    case TransformType::kMonth:
+    case TransformType::kDay:
+    case TransformType::kHour:
+    case TransformType::kUnknown:
```
wgtmac: Should we use `Result` as the return type and return error for kUnknown?

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+  // No matching field found, create a new one
+  std::string field_name = name ? *name : "";
+  return {source_id, AssignFieldId(), field_name, transform};
+}
+
+std::string UpdatePartitionSpec::GeneratePartitionName(
+    int32_t source_id, const std::shared_ptr<Transform>& transform) const {
+  // Find the source field name
+  auto field_result = schema_->FindFieldById(source_id);
+  std::string_view source_name = "unknown";
```
wgtmac: What about returning Result and handling return for `FindFieldById` and unknown name?

src/iceberg/update/update_partition_spec.cc
```diff
@@ -0,0 +1,488 @@
...
+  if (field_result.has_value() && field_result.value().has_value()) {
+    source_name = field_result.value().value().get().name();
+  }
+
+  return transform->GeneratePartitionName(std::string(source_name));
```
wgtmac:
```suggestion
  return transform->GeneratePartitionName(source_name);
```

`Transform::GeneratePartitionName` accepts string_view.

src/iceberg/transform.cc
```diff
@@ -388,6 +388,29 @@ std::string Transform::ToString() const {
   std::unreachable();
 }
 
+Result<std::string> Transform::GeneratePartitionName(std::string_view source_name) const {
+  switch (transform_type_) {
+    case TransformType::kIdentity:
+      return std::string(source_name);
+    case TransformType::kBucket:
+      // Format: sourceName_bucket_N (matching Java: sourceName + "_bucket_" + numBuckets)
```
wgtmac:
```suggestion
```

I think this is obvious enough.

src/iceberg/transform.cc
```diff
@@ -388,6 +388,29 @@ std::string Transform::ToString() const {
...
+      return std::format("{}_bucket_{}", source_name, std::get<int32_t>(param_));
+    case TransformType::kTruncate:
+      // Format: sourceName_trunc_N (matching Java: sourceName + "_trunc_" + width)
```
wgtmac:
```suggestion
```

src/iceberg/table_metadata.cc
```diff
@@ -572,6 +582,69 @@ Result<int32_t> TableMetadataBuilder::Impl::AddSortOrder(const SortOrder& order)
...
+    // update last_added_spec_id if the spec was added in this set of changes (since it
+    // is now the last)
+    bool is_new_spec =
+        last_added_spec_id_.has_value() &&
+        std::ranges::find_if(changes_, [new_spec_id](const auto& change) {
+          return change->kind() == TableUpdate::Kind::kAddPartitionSpec &&
+                 internal::checked_cast<const table::AddPartitionSpec&>(*change)
+                         .spec()
+                         ->spec_id() == new_spec_id;
+        }) != changes_.cend();
```
WZhuo, line 621:
```suggestion
        std::ranges::any_of(changes_, [new_spec_id](const auto& change) {
          return change->kind() == TableUpdate::Kind::kAddPartitionSpec &&
                 internal::checked_cast<const table::AddPartitionSpec&>(*change)
                         .spec()
                         ->spec_id() == new_spec_id;
        });
```

WZhuo (approved):

src/iceberg/table_metadata.cc
zhjwpku: Good suggestion, I do this refactor in #440, please take a look.

PR #402, opened 2025-12-07
feat: add residual evaluator

zhjwpku:


src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,567 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/expression/residual_evaluator.h"
+
+#include <cstddef>
+
```
HuaHuaY:
```suggestion
```

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,567 @@
...
+  Result<std::shared_ptr<Expression>> Or(
+      const std::shared_ptr<Expression>& left_result,
+      const std::shared_ptr<Expression>& right_result) override {
+    return Expressions::Or(left_result, right_result);
+  }
+
+  Result<std::shared_ptr<Expression>> IsNull(
+      const std::shared_ptr<Bound>& expr) override {
+    return IsNullImpl(expr);
```
HuaHuaY: I think that in source files we don't need the impl methods.

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,567 @@
...
+
+      if (bound_strict->is_bound_predicate()) {
+        // Evaluate the bound predicate against partition data
+        ICEBERG_ASSIGN_OR_RAISE(
+            strict_result, EvaluateBoundPredicate(
+                               std::dynamic_pointer_cast<BoundPredicate>(bound_strict)));
+      } else {
+        // If the result is not a predicate, then it must be a constant like alwaysTrue
+        // or alwaysFalse
+        strict_result = bound_strict;
```
HuaHuaY:
```suggestion
        strict_result = std::move(bound_strict);
```

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,567 @@
...
+      if (bound_inclusive->is_bound_predicate()) {
+        // Evaluate the bound predicate against partition data
+        ICEBERG_ASSIGN_OR_RAISE(
+            inclusive_result,
+            EvaluateBoundPredicate(
+                std::dynamic_pointer_cast<BoundPredicate>(bound_inclusive)));
+      } else {
+        // If the result is not a predicate, then it must be a constant like alwaysTrue
+        // or alwaysFalse
+        inclusive_result = bound_inclusive;
```
HuaHuaY:
```suggestion
        inclusive_result = std::move(bound_inclusive);
```

src/iceberg/expression/residual_evaluator.cc
zhjwpku: Introducing the xxxImpl mainly because they are used more than once, I'm not sure if the name is appropriate though.

src/iceberg/expression/residual_evaluator.h
```diff
@@ -0,0 +1,93 @@
...
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/partition_spec.h"
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+class Expression;
+class StructLike;
```
wgtmac:
```suggestion
```

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,565 @@
...
+class ResidualVisitor : public BoundVisitor<std::shared_ptr<Expression>> {
+ public:
+  ResidualVisitor(const std::shared_ptr<PartitionSpec>& spec,
+                  const std::shared_ptr<Schema>& schema, const StructLike& partition_data,
+                  bool case_sensitive)
+      : spec_(spec),
+        schema_(schema),
+        partition_data_(partition_data),
+        case_sensitive_(case_sensitive) {
+    ICEBERG_ASSIGN_OR_THROW(auto partition_type_, spec_->PartitionType(*schema));
```
wgtmac: We shouldn't throw in the ctor. Please check this in the `Make` function and make sure all inputs are valid to the ctor.

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,565 @@
...
+    partition_schema_ = FromStructType(std::move(*partition_type_), std::nullopt);
```
wgtmac: `partition_type_` seems to be useless?

src/iceberg/expression/residual_evaluator.h
```diff
@@ -0,0 +1,93 @@
...
+  ///
+  /// \param partition_data Partition data values
+  /// \return The residual of this evaluator's expression from the partition values
+  virtual Result<std::shared_ptr<Expression>> ResidualFor(
+      const StructLike& partition_data) const;
+
+ protected:
+  ResidualEvaluator(std::shared_ptr<Expression> expr,
+                    const std::shared_ptr<PartitionSpec>& spec,
+                    const std::shared_ptr<Schema>& schema, bool case_sensitive);
```
wgtmac: Should we just use `const PartitionSpec& spec` and `const Schema& schema`?

src/iceberg/expression/residual_evaluator.cc
wgtmac: BTW, do you think we should change `PartitionSpec::PartitionType` to return `Schema` instead of `StructType` to make such calls easier?

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,565 @@
...
+  }
+
+  Result<std::shared_ptr<Expression>> AlwaysTrue() override {
+    return Expressions::AlwaysTrue();
```
wgtmac: We cannot use `Expressions` family in this file because they may throw. Please use corresponding `Make` functions.

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,565 @@
...
+
+  const auto& str_value = std::get<std::string>(value.value());
+  const auto& str_prefix = std::get<std::string>(lit.value());
+  if (str_value.starts_with(str_prefix)) {
+    return Expressions::AlwaysFalse();
+  }
+  return Expressions::AlwaysTrue();
+}
+
+Result<std::shared_ptr<Expression>> ResidualVisitor::EvaluateBoundPredicate(
```
wgtmac: Why do you reinvent `BoundVisitor::Predicate(const std::shared_ptr<BoundPredicate>& pred)` again?

src/iceberg/expression/residual_evaluator.cc
wgtmac:
```suggestion
    ICEBERG_ASSIGN_OR_THROW(auto partition_type, spec_->PartitionType(*schema));
```

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,565 @@
...
+  bool case_sensitive_;
+};
+
+Result<std::shared_ptr<Expression>> ResidualVisitor::IsNullImpl(
+    const std::shared_ptr<Bound>& expr) {
+  ICEBERG_ASSIGN_OR_RAISE(auto value, expr->Evaluate(partition_data_));
+  if (value.IsNull()) {
+    return Expressions::AlwaysTrue();
+  }
+  return Expressions::AlwaysFalse();
```
wgtmac: We can make all these functions are a one-liner if we do:

```cpp
namespace {
std::shared_ptr<Expression> always_true() { return True::Instance(); }
std::shared_ptr<Expression> always_false() { return False::Instance(); }
}  // namespace

  Result<std::shared_ptr<Expression>> IsNull(
      const std::shared_ptr<Bound>& expr) override {
    return expr->Evaluate(partition_data_).transform([](const auto& value) {
      return value.IsNull() ? always_true() : always_false();
    });
  }

```

src/iceberg/expression/residual_evaluator.cc
```diff
@@ -0,0 +1,565 @@
...
+  // Similarly the inclusive projection returns false iff the original predicate would
+  // have returned false, so the predicate can also be eliminated if the inclusive
+  // projection evaluates to false.
+
+  // Get the field ID from the predicate's reference
+  const auto& ref = pred->reference();
+  int32_t field_id = ref->field().field_id();
+
+  // Find partition fields that match this source field ID
+  std::vector<const PartitionField*> matching_fields;
```
wgtmac: It seems worth adding a `PartitionSpec::GetFieldsBySourceId` using `lazy.h` to avoid repeated search?

src/iceberg/expression/residual_evaluator.cc
zhjwpku: ResidualVisitor override that, how can a subclass call the superclass's method when overriding the interface method itself?

zhjwpku: Let me try using the BoundVisitor::Predicate directly.

src/iceberg/expression/residual_evaluator.cc
zhjwpku: Done

src/iceberg/expression/residual_evaluator.h
zhjwpku: Changed to const ref

src/iceberg/expression/residual_evaluator.cc
zhjwpku: Removed the reinvent one, sorry for my misunderstanding.

src/iceberg/expression/residual_evaluator.cc
zhjwpku: Added, thanks for the reminder.

src/iceberg/expression/residual_evaluator.cc
zhjwpku: done.

PR #403, opened 2025-12-08
feat: add manifest evaluator

dongxiao1198:
support manifest evaluator and unittest

### DIFF
--- a/src/iceberg/expression/manifest_evaluator.cc
+++ b/src/iceberg/expression/manifest_evaluator.cc
@@ -40,9 +40,9 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   explicit ManifestEvalVisitor(const ManifestFile& manifest)
       : stats_(manifest.partitions) {}
 
-  Result<bool> AlwaysTrue() override { return true; }
+  Result<bool> AlwaysTrue() override { return kRowsMightMatch; }
 
-  Result<bool> AlwaysFalse() override { return false; }
+  Result<bool> AlwaysFalse() override { return kRowCannotMatch; }
 
   Result<bool> Not(bool child_result) override { return !child_result; }
 
@@ -57,7 +57,7 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   Result<bool> IsNull(const std::shared_ptr<Bound>& expr) override {
     // no need to check whether the field is required because binding evaluates that case
     // if the column has no null values, the expression cannot match
-    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    const auto& ref = expr->reference();
     ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
     if (!stats_.at(pos).contains_null) {
       return kRowCannotMatch;
@@ -67,7 +67,7 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   }
 
   Result<bool> NotNull(const std::shared_ptr<Bound>& expr) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    const auto& ref = expr->reference();
     ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
     if (AllValuesAreNull(stats_.at(pos), ref->type()->type_id())) {
       return kRowCannotMatch;
@@ -77,7 +77,7 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   }
 
   Result<bool> IsNaN(const std::shared_ptr<Bound>& expr) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    const auto& ref = expr->reference();
     ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
     if (stats_.at(pos).contains_nan.has_value() && !stats_.at(pos).contains_nan.value()) {
       return kRowCannotMatch;
@@ -90,7 +90,7 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   }
 
   Result<bool> NotNaN(const std::shared_ptr<Bound>& expr) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    const auto& ref = expr->reference();
     ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
     const auto& summary = stats_.at(pos);
     // if containsNaN is true, containsNull is false and lowerBound is null, all values
@@ -104,7 +104,7 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   }
 
   Result<bool> Lt(const std::shared_ptr<Bound>& expr, const Literal& lit) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    const auto& ref = expr->reference();
     ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
     const auto& summary = stats_.at(pos);
     if (!summary.lower_bound.has_value()) {
@@ -119,7 +119,7 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   }
 
   Result<bool> LtEq(const std::shared_ptr<Bound>& expr, const Literal& lit) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    const auto& ref = expr->reference();
     ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
     const auto& summary = stats_.at(pos);
     if (!summary.lower_bound.has_value()) {
@@ -134,7 +134,7 @@ class ManifestEvalVisitor : public BoundVisitor<bool> {
   }
 
   Result<bool> Gt(const std::shared_ptr<Bound>& expr, const Literal& lit) override {
-    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    const auto& ref = expr->reference();
     ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
     const auto& summary = stats_.at(pos);
     if (!summary.upper_bound.has_value()) {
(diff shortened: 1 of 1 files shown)


src/iceberg/expression/manifest_evaluator.h
```diff
@@ -0,0 +1,84 @@
...
+      const Schema& schema, bool case_sensitive = true);
+
+  /// \brief Make a manifest evaluator for PartitionFilter
+  ///
+  /// \param expr The expression to evaluate
+  /// \param spec The partition spec
+  /// \param schema The schema of the table
+  /// \param case_sensitive Whether field name matching is case-sensitive
+  static Result<std::unique_ptr<ManifestEvaluator>> MakePartitionFilter(
+      std::shared_ptr<Expression> expr, const std::shared_ptr<PartitionSpec> spec,
```
HuaHuaY: Why is the signature `const shared_ptr` instead of `shared_ptr` or `const shared_ptr&`?

src/iceberg/expression/manifest_evaluator.h
```diff
@@ -0,0 +1,84 @@
...
+/// Files are passed to #eval(ManifestFile), which returns true if the manifest may
+/// contain data files that match the partition expression. Manifest files may be
+/// skipped if and only if the return value of eval is false.
+///
+
+#include <memory>
+
+#include "iceberg/expression/expression.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/partition_spec.h"
```
HuaHuaY:
```suggestion
#include "iceberg/iceberg_export.h"
```
We can use forward declaration and move the include into source files.

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,397 @@
...
+constexpr bool kRowCannotMatch = false;
+constexpr int32_t kInPredicateLimit = 200;
+}  // namespace
+
+class ManifestEvalVisitor : public BoundVisitor<bool> {
+ public:
+  explicit ManifestEvalVisitor(const ManifestFile& manifest)
+      : stats_(manifest.partitions) {}
+
+  Result<bool> AlwaysTrue() override { return true; }
```
HuaHuaY: @wgtmac Too many subclasses override these methods but it seems meaningless. Should we open another PR to merge the implementations?

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,397 @@
...
+                         });
+
+    if (literals_view.empty()) {
+      // if all values are less than lower bound, rows cannot match.
+      return kRowCannotMatch;
+    }
+
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto upper, DeserializeBoundLiteral(summary.upper_bound.value(), ref->type()));
+    auto filtered_view = literals_view | std::views::filter([&](const Literal& lit) {
```
HuaHuaY: How about merging two `std::views::filter`?
```cpp
    auto filtered_view = literal_set | std::views::filter([&lower, &upper](const Literal& lit) {
                           return lower <= lit && lit <= upper;
                         });
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,397 @@
...
+    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
+    const auto& summary = stats_.at(pos);
+    if (!summary.lower_bound.has_value() || !summary.upper_bound.has_value()) {
+      return kRowCannotMatch;
+    }
+    if (lit.type()->type_id() != TypeId::kString) {
+      return kRowCannotMatch;
+    }
+    const auto& prefix = get<std::string>(lit.value());
```
HuaHuaY: I'm a little confused why the namespace prefix is ​​not needed here.

HuaHuaY: It seems to be related to ADL(Argument-Dependent Lookup). I suggest to add `std::`.

src/iceberg/expression/manifest_evaluator.h
dongxiao1198: typo, fixed

src/iceberg/expression/manifest_evaluator.cc
dongxiao1198: +1

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,391 @@
...
+
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto lower, DeserializeBoundLiteral(summary.lower_bound.value(), ref->type()));
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto upper, DeserializeBoundLiteral(summary.upper_bound.value(), ref->type()));
+    auto literals_view = literal_set | std::views::filter([&](const Literal& lit) {
+                           return lower <= lit && lit <= upper;
+                         });
+
+    if (literals_view.empty()) {
```
HuaHuaY: Sorry I only noticed duplicate code in the previous round of reviews. I think it would be clearer to replace `.empty()` with `std::ranges::all_of`.
```cpp
    if (std::ranges::all_of(literal_set, [&](const Literal& lit) {
          return lit < lower || lit > upper;
        })) {
```
And you need to add `#include <algorithm>`

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+constexpr bool kRowCannotMatch = false;
+constexpr int32_t kInPredicateLimit = 200;
+}  // namespace
+
+class ManifestEvalVisitor : public BoundVisitor<bool> {
+ public:
+  explicit ManifestEvalVisitor(const ManifestFile& manifest)
+      : stats_(manifest.partitions) {}
+
+  Result<bool> AlwaysTrue() override { return true; }
```
wgtmac:
```suggestion
  Result<bool> AlwaysTrue() override { return kRowsMightMatch; }
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+
+  Result<bool> AlwaysFalse() override { return false; }
```
wgtmac:
```suggestion
  Result<bool> AlwaysFalse() override { return kRowCannotMatch; }
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+      // of prefix
+      if (upper_bound.starts_with(prefix)) {
+        return kRowCannotMatch;
+      }
+    }
+    return kRowsMightMatch;
+  }
+
+ private:
+  Result<BoundReference*> ParseBoundReference(const std::shared_ptr<Bound>& expr) const {
```
wgtmac: We don't need this function. The Java impl extends `BoundExpressionVisitor` which can only deal with bound reference. However, we extend `BoundVisitor` so that we can directly call `expr->reference()` to get a bound reference.

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+      return Invalid("Invalid expression: not a BoundReference");
+    }
+    return ref;
+  }
+
+  Result<size_t> GetPosition(const BoundReference& ref) const {
+    const auto& accessor = ref.accessor();
+    const auto& position_path = accessor.position_path();
+    if (position_path.empty()) {
+      return Invalid("Invalid accessor: empty position path.");
```
wgtmac:
```suggestion
      return InvalidArgument("Invalid accessor: empty position path.");
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+    }
+    // nested accessors are not supported for partition fields
+    if (position_path.size() > 1) {
+      return Invalid("Cannot convert nested accessor to position");
```
wgtmac:
```suggestion
      return InvalidArgument("Cannot convert nested accessor to position");
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+    }
+    auto pos = position_path.at(0);
+    if (pos >= stats_.size()) {
+      return Invalid("Invalid position: out of partition filed range.");
```
wgtmac:
```suggestion
      return InvalidArgument("Position {} is out of partition field range {}", pos, stats_.size());
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+    return allNull;
+  }
+
+  Result<Literal> DeserializeBoundLiteral(const std::vector<uint8_t>& bound,
+                                          const std::shared_ptr<Type>& type) const {
+    if (!type->is_primitive()) {
+      return NotSupported("Bounds of non-primitive partition fields are not supported.");
+    }
+    auto primitive_type = internal::checked_pointer_cast<PrimitiveType>(type);
+    return Literal::Deserialize(bound, primitive_type);
```
wgtmac:
```suggestion
    return Literal::Deserialize(bound, std::move(primitive_type));
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+  Result<bool> StartsWith(const std::shared_ptr<Bound>& expr,
+                          const Literal& lit) override {
+    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
+    const auto& summary = stats_.at(pos);
+    if (!summary.lower_bound.has_value() || !summary.upper_bound.has_value()) {
+      return kRowCannotMatch;
+    }
+    if (lit.type()->type_id() != TypeId::kString) {
+      return Invalid("Invalid literal: not a string, cannot use StartsWith");
```
wgtmac:
```suggestion
      return InvalidExpression("Invalid literal: not a string, cannot use StartsWith");
```

src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -0,0 +1,390 @@
...
+                             const Literal& lit) override {
+    ICEBERG_ASSIGN_OR_RAISE(auto ref, ParseBoundReference(expr));
+    ICEBERG_ASSIGN_OR_RAISE(auto pos, GetPosition(*ref));
+    const auto& summary = stats_.at(pos);
+    if (summary.contains_null || !summary.lower_bound.has_value() ||
+        !summary.upper_bound.has_value()) {
+      return kRowsMightMatch;
+    }
+    if (lit.type()->type_id() != TypeId::kString) {
+      return Invalid("Invalid literal: not a string, cannot use notStartsWith");
```
wgtmac:
```suggestion
      return InvalidExpression("Invalid literal: not a string, cannot use notStartsWith");
```

wgtmac (approved):

PR #405, opened 2025-12-10
refactor: optimize with shared_lock/unique_lock for InMemoryCatalog

WZhuo:


src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -337,42 +337,42 @@ std::string_view InMemoryCatalog::name() const { return catalog_name_; }
 
 Status InMemoryCatalog::CreateNamespace(
     const Namespace& ns, const std::unordered_map<std::string, std::string>& properties) {
-  std::lock_guard guard(mutex_);
+  std::unique_lock lock(mutex_);
```
HuaHuaY, line 339: Keep `std::lock_guard` if you only need write guard here.

WZhuo: std::lock_guard cannot be used to acquire a write lock on a std::shared_mutex

HuaHuaY: `std::lock_guard` is a template wrapper ([cppreference link](https://en.cppreference.com/w/cpp/thread/lock_guard.html)). Any lock which meets [BasicLockable](https://en.cppreference.com/w/cpp/named_req/BasicLockable.html) can be used in `std::lock_guard`, including `std::shared_mutex`.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -443,12 +447,14 @@ Result<std::unique_ptr<Table>> InMemoryCatalog::LoadTable(
...
+  {
+    std::unique_lock lock(mutex_);
+    if (!root_namespace_->NamespaceExists(identifier.ns)) {
+      return NoSuchNamespace("table namespace does not exist.");
+    }
+    if (!root_namespace_->RegisterTable(identifier, metadata_file_location)) {
+      return UnknownError("The registry failed.");
+    }
   }
   return LoadTable(identifier);
```
HuaHuaY: Is it possible that another thread drops the table between line 458 and 459? If it's, what will happen? I think we may get an error `Table does not exist` which looks strange in `RegisterTable`. I'm not sure if there are other concurrency conflict scenarios here.
In the previous design, mutex_ is a recursive mutex. I think recursive mutex is a bad design. But when we change it into a non-recursive mutex, we need to be careful. If we think the current code is acceptable, I suggest to add some comments here.
I think a better solution is adding a new method `LoadTableImpl`.

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: I was wrong — lock_guard can indeed lock a shared_mutex, but it is still recommended to use unique_lock and shared_lock for read-write locking on a shared_mutex for better code readability.

HuaHuaY: I don't think there are some differences about code readability here. `std::unique_lock` has additional overhead but `std::scoped_lock` and `std::lock_guard` don't. It should not be compared with `std::scoped_lock` or `std::lock_guard`.

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: Yeah, It should be first load metadata from file and check the validation of the metadata file. Only after a successful validation should it be registered to the root_namespace.

HuaHuaY: How do you think about this scene?
1. thread a: line 458 register table A
2. thread b: drop A and register same name table A'
3. thread a: line 459 load A'

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: Using unique_lock and shared_lock with shared_mutex makes it clear that this is a read-write lock, as seeing unique_lock immediately indicates that exclusive (write) access is being acquired. And, in other mutex scenarios, use lock_guard.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -387,48 +388,50 @@ Result<std::unique_ptr<Table>> InMemoryCatalog::CreateTable(
...
 
 Status InMemoryCatalog::RenameTable(const TableIdentifier& from,
                                     const TableIdentifier& to) {
+  std::unique_lock lock(mutex_);
   return NotImplemented("rename table");
 }
 
 Result<std::unique_ptr<Table>> InMemoryCatalog::LoadTable(
     const TableIdentifier& identifier) {
-  if (!file_io_) [[unlikely]] {
```
wgtmac, line 425: Why this gets removed?

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -443,14 +446,20 @@ Result<std::unique_ptr<Table>> InMemoryCatalog::LoadTable(
 
 Result<std::shared_ptr<Table>> InMemoryCatalog::RegisterTable(
     const TableIdentifier& identifier, const std::string& metadata_file_location) {
-  std::lock_guard guard(mutex_);
+  ICEBERG_CHECK(file_io_, "file_io is not set for catalog {}", catalog_name_);
```
wgtmac: Shouldn't we return an error?

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -443,14 +446,20 @@ Result<std::unique_ptr<Table>> InMemoryCatalog::LoadTable(
...
+  ICEBERG_ASSIGN_OR_RAISE(auto metadata,
+                          TableMetadataUtil::Read(*file_io_, metadata_file_location));
```
wgtmac, line 454: It might be delayed until register is successful. Or we can add a new lockless `LoadTableImpl` to wrap the internal.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -21,8 +21,9 @@
 
 #include <algorithm>
 #include <iterator>
-#include <mutex>
+#include <shared_mutex>
```
wgtmac: This seems redundant.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -21,8 +21,9 @@
...
 
+#include "iceberg/exception.h"
```
wgtmac: Please not add this since we cannot throw in most of the codebase.

src/iceberg/catalog/memory/in_memory_catalog.cc
wgtmac: +1 on `std::lock_guard` if we care the performance here.

HuaHuaY: Compiler can optimize the simple scene ([godbolt](https://godbolt.org/z/9M7jaTYsr)). But I still suggest to keep the code from doing unnecessary things. The simpler the better.

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: Here's an issue: if a file is invalid, register returns failure, but the map may already contain an entry. Subsequent calls to register will then return "already exists".
Therefore, I think we should validate the file's legitimacy in advance, ensuring that only valid table metadata can be successfully registered.

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: It is an atomicity issue between Register and Load. We can discuss it based on the new code.

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: I don't think it's unnecessary, and I recommend use the unique_lock/shared_lock pair for locking shared_mutex

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: `file_io_` should not be nullptr, so I replace this with ICEBERG_CHECK.

wgtmac: ICEBERG_CHECK will throw internally. I recommend to use ICEBERG_DCHECK

wgtmac (approved):

PR #407, opened 2025-12-12
refactor: adapt table metadata to use table properties

wgtmac:


src/iceberg/table.h
zhjwpku, line 133: This field should be removed accordingly.

liurenjie1024 (approved):

PR #408, opened 2025-12-14
feat: add snapshot update

zhjwpku:


src/iceberg/snapshot.cc
```diff
@@ -75,6 +78,24 @@ std::optional<std::string_view> Snapshot::operation() const {
   return std::nullopt;
 }
 
+std::optional<int64_t> Snapshot::FirstRowId() const {
+  auto it = summary.find("first-row-id");
```
gty404: add new kFirstRowId to SnapshotSummaryFields ？

src/iceberg/snapshot.cc
```diff
@@ -75,6 +78,24 @@ std::optional<std::string_view> Snapshot::operation() const {
...
+  if (it == summary.end()) {
+    return std::nullopt;
+  }
+
+  return StringUtils::ParseInt<int64_t>(it->second);
+}
+
+std::optional<int64_t> Snapshot::AddedRows() const {
+  auto it = summary.find("added-rows");
```
gty404: SnapshotSummaryFields::kAddedRows

src/iceberg/snapshot.cc
```diff
@@ -141,4 +162,103 @@ Result<std::span<ManifestFile>> SnapshotCache::DeleteManifests(
...
+  return Builder(SnapshotRefType::kBranch, snapshot_id);
+}
+
+SnapshotRef::Builder SnapshotRef::Builder::BuilderFor(int64_t snapshot_id,
+                                                      SnapshotRefType type) {
+  return Builder(type, snapshot_id);
+}
+
+SnapshotRef::Builder SnapshotRef::Builder::BuilderFrom(const SnapshotRef& ref) {
+  Builder builder(ref.type(), ref.snapshot_id);
```
gty404:
```suggestion
  return BuilderFrom(ref, ref.snapshot_id);
```

src/iceberg/snapshot.cc
```diff
@@ -141,4 +162,103 @@ Result<std::span<ManifestFile>> SnapshotCache::DeleteManifests(
...
+  } else {
+    const auto& tag = std::get<SnapshotRef::Tag>(ref.retention);
+    builder.max_ref_age_ms_ = tag.max_ref_age_ms;
+  }
+  return builder;
+}
+
+SnapshotRef::Builder& SnapshotRef::Builder::MinSnapshotsToKeep(
+    std::optional<int32_t> value) {
+  if (type_ == SnapshotRefType::kTag && value.has_value()) {
```
gty404:
```suggestion
  ICEBERG_BUILDER_CHECK(type_ != SnapshotRefType::kTag || !value.has_value(), "Tags do not support setting minSnapshotsToKeep");
```

src/iceberg/snapshot.cc
```diff
@@ -141,4 +162,103 @@ Result<std::span<ManifestFile>> SnapshotCache::DeleteManifests(
...
+    return AddError(ErrorKind::kInvalidArgument,
+                    "Tags do not support setting minSnapshotsToKeep");
+  }
+  if (value.has_value() && value.value() <= 0) {
```
gty404: The following can also be changed to use ICEBERG_BUILDER_CHECK

src/iceberg/snapshot.h
```diff
@@ -119,6 +120,67 @@ struct ICEBERG_EXPORT SnapshotRef {
...
+    /// \return A Result containing the SnapshotRef instance, or an error if validation
+    /// failed
+    Result<SnapshotRef> Build() const;
+
+   private:
+    explicit Builder(SnapshotRefType type, int64_t snapshot_id);
+
+    SnapshotRefType type_;
+    int64_t snapshot_id_;
+    std::optional<int32_t> min_snapshots_to_keep_;
```
gty404: Should we use Branch/Tag directly ?  
std::optional<SnapshotRef::Branch> branch_;
  std::optional<SnapshotRef::Tag> tag_;

src/iceberg/table_metadata.cc
```diff
@@ -982,6 +994,158 @@ void TableMetadataBuilder::Impl::SetLocation(std::string_view location) {
...
+    const auto& branch = std::get<SnapshotRef::Branch>(ref->retention);
+    min_snapshots_to_keep = branch.min_snapshots_to_keep;
+    max_snapshot_age_ms = branch.max_snapshot_age_ms;
+    max_ref_age_ms = branch.max_ref_age_ms;
+  } else {
+    const auto& tag = std::get<SnapshotRef::Tag>(ref->retention);
+    max_ref_age_ms = tag.max_ref_age_ms;
+  }
+
+  changes_.push_back(std::make_unique<table::SetSnapshotRef>(
```
gty404: How about modifying the construction parameters of SetSnapshotRef to accept SnapshotRef?

src/iceberg/table_metadata.cc
gty404, line 1: We also need to rebuild the SnapshotLog list in TableMetadataBuilder::Impl::Build.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+  std::optional<int64_t> snapshot_id =
+      snapshot_id_ ? std::make_optional(*snapshot_id_) : std::nullopt;
+
+  // Create factory function for rolling manifest writer
+  RollingManifestWriter::ManifestWriterFactory factory =
+      [this, spec, current_schema, format_version,
+       snapshot_id]() -> Result<std::unique_ptr<ManifestWriter>> {
+    std::string manifest_path = ManifestPath();
+
+    if (format_version == 1) {
```
gty404: The related judgments for the format version (format_version) are scattered across multiple methods. Each version's logic exists in the form of if-else, encapsulating the processing logic of each version into independent implementations to avoid the expansion of if-else branches?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+  // Return the new snapshot
+  return ApplyResult{.snapshot = staged_snapshot_,
+                     .target_branch = target_branch_,
+                     .stage_only = stage_only_};
+}
+
+Status SnapshotUpdate::Finalize() {
+  // Cleanup after successful commit
+  if (cleanup_after_commit()) {
+    auto cached_snapshot = SnapshotCache(staged_snapshot_.get());
```
gty404, line 316: `staged_snapshot` might be null if an exception occurs in the `apply` method ？

src/iceberg/util/uuid.h
```diff
@@ -78,6 +78,9 @@ class ICEBERG_EXPORT Uuid : public util::Formattable {
     return lhs.data_ == rhs.data_;
   }
 
+  int64_t highbits() const;
+  int64_t lowbits() const;
```
wgtmac:
```suggestion
  int64_t high_bits() const;
  int64_t low_bits() const;
```

src/iceberg/util/string_util.h
```diff
@@ -54,6 +56,16 @@ class ICEBERG_EXPORT StringUtils {
     }
     return count;
   }
+
+  template <typename T>
+  static std::optional<T> ParseInt(std::string_view str) {
```
wgtmac: Should we return `Result<T>` and convert all errors to `InvalidArgument` with detail message for `std::errc::invalid_argument` and `std::errc::result_out_of_range`?

src/iceberg/table_metadata.h
```diff
@@ -145,6 +145,8 @@ struct ICEBERG_EXPORT TableMetadata {
   Result<std::shared_ptr<iceberg::Snapshot>> Snapshot() const;
   /// \brief Get the snapshot of this table with the given id
   Result<std::shared_ptr<iceberg::Snapshot>> SnapshotById(int64_t snapshot_id) const;
+  /// \brief Get the next sequence number for
```
```suggestion
  /// \brief Get the next sequence number
```

src/iceberg/table_metadata.cc
```diff
@@ -982,6 +994,158 @@ void TableMetadataBuilder::Impl::SetLocation(std::string_view location) {
...
+      })) {
+    metadata_.last_updated_ms = snapshot->timestamp_ms;
+  }
+
+  // If it's MAIN_BRANCH, update currentSnapshotId and add to snapshotLog
+  if (name == SnapshotRef::kMainBranch) {
+    metadata_.current_snapshot_id = ref->snapshot_id;
+    if (metadata_.last_updated_ms == kInvalidLastUpdatedMs) {
+      metadata_.last_updated_ms =
+          TimePointMs{std::chrono::duration_cast<std::chrono::milliseconds>(
```
wgtmac: Perhaps we can add an utility function to iceberg/util/timepoint.h to return now since this is too long.

src/iceberg/table_metadata.cc
wgtmac: +1

src/iceberg/table_metadata.cc
```diff
@@ -982,6 +994,158 @@ void TableMetadataBuilder::Impl::SetLocation(std::string_view location) {
...
+        "Cannot add a snapshot, first-row-id is behind table next-row-id: {} < {}",
+        first_row_id.value(), metadata_.next_row_id);
+
+    metadata_.next_row_id += snapshot->AddedRows().value_or(0);
+  }
+
+  return {};
+}
+
+Status TableMetadataBuilder::Impl::SetBranchSnapshot(int64_t snapshot_id,
```
wgtmac, line 1083: Java impl has a convenience method `setBranchSnapshot(Snapshot snapshot, String branch)` which we don't have. Do we need it?

src/iceberg/table_metadata.cc
```diff
@@ -982,6 +994,158 @@ void TableMetadataBuilder::Impl::SetLocation(std::string_view location) {
...
+          snapshot->sequence_number <= metadata_.last_sequence_number,
+      "Last sequence number {} is less than existing snapshot sequence number {}",
+      metadata_.last_sequence_number, snapshot->sequence_number);
+
+  // Create new ref: either from existing ref or create new branch ref
+  std::shared_ptr<SnapshotRef> new_ref;
+  if (ref_it != metadata_.refs.end()) {
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto ref_result,
+        SnapshotRef::Builder::BuilderFrom(*ref_it->second, snapshot_id).Build());
```
wgtmac: It seems that `SnapshotRef::Builder::Build()` should return `Result<std::unique_ptr<SnapshotRef>>`.

src/iceberg/snapshot.h
```diff
@@ -119,6 +120,67 @@ struct ICEBERG_EXPORT SnapshotRef {
...
+
+    /// \brief Set the maximum reference age in milliseconds
+    /// \param value The maximum reference age in milliseconds, or nullopt for default
+    /// \return Reference to this builder for method chaining
+    Builder& MaxRefAgeMs(std::optional<int64_t> value);
+
+    /// \brief Build the SnapshotRef
+    /// \return A Result containing the SnapshotRef instance, or an error if validation
+    /// failed
+    Result<SnapshotRef> Build() const;
```
wgtmac: As I have commented in other comment, it is better to return `Result<std::unique_ptr<SnapshotRef>>`.

src/iceberg/snapshot.h
```diff
@@ -119,6 +120,67 @@ struct ICEBERG_EXPORT SnapshotRef {
     return lhs.Equals(rhs);
   }
 
+  /// \brief Builder class for constructing SnapshotRef objects
+  class ICEBERG_EXPORT Builder : public ErrorCollector {
```
wgtmac: I'm not sure if it is simple to directly add `Result<std::unique_ptr<SnapshotRef>> SnapshotRef::MakeBranch(...)` and `Result<std::unique_ptr<SnapshotRef>> SnapshotRef::MakeTag(...)`? For `BuildFrom`, is it simpler to add a `SnapshotRef::Clone` and directly modify values on the cloned object? Introducing builder pattern seems an overkill here.

src/iceberg/snapshot.cc
```diff
@@ -75,6 +78,24 @@ std::optional<std::string_view> Snapshot::operation() const {
   return std::nullopt;
 }
 
+std::optional<int64_t> Snapshot::FirstRowId() const {
```
wgtmac: Should we return `Result<std::optional<int64_t>>` and error out on parsing failure?

src/iceberg/snapshot.cc
```diff
@@ -75,6 +78,24 @@ std::optional<std::string_view> Snapshot::operation() const {
...
+  auto it = summary.find("first-row-id");
+  if (it == summary.end()) {
+    return std::nullopt;
+  }
+
+  return StringUtils::ParseInt<int64_t>(it->second);
+}
+
+std::optional<int64_t> Snapshot::AddedRows() const {
```
wgtmac: Same question here.

src/iceberg/transaction.cc
```diff
@@ -158,6 +176,15 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
   ICEBERG_ASSIGN_OR_RAISE(auto updated_table, table_->catalog()->UpdateTable(
                                                   table_->name(), requirements, updates));
 
+  for (const auto& update : pending_updates_) {
+    if (auto update_ptr = update.lock()) {
+      if (auto snapshot_update =
+              internal::checked_cast<SnapshotUpdate*>(update_ptr.get())) {
+        std::ignore = snapshot_update->Finalize();
```
wgtmac: cc @dongxiao1198 as `ExpireSnapshots` may need to deal with the same issue.

wgtmac (commented):
I haven't reviewed `snapshot_update.h/.cc` yet. Just post my findings so far.

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+#include <atomic>
+#include <functional>
+#include <memory>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include "iceberg/catalog.h"
```
wgtmac: Let's use forward declaration as much as possible.

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+#include "iceberg/transaction.h"
+#include "iceberg/type_fwd.h"
+#include "iceberg/update/pending_update.h"
+
+namespace iceberg {
+
+/// \brief Base class for operations that produce snapshots.
+///
+/// This class provides common functionality for creating new snapshots,
+/// including manifest list writing, commit retries, and cleanup.
```
wgtmac: We don't have `retries`, right?

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+///
+class ICEBERG_EXPORT SnapshotUpdate : public PendingUpdate {
+ public:
+  /// \brief Result of applying a snapshot update
+  struct ApplyResult {
+    /// \brief The new snapshot
```
wgtmac:
```suggestion
```

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+  /// and creates a new snapshot without committing it. The snapshot
+  /// is stored internally and can be accessed after Apply() succeeds.
+  ///
+  /// \return A result containing the new snapshot, or an error
+  Result<ApplyResult> Apply();
+
+  /// \brief Finalizes the snapshot, cleaning up any uncommitted files.
+  ///
+  /// \return Status indicating success or failure
+  Status Finalize();
```
wgtmac: It seems that we can add it to the base `PendingUpdate` class if desired.

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+
+  ~SnapshotUpdate() override = default;
+
+  /// \brief Set a callback to delete files instead of the table's default.
+  ///
+  /// \param delete_func A function used to delete file locations
+  /// \return Reference to this for method chaining
+  /// \tparam T The concrete subclass type
+  template <typename T>
+    requires std::is_base_of_v<SnapshotUpdate, T>
```
wgtmac: How about using `deducing this` just like `ErrorCollector::AddError`?

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+                   const std::string& total_property, const std::string& added_property,
+                   const std::string& deleted_property);
+
+  int32_t format_version_;
+  std::shared_ptr<PartitionSpec> spec_;
+  std::shared_ptr<Schema> schema_;
+
+  std::function<Status(const std::string&)> delete_func_;
+  bool stage_only_ = false;
+  std::string target_branch_ = std::string(SnapshotRef::kMainBranch);
```
wgtmac:
```suggestion
  std::string target_branch_{SnapshotRef::kMainBranch};
```

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+                   const std::unordered_map<std::string, std::string>& previous_summary,
+                   const std::string& total_property, const std::string& added_property,
+                   const std::string& deleted_property);
+
+  int32_t format_version_;
+  std::shared_ptr<PartitionSpec> spec_;
+  std::shared_ptr<Schema> schema_;
+
+  std::function<Status(const std::string&)> delete_func_;
+  bool stage_only_ = false;
```
wgtmac:
```suggestion
  bool stage_only_{false};
```

Just be consistent

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+  std::string target_branch_ = std::string(SnapshotRef::kMainBranch);
+
+  std::optional<int64_t> snapshot_id_{std::nullopt};
+  std::atomic<int32_t> attempt_{0};
```
wgtmac: We haven't added concurrent support so it is an overkill to use atomic for now.

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+  std::atomic<int32_t> manifest_count_{0};
+  std::vector<std::string> manifest_lists_;
+  std::string commit_uuid_;
+  std::shared_ptr<Snapshot> staged_snapshot_;
+
+  int64_t target_manifest_size_bytes_;
+  // For format version > 1, inheritance is enabled by default
+  bool can_inherit_snapshot_id_{true};
```
wgtmac: Can we apply the same order of class member variables as the Java impl to make it easier for future change?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/update/snapshot_update.h"
+
+#include <charconv>
```
wgtmac:
```suggestion
```

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+  /// \brief Get the path for a manifest file
+  std::string ManifestPath();
+
+  /// \brief Update a total property in the summary
+  void UpdateTotal(std::unordered_map<std::string, std::string>& summary,
+                   const std::unordered_map<std::string, std::string>& previous_summary,
+                   const std::string& total_property, const std::string& added_property,
+                   const std::string& deleted_property);
+
+  int32_t format_version_;
```
wgtmac:
```suggestion
```

This is unused

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+#include "iceberg/util/snapshot_util_internal.h"
+#include "iceberg/util/string_util.h"
+#include "iceberg/util/uuid.h"
+
+namespace iceberg {
+
+SnapshotUpdate::SnapshotUpdate(std::shared_ptr<Transaction> transaction)
+    : PendingUpdate(std::move(transaction)) {
+  target_manifest_size_bytes_ =
+      transaction_->current().properties.Get(TableProperties::kManifestTargetSizeBytes);
```
wgtmac: It seems that `transaction_->current()` has been called for many times. How about adding `const TableMetadata& base_` as a class member variable or simply adding a function like `const TableMetadata& base() const`?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+  // For format version 1, check if snapshot ID inheritance is enabled
+  if (transaction_->current().format_version == 1) {
+    can_inherit_snapshot_id_ = transaction_->current().properties.Get(
```
wgtmac: I think `target_manifest_size_bytes_`, `can_inherit_snapshot_id_` and `commit_uuid_` can be initialized directly in the member initializer list.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+      if (manifest_list != staged_snapshot_->manifest_list) {
+        std::ignore = DeleteFile(manifest_list);
+      }
+    }
+  }
+
+  return {};
+}
+
+void SnapshotUpdate::SetTargetBranch(const std::string& branch) {
```
wgtmac:
```suggestion
Status SnapshotUpdate::SetTargetBranch(const std::string& branch) {
```

Returning `Status` would help write less code and enforce callers to fail fast.

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+
+  /// \brief Set a callback to delete files instead of the table's default.
+  ///
+  /// \param delete_func A function used to delete file locations
+  /// \return Reference to this for method chaining
+  /// \tparam T The concrete subclass type
+  template <typename T>
+    requires std::is_base_of_v<SnapshotUpdate, T>
+  T& DeleteWith(std::function<Status(const std::string&)> delete_func) {
+    delete_func_ = std::move(delete_func);
```
wgtmac: The Java impl prevents multiple assignment here. Should we do the same thing?

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+
+  const std::string& commit_uuid() const { return commit_uuid_; }
+
+  int32_t attempt() const { return attempt_.load(); }
+
+  int32_t manifest_count() const { return manifest_count_.load(); }
+
+  /// \brief Clean up any uncommitted manifests that were created.
+  ///
+  /// Manifests may not be committed if apply is called multiple times
```
wgtmac:
```suggestion
  /// Manifests may not be committed if Apply is called multiple times
```

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+  virtual std::string operation() = 0;
+
+  /// \brief Validate the current metadata.
+  ///
+  /// Child operations can override this to add custom validation.
+  ///
+  /// \param current_metadata Current table metadata to validate
+  /// \param snapshot Ending snapshot on the lineage which is being validated
+  virtual Status Validate(const TableMetadata& current_metadata,
+                          const std::shared_ptr<Snapshot>& snapshot) = 0;
```
wgtmac:
```suggestion
                          const std::shared_ptr<Snapshot>& snapshot) {}
```

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+
+  /// \brief Apply the update's changes to the given metadata and snapshot.
+  ///
+  /// \param metadata_to_update The base table metadata to apply changes to
+  /// \param snapshot Snapshot to apply the changes to
+  /// \return A vector of manifest files for the new snapshot
+  virtual std::vector<ManifestFile> Apply(const TableMetadata& metadata_to_update,
```
wgtmac: Should it return `Result<std::vector<ManifestFile>>`?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+      parent_snapshot_id = std::nullopt;
+    }
+    return std::unexpected<Error>(parent_snapshot_result.error());
+  } else {
+    parent_snapshot = *parent_snapshot_result;
+    parent_snapshot_id = parent_snapshot->snapshot_id;
+  }
+  int64_t sequence_number = transaction_->current().NextSequenceNumber();
+
+  ICEBERG_RETURN_UNEXPECTED(Validate(transaction_->current(), parent_snapshot));
```
wgtmac:
```suggestion
  if (parent_snapshot != nullptr) {
    ICEBERG_RETURN_UNEXPECTED(Validate(transaction_->current(), parent_snapshot));
  }
```

`parent_snapshot` can be nullptr in which case it shouldn't be validated.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+Result<SnapshotUpdate::ApplyResult> SnapshotUpdate::Apply() {
+  ICEBERG_RETURN_UNEXPECTED(CheckErrors());
+
+  // Get the latest snapshot for the target branch
+  std::shared_ptr<Snapshot> parent_snapshot;
+  std::optional<int64_t> parent_snapshot_id;
+  auto parent_snapshot_result =
+      SnapshotUtil::LatestSnapshot(transaction_->current(), target_branch_);
+  if (!parent_snapshot_result.has_value()) [[unlikely]] {
```
wgtmac: I'm not sure if we want to add something like below:

```cpp
Result<std::shared_ptr<Snapshot>> TransferNotFoundSnapshot(
    Result<std::shared_ptr<Snapshot>>&& snapshot_result) {
  if (!snapshot_result.has_value()) {
    if (snapshot_result.error().kind == ErrorKind::kNotFound) {
      return nullptr;
    }
    return std::unexpected<Error>(std::move(snapshot_result.error()));
  }
  return std::move(snapshot_result.value());
}

}  // namespace
```

Or we can add a `SnapshotUtil::OptionalLatestSnapshot` to return nullptr instead of error if not found?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+  std::string manifest_list_path = ManifestListPath();
+  manifest_lists_.push_back(manifest_list_path);
+
+  // Create manifest list writer based on format version
+  int8_t format_version = transaction_->current().format_version;
+  int64_t snapshot_id = SnapshotId();
+  std::unique_ptr<ManifestListWriter> writer;
+
+  if (format_version == 1) {
```
wgtmac: Do we need a factory function to create the writer of different versions? It has been scattered all around and not easy to maintain if we add v4, v5 in the future.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+Status SnapshotUpdate::DeleteFile(const std::string& path) { return delete_func_(path); }
+
+std::string SnapshotUpdate::ManifestListPath() {
+  // Generate manifest list path
+  // Format: {metadata_location}/snap-{snapshot_id}-{attempt}-{uuid}.avro
+  int64_t snapshot_id = SnapshotId();
+  std::string filename = std::format("snap-{}-{}-{}.avro", snapshot_id,
+                                     attempt_.fetch_add(1) + 1, commit_uuid_);
+  return std::format("{}/metadata/{}", transaction_->table()->location(), filename);
```
wgtmac: We should follow the Java impl in `BaseMetastoreTableOperations.metadataFileLocation` as below:

```java
  protected String metadataFileLocation(TableMetadata metadata, String filename) {
    String metadataLocation = metadata.properties().get(TableProperties.WRITE_METADATA_LOCATION);

    if (metadataLocation != null) {
      return String.format("%s/%s", LocationUtil.stripTrailingSlash(metadataLocation), filename);
    } else {
      return String.format("%s/%s/%s", metadata.location(), METADATA_FOLDER_NAME, filename);
    }
  }
```

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+}
+
+std::string SnapshotUpdate::ManifestPath() {
+  // Generate manifest path
+  // Format: {metadata_location}/{uuid}-m{manifest_count}.avro
+  int32_t count = manifest_count_.fetch_add(1);
+  std::string filename = std::format("{}-m{}.avro", commit_uuid_, count);
+  return std::format("{}/metadata/{}", transaction_->table()->location(), filename);
```
wgtmac: ditto

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+}
+
+Status SnapshotUpdate::DeleteFile(const std::string& path) { return delete_func_(path); }
+
+std::string SnapshotUpdate::ManifestListPath() {
+  // Generate manifest list path
+  // Format: {metadata_location}/snap-{snapshot_id}-{attempt}-{uuid}.avro
+  int64_t snapshot_id = SnapshotId();
+  std::string filename = std::format("snap-{}-{}-{}.avro", snapshot_id,
+                                     attempt_.fetch_add(1) + 1, commit_uuid_);
```
wgtmac:
```suggestion
                                     ++attempt_, commit_uuid_);
```

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+  return std::format("{}/metadata/{}", transaction_->table()->location(), filename);
+}
+
+std::string SnapshotUpdate::ManifestPath() {
+  // Generate manifest path
+  // Format: {metadata_location}/{uuid}-m{manifest_count}.avro
+  int32_t count = manifest_count_.fetch_add(1);
+  std::string filename = std::format("{}-m{}.avro", commit_uuid_, count);
```
wgtmac:
```suggestion
  std::string filename = std::format("{}-m{}.avro", commit_uuid_, manifest_count_++);
```

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+    return std::unexpected<Error>(parent_snapshot_result.error());
+  } else {
+    parent_snapshot = *parent_snapshot_result;
+    parent_snapshot_id = parent_snapshot->snapshot_id;
+  }
+  int64_t sequence_number = transaction_->current().NextSequenceNumber();
+
+  ICEBERG_RETURN_UNEXPECTED(Validate(transaction_->current(), parent_snapshot));
+
+  std::vector<ManifestFile> manifests = Apply(transaction_->current(), parent_snapshot);
```
wgtmac: From the Java impl, some `ManifestFile`s here may not have set `added_snapshot_id` (i.e. which is `kInvalidSnapshotId`. Therefore it uses a `manifestsWithMetadata` cache and opens a ManifestReader to fill some missing fields to it (see `addMetadata` function). We might need to do the same thing here.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+  ICEBERG_RETURN_UNEXPECTED(writer->AddAll(manifests));
+  ICEBERG_RETURN_UNEXPECTED(writer->Close());
+
+  // Get nextRowId and assignedRows for format version 3
+  std::optional<int64_t> next_row_id;
+  std::optional<int64_t> assigned_rows;
+  if (format_version >= 3) {
+    next_row_id = transaction_->current().next_row_id;
+    if (writer->next_row_id().has_value()) {
```
wgtmac: Use `ICEBERG_CHECK` to return an error when `writer->next_row_id().has_value()` is false.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+      assigned_rows = writer->next_row_id().value() - next_row_id.value();
+    }
+  }
+
+  std::unordered_map<std::string, std::string> summary = Summary();
+  std::string op = operation();
+
+  if (!op.empty() && op == DataOperation::kReplace) {
```
wgtmac: Return error when `!op.empty()`?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+    auto added_records_it = summary.find(SnapshotSummaryFields::kAddedRecords);
+    auto replaced_records_it = summary.find(SnapshotSummaryFields::kDeletedRecords);
+    if (added_records_it != summary.end() && replaced_records_it != summary.end()) {
+      auto added_records = StringUtils::ParseInt<int64_t>(added_records_it->second);
```
wgtmac: We should return error if parsing fails.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+  std::optional<int64_t> next_row_id;
+  std::optional<int64_t> assigned_rows;
+  if (format_version >= 3) {
+    next_row_id = transaction_->current().next_row_id;
+    if (writer->next_row_id().has_value()) {
+      assigned_rows = writer->next_row_id().value() - next_row_id.value();
+    }
+  }
+
+  std::unordered_map<std::string, std::string> summary = Summary();
```
wgtmac: Can we call this inside the `if (!op.empty() && op == DataOperation::kReplace)` branch?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+    }
+  }
+
+  summary = ComputeSummary(transaction_->current());
+
+  // Get current time
+  auto now = std::chrono::system_clock::now();
+  auto duration_since_epoch = now.time_since_epoch();
+  TimePointMs timestamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(
+      std::chrono::system_clock::time_point(duration_since_epoch));
```
wgtmac: nit: wrap this function in the timepoint util.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+  // Get schema ID
+  std::optional<int32_t> schema_id = transaction_->current().current_schema_id;
```
wgtmac: We have removed optional from schema_id, so you can directly pass `transaction_->current().current_schema_id` to Snapshot now.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+  // Create snapshot
+  staged_snapshot_ =
+      std::make_shared<Snapshot>(Snapshot{.snapshot_id = snapshot_id,
```
wgtmac: We need to set nextRowId and assignedRows as well.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+  target_branch_ = branch;
+}
+
+std::unordered_map<std::string, std::string> SnapshotUpdate::ComputeSummary(
+    const TableMetadata& previous) {
+  std::unordered_map<std::string, std::string> summary = Summary();
+
+  if (summary.empty()) {
+    return std::unordered_map<std::string, std::string>{};
```
wgtmac:
```suggestion
    return summary;
```

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+  }
+
+  // Get previous summary from the target branch
+  std::unordered_map<std::string, std::string> previous_summary;
+  if (auto ref_it = previous.refs.find(target_branch_); ref_it != previous.refs.end()) {
+    auto snapshot_result = previous.SnapshotById(ref_it->second->snapshot_id);
```
wgtmac: We can call `TransferNotFoundSnapshot` as suggested above.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+              SnapshotSummaryFields::kAddedDeleteFiles,
+              SnapshotSummaryFields::kRemovedDeleteFiles);
+  UpdateTotal(summary, previous_summary, SnapshotSummaryFields::kTotalPosDeletes,
+              SnapshotSummaryFields::kAddedPosDeletes,
+              SnapshotSummaryFields::kRemovedPosDeletes);
+  UpdateTotal(summary, previous_summary, SnapshotSummaryFields::kTotalEqDeletes,
+              SnapshotSummaryFields::kAddedEqDeletes,
+              SnapshotSummaryFields::kRemovedEqDeletes);
+
+  return summary;
```
wgtmac, line 396: Add a `TODO` comment to notify that we can add custom summary fields like engine info in the future here.

src/iceberg/update/snapshot_update.h
```diff
@@ -0,0 +1,221 @@
...
+  Status DeleteFile(const std::string& path);
+
+  /// \brief Get the path for a manifest list file
+  std::string ManifestListPath();
+
+  /// \brief Get the path for a manifest file
+  std::string ManifestPath();
+
+  /// \brief Update a total property in the summary
+  void UpdateTotal(std::unordered_map<std::string, std::string>& summary,
```
wgtmac: Make it a static function or even only define it in the anonymous namespace in the cc file?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+}
+
+void SnapshotUpdate::UpdateTotal(
+    std::unordered_map<std::string, std::string>& summary,
+    const std::unordered_map<std::string, std::string>& previous_summary,
+    const std::string& total_property, const std::string& added_property,
+    const std::string& deleted_property) {
+  auto total_it = previous_summary.find(total_property);
+  if (total_it != previous_summary.end()) {
+    auto new_total_opt = StringUtils::ParseInt<int64_t>(total_it->second);
```
wgtmac: The Java impl catches `NumberFormatException` when parsing integers from the map and do not update the total value. I'm not sure if we should be strict with it by returning `Status` and do not allow committing new snapshots with malformed summary.

src/iceberg/transaction.cc
```diff
@@ -113,6 +114,23 @@ Status Transaction::Apply(PendingUpdate& update) {
       metadata_builder_->SetCurrentSchema(std::move(result.schema),
                                           result.new_last_column_id);
     } break;
+    case PendingUpdate::Kind::kUpdateSnapshot: {
+      auto& update_snapshot = internal::checked_cast<SnapshotUpdate&>(update);
+      ICEBERG_ASSIGN_OR_RAISE(auto result, update_snapshot.Apply());
+      if (metadata_builder_->current()
```
wgtmac: Should we use a temp metadata builder to build from current and check if it is an empty update just like the java impl? If not empty, we can iterate `temp_metadata_builder.changes()` and call `change->ApplyTo(*metadata_builder_)` for each change.

wgtmac: BTW, the Java impl also called `updated.withUUID()` to make sure UUID is assigned when missing. It would be good to follow this for backward compatibility.

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+  // Cleanup after successful commit
+  if (cleanup_after_commit()) {
+    auto cached_snapshot = SnapshotCache(staged_snapshot_.get());
+    ICEBERG_ASSIGN_OR_RAISE(auto manifests,
+                            cached_snapshot.Manifests(transaction_->table()->io()));
+    std::unordered_set<std::string> manifest_paths;
+    for (const auto& manifest : manifests) {
+      manifest_paths.insert(manifest.manifest_path);
+    }
+    CleanUncommitted(manifest_paths);
```
wgtmac: Should we also add `protected boolean cleanupAfterCommit()` and check it here before cleaning up?

src/iceberg/update/snapshot_update.cc
```diff
@@ -0,0 +1,438 @@
...
+
+  // Initialize delete function if not set
+  if (!delete_func_) {
+    delete_func_ = [this](const std::string& path) {
+      return transaction_->table()->io()->DeleteFile(path);
+    };
+  }
+}
+
+Result<std::vector<ManifestFile>> SnapshotUpdate::WriteDataManifests(
```
wgtmac, line 168: `WriteDataManifests` and `WriteDeleteManifests` look so similar. Can we use a single function to accept different `ManifestContent` values?

wgtmac: BTW, `WriteDataManifests` should also accept an optional `data_sequence_number` just like the Java `List<ManifestFile> writeDataManifests(Collection<DataFile> files, Long dataSeq, PartitionSpec spec)`.

For `WriteDeleteManifests`, data_sequence_number from the file should be respected and `PendingDeleteFile` is applied. We need to add a TODO here to implement this when `MergingSnapshotProducer` has been added.

src/iceberg/update/snapshot_update.cc
zhjwpku: Yeah, I think this deserves a separate PR, so #493

src/iceberg/update/snapshot_update.cc
zhjwpku: Yeah, there is a cleanup_after_commit, let me change it to CleanupAfterCommit

src/iceberg/update/snapshot_update.cc
zhjwpku: Yeah, I was just targeting FastAppend, let's refactor this later in a separate PR.

src/iceberg/table_metadata.cc
zhjwpku: done

src/iceberg/update/snapshot_update.cc
zhjwpku: I add MakeWriter helpers to wrap that, see #493

src/iceberg/snapshot.h
zhjwpku: Agreed. Introduce MakeBranch and MakeTag, and remove the builder pattern.

src/iceberg/update/snapshot_update.h
zhjwpku: Add Finalize as a `PendingUpdate` method and default to just return success.

src/iceberg/update/snapshot_update.h
zhjwpku: removed the atomic.

src/iceberg/update/snapshot_update.h
zhjwpku: Reasonable, done.

src/iceberg/update/snapshot_update.cc
zhjwpku: add a base() method to `PendingUpdate`

src/iceberg/update/snapshot_update.cc
zhjwpku: I choose to add `SnapshotUtil::OptionalLatestSnapshot`, we can refactor if we prefer `TransferNotFoundSnapshot` later.

src/iceberg/update/snapshot_update.cc
zhjwpku: I add a `MetadataFileLocation` method to `Transaction` instead of `Table`, because we can always get the current metadata here.

src/iceberg/update/snapshot_update.cc
zhjwpku: Add a std::unordered_map<std::string, ManifestFile> as a cache, using ManifestFile::manifest_path as the key, it remains to be evaluated whether this key choice is sufficient, please help to review.

src/iceberg/update/snapshot_update.cc
zhjwpku: Yeah, let's be strict.

src/iceberg/update/snapshot_update.cc
zhjwpku: Done. btw. first_row_id and added_rows are currently wrapped in summaries as string. I am wondering whether we should promote them to direct fields of Snapshot for convenience, similar to Java’s BaseSnapshot.

src/iceberg/update/snapshot_update.cc
zhjwpku: Let's be strict for now.

src/iceberg/transaction.cc
zhjwpku: Yeah, fixed.

src/iceberg/table_metadata.cc
zhjwpku: I think it's pretty much the same as AddSnapshot + SetBranchSnapshot(int64_t snapshot_id, const std::string& branch). We don't have TableMetadataBuilder::SetBranchSnapshot(std::shared_ptr<Snapshot> snapshot, const std::string& branch), so I guess we are good here.

src/iceberg/util/string_util.h
```diff
@@ -54,6 +57,20 @@ class ICEBERG_EXPORT StringUtils {
...
+  template <typename T>
+  static Result<T> ParseInt(std::string_view str) {
+    T value = 0;
+    auto [ptr, ec] = std::from_chars(str.data(), str.data() + str.size(), value);
+    if (ec == std::errc::invalid_argument) [[unlikely]] {
+      return InvalidArgument("Failed to parse integer from string '{}': invalid argument",
+                             str);
+    } else if (ec == std::errc::result_out_of_range) [[unlikely]] {
+      return InvalidArgument(
+          "Failed to parse integer from string '{}': value out of range", str);
```
wgtmac: nit: use type_info to print the type name of T

src/iceberg/json_internal.cc
```diff
@@ -1553,9 +1553,17 @@ Result<std::unique_ptr<TableUpdate>> TableUpdateFromJson(const nlohmann::json& j
                             GetJsonValueOptional<int64_t>(json, kMaxSnapshotAgeMs));
     ICEBERG_ASSIGN_OR_RAISE(auto max_ref_age,
                             GetJsonValueOptional<int64_t>(json, kMaxRefAgeMs));
-    return std::make_unique<table::SetSnapshotRef>(std::move(ref_name), snapshot_id, type,
-                                                   min_snapshots, max_snapshot_age,
-                                                   max_ref_age);
+    if (type == SnapshotRefType::kTag) {
```
wgtmac: The original implementation looks better because it avoids an indirection of the temporary SnapshotRef instance.

src/iceberg/util/snapshot_util_internal.h
```diff
@@ -255,6 +256,27 @@ class ICEBERG_EXPORT SnapshotUtil {
...
 
+  /// \brief Fetch the snapshot at the head of the given branch in the given table.
+  ///
+  /// Like LatestSnapshot, but returns nullopt if the branch does not exist.
+  ///
+  /// \param metadata The table metadata
+  /// \param branch Branch name of the table metadata (empty string means main
+  /// branch)
+  /// \return The latest snapshot for the given branch, or nullopt if the branch does not
+  /// exist
```
wgtmac:
```suggestion
  /// \brief Fetch the snapshot at the head of the given branch in the given table.
  ///
  /// Like LatestSnapshot, but returns nullptr if the snapshot does not exist.
  ///
  /// \param metadata The table metadata
  /// \param branch Branch name of the table metadata (empty string means main
  /// branch)
  /// \return The latest snapshot for the given branch, or nullptr if the snapshot
  /// does not exist.
```

HeartLinked (approved):

wgtmac (approved):

src/iceberg/transaction.cc
```diff
@@ -69,6 +73,16 @@ const TableMetadata* Transaction::base() const { return metadata_builder_->base(
 
 const TableMetadata& Transaction::current() const { return metadata_builder_->current(); }
 
+std::string Transaction::MetadataFileLocation(std::string_view filename) const {
```
WZhuo, line 76: It's also used in `TableMetadataUtil::NewTableMetadataFilePath`, I think it can be a public static method of TableMetadataUtil. Maybe do it in another PR.

wgtmac:
Thanks @zhjwpku for working on this large and complex feature and others for the review! Let me merge this to unblock other conflicting PRs.

PR #412, opened 2025-12-15
feat: replace sort order for pending update

HeartLinked:


src/iceberg/meson.build
```diff
@@ -196,6 +197,7 @@ install_headers(
         'transform.h',
         'type_fwd.h',
         'type.h',
+        'update/replace_sort_order.h',
```
wgtmac: We need to add `iceberg/update/meson.build` to install update/*.h. Could you fix this?

src/iceberg/type_fwd.h
```diff
@@ -163,6 +163,7 @@ class PendingUpdate;
 template <typename T>
 class PendingUpdateTyped;
```
wgtmac:
```suggestion
```

Please help remove this.

src/iceberg/update/replace_sort_order.cc
```diff
@@ -0,0 +1,150 @@
...
+    : identifier_(std::move(identifier)),
+      catalog_(std::move(catalog)),
+      base_metadata_(std::move(base)) {}
+
+ReplaceSortOrder& ReplaceSortOrder::Asc(std::shared_ptr<Term> term,
+                                        NullOrder null_order) {
+  if (!term) {
+    return AddError(ErrorKind::kInvalidArgument, "Term cannot be null");
+  }
+  if (term->kind() != BoundTerm::Kind::kTransform) {
```
wgtmac:
```suggestion
  if (term->kind() != Term::Kind::kTransform) {
```

wgtmac: It is weird to use `BoundTerm` here.

wgtmac: BTW, it is totally valid to see `Term::Kind::kReference` here.

src/iceberg/update/replace_sort_order.cc
```diff
@@ -0,0 +1,150 @@
...
+  auto unbound_transform = internal::checked_pointer_cast<UnboundTransform>(term);
+
+  BUILDER_ASSIGN_OR_RETURN(auto schema, base_metadata_->Schema());
+  BUILDER_ASSIGN_OR_RETURN(auto bound_term,
+                           unbound_transform->Bind(*schema, case_sensitive_));
+  return AddSortField(bound_term->reference(), unbound_transform->transform(),
+                      SortDirection::kAscending, null_order);
+}
+
+ReplaceSortOrder& ReplaceSortOrder::Desc(std::shared_ptr<Term> term,
```
wgtmac: This looks so similar to the above `Asc` call. Can we just make `SortDirection` as a parameter to combine these two functions?

src/iceberg/update/replace_sort_order.cc
```diff
@@ -0,0 +1,150 @@
...
+}
+
+Status ReplaceSortOrder::Apply() {
+  ICEBERG_RETURN_UNEXPECTED(CheckErrors());
+  // Note: We use kInitialSortOrderId (1) here like the Java implementation.
+  // The actual sort order ID will be assigned by TableMetadataBuilder when
+  // the AddSortOrder update is applied.
+  ICEBERG_ASSIGN_OR_RAISE(auto order,
+                          SortOrder::Make(SortOrder::kInitialSortOrderId, sort_fields_));
+  ICEBERG_RETURN_UNEXPECTED(order->Validate(*base_metadata_->Schema().value()));
```
wgtmac: `base_metadata_->Schema()` has been called several times in this class but it is not cheap. Please consider caching and reusing the schema.

src/iceberg/update/replace_sort_order.cc
```diff
@@ -0,0 +1,150 @@
...
+
+ReplaceSortOrder& ReplaceSortOrder::AddSortField(std::shared_ptr<BoundReference> ref,
+                                                 std::shared_ptr<Transform> transform,
+                                                 SortDirection direction,
+                                                 NullOrder null_order) {
+  int32_t source_id = ref->field_id();
+  const auto& field = ref->field();
+
+  // Validate that the transform can be applied to the field type  、
+  if (!transform->CanTransform(*field.type())) {
```
wgtmac: We can skip this check because you have called `order->Validate` in the `Apply()`. They are doing the same thing.

wgtmac: If you remove this check, we don't need to pass `std::shared_ptr<BoundReference> ref` and just need an extra `int32_t source_id`.

PR #413, opened 2025-12-16
chore: disable timeout in meson test cases

zhjwpku:
The rest_integration_test can sometimes run longer that 30s, so disable meson test timeout[1] and rely on the GH actions job–level timeout.

[1] https://mesonbuild.com/Unit-tests.html#other-test-options


wgtmac (approved):
cc @WillAyd

wgtmac:
Thanks @zhjwpku for improving this and @WillAyd for the review!

PR #414, opened 2025-12-16
refactor: optimize Table accessors and TimePoint passing

wgtmac:
- Update `Table::location()` and `Table::metadata_file_location()` to return `std::string_view` instead of `const std::string&` to avoid unnecessary reference indirection.
- Change `TimePointMs` and `TimePointNs` arguments and return types to pass-by-value in `Table` and utility functions, as they are small types.
- Expose `Table::metadata()` accessor to allow access to the underlying table metadata.
- Clean up unused variable declarations in `avro_schema_util.cc`.

### DIFF
--- a/src/iceberg/avro/avro_schema_util.cc
+++ b/src/iceberg/avro/avro_schema_util.cc
@@ -734,7 +734,7 @@ Result<FieldProjection> ProjectMap(const MapType& map_type,
   const auto& expected_key_field = map_type.key();
   const auto& expected_value_field = map_type.value();
 
-  FieldProjection result, key_projection, value_projection;
+  FieldProjection result;
   int32_t avro_key_id, avro_value_id;
   ::avro::NodePtr map_node;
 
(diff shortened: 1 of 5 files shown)


Fokko (approved):

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -734,7 +734,7 @@ Result<FieldProjection> ProjectMap(const MapType& map_type,
   const auto& expected_key_field = map_type.key();
   const auto& expected_value_field = map_type.value();
 
-  FieldProjection result, key_projection, value_projection;
+  FieldProjection result;
```
HuaHuaY, line 737: How about adding `readability-isolate-declaration` in `.clang-tidy`? We can do this in next PR.

wgtmac: Sounds good.

PR #416, opened 2025-12-17
feat: InMemoryCatalog::CreateTable & StageCreateTable api

WZhuo:


src/iceberg/table_metadata.h
```diff
@@ -124,6 +124,12 @@ struct ICEBERG_EXPORT TableMetadata {
   /// A `long` higher than all assigned row IDs
   int64_t next_row_id;
 
+  static Result<std::unique_ptr<TableMetadata>> Make(
+      const iceberg::Schema& schema, const iceberg::PartitionSpec& spec,
```
wgtmac, line 131: Do you think we should rename functions below to `GetSchema`, `GetSnapshot`, etc. so we don't need to write `iceberg::` prefix here and elsewhere?

src/iceberg/table_identifier.h
```diff
@@ -53,6 +64,27 @@ struct ICEBERG_EXPORT TableIdentifier {
...
+
+  std::string ToString() const { return ns.ToString() + '.' + name; }
 };
 
 }  // namespace iceberg
+
+namespace std {
+
+template <>
+struct formatter<iceberg::Namespace> : std::formatter<std::string> {
```
wgtmac: You don't need to explicitly add these as they are automatically supported by including `iceberg/util/formatter.h`.

src/iceberg/table_identifier.h
```diff
@@ -35,6 +37,15 @@ struct ICEBERG_EXPORT Namespace {
   std::vector<std::string> levels;
 
   bool operator==(const Namespace& other) const { return levels == other.levels; }
+
+  std::string ToString() const {
+    std::ostringstream oss;
+    for (size_t i = 0; i < levels.size(); ++i) {
```
wgtmac: `formatter_internal.h` has a `FormatRange` to implement this. Using it requires adding a table_identifiler.cc since we cannot export internal header here.

src/iceberg/table_metadata.cc
```diff
@@ -65,6 +68,106 @@ std::string ToString(const MetadataLogEntry& entry) {
...
+    const std::unordered_map<std::string, std::string>& properties, int format_version) {
+  for (const auto& [key, _] : properties) {
+    if (TableProperties::reserved_properties().contains(key)) {
+      return InvalidArgument(
+          "Table properties should not contain reserved properties, but got {}", key);
+    }
+  }
+
+  // Reassign all column ids to ensure consistency
+  std::atomic<int32_t> last_column_id = 0;
```
wgtmac: Why do we need to use atomic here?

src/iceberg/table_metadata.cc
```diff
@@ -65,6 +68,106 @@ std::string ToString(const MetadataLogEntry& entry) {
...
+
+  // Validata the metrics configuration.
+  ICEBERG_RETURN_UNEXPECTED(
+      MetricsConfig::VerifyReferencedColumns(properties, *fresh_schema));
+
+  // TODO(anyone) Validate commit properties
+
+  return TableMetadataBuilder::BuildFromEmpty(format_version)
+      ->SetLocation(location)
+      .AddSchema(std::move(fresh_schema))
```
wgtmac: `AddXXX` does not set it as the default. We may need to use `SetCurrentSchema`, `SetDefaultPartitionSpec` and `SetDefaultSortOrder` here.

src/iceberg/table_metadata.cc
```diff
@@ -408,6 +518,10 @@ class TableMetadataBuilder::Impl {
   const TableMetadata* base() const { return base_; }
   const TableMetadata& metadata() const { return metadata_; }
 
+  void SetLocation(std::string_view location) {
+    metadata_.location = std::string(location);
```
wgtmac: We might need to follow the java impl to validate the input and then produce a update to the changes list.

```java
We might need to follow the Java impl to validate the input location and then produce a TableUpdate to the changes list.

```java
    public Builder setLocation(String newLocation) {
      if (location != null && location.equals(newLocation)) {
        return this;
      }

      this.location = newLocation;
      changes.add(new MetadataUpdate.SetLocation(newLocation));

      return this;
    }
```
```

File: src/iceberg/table_metadata.cc [hunk 7]
```diff
@@ -50,12 +52,73 @@
...
+#include "iceberg/util/type_util.h"
 #include "iceberg/util/uuid.h"
 namespace iceberg {
 namespace {
 const TimePointMs kInvalidLastUpdatedMs = TimePointMs::min();
 constexpr int32_t kLastAdded = -1;
 constexpr std::string_view kMetadataFolderName = "metadata";
+
+// TableMetadata private static methods
+Result<std::shared_ptr<PartitionSpec>> FreshPartitionSpec(
```
wgtmac: I think `FreshPartitionSpec` and `FreshSortOrder` should also be called in the `SetDefaultPartitionSpec` and `SetDefaultSortOrder`, respectively.

src/iceberg/table_requirements.cc
```diff
@@ -134,4 +136,24 @@ Result<std::vector<std::unique_ptr<TableRequirement>>> TableRequirements::ForUpd
   return context.Build();
 }
 
+Result<bool> TableRequirements::IsCreate(
+    const std::vector<std::unique_ptr<TableRequirement>>& requirements) {
+  bool is_create = std::ranges::any_of(requirements, [](const auto& req) {
+    return dynamic_cast<table::AssertDoesNotExist*>(req.get()) != nullptr;
```
wgtmac: Please add an enum type to `TableRequirement` like `TableUpdate::kind()` and use it here.

src/iceberg/table_metadata.cc
WZhuo: Unlike Java impl, I think no need to fresh field id in the TableMetadataBuilder::AddXXX.

src/iceberg/table_metadata.cc
WZhuo: Yeah, It's convert from java, I think it's not really needed, fix it.

src/iceberg/table_metadata.h
WZhuo: 1. Rename the function name like `GetSchema`.
2. use `class` keyword replace of the namespace specified, like `const class Schema& schema, const class PartitionSpec& spec`.
There's no need to write the namespace `iceberg` prefix other than TableMetadata's class scope. So I don't think rename function is well, we'll write `GetSchema` everywhere. I prefer solution 2.

src/iceberg/test/formatter_test.cc
```diff
@@ -160,4 +161,19 @@ TEST(FormatterTest, StatisticsFileFormat) {
   EXPECT_EQ(expected, std::format("{}", statistics_file));
 }
 
+// For Types that has a ToString function
+TEST(FormatterTest, TableIdentifierFormat) {
+  TableIdentifier empty_ns_table{
+      .ns = Namespace({}),
+      .name = "table_name",
+  };
+  EXPECT_EQ(".table_name", std::format("{}", empty_ns_table));
```
wgtmac: This looks weird. Shouldn't it be `"table_name"`?

src/iceberg/table_identifier.cc
```diff
@@ -0,0 +1,32 @@
...
+#include "iceberg/table_identifier.h"
+
+#include "iceberg/util/formatter_internal.h"
+
+namespace iceberg {
+
+std::string Namespace::ToString() const { return FormatRange(levels, ".", "", ""); }
+
+std::string TableIdentifier::ToString() const {
+  return std::format("{}.{}", ns.ToString(), name);
```
wgtmac: We should directly print the name if ns is empty.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -405,32 +407,68 @@ Result<std::shared_ptr<Table>> InMemoryCatalog::CreateTable(
...
+  ICEBERG_ASSIGN_OR_RAISE(auto is_create, TableRequirements::IsCreate(requirements));
+  if (is_create) {
+    if (base_metadata_location.has_value()) {
+      return AlreadyExists("Table already exists: {}", identifier);
+    }
+    int8_t format_version = TableMetadata::kDefaultTableFormatVersion;
+    for (const auto& update : updates) {
+      if (update->kind() == TableUpdate::Kind::kUpgradeFormatVersion) {
+        format_version =
+            dynamic_cast<const table::UpgradeFormatVersion&>(*update).format_version();
```
wgtmac:
```suggestion
            internal::checked_cast<const table::UpgradeFormatVersion&>(*update).format_version();
```

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -405,32 +407,68 @@ Result<std::shared_ptr<Table>> InMemoryCatalog::CreateTable(
...
+                                                                   location, properties));
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto metadata_file_location,
+      TableMetadataUtil::Write(*file_io_, nullptr, "", *table_metadata));
+  ICEBERG_RETURN_UNEXPECTED(
+      root_namespace_->UpdateTableMetadataLocation(identifier, metadata_file_location));
+  return Table::Make(identifier, std::move(table_metadata),
+                     std::move(metadata_file_location), file_io_,
+                     std::static_pointer_cast<Catalog>(shared_from_this()));
```
wgtmac:
```suggestion
                     internal::checked_pointer_cast<Catalog>(shared_from_this()));
```

src/iceberg/table_metadata.h
```diff
@@ -210,8 +219,13 @@ class ICEBERG_EXPORT TableMetadataBuilder : public ErrorCollector {
   /// \brief Create a builder from existing table metadata
   ///
   /// \param base The base table metadata to build from
-  /// \return A new TableMetadataBuilder instance initialized with base metadata
-  static std::unique_ptr<TableMetadataBuilder> BuildFrom(const TableMetadata* base);
+  /// \param is_create Whether the builder is for creating a new table. It will call
```
wgtmac: Can we revert this change to keep the behavior that `BuildFromEmpty` is for create and `BuildFrom` is for update?

To achieve this, we can refactor `ChangesForCreate` function above to make it a static utility function and rename it to `ApplyCreateChanges` like below to figure out required changes from a `TableMetadata` and directly apply changes to a `TableMetadataBuilder`:

```cpp
Status TableMetadataUtil::ApplyCreateChanges(const TableMetadata& metadata, TableMetadataBuilder& builder);
```

In this way, we can still call `BuildFrom` to create a builder for create mode and then apply changes to the current metadata. This function can also be reused for the rest catalog implementation.

wgtmac: If we use this approach, perhaps `ChangesForCreate()` is no longer needed as well.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -445,7 +483,20 @@ Result<std::shared_ptr<Transaction>> InMemoryCatalog::StageCreateTable(
...
+  std::string base_location =
+      location.empty() ? warehouse_location_ + "/" + identifier.ToString() : location;
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto table_metadata,
+      TableMetadata::Make(*schema, *spec, *order, base_location, properties));
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto table, StagedTable::Make(identifier, std::move(table_metadata), "", file_io_,
+                                    shared_from_this()));
+  return Transaction::Make(std::move(table), Transaction::Kind::kCreate, false);
```
wgtmac:
```suggestion
  return Transaction::Make(std::move(table), Transaction::Kind::kCreate, /*auto_commit=*/false);
```

Let's enhance the readability.

src/iceberg/table_metadata.h
wgtmac: BTW, I didn't see any place that uses `is_create=true`. Did I miss something?

wgtmac (changes requested):

src/iceberg/table_metadata.h
WZhuo: It's used in the Transaction constructor.
https://github.com/apache/iceberg-cpp/pull/416/changes#diff-669b44489f32d2b272a6654f1eb484d9626ca9a1884f7829244fb34b8ade96c0L43

wgtmac: Then can we do something like this?

```cpp
Transaction::Transaction(std::shared_ptr<Table> table, Kind kind, bool auto_commit,
                         std::unique_ptr<TableMetadataBuilder> metadata_builder)
    : table_(std::move(table)),
      kind_(kind),
      auto_commit_(auto_commit),
      metadata_builder_(std::move(metadata_builder)) {}

Result<std::shared_ptr<Transaction>> Transaction::Make(std::shared_ptr<Table> table,
                                                       Kind kind, bool auto_commit) {
  if (!table || !table->catalog()) [[unlikely]] {
    return InvalidArgument("Table and catalog cannot be null");
  }

  const auto* metadata = table->metadata().get();
  std::unique_ptr<TableMetadataBuilder> metadata_builder;
  switch (kind) {
    case Kind::kCreate: {
      metadata_builder = TableMetadataBuilder::BuildFromEmpty();
      ICEBERG_RETURN_UNEXPECTED(
          TableMetadataUtil::ApplyCreateChanges(metadata, *metadata_builder));
    } break;
    case Kind::kUpdate: {
      metadata_builder = TableMetadataBuilder::BuildFrom(metadata);
    } break;
    default:
      return NotSupported("Unsupported transaction kind: {}", static_cast<int>(kind));
  }

  return std::shared_ptr<Transaction>(
      new Transaction(std::move(table), kind, auto_commit, std::move(metadata_builder)));
}
```

src/iceberg/catalog/memory/in_memory_catalog.cc
WZhuo: Copy from old code, no need cast any more.

src/iceberg/table_metadata.h
WZhuo: Yes, that's indeed how it should be.

src/iceberg/table_identifier.cc
WZhuo: You're right. Align with java, return `name` if `ns` is empty

src/iceberg/table_metadata.cc
```diff
@@ -68,6 +130,100 @@ std::string ToString(const MetadataLogEntry& entry) {
...
+  return TableMetadataBuilder::BuildFromEmpty(format_version)
+      ->SetLocation(location)
+      .SetCurrentSchema(std::move(fresh_schema), last_column_id)
+      .SetDefaultPartitionSpec(std::move(fresh_spec))
+      .SetDefaultSortOrder(std::move(fresh_order))
+      .SetProperties(properties)
+      .Build();
+}
+
+std::vector<std::unique_ptr<TableUpdate>> TableMetadata::ChangesForCreate() const {
```
wgtmac: I still prefer this function to be a static utility function instead of a member function. The reason is to keep minimal functions in the `TableMetadata` (only necessary convenience accessors) and this function is not supposed to be used by downstream users.

WZhuo: Ok, it make sense, move `ChangesForCreate` to anonymous namespace in table_metadata.cc, as a private static function

PR #418, opened 2025-12-17
feat: implement transaction api

wgtmac:
- Introduce `Transaction` class to manage multi-operation table updates.
- Add `Table::NewTransaction()` and `Table::NewUpdateProperties()` to initiate updates.
- Move `PendingUpdate` and `UpdateProperties` to `src/iceberg/update/` and refactor them to use the transaction mechanism.
- Add `StagedTable` to represent tables with uncommitted changes.
- Update `InMemoryCatalog` and `RestCatalog` to align with the new update flow.
- Refactor `Table` to support metadata refresh and location management within transactions.
- Add comprehensive tests for transactions and property updates.

### DIFF
--- a/src/iceberg/table_metadata.cc
+++ b/src/iceberg/table_metadata.cc
@@ -796,6 +796,6 @@ const std::vector<std::unique_ptr<TableUpdate>>& TableMetadataBuilder::changes()
 
 const TableMetadata* TableMetadataBuilder::base() const { return impl_->base; }
 
-const TableMetadata* TableMetadataBuilder::current() const { return &impl_->metadata; }
+const TableMetadata& TableMetadataBuilder::current() const { return impl_->metadata; }
 
 }  // namespace iceberg
--- a/src/iceberg/transaction.cc
+++ b/src/iceberg/transaction.cc
@@ -49,7 +49,7 @@ Result<std::shared_ptr<Transaction>> Transaction::Make(std::shared_ptr<Table> ta
 
 const TableMetadata* Transaction::base() const { return metadata_builder_->base(); }
 
-const TableMetadata* Transaction::current() const { return metadata_builder_->current(); }
+const TableMetadata& Transaction::current() const { return metadata_builder_->current(); }
 
 Status Transaction::AddUpdate(const std::shared_ptr<PendingUpdate>& update) {
   if (!last_update_committed_) {
--- a/src/iceberg/transaction.h
+++ b/src/iceberg/transaction.h
@@ -47,7 +47,7 @@ class ICEBERG_EXPORT Transaction : public std::enable_shared_from_this<Transacti
   const TableMetadata* base() const;
 
   /// \brief Return the current metadata with staged changes applied
-  const TableMetadata* current() const;
+  const TableMetadata& current() const;
 
   /// \brief Apply the pending changes from all actions and commit.
   ///
--- a/src/iceberg/update/update_properties.cc
+++ b/src/iceberg/update/update_properties.cc
@@ -78,11 +78,6 @@ UpdateProperties& UpdateProperties::Remove(const std::string& key) {
 Result<PendingUpdate::ApplyResult> UpdateProperties::Apply() {
   ICEBERG_RETURN_UNEXPECTED(CheckErrors());
 
-  const auto* base_metadata = transaction_->current();
-  if (!base_metadata) {
-    return InvalidArgument("Base table metadata is required to apply property updates");
-  }
-
   auto iter = updates_.find(TableProperties::kFormatVersion.key());
   if (iter != updates_.end()) {
     int parsed_version = 0;
@@ -105,7 +100,7 @@ Result<PendingUpdate::ApplyResult> UpdateProperties::Apply() {
     updates_.erase(iter);
   }
 
-  if (auto schema = base_metadata->Schema(); schema.has_value()) {
+  if (auto schema = transaction_->current().Schema(); schema.has_value()) {
     ICEBERG_RETURN_UNEXPECTED(
         MetricsConfig::VerifyReferencedColumns(updates_, *schema.value()));
   }
--- a/src/iceberg/table_metadata.h
+++ b/src/iceberg/table_metadata.h
@@ -426,7 +426,7 @@ class ICEBERG_EXPORT TableMetadataBuilder : public ErrorCollector {
   const TableMetadata* base() const;
 
   /// \brief Returns the current metadata with staged changes applied
-  const TableMetadata* current() const;
+  const TableMetadata& current() const;
 
   /// \brief Destructor
   ~TableMetadataBuilder() override;


src/iceberg/update/update_properties.cc
```diff
@@ -70,65 +75,56 @@ UpdateProperties& UpdateProperties::Remove(const std::string& key) {
...
 
-Status UpdateProperties::Apply() {
-  if (!catalog_) {
-    return InvalidArgument("Catalog is required to apply property updates");
-  }
-  if (!base_metadata_) {
+Result<PendingUpdate::ApplyResult> UpdateProperties::Apply() {
+  ICEBERG_RETURN_UNEXPECTED(CheckErrors());
+
+  const auto* base_metadata = transaction_->base();
```
lishuxu: In a transaction, earlier updates may have updated table_meta in transaction, I think we should user transaction_->current() as base_metadata here

src/iceberg/update/update_properties.cc
```diff
@@ -70,65 +75,56 @@ UpdateProperties& UpdateProperties::Remove(const std::string& key) {
...
-
-Status UpdateProperties::Commit() {
-  ICEBERG_RETURN_UNEXPECTED(Apply());
 
-  std::vector<std::unique_ptr<TableUpdate>> updates;
+  ApplyResult result;
   if (!updates_.empty()) {
-    updates.emplace_back(std::make_unique<table::SetProperties>(std::move(updates_)));
+    result.updates.emplace_back(
+        std::make_unique<table::SetProperties>(std::move(updates_)));
```
lishuxu: Should we keep updates_/removals_ intact when producing ApplyResult?

src/iceberg/transaction.cc
```diff
@@ -0,0 +1,111 @@
...
+    } break;
+    case Kind::kUpdate: {
+      ICEBERG_ASSIGN_OR_RAISE(requirements, TableRequirements::ForUpdateTable(
+                                                *metadata_builder_->base(), updates));
+
+    } break;
+  }
+
+  // XXX: we should handle commit failure and retry here.
+  return table_->catalog()->UpdateTable(table_->name(), requirements, updates);
```
WZhuo, line 101: CreateTable or UpdateTable according to `kind_`

src/iceberg/transaction.cc
```diff
@@ -0,0 +1,111 @@
...
+Result<std::shared_ptr<Table>> Transaction::Commit() {
+  if (!last_update_committed_) {
+    return InvalidArgument(
+        "Cannot commit transaction when previous update is not committed");
+  }
+
+  const auto& updates = metadata_builder_->changes();
+  if (updates.empty()) {
+    return table_;
+  }
```
WZhuo, line 86: `updates` can be empty if it's a StagedTable

src/iceberg/transaction.cc
```diff
@@ -0,0 +1,111 @@
...
+
+Transaction::~Transaction() = default;
+
+Result<std::shared_ptr<Transaction>> Transaction::Make(std::shared_ptr<Table> table,
+                                                       Kind kind, bool auto_commit) {
+  if (!table || !table->catalog()) [[unlikely]] {
+    return InvalidArgument("Table and catalog cannot be null");
+  }
+  return std::shared_ptr<Transaction>(
+      new Transaction(std::move(table), kind, auto_commit));
```
HuaHuaY:
```suggestion
  return std::make_shared<Transaction>(std::move(table), kind, auto_commit);
```

src/iceberg/table.cc
```diff
@@ -110,18 +130,78 @@ const std::vector<SnapshotLogEntry>& Table::history() const {
...
 
-std::unique_ptr<UpdateProperties> Table::UpdateProperties() const {
-  return std::make_unique<iceberg::UpdateProperties>(identifier_, catalog_, metadata_);
+const std::shared_ptr<FileIO>& Table::io() const { return io_; }
+
+const std::shared_ptr<TableMetadata>& Table::metadata() const { return metadata_; }
+
+const std::shared_ptr<Catalog>& Table::catalog() const { return catalog_; }
+
+Result<std::unique_ptr<TableScanBuilder>> Table::NewScan() const {
```
zhjwpku, line 139: Why use Result to wrap this? It seems won't return any error.

src/iceberg/table_metadata.cc
```diff
@@ -790,4 +790,12 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataBuilder::Build() {
   return std::make_unique<TableMetadata>(std::move(impl_->metadata));
 }
 
+const std::vector<std::unique_ptr<TableUpdate>>& TableMetadataBuilder::changes() const {
+  return impl_->changes;
+}
+
+const TableMetadata* TableMetadataBuilder::base() const { return impl_->base; }
+
+const TableMetadata* TableMetadataBuilder::current() const { return &impl_->metadata; }
```
HuaHuaY: Return reference instead of pointer so caller doesn't need to think about whether result will be nullptr. `Transaction::current() const` also need to be modified.

src/iceberg/transaction.h
```diff
@@ -29,28 +30,61 @@
...
-  ///
-  /// \return this transaction's table
-  virtual const std::shared_ptr<Table>& table() const = 0;
+  const std::shared_ptr<Table>& table() const { return table_; }
 
-  /// \brief Create a new append API to add files to this table
-  ///
-  /// \return a new AppendFiles
-  virtual std::shared_ptr<AppendFiles> NewAppend() = 0;
+  /// \brief Return the base metadata without any changes
```
zhjwpku: nit: %s/Return/Returns/g

src/iceberg/table.cc
```diff
@@ -21,16 +21,40 @@
...
 #include "iceberg/util/macros.h"
 
 namespace iceberg {
 
+Result<std::shared_ptr<Table>> Table::Make(TableIdentifier identifier,
+                                           std::shared_ptr<TableMetadata> metadata,
+                                           std::string metadata_location,
+                                           std::shared_ptr<FileIO> io,
+                                           std::shared_ptr<Catalog> catalog) {
+  if (metadata == nullptr) [[unlikely]] {
```
WZhuo, line 41: It's better to define a macro for checking this parameter as well as the following.

src/iceberg/update/update_properties.h
```diff
@@ -20,28 +20,24 @@
 #pragma once
 
 #include <memory>
+#include <optional>
```
zhjwpku, line 23:
```suggestion
```
not used?

src/iceberg/table.h
```diff
@@ -143,4 +145,39 @@ class ICEBERG_EXPORT Table {
   std::unique_ptr<class TableMetadataCache> metadata_cache_;
 };
 
+class ICEBERG_EXPORT StagedTable : public Table {
```
HuaHuaY: add `final`?

src/iceberg/update/update_properties.cc
```diff
@@ -70,65 +75,56 @@ UpdateProperties& UpdateProperties::Remove(const std::string& key) {
...
     }
+    format_version_ = static_cast<int8_t>(parsed_version);
 
     updates_.erase(iter);
   }
 
-  if (auto schema = base_metadata_->Schema(); schema.has_value()) {
+  if (auto schema = base_metadata->Schema(); schema.has_value()) {
     ICEBERG_RETURN_UNEXPECTED(
         MetricsConfig::VerifyReferencedColumns(updates_, *schema.value()));
```
HeartLinked, line 105: As discussed offline, I think there are some errors since #363. I'll open another PR to modify this like #415 after this  PR merged.

src/iceberg/table.cc
```diff
@@ -110,18 +130,78 @@ const std::vector<SnapshotLogEntry>& Table::history() const {
...
+    std::string metadata_location, std::shared_ptr<FileIO> io) {
+  if (metadata == nullptr) [[unlikely]] {
+    return InvalidArgument("Metadata cannot be null");
+  }
+  if (io == nullptr) [[unlikely]] {
+    return InvalidArgument("FileIO cannot be null");
+  }
+  return std::shared_ptr<StaticTable>(
+      new StaticTable(std::move(identifier), std::move(metadata),
+                      std::move(metadata_location), std::move(io), /*catalog=*/nullptr));
```
HuaHuaY:
```suggestion
  return std::make_shared<StaticTable>(std::move(identifier), std::move(metadata),
                                       std::move(metadata_location), std::move(io),
                                       /*catalog=*/nullptr);
```

src/iceberg/update/update_properties.cc
```diff
@@ -19,28 +19,33 @@
...
-                                   std::shared_ptr<TableMetadata> base)
-    : identifier_(std::move(identifier)),
-      catalog_(std::move(catalog)),
-      base_metadata_(std::move(base)) {}
+Result<std::shared_ptr<UpdateProperties>> UpdateProperties::Make(
+    std::shared_ptr<Transaction> transaction) {
+  if (!transaction) [[unlikely]] {
+    return InvalidArgument("Cannot create UpdateProperties without a transaction");
+  }
+  return std::shared_ptr<UpdateProperties>(new UpdateProperties(std::move(transaction)));
```
HuaHuaY: I suggest to add some comments if we prefer to use `std::shared_ptr` rather than `std::make_shared` here.

src/iceberg/table.cc
wgtmac: IMO, we should not call this on a `StagedTable`.

src/iceberg/update/update_properties.cc
wgtmac: Good catch!

src/iceberg/transaction.cc
wgtmac: This is by design: https://github.com/apache/iceberg/blob/9ca8029b4f7932ca2c22b034a0abc0edf8972975/open-api/rest-catalog-open-api.yaml#L566-L571.

In short, stage create is a two-pass process that `StageCreateTable` is called first to get its initial metadata and then `UpdateTable` should be called to finish the creation together with other metadata updates in a batch.

src/iceberg/update/update_properties.h
wgtmac: It is used by `std::optional<int8_t> format_version_;` and cannot be removed to avoid windows ci failure.

src/iceberg/table_metadata.cc
wgtmac: Actually `base` can be nullptr, so `current` uses same pattern to be consistent.

HuaHuaY: Yeah, I know that `base()` can be nullptr. So I prefer to let `current()` return a reference to distinguish it from `base()`.

src/iceberg/table.cc
wgtmac: How about a dedicated PR for this? I think there are other places using the same pattern.

src/iceberg/transaction.cc
wgtmac: I'm not sure if this can happen. We can explore this later.

wgtmac: Now I have a clearer picture of it. For the rest catalog, committing the staged table requires adding all initial updates to it: https://github.com/apache/iceberg/blob/05998edb8808ae21739072cca17b621925a2081a/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L890. We can implement this later before committing the create transaction.

lishuxu (approved):

HeartLinked (approved):

WZhuo (approved):

liurenjie1024 (approved):

PR #419, opened 2025-12-17
chore: add readability-isolate-declaration to .clang-tidy

HuaHuaY:
Add this rule to avoid defining multiple variables on the same line.

### DIFF
--- a/src/iceberg/test/mock_catalog.h
+++ b/src/iceberg/test/mock_catalog.h
@@ -23,7 +23,6 @@
 #include <gtest/gtest.h>
 
 #include "iceberg/catalog.h"
-#include "iceberg/table.h"
 
 namespace iceberg {
 


src/iceberg/test/mock_catalog.h
```diff
@@ -23,6 +23,7 @@
 #include <gtest/gtest.h>
 
 #include "iceberg/catalog.h"
+#include "iceberg/table.h"
```
HuaHuaY: Fix a bug. Otherwise clang-tidy checks would fail when someone else modified this file. 
The reasion is `MOCK_METHOD((Result<std::unique_ptr<Table>>), ...` will instantiate `std::expected<std::unique_ptr<Table>>` but `Table` here is just a forward declaration so `std::unique_ptr` will fail to generate destructor.

zhjwpku: Yeah, I remember in the test file if include mock_catalog.h without including table.h, it will fail to compile.

HuaHuaY: If we change `std::unique_ptr<Table>` to `std::shared_ptr<Table>` like https://github.com/apache/iceberg-cpp/pull/418, we don't need to get the complete definition.

PR #420, opened 2025-12-17
feat: add snapshot util

zhjwpku:


src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+
+namespace iceberg {
+
+namespace {}  // namespace
+
+Result<std::vector<std::shared_ptr<Snapshot>>> SnapshotUtil::AncestorsOf(
+    const Table& table, int64_t snapshot_id) {
+  ICEBERG_ASSIGN_OR_RAISE(auto start, table.SnapshotById(snapshot_id));
+  if (!start) {
+    return InvalidArgument("Cannot find snapshot: {}", snapshot_id);
```
wgtmac:
```suggestion
    return NotFound("Cannot find snapshot: {}", snapshot_id);
```

wgtmac: BTW, perhaps this should be a ICEBERG_DCHECK.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  std::shared_ptr<Snapshot> current = snapshot;
+  while (current) {
+    result.push_back(current);
+    if (!current->parent_snapshot_id.has_value()) {
+      break;
+    }
+    auto parent_result = table.SnapshotById(current->parent_snapshot_id.value());
+    if (!parent_result.has_value()) {
```
wgtmac: We might need to check if the error is `NotFound` and return the original error for other kind of errors.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+      // Parent snapshot not found (e.g., expired), stop traversal
+      break;
+    }
+    current = parent_result.value();
```
wgtmac:
```suggestion
    current = std::move(parent_result.value());
```

src/iceberg/util/snapshot_util_internal.h
```diff
@@ -0,0 +1,274 @@
...
+  /// \param table The table
+  /// \return A vector of snapshot IDs of the known ancestor snapshots, including the
+  /// current ID
+  static Result<std::vector<int64_t>> CurrentAncestorIds(const Table& table);
+
+  /// \brief Traverses the history of the table's current snapshot and finds the oldest
+  /// Snapshot.
+  ///
+  /// \param table The table
+  /// \return The oldest snapshot, or nullopt if there is no current snapshot
```
wgtmac, line 95: It would be good to document that all returned `std::shared_ptr<Snapshot>` cannot be nullptr.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto ancestors, AncestorsOf(table, snapshot_id));
+  if (ancestors.empty()) {
+    return std::nullopt;
+  }
+  return ancestors.back();
+}
+
+Result<std::optional<std::shared_ptr<Snapshot>>> SnapshotUtil::OldestAncestorAfter(
+    const Table& table, TimePointMs timestamp_ms) {
+  ICEBERG_ASSIGN_OR_RAISE(auto current, table.current_snapshot());
```
wgtmac: Actually it is not an error if current snapshot is not found. We can return std::nullopt when the result error is `NotFound`.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+  if (!current) {
```
wgtmac: `table.current_snapshot()` should not return a nullptr, so this should be an error instead.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+  if (!start) {
+    return InvalidArgument("Cannot find snapshot: {}", snapshot_id);
+  }
+  return AncestorsOf(table, start);
+}
+
+Result<bool> SnapshotUtil::IsAncestorOf(const Table& table, int64_t snapshot_id,
+                                        int64_t ancestor_snapshot_id) {
+  ICEBERG_ASSIGN_OR_RAISE(auto ancestors, AncestorsOf(table, snapshot_id));
+  for (const auto& snapshot : ancestors) {
```
wgtmac: nit: use std::ranges for this kind of processing.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+
+Result<std::optional<std::shared_ptr<Snapshot>>> SnapshotUtil::OldestAncestorAfter(
+    const Table& table, TimePointMs timestamp_ms) {
+  ICEBERG_ASSIGN_OR_RAISE(auto current, table.current_snapshot());
+  if (!current) {
+    // there are no snapshots or ancestors
+    return std::nullopt;
+  }
+
+  std::shared_ptr<Snapshot> last_snapshot = nullptr;
```
wgtmac: What about using `std::optional<std::shared_ptr<Snapshot>>` so we don't need to check at line 111.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+    }
+    last_snapshot = snapshot;
+  }
+
+  if (last_snapshot && !last_snapshot->parent_snapshot_id.has_value()) {
+    // this is the first snapshot in the table, return it
+    return last_snapshot;
+  }
+
+  return ValidationFailed("Cannot find snapshot older than {}",
```
wgtmac:
```suggestion
  return NotFound("Cannot find snapshot older than {}",
```

wgtmac: I'm undecided which is better, ValidationFailed, NotFound or Invalid.

src/iceberg/util/snapshot_util.cc
```diff
@@ -0,0 +1,373 @@
...
+                          FormatTimestamp(timestamp_ms));
+}
+
+Result<std::vector<int64_t>> SnapshotUtil::SnapshotIdsBetween(const Table& table,
+                                                              int64_t from_snapshot_id,
+                                                              int64_t to_snapshot_id) {
+  ICEBERG_ASSIGN_OR_RAISE(auto to_snapshot, table.SnapshotById(to_snapshot_id));
+  if (!to_snapshot) {
+    return InvalidArgument("Cannot find snapshot: {}", to_snapshot_id);
```
wgtmac: `InvalidArgument` should be used for input argument. This is an invalid state so perhaps return `Invalid`?

wgtmac (commented):
I haven't finished my review yet. Just post my findings so far. Will review it later.

PR #426, opened 2025-12-21
fix(transaction): move flag assignment after auto-commit

shangxinli:
Move last_update_committed_ assignment to after auto-commit attempt.

If Commit() fails, ICEBERG_RETURN_UNEXPECTED returns early and the flag remains false, preventing the transaction from incorrectly reporting the update as committed when it actually failed.

### DIFF
--- a/src/iceberg/transaction.cc
+++ b/src/iceberg/transaction.cc
@@ -70,8 +70,6 @@ Status Transaction::Apply(std::vector<std::unique_ptr<TableUpdate>> updates) {
   if (auto_commit_) {
     auto result = Commit();
     if (!result.has_value()) {
-      // Commit failed, revert the flag
-      last_update_committed_ = false;
       return std::unexpected(result.error());
     }
   }


src/iceberg/transaction.cc
```diff
@@ -68,7 +68,12 @@ Status Transaction::Apply(std::vector<std::unique_ptr<TableUpdate>> updates) {
   last_update_committed_ = true;
```
wgtmac, line 68: Perhaps it is simpler to move line 68 to be after the `if (auto_commit_)` block?

shangxinli: make sense! it makes simpler.

shangxinli: It seems the test failed after we move this line. The problem is  Commit() requires this flag to be true at the start. So we need to set it to be true before that.

wgtmac: I think the name `last_update_committed_` might confuse readers a little bit because it only means whether previous created pending update has called commit to apply changes to the transaction. We use this flag to make sure updates are created and committed sequentially to make the state of transaction simpler. Since it has nothing to do with the transaction object itself, I'm hesitant to set `last_update_committed_` to false when transaction.Commit fails. Perhaps we should rename it to `last_update_applied` or something more precise. WDYT?

shangxinli: Yeah, it seem in java the flag hasLastOpCommitted (the equivalent flag) tracks whether PendingUpdate.commit() was called, not whether Transaction.commitTransaction()
  succeeded.

PR #427, opened 2025-12-22
fix: prevent double-commit in Transaction

shangxinli:
Prevents Transaction::Commit() from being called multiple times by:
- Adding committed_ flag to track transaction state
- Checking flag at start of Commit() and returning error if already committed
- Setting flag after successful commit (both empty and non-empty updates)
- Updating table_ reference after successful catalog update

This ensures transactions can only be committed once, preventing unintended side effects and maintaining transaction semantics.

### DIFF
--- a/src/iceberg/transaction.cc
+++ b/src/iceberg/transaction.cc
@@ -76,7 +76,7 @@ Status Transaction::Apply(std::vector<std::unique_ptr<TableUpdate>> updates) {
 
 Result<std::shared_ptr<Table>> Transaction::Commit() {
   if (committed_) {
-    return InvalidArgument("Transaction already committed");
+    return Invalid("Transaction already committed");
   }
   if (!last_update_committed_) {
     return InvalidArgument(


src/iceberg/transaction.cc
```diff
@@ -98,7 +102,14 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
   }
 
   // XXX: we should handle commit failure and retry here.
-  return table_->catalog()->UpdateTable(table_->name(), requirements, updates);
+  ICEBERG_ASSIGN_OR_RAISE(auto updated_table, table_->catalog()->UpdateTable(
+                                                  table_->name(), requirements, updates));
+
+  // Mark as committed and update table reference
+  committed_ = true;
+  table_ = updated_table;
```
wgtmac:
```suggestion
  table_ = std::move(updated_table);
```

src/iceberg/transaction.cc
```diff
@@ -98,7 +102,14 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
+
+  return updated_table;
```
wgtmac:
```suggestion
  return table_;
```

src/iceberg/transaction.cc
```diff
@@ -75,13 +75,17 @@ Status Transaction::Apply(std::vector<std::unique_ptr<TableUpdate>> updates) {
 }
 
 Result<std::shared_ptr<Table>> Transaction::Commit() {
+  if (committed_) {
+    return InvalidArgument("Transaction already committed");
```
wgtmac:
```suggestion
    return Invalid("Transaction already committed");
```

This is no argument so we can simply use `Invalid`.

wgtmac (approved):

PR #428, opened 2025-12-22
feat: manifest evaluator support build for row filter

dongxiao1198:
feat: manifest evaluator support build for row filter


src/iceberg/expression/manifest_evaluator.cc
```diff
@@ -355,8 +356,9 @@ Result<std::unique_ptr<ManifestEvaluator>> ManifestEvaluator::MakeRowFilter(
     [[maybe_unused]] std::shared_ptr<Expression> expr,
```
wgtmac: Remove these `[[maybe_unused]]`?

PR #429, opened 2025-12-22
fix: correct errors in update properties implementation

HeartLinked:


src/iceberg/test/update_properties_test.cc
```diff
@@ -28,11 +27,6 @@ namespace iceberg {
 
 class UpdatePropertiesTest : public UpdateTestBase {};
 
-TEST_F(UpdatePropertiesTest, EmptyUpdates) {
```
wgtmac, line 31: Why this gets removed? It should be a valid case.

src/iceberg/update/update_properties.cc
```diff
@@ -92,21 +103,29 @@ Result<PendingUpdate::ApplyResult> UpdateProperties::Apply() {
     }
     format_version_ = static_cast<int8_t>(parsed_version);
 
-    updates_.erase(iter);
+    updates_.erase(TableProperties::kFormatVersion.key());
```
wgtmac, line 107: Please revert this line since the original implementation does not search the same key again.

src/iceberg/update/update_properties.cc
```diff
@@ -92,21 +103,29 @@ Result<PendingUpdate::ApplyResult> UpdateProperties::Apply() {
...
   }
 
   ApplyResult result;
   if (!updates_.empty()) {
     result.updates.emplace_back(std::make_unique<table::SetProperties>(updates_));
   }
   if (!removals_.empty()) {
-    result.updates.emplace_back(std::make_unique<table::RemoveProperties>(
-        std::vector<std::string>{removals_.begin(), removals_.end()}));
+    std::vector<std::string> existing_removals;
```
wgtmac: I think you can build `existing_removals` inside the for loop at line 77. Perhaps `existing_removals` can be simply renamed to `removals` for short.

HeartLinked: ok

wgtmac (approved):

PR #431, opened 2025-12-22
feat: enhance ManifestReader with projection and filtering support

wgtmac:
Refactors the `ManifestReader` implementation to support advanced data skipping
and projection pushdown, and adds full test coverage ported from the Java implementation.

### Implementation Changes:
- Consolidate `ManifestReader` logic: Move implementation from `manifest_reader_internal.cc`
  to `manifest_reader.cc` and remove the obsolete internal class.
- Fluent API: Add `Select()`, `FilterPartitions()`, and `FilterRows()` to support
  column projection and expression-based filtering (partition & metrics).
- Optimization: Support lazy initialization of the underlying Avro reader.
- ManifestEntry: Update schema handling to support reading specific columns and stats.

### Test Coverage:
- Add `manifest_reader_test.cc`: Ports `TestManifestReader` from Java, covering:
  - Basic reading and partition filtering.
  - V2/V3 specific features: Delete Files and Deletion Vectors (DVs).
  - Invalid usage validation (missing snapshot ID).
- Add `manifest_reader_stats_test.cc`: Ports `TestManifestReaderStats`, verifying:
  - Metrics projection logic (stats columns included/dropped based on filter/select).
  - Exact match of stats parsing against expected values.
- Parameterized Testing: All tests run against Format Versions 1, 2, and 3 to ensure cross-version compatibility.

### DIFF
--- a/src/iceberg/manifest/manifest_entry.h
+++ b/src/iceberg/manifest/manifest_entry.h
@@ -326,7 +326,7 @@ struct ICEBERG_EXPORT ManifestEntry {
   inline static const SchemaField kSnapshotId =
       SchemaField::MakeOptional(kSnapshotIdFieldId, "snapshot_id", int64());
 
-  inline static const int32_t kDataFileFieldId = 2;
+  static constexpr int32_t kDataFileFieldId = 2;
   inline static const std::string kDataFileField = "data_file";
 
   static constexpr int32_t kSequenceNumberFieldId = 3;
--- a/src/iceberg/manifest/manifest_list.h
+++ b/src/iceberg/manifest/manifest_list.h
@@ -153,63 +153,63 @@ struct ICEBERG_EXPORT ManifestFile {
   /// \brief Checks if this manifest file contains entries with DELETED status
   bool has_deleted_files() const { return deleted_files_count.value_or(1) > 0; }
 
-  inline static const int32_t kManifestPathFieldId = 500;
+  static constexpr int32_t kManifestPathFieldId = 500;
   inline static const SchemaField kManifestPath = SchemaField::MakeRequired(
       kManifestPathFieldId, "manifest_path", string(), "Location URI with FS scheme");
 
-  inline static const int32_t kManifestLengthFieldId = 501;
+  static constexpr int32_t kManifestLengthFieldId = 501;
   inline static const SchemaField kManifestLength = SchemaField::MakeRequired(
       kManifestLengthFieldId, "manifest_length", int64(), "Total file size in bytes");
 
-  inline static const int32_t kPartitionSpecIdFieldId = 502;
+  static constexpr int32_t kPartitionSpecIdFieldId = 502;
   inline static const SchemaField kPartitionSpecId = SchemaField::MakeRequired(
       kPartitionSpecIdFieldId, "partition_spec_id", int32(), "Spec ID used to write");
 
-  inline static const int32_t kContentFieldId = 517;
+  static constexpr int32_t kContentFieldId = 517;
   inline static const SchemaField kContent = SchemaField::MakeOptional(
       kContentFieldId, "content", int32(), "Contents of the manifest: 0=data, 1=deletes");
 
-  inline static const int32_t kSequenceNumberFieldId = 515;
+  static constexpr int32_t kSequenceNumberFieldId = 515;
   inline static const SchemaField kSequenceNumber =
       SchemaField::MakeOptional(kSequenceNumberFieldId, "sequence_number", int64(),
                                 "Sequence number when the manifest was added");
 
-  inline static const int32_t kMinSequenceNumberFieldId = 516;
+  static constexpr int32_t kMinSequenceNumberFieldId = 516;
   inline static const SchemaField kMinSequenceNumber =
       SchemaField::MakeOptional(kMinSequenceNumberFieldId, "min_sequence_number", int64(),
                                 "Lowest sequence number in the manifest");
 
-  inline static const int32_t kAddedSnapshotIdFieldId = 503;
+  static constexpr int32_t kAddedSnapshotIdFieldId = 503;
   inline static const SchemaField kAddedSnapshotId =
       SchemaField::MakeRequired(kAddedSnapshotIdFieldId, "added_snapshot_id", int64(),
                                 "Snapshot ID that added the manifest");
 
-  inline static const int32_t kAddedFilesCountFieldId = 504;
+  static constexpr int32_t kAddedFilesCountFieldId = 504;
   inline static const SchemaField kAddedFilesCount = SchemaField::MakeOptional(
       kAddedFilesCountFieldId, "added_files_count", int32(), "Added entry count");
 
-  inline static const int32_t kExistingFilesCountFieldId = 505;
+  static constexpr int32_t kExistingFilesCountFieldId = 505;
   inline static const SchemaField kExistingFilesCount =
       SchemaField::MakeOptional(kExistingFilesCountFieldId, "existing_files_count",
                                 int32(), "Existing entry count");
 
-  inline static const int32_t kDeletedFilesCountFieldId = 506;
+  static constexpr int32_t kDeletedFilesCountFieldId = 506;
   inline static const SchemaField kDeletedFilesCount = SchemaField::MakeOptional(
       kDeletedFilesCountFieldId, "deleted_files_count", int32(), "Deleted entry count");
 
-  inline static const int32_t kAddedRowsCountFieldId = 512;
+  static constexpr int32_t kAddedRowsCountFieldId = 512;
   inline static const SchemaField kAddedRowsCount = SchemaField::MakeOptional(
       kAddedRowsCountFieldId, "added_rows_count", int64(), "Added rows count");
 
-  inline static const int32_t kExistingRowsCountFieldId = 513;
+  static constexpr int32_t kExistingRowsCountFieldId = 513;
   inline static const SchemaField kExistingRowsCount = SchemaField::MakeOptional(
       kExistingRowsCountFieldId, "existing_rows_count", int64(), "Existing rows count");
 
-  inline static const int32_t kDeletedRowsCountFieldId = 514;
+  static constexpr int32_t kDeletedRowsCountFieldId = 514;
   inline static const SchemaField kDeletedRowsCount = SchemaField::MakeOptional(
       kDeletedRowsCountFieldId, "deleted_rows_count", int64(), "Deleted rows count");
 
-  inline static const int32_t kPartitionSummaryFieldId = 507;
+  static constexpr int32_t kPartitionSummaryFieldId = 507;
   inline static const SchemaField kPartitions = SchemaField::MakeOptional(
       kPartitionSummaryFieldId, "partitions",
       list(SchemaField::MakeRequired(508, std::string(ListType::kElementName),
@@ -221,11 +221,11 @@ struct ICEBERG_EXPORT ManifestFile {
                                      }))),
       "Summary for each partition");
 
-  inline static const int32_t kKeyMetadataFieldId = 519;
+  static constexpr int32_t kKeyMetadataFieldId = 519;
   inline static const SchemaField kKeyMetadata = SchemaField::MakeOptional(
       kKeyMetadataFieldId, "key_metadata", binary(), "Encryption key metadata blob");
 
-  inline static const int32_t kFirstRowIdFieldId = 520;
+  static constexpr int32_t kFirstRowIdFieldId = 520;
   inline static const SchemaField kFirstRowId = SchemaField::MakeOptional(
       kFirstRowIdFieldId, "first_row_id", int64(),
       "Starting row ID to assign to new rows in ADDED data files");


src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,870 @@
...
+#include "iceberg/schema_field.h"
 #include "iceberg/type.h"
+#include "iceberg/util/checked_cast.h"
 #include "iceberg/util/macros.h"
 
 namespace iceberg {
 
+namespace {
+
+#define PARSE_PRIMITIVE_FIELD(item, array_view, type)                                   \
```
wgtmac, line 49: Github is not clever enough to figure out changes in this file. Most parts come from combining `manifest_reader.cc` and `manifest_reader_internal.cc`.

src/iceberg/manifest/manifest_reader_internal.h
```diff
@@ -22,33 +22,105 @@
...
+
+  /// \brief Check if there's a non-trivial partition filter.
+  bool HasPartitionFilter() const;
+
+  /// \brief Check if there's a non-trivial row filter.
+  bool HasRowFilter() const;
+
+  /// \brief Add stats columns to the column list if needed.
+  std::vector<std::string> WithStatsColumns(
+      const std::vector<std::string>& columns) const;
```
WZhuo: It's a static function in ManifestReader

src/iceberg/manifest/manifest_reader.h
```diff
@@ -23,43 +23,71 @@
...
-#include "iceberg/type.h"
 #include "iceberg/type_fwd.h"
 
 namespace iceberg {
 
 /// \brief Read manifest entries from a manifest file.
 class ICEBERG_EXPORT ManifestReader {
  public:
+  /// \brief Special value to select all columns from manifest files.
+  inline static const std::vector<std::string> kAllColumns{"*"};
```
WZhuo: There's no need to be a vector, and it's also defined in Schema::Select

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+}
+
+}  // namespace
+
+std::vector<std::string> ManifestReader::WithStatsColumns(
+    const std::vector<std::string>& columns) {
+  if (std::ranges::all_of(ManifestReader::kAllColumns,
+                          [&columns](const std::string& col) {
+                            return std::ranges::find(columns, col) != columns.cend();
+                          })) {
```
WZhuo:
```suggestion
  if (std::ranges::find(columns, ManifestReader::kAllColumns) != columns.cend()) {
```

src/iceberg/manifest/manifest_entry.h
```diff
@@ -178,94 +178,114 @@ struct ICEBERG_EXPORT DataFile {
   /// present
   std::optional<int64_t> content_size_in_bytes;
 
+  inline static constexpr int32_t kContentFieldId = 134;
```
zhjwpku: nit: constexpr implies inline, maybe remove the inline, not a strong opinion

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+  }
+  if (columns.empty()) {
+    return false;
+  }
+  const std::unordered_set<std::string> selected(columns.cbegin(), columns.cend());
+  if (std::ranges::all_of(
+          ManifestReader::kAllColumns,
+          [&selected](const std::string& col) { return selected.contains(col); })) {
+    return false;
+  }
```
WZhuo:
```suggestion
  if (selected.contains(ManifestReader::kAllColumns)) {
    return false;
  }
```
If kAllColumns is a string

src/iceberg/manifest/manifest_entry.h
HuaHuaY: We can remove `inline` on `static constexpr`.

src/iceberg/manifest/manifest_list.cc
```diff
@@ -19,22 +19,24 @@
...
-#include "iceberg/schema.h"
+#include <memory>
+
+#include "iceberg/type.h"
 
 namespace iceberg {
 
-const StructType& PartitionFieldSummary::Type() {
-  static const StructType kInstance{{
+const std::shared_ptr<StructType>& PartitionFieldSummary::Type() {
```
HuaHuaY, line 28: Why do we need to change the return type from `const T&` to `const std::shared_ptr<T>&`?

src/iceberg/manifest/manifest_reader.cc
HuaHuaY: Can we change these macros to template functions?

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+  return *this;
+}
+
+ManifestReader& ManifestReaderImpl::CaseSensitive(bool case_sensitive) {
+  case_sensitive_ = case_sensitive;
+  return *this;
+}
+
+bool ManifestReaderImpl::HasPartitionFilter() const {
+  return part_filter_ && part_filter_->op() != Expression::Operation::kTrue;
```
WZhuo: part_filter_ always true
```suggestion
  return part_filter_->op() != Expression::Operation::kTrue;
```

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+}
+
+bool ManifestReaderImpl::HasRowFilter() const {
+  return row_filter_ && row_filter_->op() != Expression::Operation::kTrue;
```
WZhuo:
```suggestion
  return row_filter_->op() != Expression::Operation::kTrue;
```

src/iceberg/manifest/manifest_reader.cc
HuaHuaY: Use `std::ranges::contains` instead.

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+      default:
+        return InvalidManifestList("Unsupported field: {} in manifest file.", field_name);
+    }
+  }
+  return manifest_files;
+}
+
+Status ParseLiteral(ArrowArrayView* view_of_partition, int64_t row_idx,
+                    std::vector<ManifestEntry>& manifest_entries) {
+  if (view_of_partition->storage_type == ArrowType::NANOARROW_TYPE_BOOL) {
```
HuaHuaY: Use `switch` instead of multiple `if`?

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+        break;
+      }
+      default:
+        return InvalidManifest("Unsupported field: {} in manifest entry.", field_name);
+    }
+  }
+  return manifest_entries;
+}
+
+const std::unordered_set<std::string> kStatsColumns = {
```
HuaHuaY: I think `std::vector<std::string>` is enough.

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+      internal::ArrowArrayGuard array_guard(&result.value());
+      ICEBERG_ASSIGN_OR_RAISE(
+          auto parse_result, ParseManifestList(&arrow_schema, &result.value(), *schema_));
+      manifest_files.insert(manifest_files.end(),
+                            std::make_move_iterator(parse_result.begin()),
+                            std::make_move_iterator(parse_result.end()));
+    } else {
+      // eof
+      break;
+    }
```
HuaHuaY:
```suggestion
    if (!result.has_value()) {
      // eof
      break;
    }
    internal::ArrowArrayGuard array_guard(&result.value());
    ICEBERG_ASSIGN_OR_RAISE(
        auto parse_result, ParseManifestList(&arrow_schema, &result.value(), *schema_));
    manifest_files.insert(manifest_files.end(),
                          std::make_move_iterator(parse_result.begin()),
                          std::make_move_iterator(parse_result.end()));
```
I prefer to write this way, with one less tab.

src/iceberg/manifest/manifest_reader_internal.h
HuaHuaY: It seems to be a function declaration without implementation. I guess there is a careless mistake.

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+}
+
+Status ParsePartitionFieldSummaryList(ArrowArrayView* view_of_column,
+                                      std::vector<ManifestFile>& manifest_files) {
+  auto manifest_count = view_of_column->length;
+  // view_of_column is list<struct<PartitionFieldSummary>>
+  if (view_of_column->storage_type != ArrowType::NANOARROW_TYPE_LIST) {
+    return InvalidManifestList("partitions field should be a list.");
+  }
+  auto view_of_list_iterm = view_of_column->children[0];
```
zhjwpku:
```suggestion
  auto view_of_list_item = view_of_column->children[0];
```
typo?

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+  // view_of_list_iterm is struct<PartitionFieldSummary>
+  if (view_of_list_iterm->storage_type != ArrowType::NANOARROW_TYPE_STRUCT) {
+    return InvalidManifestList("partitions list field should be a list.");
```
zhjwpku:
```suggestion
    return InvalidManifestList("partitions list item should be a struct.");
```

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,867 @@
...
+                              int64_t);
+        break;
+      default:
+        return InvalidManifestList("Unsupported field: {} in manifest file.", field_name);
+    }
+  }
+  return manifest_files;
+}
+
+Status ParseLiteral(ArrowArrayView* view_of_partition, int64_t row_idx,
```
zhjwpku: Should we rename this to ParsePartitionValues, the current name is a little bit confusing to me.

src/iceberg/manifest/manifest_reader.cc
wgtmac: I agree with you. But I prefer to do it in a separate PR.

src/iceberg/manifest/manifest_reader_internal.h
wgtmac: It is due to a half-way refactoring that I forgot to delete it from the original place.

src/iceberg/manifest/manifest_list.cc
wgtmac: Now it is a singleton that can be reused.

src/iceberg/manifest/manifest_reader.h
zhjwpku: Why this is a vector, a single string should be enough?

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,872 @@
...
+    : manifest_path_(std::move(manifest_path)),
+      manifest_length_(manifest_length),
+      file_io_(std::move(file_io)),
+      schema_(std::move(schema)),
+      spec_(std::move(spec)),
+      inheritable_metadata_(std::move(inheritable_metadata)),
+      first_row_id_(first_row_id) {}
+
+ManifestReader& ManifestReaderImpl::Select(const std::vector<std::string>& columns) {
+  columns_ = columns;
```
shangxinli, line 623: I think Java maps user-requested names to immutable Field IDs from the manifest's actual schema.

shangxinli: Can we do something like this? 
 ```
 ManifestReader& ManifestReaderImpl::Select(const std::vector<std::string>& columns) {
    // Validate columns exist in data file schema before storing
    for (const auto& col_name : columns) {
      if (col_name != kAllColumns) {
        ICEBERG_CHECK(DataFile::GetSchemaByName(col_name, case_sensitive_).has_value(),
                      "Column '{}' not found in data file schema", col_name);
      }
    }
    columns_ = columns;
    return *this;
  }
```

File: src/iceberg/manifest/manifest_reader.cc [hunk 17]
```diff
@@ -19,61 +19,872 @@
...
+  return *this;
+}
+
+ManifestReader& ManifestReaderImpl::FilterPartitions(std::shared_ptr<Expression> expr) {
+  part_filter_ = expr ? std::move(expr) : True::Instance();
```
shangxinli, line 628: The java implementation have 'AND': https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestReader.java

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,872 @@
...
+}
+
+ManifestReader& ManifestReaderImpl::FilterPartitions(
+    std::shared_ptr<PartitionSet> partition_set) {
+  partition_set_ = std::move(partition_set);
+  return *this;
+}
+
+ManifestReader& ManifestReaderImpl::FilterRows(std::shared_ptr<Expression> expr) {
+  row_filter_ = expr ? std::move(expr) : True::Instance();
```
shangxinli, line 639: It is missing AND combination comparing with Java.

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,872 @@
...
+const std::vector<std::string> kStatsColumns = {"value_counts",     "null_value_counts",
+                                                "nan_value_counts", "lower_bounds",
+                                                "upper_bounds",     "record_count"};
+
+bool RequireStatsProjection(const std::shared_ptr<Expression>& row_filter,
+                            const std::vector<std::string>& columns) {
+  if (!row_filter || row_filter->op() == Expression::Operation::kTrue) {
+    return false;
+  }
+  if (columns.empty()) {
```
shangxinli, line 569: Should this function return true when stats columns are needed for filtering but NOT all selected. Currently it returns true when stats are NOT all selected, which is backwards. This will cause unnecessary stats projection when stats columns ARE already selected.

Something like below: 
```
  bool RequireStatsProjection(const std::shared_ptr<Expression>& row_filter,
                              const std::vector<std::string>& columns) {
    if (!row_filter || row_filter->op() == Expression::Operation::kTrue) {
      return false;  // No row filter, no stats needed
    }
    if (columns.empty() || std::ranges::contains(columns, ManifestReader::kAllColumns)) {
      return false;  // All columns selected, stats already included
    }
    // Return true if ANY stats column is missing from selection
    const std::unordered_set<std::string> selected(columns.cbegin(), columns.cend());
    return !std::ranges::all_of(kStatsColumns, [&selected](const std::string& col) {
      return selected.contains(col);
    });
  }

```

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -19,61 +19,872 @@
...
+  return part_filter_->op() != Expression::Operation::kTrue;
+}
+
+bool ManifestReaderImpl::HasRowFilter() const {
+  ICEBERG_DCHECK(row_filter_, "Row filter is not set");
+  return row_filter_->op() != Expression::Operation::kTrue;
+}
+
+Result<Evaluator*> ManifestReaderImpl::GetEvaluator() {
+  if (!evaluator_) {
```
shangxinli, line 659: Add null checks for  row_filter_ and part_filter_ before us?

src/iceberg/manifest/manifest_reader.h
```diff
@@ -23,43 +23,71 @@
...
-#include "iceberg/type.h"
 #include "iceberg/type_fwd.h"
 
 namespace iceberg {
 
 /// \brief Read manifest entries from a manifest file.
 class ICEBERG_EXPORT ManifestReader {
  public:
+  /// \brief Special value to select all columns from manifest files.
+  inline static const std::string kAllColumns = "*";
```
shangxinli: Consider to use constexpr std::string_view for better performance?

src/iceberg/manifest/manifest_reader.cc
wgtmac: This check is delayed to `ProjectSchema` where `schema->Select(columns, case_sensitive)` is called. It matches the Java impl that unmatched columns are regarded as missing columns.

src/iceberg/manifest/manifest_reader.cc
wgtmac: I'm hesitant to introduce such behavior because `ManifestReader` requires to extend `ErrorCollector` to collect errors in these APIs. I'm inclined to add a comment to explain this explicitly.

src/iceberg/manifest/manifest_reader.cc
wgtmac: Same as above.

src/iceberg/manifest/manifest_reader.cc
wgtmac: They are not null at all times so we don't need to check this.

src/iceberg/manifest/manifest_reader.cc
wgtmac: `Schema::Select` has guaranteed that duplicate selections will only appear once:

```cpp
  std::unordered_set<int32_t> selected_ids;
  for (const auto& name : names) {
    ICEBERG_ASSIGN_OR_RAISE(auto result, FindFieldByName(name, case_sensitive));
    if (result.has_value()) {
      selected_ids.insert(result.value().get().field_id());
    }
  }
```

zhjwpku (approved):

src/iceberg/manifest/manifest_list.h
```diff
@@ -153,51 +153,86 @@ struct ICEBERG_EXPORT ManifestFile {
   /// \brief Checks if this manifest file contains entries with DELETED status
   bool has_deleted_files() const { return deleted_files_count.value_or(1) > 0; }
 
+  inline static const int32_t kManifestPathFieldId = 500;
```
HuaHuaY: use `static constexpr`?

wgtmac:
Thanks all for the review! Let me merge it.

PR #434, opened 2025-12-24
refactor: move TableMetadataBuilder's methods into impl class

WZhuo:
Refactor TableMetadataBuilder, so that all updates implement in TableMetadataBuilder::Impl, and no more private method in header file for TableMetadataBuilder class.


wgtmac (approved):
Make sense!

PR #435, opened 2025-12-24
feat: implement delete file index

wgtmac:
Implemented the DeleteFileIndex and Builder to manage and efficiently filter
delete files (equality deletes, position deletes, and deletion vectors)
based on sequence numbers and partitions.

#### Key changes:
- Added `DeleteFileIndex` and `DeleteFileIndex::Builder` in `src/iceberg/delete_file_index.{h,cc}`.
- Added `ContentFileUtil` for helper functions related to content files and DVs.
- Updated `ManifestReader` to support dropping stats via `TryDropStats()`.
- Added comprehensive unit tests in `src/iceberg/test/delete_file_index_test.cc`.

### DIFF
--- a/src/iceberg/delete_file_index.cc
+++ b/src/iceberg/delete_file_index.cc
@@ -32,6 +32,7 @@
 #include "iceberg/manifest/manifest_entry.h"
 #include "iceberg/manifest/manifest_list.h"
 #include "iceberg/manifest/manifest_reader.h"
+#include "iceberg/metadata_columns.h"
 #include "iceberg/partition_spec.h"
 #include "iceberg/schema.h"
 #include "iceberg/util/checked_cast.h"
@@ -524,11 +525,6 @@ DeleteFileIndex::Builder& DeleteFileIndex::Builder::IgnoreResiduals() {
 }
 
 Result<std::vector<ManifestEntry>> DeleteFileIndex::Builder::LoadDeleteFiles() {
-  ICEBERG_PRECHECK(io_ != nullptr, "FileIO is required to load delete files");
-  ICEBERG_PRECHECK(!specs_by_id_.empty(),
-                   "Partition specs are required to load delete files");
-  ICEBERG_PRECHECK(schema_ != nullptr, "Schema is required to load delete files");
-
   // Build expression caches per spec ID
   std::unordered_map<int32_t, std::shared_ptr<Expression>> part_expr_cache;
   std::unordered_map<int32_t, std::unique_ptr<ManifestEvaluator>> eval_cache;
@@ -615,6 +611,14 @@ Result<std::vector<ManifestEntry>> DeleteFileIndex::Builder::LoadDeleteFiles() {
                     "Missing sequence number for delete file: {}",
                     entry.data_file->file_path);
       if (entry.sequence_number.value() > min_sequence_number_) {
+        auto& file = *entry.data_file;
+        // keep minimum stats to avoid memory pressure
+        std::unordered_set<int32_t> columns =
+            file.content == DataFile::Content::kPositionDeletes
+                ? std::unordered_set<int32_t>{MetadataColumns::kDeleteFilePathColumnId}
+                : std::unordered_set<int32_t>(file.equality_ids.begin(),
+                                              file.equality_ids.end());
+        ContentFileUtil::DropUnselectedStats(*entry.data_file, columns);
         files.emplace_back(std::move(entry));
       }
     }
@@ -717,7 +721,10 @@ Status DeleteFileIndex::Builder::AddEqualityDelete(
 
 Result<std::unique_ptr<DeleteFileIndex>> DeleteFileIndex::Builder::Build() {
   ICEBERG_RETURN_UNEXPECTED(CheckErrors());
-  ICEBERG_PRECHECK(schema_ != nullptr, "Schema is required to build DeleteFileIndex");
+  ICEBERG_PRECHECK(io_ != nullptr, "FileIO is required to load delete files");
+  ICEBERG_PRECHECK(schema_ != nullptr, "Schema is required to load delete files");
+  ICEBERG_PRECHECK(!specs_by_id_.empty(),
+                   "Partition specs are required to load delete files");
 
   std::vector<ManifestEntry> entries;
   if (!delete_manifests_.empty()) {
--- a/src/iceberg/manifest/manifest_reader.cc
+++ b/src/iceberg/manifest/manifest_reader.cc
@@ -844,7 +844,7 @@ Result<std::vector<ManifestEntry>> ManifestReaderImpl::ReadEntries(bool only_liv
       }
 
       if (drop_stats) {
-        ContentFileUtil::DropStats(*entry.data_file);
+        ContentFileUtil::DropAllStats(*entry.data_file);
       }
 
       manifest_entries.push_back(std::move(entry));
--- a/src/iceberg/test/delete_file_index_test.cc
+++ b/src/iceberg/test/delete_file_index_test.cc
@@ -35,6 +35,7 @@
 #include "iceberg/manifest/manifest_list.h"
 #include "iceberg/manifest/manifest_reader.h"
 #include "iceberg/manifest/manifest_writer.h"
+#include "iceberg/metadata_columns.h"
 #include "iceberg/partition_spec.h"
 #include "iceberg/schema.h"
 #include "iceberg/test/matchers.h"
@@ -1065,11 +1066,13 @@ TEST_P(DeleteFileIndexTest, TestDeletedStatusExcluded) {
                                             "/path/to/pos-delete-added.parquet"));
 }
 
-// TODO(gangwu): enable this test after ManifestGroup has been added.
-TEST_P(DeleteFileIndexTest, DISABLED_TestPositionDeleteDiscardMetrics) {
+TEST_P(DeleteFileIndexTest, TestPositionDeleteDiscardMetrics) {
   int version = GetParam();
 
   auto partition_a = PartitionValues({Literal::Int(0)});
+
+  constexpr int32_t kDeleteFilePathFieldId = MetadataColumns::kDeleteFilePathColumnId;
+  constexpr int32_t kPositionFieldId = MetadataColumns::kFilePositionColumnId;
 
   // Create a position delete file with full metrics
   auto pos_delete = std::make_shared<DataFile>(DataFile{
@@ -1080,12 +1083,12 @@ TEST_P(DeleteFileIndexTest, DISABLED_TestPositionDeleteDiscardMetrics) {
       .record_count = 100,
       .file_size_in_bytes = 1024,
       // Add stats for multiple columns
-      .column_sizes = {{1, 100}, {2, 200}, {3, 300}},
-      .value_counts = {{1, 10}, {2, 20}, {3, 30}},
-      .null_value_counts = {{1, 1}, {2, 2}, {3, 3}},
-      .nan_value_counts = {{1, 0}, {2, 0}, {3, 0}},
-      .lower_bounds = {{1, {0x01}}, {2, {0x02}}, {3, {0x03}}},
-      .upper_bounds = {{1, {0xFF}}, {2, {0xFE}}, {3, {0xFD}}},
+      .column_sizes = {{kDeleteFilePathFieldId, 100}, {kPositionFieldId, 200}},
+      .value_counts = {{kDeleteFilePathFieldId, 10}, {kPositionFieldId, 20}},
+      .null_value_counts = {{kDeleteFilePathFieldId, 1}, {kPositionFieldId, 2}},
+      .nan_value_counts = {{kDeleteFilePathFieldId, 0}, {kPositionFieldId, 0}},
+      .lower_bounds = {{kDeleteFilePathFieldId, {0x01}}, {kPositionFieldId, {0x02}}},
+      .upper_bounds = {{kDeleteFilePathFieldId, {0xFF}}, {kPositionFieldId, {0xFE}}},
       .partition_spec_id = partitioned_spec_->spec_id(),
   });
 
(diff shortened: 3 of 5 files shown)


src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,775 @@
...
+
+std::vector<std::shared_ptr<DataFile>> PositionDeletes::Filter(int64_t seq) {
+  IndexIfNeeded();
+
+  size_t start = FindStartIndex(seqs_, seq);
+  if (start >= files_.size()) {
+    return {};
+  }
+
+  return std::ranges::transform_view(files_ | std::views::drop(start),
```
HuaHuaY:
```cpp
return files_ 
  | std::views::drop(start) 
  | std::views::transform(&EqualityDeleteFile::data_file) 
  | std::ranges::to<std::vector<std::shared_ptr<DataFile>>>();
```

src/iceberg/delete_file_index.h
```diff
@@ -0,0 +1,406 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/delete_file_index.h
+/// An index of delete files by sequence number.
+
+#include <algorithm>
```
HuaHuaY:
```suggestion
```

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+  return {};
+}
+
+std::vector<std::shared_ptr<DataFile>> PositionDeletes::Filter(int64_t seq) {
+  IndexIfNeeded();
+
+  size_t start = FindStartIndex(seqs_, seq);
+  if (start >= files_.size()) {
+    return {};
+  }
```
HuaHuaY:
```suggestion
  if (std::ranges::lower_bound(seqs_, seq) == seqs_.end()) {
    return {};
  }
```
I think we don't need to add `FindStartIndex`.

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+
+  return files_ | std::views::drop(start) |
+         std::views::transform(&ManifestEntry::data_file) |
+         std::ranges::to<std::vector<std::shared_ptr<DataFile>>>();
+}
+
+std::vector<std::shared_ptr<DataFile>> PositionDeletes::ReferencedDeleteFiles() {
+  IndexIfNeeded();
+  return std::ranges::transform_view(files_,
```
HuaHuaY: Use `std::views::transform`

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+
+void PositionDeletes::IndexIfNeeded() {
+  if (indexed_) {
+    return;
+  }
+
+  // Sort by data sequence number
+  std::ranges::sort(files_, [](const auto& a, const auto& b) {
+    return a.sequence_number.value() < b.sequence_number.value();
+  });
```
HuaHuaY:
```suggestion
  std::ranges::sort(files_, std::ranges::less{}, &ManifestEntry::sequence_number);
```

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+
+  // Build sequence number array for binary search
+  seqs_ = std::ranges::transform_view(
```
HuaHuaY: ditto

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+  IndexIfNeeded();
+
+  size_t start = FindStartIndex(seqs_, seq);
+  if (start >= files_.size()) {
+    return {};
+  }
+
+  return files_ | std::views::drop(start) |
+         std::views::transform(&ManifestEntry::data_file) |
+         std::ranges::to<std::vector<std::shared_ptr<DataFile>>>();
```
HuaHuaY:
```suggestion
  auto iter = std::ranges::lower_bound(seqs_, seq);
  if (iter == seqs_.end()) {
    return {};
  }

  return files_ | std::views::drop(iter - seqs_.begin()) |
         std::views::transform(&ManifestEntry::data_file) |
         std::ranges::to<std::vector<std::shared_ptr<DataFile>>>();
```
I think adding `FindStartIndex` is not needed.

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+  std::vector<std::shared_ptr<DataFile>> result;
+  result.reserve(files_.size() - start);
+  for (size_t i = start; i < files_.size(); ++i) {
+    const auto& delete_file = files_[i];
+    ICEBERG_ASSIGN_OR_RAISE(bool may_contain,
+                            CanContainEqDeletesForFile(data_file, delete_file));
+    if (may_contain) {
+      result.push_back(delete_file.wrapped.data_file);
+    }
+  }
```
HuaHuaY:
```suggestion
  auto iter = std::ranges::lower_bound(seqs_, seq);
  if (iter == seqs_.end()) {
    return {};
  }

  std::vector<std::shared_ptr<DataFile>> result;
  result.reserve(seqs_.end() - iter);
  for (auto& delete_file : files_ | std::views::drop(iter - seqs_.begin())) {
    ICEBERG_ASSIGN_OR_RAISE(bool may_contain,
                            CanContainEqDeletesForFile(data_file, delete_file));
    if (may_contain) {
      result.push_back(delete_file.wrapped.data_file);
    }
  }
```

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+
+  return result;
+}
+
+std::vector<std::shared_ptr<DataFile>> EqualityDeletes::ReferencedDeleteFiles() {
+  IndexIfNeeded();
+  return std::ranges::transform_view(
```
HuaHuaY: ditto

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+         std::ranges::to<std::vector<std::shared_ptr<DataFile>>>();
+}
+
+void EqualityDeletes::IndexIfNeeded() {
+  if (indexed_) {
+    return;
+  }
+
+  // Sort by apply sequence number
+  std::ranges::sort(files_, [](const auto& a, const auto& b) {
```
HuaHuaY: ditto

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,776 @@
...
+    return a.apply_sequence_number < b.apply_sequence_number;
+  });
+
+  // Build sequence number array for binary search
+  seqs_ = std::ranges::transform_view(
```
HuaHuaY: ditto

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -592,6 +593,29 @@ Result<std::shared_ptr<Schema>> ProjectSchema(std::shared_ptr<Schema> schema,
...
 
+bool ManifestReader::ShouldDropStats(const std::vector<std::string>& columns) {
+  // Make sure we only drop all stats if we had projected all stats.
+  // We do not drop stats even if we had partially added some stats columns, except for
+  // record_count column.
+  // Since we don't want to keep stats map which could be huge in size just because we
+  // select record_count, which is a primitive type.
+  if (!columns.empty()) {
+    const std::unordered_set<std::string_view> selected(columns.cbegin(), columns.cend());
+    if (selected.contains(ManifestReader::kAllColumns)) {
```
zhjwpku:
```suggestion
    if (selected.contains(Schema::kAllColumns)) {
```

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -557,9 +558,9 @@ Result<std::vector<ManifestEntry>> ParseManifestEntry(
   return manifest_entries;
 }
 
-const std::vector<std::string> kStatsColumns = {"value_counts",     "null_value_counts",
-                                                "nan_value_counts", "lower_bounds",
-                                                "upper_bounds",     "record_count"};
+const std::vector<std::string> kStatsColumns = {
+    "value_counts", "null_value_counts", "nan_value_counts", "lower_bounds",
+    "upper_bounds", "column_sizes",      "record_count"};
```
zhjwpku, line 563: The newly added `column_sizes` seems not aligned with java impl, is that a bug in java or our implementation is different?

zhjwpku (approved):

src/iceberg/manifest/manifest_reader.cc
wgtmac: Yes, I think this is an oversight in the Java impl if I understand it correctly.

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+#include "iceberg/schema.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/content_file_util.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg {
+
+namespace internal {
+
+Status EqualityDeleteFile::ConvertBoundsIfNeeded() const {
```
shangxinli, line 46: if the same EqualityDeleteFile is accessed from multiple threads during filtering, we might have race condition. Java has synchronization. 

```
 private Map<Integer, Object> lowerBounds() {
    if (convertedLowerBounds == null) {
      synchronized (this) {
        if (convertedLowerBounds == null) {
          this.convertedLowerBounds = convertBounds(wrapped.lowerBounds());
        }
      }
    }
    return convertedLowerBounds;
  }
```

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+}
+
+std::vector<std::shared_ptr<DataFile>> PositionDeletes::ReferencedDeleteFiles() {
+  IndexIfNeeded();
+  return files_ | std::views::transform(&ManifestEntry::data_file) |
+         std::ranges::to<std::vector<std::shared_ptr<DataFile>>>();
+}
+
+void PositionDeletes::IndexIfNeeded() {
+  if (indexed_) {
```
shangxinli, line 197: The Java implementation uses double-checked locking with synchronized blocks:

```
  // Java
  private void indexIfNeeded() {
    if (buffer != null) {
      synchronized (this) {
        if (buffer != null) {
          this.files = indexFiles(buffer);
          this.seqs = indexSeqs(files);
          this.buffer = null;
        }
      }
    }
  }
```

  The C++ version has no synchronization

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+
+std::vector<std::shared_ptr<DataFile>> EqualityDeletes::ReferencedDeleteFiles() {
+  IndexIfNeeded();
+  return files_ |
+         std::views::transform([](const auto& f) { return f.wrapped.data_file; }) |
+         std::ranges::to<std::vector<std::shared_ptr<DataFile>>>();
+}
+
+void EqualityDeletes::IndexIfNeeded() {
+  if (indexed_) {
```
shangxinli, line 252: Same comments for thread safety.

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+  case_sensitive_ = case_sensitive;
+  return *this;
+}
+
+DeleteFileIndex::Builder& DeleteFileIndex::Builder::IgnoreResiduals() {
+  ignore_residuals_ = true;
+  return *this;
+}
+
+Result<std::vector<ManifestEntry>> DeleteFileIndex::Builder::LoadDeleteFiles() {
```
shangxinli, line 527: The Java implementation supports optional parallel manifest loading via ExecutorService. The C++ version currently only supports sequential loading, Consider adding: Builder& PlanWith(std::shared_ptr<ThreadPool> executor)

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+
+Result<DeleteFileIndex::Builder> DeleteFileIndex::BuilderFor(
+    std::shared_ptr<FileIO> io, std::vector<ManifestFile> delete_manifests) {
+  ICEBERG_PRECHECK(io != nullptr, "FileIO cannot be null");
+  return Builder(std::move(io), std::move(delete_manifests));
+}
+
+// Builder implementation
+
+DeleteFileIndex::Builder::Builder(std::shared_ptr<FileIO> io,
```
shangxinli, line 463: Java tracks and reports metrics throughout the indexing process.

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+  if (!dv_by_path_) {
+    return nullptr;
+  }
+
+  auto it = dv_by_path_->find(data_file.file_path);
+  if (it == dv_by_path_->end()) {
+    return nullptr;
+  }
+
+  ICEBERG_CHECK(it->second.sequence_number.value() >= seq,
```
shangxinli, line 447: ICEBERG_CHECK() will crash the prgram, right? ValidationException.check() will just throw exception which is catchable. Should we do? 

```
if (it->second.sequence_number.value() < seq) {
    return InvalidArgument(
        "DV data sequence number {} must be greater than or equal to data file sequence number {}",
        it->second.sequence_number.value(), seq);
  }
```

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+    ICEBERG_ASSIGN_OR_RAISE(auto entries, reader->LiveEntries());
+    files.reserve(files.size() + entries.size());
+
+    for (auto& entry : entries) {
+      ICEBERG_CHECK(entry.data_file != nullptr, "ManifestEntry must have a data file");
+      ICEBERG_CHECK(entry.sequence_number.has_value(),
+                    "Missing sequence number for delete file: {}",
+                    entry.data_file->file_path);
+      if (entry.sequence_number.value() > min_sequence_number_) {
+        files.emplace_back(std::move(entry));
```
shangxinli, line 622: Java keeps stats only for columns needed for filtering:
  - Position deletes: Only DELETE_FILE_PATH column
  - Equality deletes: Only columns in equalityFieldIds()

src/iceberg/delete_file_index.cc
```diff
@@ -0,0 +1,769 @@
...
+      auto deletes = std::make_unique<internal::EqualityDeletes>(*schema_);
+      ICEBERG_RETURN_UNEXPECTED(deletes->Add(std::move(entry)));
+      deletes_by_partition.put(spec_id, partition, std::move(deletes));
+    }
+  }
+
+  return {};
+}
+
+Result<std::unique_ptr<DeleteFileIndex>> DeleteFileIndex::Builder::Build() {
```
shangxinli, line 722: Add  io_, specs_by_id_ validation?

src/iceberg/delete_file_index.cc
wgtmac: We don't have multithreading support yet so I don't think we need this at this moment. We can revisit this later once we try to add an executor implementation.

src/iceberg/delete_file_index.cc
wgtmac: Same comment as above.

src/iceberg/delete_file_index.cc
wgtmac: Perhaps we can consider https://en.cppreference.com/w/cpp/experimental/execution.html once we bump to C++26?

src/iceberg/delete_file_index.cc
wgtmac: Metrics reporting is not planned at this moment. I think we can include it in later releases.

src/iceberg/delete_file_index.cc
wgtmac: It is defined as below:

```cpp
#define ICEBERG_CHECK(expr, ...)            \
  do {                                      \
    if (!(expr)) [[unlikely]] {             \
      return ValidationFailed(__VA_ARGS__); \
    }                                       \
  } while (0)
```

shangxinli: I misread it as ICEBERG_CHECK_OR_DIE. It is good.

src/iceberg/delete_file_index.cc
wgtmac: Ah, I was checking this in the `DeleteFileIndex::LoadDeleteFiles` but I agree that here is a better place. Fixed as requested.

src/iceberg/delete_file_index.cc
wgtmac: Good catch! This was my oversight and now I have fixed it.

HuaHuaY (approved):

liurenjie1024 (approved):

PR #437, opened 2025-12-24
feat: add schema update to table metadata builder

gty404:


src/iceberg/partition_spec.cc
```diff
@@ -155,6 +155,42 @@ Status PartitionSpec::Validate(const Schema& schema, bool allow_missing_fields)
   return {};
 }
 
+Status PartitionSpec::ValidatePartitionName(const Schema& schema) const {
+  std::unordered_set<std::string> partition_names;
+  for (const auto& partition_field : fields_) {
+    auto name = std::string(partition_field.name());
+    if (name.empty()) {
+      return InvalidArgument("Cannot use empty partition name: {}", name);
+    }
```
WZhuo:
```suggestion
    ICEBERG_PRECHECK(!partition_field.name().empty(), "Cannot use empty partition name: {}", name);
    auto name = std::string(partition_field.name());
```

src/iceberg/schema.cc
```diff
@@ -30,6 +30,7 @@
 #include "iceberg/util/macros.h"
 #include "iceberg/util/type_util.h"
 #include "iceberg/util/visit_type.h"
+#include "table_metadata.h"
```
WZhuo:
```suggestion
#include "iceberg/table_metadata.h"
```

src/iceberg/schema.cc
```diff
@@ -228,4 +229,43 @@ Result<std::vector<std::string>> Schema::IdentifierFieldNames() const {
...
+Result<int32_t> Schema::HighestFieldId() const {
+  ICEBERG_ASSIGN_OR_RAISE(auto id_to_field, id_to_field_.Get(*this));
+
+  if (id_to_field.get().empty()) {
+    return kInitialColumnId;
+  }
+
+  auto max_it = std::ranges::max_element(
+      id_to_field.get(),
+      [](const auto& lhs, const auto& rhs) { return lhs.first < rhs.first; });
```
WZhuo: Wrapper with a Lazy or a simple std::call_once? Avoid search every time.

src/iceberg/table_update.h
```diff
@@ -216,10 +216,10 @@ class ICEBERG_EXPORT RemovePartitionSpecs : public TableUpdate {
...
 class ICEBERG_EXPORT RemoveSchemas : public TableUpdate {
  public:
-  explicit RemoveSchemas(std::vector<int32_t> schema_ids)
+  explicit RemoveSchemas(std::unordered_set<int32_t> schema_ids)
       : schema_ids_(std::move(schema_ids)) {}
 
-  const std::vector<int32_t>& schema_ids() const { return schema_ids_; }
+  const std::unordered_set<int32_t>& schema_ids() const { return schema_ids_; }
 
   void ApplyTo(TableMetadataBuilder& builder) const override;
```
HeartLinked, line 224: What about implement `applyTo` here? Although the table update doesn’t seem to be invoked in the current architecture and may need refactoring , I think it makes sense to implement it here.

src/iceberg/schema.cc
```diff
@@ -228,4 +229,43 @@ Result<std::vector<std::string>> Schema::IdentifierFieldNames() const {
...
+
+  return max_it->first;
+}
+
+bool Schema::SameSchema(const Schema& other) const { return fields_ == other.fields_; }
```
WZhuo: Also need compare identifier_fields

src/iceberg/table_metadata.cc
```diff
@@ -608,7 +631,124 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+
+  return {};
+}
+
+Status TableMetadataBuilder::Impl::RemoveSchemas(const std::vector<int32_t>& schema_ids) {
+  std::unordered_set<int32_t> schema_ids_to_remove(schema_ids.begin(), schema_ids.end());
+  auto current_schema_id = metadata_.current_schema_id.value_or(Schema::kInitialSchemaId);
+  if (!schema_ids_to_remove.contains(current_schema_id)) {
+    return InvalidArgument("Cannot remove current schema: {}", current_schema_id);
+  }
```
WZhuo:
```suggestion
ICEBERG_PRECHECK(!schema_ids_to_remove.contains(current_schema_id), "Cannot remove current schema: {}", current_schema_id);
```

src/iceberg/table_metadata.cc
```diff
@@ -608,7 +631,124 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+    changes_.push_back(std::make_unique<table::SetCurrentSchema>(schema_id));
+  }
+
+  return {};
+}
+
+Status TableMetadataBuilder::Impl::RemoveSchemas(const std::vector<int32_t>& schema_ids) {
+  std::unordered_set<int32_t> schema_ids_to_remove(schema_ids.begin(), schema_ids.end());
+  auto current_schema_id = metadata_.current_schema_id.value_or(Schema::kInitialSchemaId);
+  if (!schema_ids_to_remove.contains(current_schema_id)) {
```
WZhuo: `schema_ids_to_remove` should not contain `current_schema_id`, otherwise return error.

src/iceberg/table_metadata.cc
```diff
@@ -608,7 +631,124 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+
+  return {};
+}
+
+Result<int32_t> TableMetadataBuilder::Impl::AddSchema(const Schema& schema,
+                                                      int32_t new_last_column_id) {
+  if (new_last_column_id < metadata_.last_column_id) {
+    return InvalidArgument("Invalid last column ID: {} < {} (previous last column ID)",
+                           new_last_column_id, metadata_.last_column_id);
+  }
```
WZhuo: It can be replaced with `ICEBERG_PRECHECK` macro

src/iceberg/table_metadata.cc
```diff
@@ -608,7 +631,124 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+    bool is_new_schema =
+        last_added_schema_id_.has_value() &&
+        std::ranges::find_if(changes_, [new_schema_id](const auto& change) {
+          if (change->kind() != TableUpdate::Kind::kAddSchema) {
+            return false;
+          }
+          auto* add_schema = dynamic_cast<table::AddSchema*>(change.get());
+          return add_schema->schema()->schema_id().value_or(Schema::kInitialSchemaId) ==
+                 new_schema_id;
+        }) != changes_.cend();
```
WZhuo:
```suggestion
        std::ranges::any_of(changes_, [new_schema_id](const auto& change) {
          if (change->kind() != TableUpdate::Kind::kAddSchema) {
            return false;
          }
          auto* add_schema = dynamic_cast<table::AddSchema*>(change.get());
          return add_schema->schema()->schema_id().value_or(Schema::kInitialSchemaId) ==
                 new_schema_id;
        });
```

src/iceberg/table_metadata.cc
```diff
@@ -608,7 +631,124 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+    last_added_schema_id_ =
+        is_new_schema ? std::make_optional(new_schema_id) : std::nullopt;
+    return new_schema_id;
+  }
+
+  auto new_schema = std::make_shared<Schema>(
+      std::vector<SchemaField>(schema.fields().begin(), schema.fields().end()),
+      new_schema_id);
```
WZhuo: Lack of `identifier_field_ids`
```suggestion
  auto new_schema = std::make_shared<Schema>(
      schema.fields() | std::ranges::to<std::vector>(),
      new_schema_id, schema.IdentifierFieldIds());
```

src/iceberg/table_metadata.cc
```diff
@@ -608,7 +631,124 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+                                                      int32_t new_last_column_id) {
+  if (new_last_column_id < metadata_.last_column_id) {
+    return InvalidArgument("Invalid last column ID: {} < {} (previous last column ID)",
+                           new_last_column_id, metadata_.last_column_id);
+  }
+
+  ICEBERG_RETURN_UNEXPECTED(schema.Validate(metadata_.format_version));
+
+  auto new_schema_id = ReuseOrCreateNewSchemaId(schema);
+  if (schemas_by_id_.find(new_schema_id) != schemas_by_id_.end()) {
```
WZhuo: It also check the `metadata_.last_column_id` equals to `new_last_column_id` in Java implement, does it matter?

src/iceberg/table_metadata.cc
```diff
@@ -709,16 +919,24 @@ TableMetadataBuilder& TableMetadataBuilder::UpgradeFormatVersion(
...
 }
 
-TableMetadataBuilder& TableMetadataBuilder::AddSchema(std::shared_ptr<Schema> schema) {
-  throw IcebergError(std::format("{} not implemented", __FUNCTION__));
+TableMetadataBuilder& TableMetadataBuilder::AddSchema(
+    std::shared_ptr<Schema> const& schema) {
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto highest_field_id, schema->HighestFieldId());
+  auto new_last_column_id = std::max(impl_->metadata().last_column_id, highest_field_id);
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto schema_id,
+                                   impl_->AddSchema(*schema, new_last_column_id));
```
WZhuo:
```suggestion
  ICEBERG_BUILDER_RETURN_IF_ERROR(impl_->AddSchema(*schema, new_last_column_id));
```
schema_id is not used

src/iceberg/table_metadata.h
```diff
@@ -73,10 +73,13 @@ struct ICEBERG_EXPORT TableMetadata {
   static constexpr int8_t kDefaultTableFormatVersion = 2;
   static constexpr int8_t kSupportedTableFormatVersion = 3;
   static constexpr int8_t kMinFormatVersionRowLineage = 3;
+  static constexpr int8_t kMinFormatVersionDefaultValues = 3;
   static constexpr int64_t kInitialSequenceNumber = 0;
   static constexpr int64_t kInvalidSequenceNumber = -1;
   static constexpr int64_t kInitialRowId = 0;
 
+  static inline const std::unordered_map<TypeId, int8_t> kMinFormatVersions = {};
```
WZhuo, line 81: Not inited?

src/iceberg/table_metadata.cc
```diff
@@ -608,7 +631,124 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
   return {};
 }
 
-std::unique_ptr<TableMetadata> TableMetadataBuilder::Impl::Build() {
+Status TableMetadataBuilder::Impl::SetCurrentSchema(int32_t schema_id) {
+  if (schema_id == kLastAdded) {
+    if (!last_added_schema_id_.has_value()) {
+      return InvalidArgument("Cannot set last added schema: no schema has been added");
```
WZhuo:
```suggestion
      return ValidationFailed("Cannot set last added schema: no schema has been added");
```

src/iceberg/table_metadata.h
gty404: This is reserved for new types. Iceberg v3 added new types, but they are currently not defined in iceberg-cpp.

WZhuo (approved):

src/iceberg/table_metadata.h
```diff
@@ -246,7 +249,7 @@ class ICEBERG_EXPORT TableMetadataBuilder : public ErrorCollector {
   /// \param schema The schema to set as current
   /// \param new_last_column_id The highest column ID in the schema
   /// \return Reference to this builder for method chaining
-  TableMetadataBuilder& SetCurrentSchema(std::shared_ptr<Schema> schema,
+  TableMetadataBuilder& SetCurrentSchema(std::shared_ptr<Schema> const& schema,
```
```suggestion
  TableMetadataBuilder& SetCurrentSchema(const std::shared_ptr<Schema>& schema,
```

src/iceberg/table_metadata.h
```diff
@@ -259,7 +262,7 @@ class ICEBERG_EXPORT TableMetadataBuilder : public ErrorCollector {
   ///
   /// \param schema The schema to add
   /// \return Reference to this builder for method chaining
-  TableMetadataBuilder& AddSchema(std::shared_ptr<Schema> schema);
+  TableMetadataBuilder& AddSchema(std::shared_ptr<Schema> const& schema);
```
```suggestion
  TableMetadataBuilder& AddSchema(const std::shared_ptr<Schema>& schema);
```

src/iceberg/schema.cc
```diff
@@ -228,4 +243,33 @@ Result<std::vector<std::string>> Schema::IdentifierFieldNames() const {
...
+    // Check if the field's type requires a minimum format version
+    if (auto it = TableMetadata::kMinFormatVersions.find(field.type()->type_id());
+        it != TableMetadata::kMinFormatVersions.end()) {
+      if (int32_t min_format_version = it->second; format_version < min_format_version) {
+        return InvalidSchema("Invalid type for {}: {} is not supported until v{}",
+                             field.name(), *field.type(), min_format_version);
+      }
+    }
+
+    // TODO(GuoTao.yu): Check default values when they are supported
```
wgtmac, line 269: Do you think we need to add the Java `Schema.validateIdentifierField` and call it from here?

cc @WZhuo

src/iceberg/partition_spec.cc
```diff
@@ -177,6 +177,41 @@ Status PartitionSpec::Validate(const Schema& schema, bool allow_missing_fields)
   return {};
 }
 
+Status PartitionSpec::ValidatePartitionName(const Schema& schema) const {
```
wgtmac: Do you think we need to call this function in the `PartitionSpec::Validate` as well?

src/iceberg/table_metadata.cc
```diff
@@ -681,7 +705,122 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
   return {};
 }
 
-std::unique_ptr<TableMetadata> TableMetadataBuilder::Impl::Build() {
+Status TableMetadataBuilder::Impl::SetCurrentSchema(int32_t schema_id) {
+  if (schema_id == kLastAdded) {
```
wgtmac, line 694: Could you help fix `if (order_id == -1)` and `if (spec_id == -1)` in this file to use `kLastAdded` to replace -1?

src/iceberg/table_metadata.cc
```diff
@@ -681,7 +705,122 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+Status TableMetadataBuilder::Impl::RemoveSchemas(
+    const std::unordered_set<int32_t>& schema_ids) {
+  auto current_schema_id = metadata_.current_schema_id.value_or(Schema::kInitialSchemaId);
+  ICEBERG_PRECHECK(!schema_ids.contains(current_schema_id),
+                   "Cannot remove current schema: {}", current_schema_id);
+
+  if (!schema_ids.empty()) {
+    metadata_.schemas = metadata_.schemas | std::views::filter([&](const auto& schema) {
+                          return !schema_ids.contains(
+                              schema->schema_id().value_or(Schema::kInitialSchemaId));
```
```suggestion
                          return schema->schema_id().has_value() &&
                              !schema_ids.contains(schema->schema_id().value());
```

src/iceberg/table_metadata.cc
```diff
@@ -681,7 +705,122 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+                                                      int32_t new_last_column_id) {
+  ICEBERG_PRECHECK(new_last_column_id >= metadata_.last_column_id,
+                   "Invalid last column ID: {} < {} (previous last column ID)",
+                   new_last_column_id, metadata_.last_column_id);
+
+  ICEBERG_RETURN_UNEXPECTED(schema.Validate(metadata_.format_version));
+
+  auto new_schema_id = ReuseOrCreateNewSchemaId(schema);
+  if (schemas_by_id_.find(new_schema_id) != schemas_by_id_.end() &&
+      new_last_column_id == metadata_.last_column_id) {
```
```suggestion
  if (schemas_by_id_.contains(new_schema_id) &&
      new_last_column_id == metadata_.last_column_id) {
```

src/iceberg/test/table_metadata_builder_test.cc
```diff
@@ -386,4 +400,149 @@ TEST(TableMetadataBuilderTest, SetDefaultSortOrderInvalid) {
...
+  builder->AddSchema(schema3);
+  builder->AddSchema(schema4);
+  ICEBERG_UNWRAP_OR_FAIL(metadata, builder->Build());
+  ASSERT_EQ(metadata->schemas.size(), 3);
+  EXPECT_EQ(metadata->schemas[1]->schema_id().value(), 1);
+  EXPECT_EQ(metadata->schemas[2]->schema_id().value(), 2);
+  EXPECT_EQ(metadata->last_column_id, 6);
+}
+
+TEST(TableMetadataBuilderTest, AddSchemaInvalid) {
```
wgtmac: Should we remove this test since it tests nothing?

src/iceberg/table_metadata.cc
```diff
@@ -681,7 +705,122 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+    // it is now the last)
+    bool is_new_schema =
+        last_added_schema_id_.has_value() &&
+        std::ranges::any_of(changes_, [new_schema_id](const auto& change) {
+          if (change->kind() != TableUpdate::Kind::kAddSchema) {
+            return false;
+          }
+          auto* add_schema = dynamic_cast<table::AddSchema*>(change.get());
+          return add_schema->schema()->schema_id().value_or(Schema::kInitialSchemaId) ==
+                 new_schema_id;
```
```suggestion
          auto* add_schema = internal::checked_cast<table::AddSchema*>(change.get());
          return add_schema->schema()->schema_id() == std::make_optional(new_schema_id);
```

src/iceberg/table_metadata.cc
```diff
@@ -681,7 +705,122 @@ Status TableMetadataBuilder::Impl::RemoveProperties(
...
+    last_added_schema_id_ =
+        is_new_schema ? std::make_optional(new_schema_id) : std::nullopt;
+    return new_schema_id;
+  }
+
+  auto new_schema =
+      std::make_shared<Schema>(schema.fields() | std::ranges::to<std::vector>(),
+                               new_schema_id, schema.IdentifierFieldIds());
+
+  metadata_.schemas.push_back(new_schema);
```
wgtmac: If `new_schema_id` is found in the `metadata_.schemas`, isn't it added twice?

src/iceberg/table_metadata.cc
```diff
@@ -691,6 +830,25 @@ std::unique_ptr<TableMetadata> TableMetadataBuilder::Impl::Build() {
             std::chrono::system_clock::now().time_since_epoch())};
   }
 
+  auto current_schema_id = metadata_.current_schema_id.value_or(Schema::kInitialSchemaId);
```
wgtmac, line 832: What about returning an error if `metadata_.current_schema_id.has_value() == false`?

src/iceberg/table_metadata.cc
```diff
@@ -691,6 +830,25 @@ std::unique_ptr<TableMetadata> TableMetadataBuilder::Impl::Build() {
...
+  auto schema_it = schemas_by_id_.find(current_schema_id);
+  ICEBERG_PRECHECK(schema_it != schemas_by_id_.end(),
+                   "Current schema ID {} not found in schemas", current_schema_id);
+  const auto& current_schema = schema_it->second;
+  {
+    auto spec_it = specs_by_id_.find(metadata_.default_spec_id);
+    // FIXME(GuoTao.yu): Default spec must exist after we support update partition spec
```
wgtmac: I think we should return errors if `spec_it` is not found just like `sort_order_it`.

src/iceberg/table_metadata.cc
```diff
@@ -740,6 +898,57 @@ int32_t TableMetadataBuilder::Impl::ReuseOrCreateNewPartitionSpecId(
   return new_spec_id;
 }
 
+int32_t TableMetadataBuilder::Impl::ReuseOrCreateNewSchemaId(
+    const Schema& new_schema) const {
+  // if the schema already exists, use its id; otherwise use the highest id + 1
+  auto new_schema_id = metadata_.current_schema_id.value_or(Schema::kInitialSchemaId);
+  for (auto& schema : metadata_.schemas) {
+    auto schema_id = schema->schema_id().value_or(Schema::kInitialSchemaId);
```
wgtmac, line 907: Do you think we need to change `Schema::schema_id` to use `int32_t` instead of `std::optional<int32_t>`? It seems that we have to deal with nullopt in all places. `PartitionSpec` and `SortOrder` do not use optional for ids and Java schema uses 0 as the default. Of course we can do this in a separate PR.

src/iceberg/table_metadata.cc
```diff
@@ -740,6 +898,57 @@ int32_t TableMetadataBuilder::Impl::ReuseOrCreateNewPartitionSpecId(
...
+
+Result<std::shared_ptr<PartitionSpec>> TableMetadataBuilder::Impl::UpdateSpecSchema(
+    const Schema& schema, const PartitionSpec& partition_spec) {
+  // UpdateSpecSchema: Update partition spec to use the new schema
+  // This preserves the partition spec structure but rebinds it to the new schema
+
+  // Copy all fields from the partition spec. IDs should not change.
+  std::vector<PartitionField> fields;
+  fields.reserve(partition_spec.fields().size());
+  int32_t last_assigned_field_id = PartitionSpec::kLegacyPartitionDataIdStart;
```
wgtmac: We don't need to maintain `last_assigned_field_id` here because `PartitionSpec::Make` internally will handle this.

src/iceberg/table_metadata.cc
```diff
@@ -740,6 +898,57 @@ int32_t TableMetadataBuilder::Impl::ReuseOrCreateNewPartitionSpecId(
...
+
+  // Build without validation because the schema may have changed in a way that
+  // makes this spec invalid. The spec should still be preserved so that older
+  // metadata can be interpreted.
+  ICEBERG_ASSIGN_OR_RAISE(auto new_partition_spec,
+                          PartitionSpec::Make(partition_spec.spec_id(), std::move(fields),
+                                              last_assigned_field_id));
+
+  // Validate the new partition name against the new schema
+  ICEBERG_RETURN_UNEXPECTED(new_partition_spec->ValidatePartitionName(schema));
```
wgtmac: It seems that we can always call `ValidatePartitionName` in the `PartitionSpec::Make`. We can consider making it a static function so that `Validate` can call it as well.

src/iceberg/table_metadata.cc
```diff
@@ -740,6 +898,57 @@ int32_t TableMetadataBuilder::Impl::ReuseOrCreateNewPartitionSpecId(
...
+  return new_partition_spec;
+}
+
+Result<std::unique_ptr<SortOrder>> TableMetadataBuilder::Impl::UpdateSortOrderSchema(
```
wgtmac: I think the key difference between C++ and Java impls is that we don't store a `Schema` along with `PartitionSpec` and `SortOrder`. So we even don't need `UpdateSortOrderSchema` and `UpdateSpecSchema` functions, right?

src/iceberg/table_metadata.cc
gty404: Yes, we should give a default value when parsing the metadata JSON for an iceberg v1 table, and other fields should be handled similarly.

src/iceberg/schema.cc
WZhuo: I think `Schema.validateIdentifierField` should be called in Schema::Make, I can do it in a seperate PR

gty404: This method is equivalent to java's checkCompatibility, which only checks types and default values.

src/iceberg/table_metadata.cc
gty404: In the v1 spec, it is optional and cannot be reported as an error directly.

src/iceberg/partition_spec.cc
```diff
@@ -177,9 +179,10 @@ Status PartitionSpec::Validate(const Schema& schema, bool allow_missing_fields)
   return {};
 }
 
-Status PartitionSpec::ValidatePartitionName(const Schema& schema) const {
+Status PartitionSpec::ValidatePartitionName(const Schema& schema,
+                                            const PartitionSpec& spec) {
```
wgtmac, line 183: Do we need to add an extra parameter `case_sensitive`?

wgtmac: BTW, Java impl also has `checkForRedundantPartitions`, do we want to add that check? We can add a TODO somewhere if necessary.

gty404: Okay, I will align the check of partition spec with java in a separate pr

PR #443, opened 2025-12-27
feat: add rolling manifest writer

zhjwpku:


src/iceberg/manifest/rolling_manifest_writer.h
```diff
@@ -0,0 +1,129 @@
...
+  /// will be created.
+  RollingManifestWriter(ManifestWriterFactory manifest_writer_factory,
+                        int64_t target_file_size_in_bytes);
+
+  ~RollingManifestWriter();
+
+  /// \brief Add an added entry for a file.
+  ///
+  /// \param file a data file
+  /// \return Status::OK() if the entry was written successfully
```
wgtmac: Unrelated to this PR: I think we don't have `Status::OK()` method. We may need to fix all these comments to not confuse readers.

src/iceberg/manifest/rolling_manifest_writer.h
```diff
@@ -0,0 +1,129 @@
...
+  /// \brief Check if we should roll to a new file.
+  ///
+  /// This method checks if the current file has reached the target size
+  /// or the number of rows has reached the threshold. If so, it rolls to a new file.
+  bool ShouldRollToNewFile() const;
+
+  /// \brief Close the current writer and add its ManifestFile to the list.
+  Status CloseCurrentWriter();
+
+  static constexpr int64_t kRowsDivisor = 250;
```
wgtmac, line 117: Please add a comment to explain it.

src/iceberg/manifest/rolling_manifest_writer.h
```diff
@@ -0,0 +1,129 @@
...
+#include "iceberg/manifest/manifest_list.h"
+#include "iceberg/manifest/manifest_writer.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+/// \brief A rolling manifest writer that can produce multiple manifest files.
+///
+/// As opposed to ManifestWriter, a rolling writer could produce multiple manifest
+/// files.
```
wgtmac:
```suggestion
```

I think they are unnecessary.

src/iceberg/manifest/rolling_manifest_writer.cc
```diff
@@ -0,0 +1,121 @@
...
+
+RollingManifestWriter::RollingManifestWriter(
+    ManifestWriterFactory manifest_writer_factory, int64_t target_file_size_in_bytes)
+    : manifest_writer_factory_(std::move(manifest_writer_factory)),
+      target_file_size_in_bytes_(target_file_size_in_bytes) {}
+
+RollingManifestWriter::~RollingManifestWriter() {
+  // Ensure we close the current writer if not already closed
+  if (!closed_) {
+    (void)CloseCurrentWriter();
```
wgtmac:
```suggestion
    std::ignore = CloseCurrentWriter();
```

wgtmac: BTW, why not directly calling `Close()` so you don't need to check `closed_`?

src/iceberg/manifest/rolling_manifest_writer.cc
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  return current_writer_.get();
+}
+
+bool RollingManifestWriter::ShouldRollToNewFile() const {
+  if (current_writer_ == nullptr) {
+    return false;
+  }
+  // Roll when row count is a multiple of ROWS_DIVISOR and file size >= target
```
wgtmac:
```suggestion
  // Roll when row count is a multiple of the divisor and file size >= target
```

src/iceberg/manifest/rolling_manifest_writer.cc
```diff
@@ -0,0 +1,121 @@
...
+  if (current_file_rows_ % kRowsDivisor == 0) {
+    auto length_result = current_writer_->length();
+    if (length_result.has_value()) {
+      return length_result.value() >= target_file_size_in_bytes_;
+    }
+    // If we can't get the length, don't roll
```
wgtmac: Not sure if this is a good idea. What if it is a fatal error returned by the underlying writer?

wgtmac (approved):

src/iceberg/manifest/rolling_manifest_writer.cc
zhjwpku: I left this as a TODO, let's revisit this later.

src/iceberg/manifest/rolling_manifest_writer.cc
zhjwpku: Changed to `Close()`

src/iceberg/manifest/rolling_manifest_writer.h
zhjwpku: Changed to `\return Status indicating success or failure`

PR #444, opened 2025-12-28
feat: add snapshot cached manifests

zhjwpku:


src/iceberg/snapshot.h
```diff
@@ -250,14 +254,57 @@ struct ICEBERG_EXPORT Snapshot {
...
+  /// \brief Destructor
+  ~Snapshot();
+
  private:
   /// \brief Compare two snapshots for equality.
   bool Equals(const Snapshot& other) const;
+
+  /// \brief Cache structure for storing loaded manifests
+  struct ManifestsCache {
+    std::vector<ManifestFile> all_manifests;
```
wgtmac: Can we avoid caching 2X manifest files here? As we return std::span for different manifests, perhaps we can use a single vector to store data manifests in the head and then delete manifests in the tail, and then use the split point to construct different spans to return.

src/iceberg/snapshot.h
```diff
@@ -250,14 +254,57 @@ struct ICEBERG_EXPORT Snapshot {
...
+  /// \return A span of ManifestFile instances, or an error
+  Result<std::span<ManifestFile>> DeleteManifests(std::shared_ptr<FileIO> file_io) const;
+
   /// \brief Compare two snapshots for equality.
   friend bool operator==(const Snapshot& lhs, const Snapshot& rhs) {
     return lhs.Equals(rhs);
   }
 
+  /// \brief Default constructor
+  Snapshot();
```
wgtmac: If we have to add ctor to it, what about removing the default ctor and adding ctor for all member variables and `Snapshot::Make` for validated object creation? The Java `BaseSnapshot` has two ctors for v1 and others respectively.

src/iceberg/manifest/manifest_list.h
```diff
@@ -107,7 +106,7 @@ struct ICEBERG_EXPORT ManifestFile {
   int64_t min_sequence_number = TableMetadata::kInitialSequenceNumber;
   /// Field id: 503
   /// ID of the snapshot where the manifest file was added
-  int64_t added_snapshot_id = Snapshot::kInvalidSnapshotId;
+  int64_t added_snapshot_id = -1;  // Snapshot::kInvalidSnapshotId
```
wgtmac, line 109: nit: we can move `kInvalidSnapshotId` to `iceberg/constants.h`.

zhjwpku: Agreed, fixed in #456

PR #445, opened 2025-12-28
feat: implement direct Avro encoder for performance

shangxinli:
Implement direct Avro encoder to eliminate GenericDatum intermediate layer, matching the decoder approach for better performance.

Implementation:
- Add avro_direct_encoder_internal.h with EncodeArrowToAvro API
- Add avro_direct_encoder.cc implementing direct Arrow→Avro encoding
  - All primitive types: bool, int, long, float, double, string, binary
  - Temporal types: date, time, timestamp
  - Logical types: uuid, decimal
  - Nested types: struct, list, map (both string and non-string keys)
  - Union type handling for optional fields
- Modify avro_writer.cc to use DataFileWriterBase with direct encoder
- Add EncodeContext to reuse scratch buffers and avoid allocations

This matches Java Iceberg implementation using Encoder interface directly, avoiding intermediate object allocation overhead.

### DIFF
--- a/src/iceberg/avro/avro_direct_encoder.cc
+++ b/src/iceberg/avro/avro_direct_encoder.cc
@@ -223,7 +223,7 @@ Status EncodeArrowToAvro(const ::avro::NodePtr& avro_node, ::avro::Encoder& enco
       const size_t num_fields = avro_node->leaves();
 
       ICEBERG_PRECHECK(
-          static_cast<size_t>(struct_array.num_fields()) == num_fields,
+          struct_array.num_fields() == static_cast<int>(num_fields),
           "Field count mismatch: Arrow struct has {} fields, Avro node has {} fields",
           struct_array.num_fields(), num_fields);
       ICEBERG_PRECHECK(


src/iceberg/avro/avro_direct_encoder_internal.h
```diff
@@ -0,0 +1,69 @@
...
+/// Features:
+/// - All primitive, temporal, and logical types
+/// - Nested types (struct, list, map)
+/// - Union types for optional fields
+///
+/// Error Handling:
+/// - Type mismatches → InvalidArgument
+/// - Null in non-nullable field → InvalidArgument
+/// - Invalid decimal precision → InvalidArgument
+///
```
wgtmac:
```suggestion
/// methods from Arrow arrays.
///
```

src/iceberg/file_writer.h
```diff
@@ -49,6 +49,10 @@ class WriterProperties : public ConfigBase<WriterProperties> {
   /// \brief The sync interval used by Avro writer.
   inline static Entry<int64_t> kAvroSyncInterval{"write.avro.sync-interval", 16 * 1024};
 
+  /// \brief Whether to skip GenericDatum and use direct encoder for Avro writing.
+  /// When true, uses direct encoder (faster). When false, uses GenericDatum (legacy).
```
wgtmac:
```suggestion
  /// When true, uses direct encoder (faster). When false, uses GenericDatum.
```

src/iceberg/avro/avro_direct_encoder.cc
```diff
@@ -0,0 +1,417 @@
...
+    return InvalidArgument("Union must have exactly one null branch");
+  }
+}
+
+}  // namespace
+
+Status EncodeArrowToAvro(const ::avro::NodePtr& avro_node, ::avro::Encoder& encoder,
+                         const Type& type, const ::arrow::Array& array, int64_t row_index,
+                         EncodeContext* ctx) {
+  if (!ctx) {
```
wgtmac: I just noticed that we haven't checked null for `DecodeContext* ctx` in the decoder implementation. Can we use `DecodeContext&` and `EncodeContext& ctx` so we don't need to bother with null check?

src/iceberg/avro/avro_direct_encoder.cc
```diff
@@ -0,0 +1,417 @@
...
+      // AVRO_RECORD corresponds to Iceberg StructType (including Schema which extends
+      // StructType). Note: ListType and MapType are encoded as AVRO_ARRAY and AVRO_MAP
+      // respectively, not AVRO_RECORD.
+      if (type.type_id() != TypeId::kStruct) {
+        return InvalidArgument("AVRO_RECORD expects StructType, got type {}",
+                               type.ToString());
+      }
+
+      // Safe cast: type_id() == kStruct guarantees this is StructType or Schema
+      // (Schema extends StructType)
```
wgtmac:
```suggestion
      if (type.type_id() != TypeId::kStruct) {
        return InvalidArgument("AVRO_RECORD expects StructType, got type {}",
                               type.ToString());
      }

```

I don't think we need these obvious comments.

src/iceberg/avro/avro_direct_encoder.cc
```diff
@@ -0,0 +1,417 @@
...
+      const auto& struct_type = static_cast<const StructType&>(type);
+      const size_t num_fields = avro_node->leaves();
+
+      for (size_t i = 0; i < num_fields; ++i) {
+        const auto& field_node = avro_node->leafAt(i);
```
wgtmac, line 235: Do we need to check if num of fields mismatch between arrow array and avro node?

wgtmac (commented):
I haven't finished my review yet, just post my findings so far.

src/iceberg/avro/avro_direct_encoder.cc
shangxinli: sure

src/iceberg/avro/avro_direct_encoder.cc
shangxinli: sure

wgtmac (approved):

src/iceberg/avro/avro_direct_encoder.cc
```diff
@@ -0,0 +1,377 @@
...
+      ICEBERG_PRECHECK(type.type_id() == TypeId::kStruct,
+                       "AVRO_RECORD expects struct type, got type {}", type.ToString());
+
+      const auto& struct_array =
+          internal::checked_cast<const ::arrow::StructArray&>(array);
+      const auto& struct_type = internal::checked_cast<const StructType&>(type);
+      const size_t num_fields = avro_node->leaves();
+
+      ICEBERG_PRECHECK(
+          static_cast<size_t>(struct_array.num_fields()) == num_fields,
```
wgtmac:
```suggestion
          struct_array.num_fields() == static_cast<int>(num_fields),
```

PR #446, opened 2025-12-28
feat: add FileWriter base interface for data file writers

shangxinli:
Add iceberg/data subdirectory with FileWriter base interface that defines common operations for writing Iceberg data files, including data files, equality delete files, and position delete files.

### DIFF
--- a/src/iceberg/data/writer.h
+++ b/src/iceberg/data/writer.h
@@ -23,6 +23,7 @@
 /// Base interface for Iceberg data file writers.
 
 #include <cstdint>
+#include <memory>
 #include <vector>
 
 #include "iceberg/arrow_c_data.h"


src/iceberg/test/CMakeLists.txt
```diff
@@ -113,6 +113,8 @@ add_iceberg_test(util_test
 
 add_iceberg_test(roaring_test SOURCES roaring_test.cc)
 
+add_iceberg_test(data_writer_test SOURCES data_writer_test.cc)
```
wgtmac: I think we need to move it to the `if(ICEBERG_BUILD_BUNDLE)` block below since it depends on Avro and Parquet libraries.

src/iceberg/test/data_writer_test.cc
```diff
@@ -0,0 +1,218 @@
...
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/data/writer.h"
+#include "iceberg/manifest/manifest_entry.h"
+#include "iceberg/result.h"
+#include "iceberg/test/matchers.h"
+
+namespace iceberg {
+
+// Mock implementation of FileWriter for testing
+class MockFileWriter : public FileWriter {
```
wgtmac, line 35: IMHO, we don't need to add test cases for a pure interface classes. We can keep it for now and then remove all these cases once we have implemented `DataWriter`.

src/iceberg/data/writer.h
```diff
@@ -0,0 +1,91 @@
...
+/// 3. Call Close() to finalize the file
+/// 4. Call Metadata() to get file metadata (only valid after Close())
+///
+/// \note This interface is not thread-safe. Concurrent calls to Write()
+/// from multiple threads on the same instance are not supported.
+///
+/// \note This interface uses PascalCase method naming (Write, Length, Close, Metadata)
+/// to distinguish it from the lower-level iceberg/file_writer.h::Writer interface which
+/// uses lowercase naming. FileWriter is the Iceberg-specific data file writer
+/// abstraction, while Writer is the file format-level abstraction.
```
wgtmac:
```suggestion
```

src/iceberg/test/CMakeLists.txt
shangxinli: sure

src/iceberg/test/data_writer_test.cc
shangxinli: Sure

src/iceberg/data/writer.h
```diff
@@ -0,0 +1,82 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/data/writer.h
+/// Base interface for Iceberg data file writers.
+
+#include <cstdint>
+#include <vector>
```
wgtmac, line 27:
```suggestion
#include <memory>
#include <vector>
```

PR #450, opened 2025-12-29
fix: Validate identifier fields in Schema::Make

WZhuo:


src/iceberg/schema.cc
```diff
@@ -34,11 +35,80 @@
 
 namespace iceberg {
 
-Schema::Schema(std::vector<SchemaField> fields, std::optional<int32_t> schema_id,
-               std::vector<int32_t> identifier_field_ids)
-    : StructType(std::move(fields)),
-      schema_id_(schema_id),
-      identifier_field_ids_(std::move(identifier_field_ids)) {}
+namespace {
+Status ValidateIdentifierFields(
```
wgtmac: How about defining it as a class static function? I think it will be used in the SchemaUpdate action.

image

src/iceberg/schema.h
```diff
@@ -52,19 +52,33 @@ class ICEBERG_EXPORT Schema : public StructType {
...
   explicit Schema(std::vector<SchemaField> fields,
-                  std::optional<int32_t> schema_id = std::nullopt,
-                  std::vector<int32_t> identifier_field_ids = {});
+                  std::optional<int32_t> schema_id = std::nullopt);
+
+  /// \brief Create a schema.
+  ///
+  /// \param fields The fields that make up the schema.
+  /// \param schema_id The unique identifier for this schema (default:
+  /// kInitialSchemaId). \param identifier_field_ids Field IDs that uniquely identify
```
wgtmac: The alignment is a little bit weird.

src/iceberg/schema.h
```diff
@@ -52,19 +52,33 @@ class ICEBERG_EXPORT Schema : public StructType {
   /// \brief Special value to select all columns from manifest files.
   static constexpr std::string_view kAllColumns = "*";
 
+  /// \brief An empty schema instance.
+  static const std::shared_ptr<Schema> kEmptySchema;
+
   explicit Schema(std::vector<SchemaField> fields,
```
wgtmac: I think eventually we need to make it a private constructor.

PR #457, opened 2025-12-30
feat: add simple url encoder & decoder

WZhuo:
The url encode and decode function will be used for generating partitions' path, so move it to the common util for later use.

### DIFF
--- a/src/iceberg/util/url_encoder.cc
+++ b/src/iceberg/util/url_encoder.cc
@@ -41,7 +41,7 @@ constexpr int8_t FromHex(char c) {
 }  // namespace
 
 std::string UrlEncoder::Encode(std::string_view str_to_encode) {
-  static const char* hex_chars = "0123456789ABCDEF";
+  static const char* kHexChars = "0123456789ABCDEF";
   std::string result;
   result.reserve(str_to_encode.size() * 3 / 2 /* Heuristic reservation */);
 
@@ -50,8 +50,8 @@ std::string UrlEncoder::Encode(std::string_view str_to_encode) {
       result += c;
     } else {
       result += '%';
-      result += hex_chars[c >> 4];
-      result += hex_chars[c & 0xF];
+      result += kHexChars[c >> 4];
+      result += kHexChars[c & 0xF];
     }
   }
 


src/iceberg/CMakeLists.txt
```diff
@@ -103,26 +104,30 @@ set(ICEBERG_SHARED_INSTALL_INTERFACE_LIBS)
 
 list(APPEND
      ICEBERG_STATIC_BUILD_INTERFACE_LIBS
+     cpr::cpr
```
wgtmac: Can we avoid adding `cpr` to the core iceberg library? Why is it needed to generate partition path?

WZhuo:
```
Result<std::string>  LocationProvider::NewDataLocation(const PartitionSpec& spec,
                                              const PartitionValues& partition_data,
                                              const std::string& filename) = 0;
```
In LocationProvider::NewDataLocation, we must do url encoding for partition names and partition values.
If not import `cpr` to iceberg core, maybe we need implement Url encode/decode by ourselves.

wgtmac: Got it. According to prior discussions in the community, we should keep minimal dependencies of `libiceberg`. If the only reason is to escape partition path, should we implement our own utility function? IIUC, partition name and value can only use a limited set of characters.

src/iceberg/util/url_encoder.cc
```diff
@@ -0,0 +1,69 @@
...
+namespace iceberg {
+
+std::string UrlEncoder::Encode(std::string_view str_to_encode) {
+  std::stringstream escaped;
+  escaped.fill('0');
+  escaped << std::hex;
+
+  for (unsigned char c : str_to_encode) {
+    // reserve letters, numbers and -._~
+    if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
```
HuaHuaY: I suggest to use `template< class CharT > bool isalnum( CharT ch, const locale& loc )` in `<locale>` instead of `int isalnum( int ch )` in `<cctype>`.

src/iceberg/util/url_encoder.cc
```diff
@@ -0,0 +1,69 @@
...
+std::string UrlEncoder::Decode(std::string_view str_to_decode) {
+  std::string result;
+  result.reserve(str_to_decode.size());
+
+  for (size_t i = 0; i < str_to_decode.size(); ++i) {
+    char c = str_to_decode[i];
+    if (c == '%' && i + 2 < str_to_decode.size()) {
+      std::string hex(str_to_decode.substr(i + 1, 2));
+      try {
+        char decoded = static_cast<char>(std::stoi(hex, nullptr, 16));
```
HuaHuaY: Use `std::from_chars` instead.

src/iceberg/util/url_encoder.cc
wgtmac: On the contrary, I think we should stick to the one in `cctype` as per the RFC 3986. Other locales may result in unencoded non-ASCII characters.

src/iceberg/util/url_encoder.cc
```diff
@@ -0,0 +1,69 @@
...
+
+#include "iceberg/util/url_encoder.h"
+
+#include <iomanip>
+#include <sstream>
+
+namespace iceberg {
+
+std::string UrlEncoder::Encode(std::string_view str_to_encode) {
+  std::stringstream escaped;
```
wgtmac: `std::stringstream` may be slow. Can we directly use a string with a maximum preserved length?

src/iceberg/util/url_encoder.cc
```diff
@@ -0,0 +1,69 @@
...
+  escaped.fill('0');
+  escaped << std::hex;
+
+  for (unsigned char c : str_to_encode) {
+    // reserve letters, numbers and -._~
+    if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
+      escaped << c;
+    } else {
+      escaped << '%' << std::setw(2) << static_cast<int>(c) << std::setfill('0');
```
wgtmac: What about using a prebuilt lookup table for the hex conversion?

src/iceberg/util/url_encoder.cc
wgtmac: +1. `from_chars` can also help eliminate the try-catch flow which might be slow.

src/iceberg/util/url_encoder.cc
```diff
@@ -0,0 +1,69 @@
...
+}
+
+std::string UrlEncoder::Decode(std::string_view str_to_decode) {
+  std::string result;
+  result.reserve(str_to_decode.size());
+
+  for (size_t i = 0; i < str_to_decode.size(); ++i) {
+    char c = str_to_decode[i];
+    if (c == '%' && i + 2 < str_to_decode.size()) {
+      std::string hex(str_to_decode.substr(i + 1, 2));
```
wgtmac: This line may introduce a lot of string allocations and hurt the performance. Can we directly manipulate the string_view?

wgtmac (commented):
Putting together all my comments, the code may look like this (disclaimer: gemini wrote it):

```cpp

namespace {

bool IsUnreserved(unsigned char c) {
  return (c >= '0' && c <= '9') || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
         c == '-' || c == '.' || c == '_' || c == '~';
}

char ToHex(unsigned char v) {
  static constexpr std::array<char, 16> kHexChars = {
      '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
  return kHexChars[v & 0x0F];
}

int FromHex(char c) {
  if (c >= '0' && c <= '9') return c - '0';
  if (c >= 'a' && c <= 'f') return c - 'a' + 10;
  if (c >= 'A' && c <= 'F') return c - 'A' + 10;
  return -1;
}

}  // namespace

std::string UrlEncoder::Encode(std::string_view str_to_encode) {
  std::string escaped;
  escaped.reserve(str_to_encode.size() * 3 / 2);  // Heuristic reservation

  for (unsigned char c : str_to_encode) {
    if (IsUnreserved(c)) {
      escaped += static_cast<char>(c);
    } else {
      escaped += '%';
      escaped += ToHex(c >> 4);
      escaped += ToHex(c);
    }
  }
  return escaped;
}

std::string UrlEncoder::Decode(std::string_view str_to_decode) {
  std::string result;
  result.reserve(str_to_decode.size());

  for (size_t i = 0; i < str_to_decode.size(); ++i) {
    char c = str_to_decode[i];
    if (c == '%' && i + 2 < str_to_decode.size()) {
      int h1 = FromHex(str_to_decode[i + 1]);
      int h2 = FromHex(str_to_decode[i + 2]);

      if (h1 >= 0 && h2 >= 0) {
        result += static_cast<char>((h1 << 4) | h2);
        i += 2;
      } else {
        result += c;
      }
    } else if (c == '+') {
      result += ' ';
    } else {
      result += c;
    }
  }

  return result;
}
```

src/iceberg/util/url_encoder.cc
HuaHuaY: Results of `isalnum` in `<cctype>` will also be effected by `std::setlocale`. I suggest to use this:
```cpp
std::locale loc1("C");
std::isalnum(c, loc1);
```

HuaHuaY: Sorry. After I looked into their implementations, I think manually writing an isalnum might be best.

src/iceberg/util/url_encoder.cc
```diff
@@ -0,0 +1,84 @@
...
+      result += c;
+    } else {
+      result += '%';
+      result += hex_chars[c >> 4];
+      result += hex_chars[c & 0xF];
+    }
+  }
+
+  return result;
+}
```
wgtmac, line 59:
```suggestion
std::string UrlEncoder::Encode(std::string_view str_to_encode) {
  static const char* kHexChars = "0123456789ABCDEF";
  std::string result;
  result.reserve(str_to_encode.size() * 3 / 2 /* Heuristic reservation */);

  for (char c : str_to_encode) {
    if (IsUnreserved(c)) {
      result += c;
    } else {
      result += '%';
      result += kHexChars[c >> 4];
      result += kHexChars[c & 0xF];
    }
  }

  return result;
}
```

PR #460, opened 2025-12-30
feat: Add UpdateSchema interface skeleton

gty404:
- Define `UpdateSchema` class interface with full API documentation

### DIFF
--- a/src/iceberg/update/update_schema.cc
+++ b/src/iceberg/update/update_schema.cc
@@ -112,13 +112,6 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::string_view name,
 
 UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> parent,
                                               std::string_view name,
-                                              std::shared_ptr<Type> type) {
-  return AddColumnInternal(std::move(parent), name, /*is_optional=*/false,
-                           std::move(type), "");
-}
-
-UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> parent,
-                                              std::string_view name,
                                               std::shared_ptr<Type> type,
                                               std::string_view doc) {
   return AddColumnInternal(std::move(parent), name, /*is_optional=*/false,


src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+  ///
+  /// If type is a nested type, its field IDs are reassigned when added to the
+  /// existing schema.
+  ///
+  /// The added column will be optional with a null default value.
+  ///
+  /// \param name Name for the new column.
+  /// \param type Type for the new column.
+  /// \return Reference to this for method chaining.
+  /// \throws InvalidArgument If name contains ".".
```
wgtmac:
```suggestion
  /// \return Reference to this for method chaining.
  /// \note InvalidArgument will be reported if name contains ".".
```

Let's fix all `\throws` like this.

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+  ///
+  /// The added column will be optional with a null default value.
+  ///
+  /// \param name Name for the new column.
+  /// \param type Type for the new column.
+  /// \param doc Documentation string for the new column.
+  /// \return Reference to this for method chaining.
+  /// \throws InvalidArgument If name contains ".".
+  UpdateSchema& AddColumn(std::string_view name, std::shared_ptr<Type> type,
+                          std::string_view doc);
```
wgtmac: nit: add a default value to `doc`

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+  /// existing schema.
+  ///
+  /// The added column will be optional with a null default value.
+  ///
+  /// \param parent Name of the parent struct to which the column will be added.
+  /// \param name Name for the new column.
+  /// \param type Type for the new column.
+  /// \return Reference to this for method chaining.
+  /// \throws InvalidArgument If parent doesn't identify a struct.
+  UpdateSchema& AddColumn(std::optional<std::string> parent, std::string_view name,
```
wgtmac: Should it be `std::optional<std::string_view> parent`?

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+  /// existing schema.
+  ///
+  /// The added column will be optional with a null default value.
+  ///
+  /// \param name Name for the new column.
+  /// \param type Type for the new column.
+  /// \param doc Documentation string for the new column.
+  /// \return Reference to this for method chaining.
+  /// \throws InvalidArgument If name contains ".".
+  UpdateSchema& AddColumn(std::string_view name, std::shared_ptr<Type> type,
```
wgtmac, line 87: Should we unify this by removing the above one and adding a default parameter to `doc`?

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+  ///
+  /// The added column will be optional with a null default value.
+  ///
+  /// \param parent Name of the parent struct to which the column will be added.
+  /// \param name Name for the new column.
+  /// \param type Type for the new column.
+  /// \param doc Documentation string for the new column.
+  /// \return Reference to this for method chaining.
+  /// \throws InvalidArgument If parent doesn't identify a struct.
+  UpdateSchema& AddColumn(std::optional<std::string> parent, std::string_view name,
```
wgtmac: Same question for unifying this and the above one.

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+  /// Columns may be updated and renamed in the same schema update.
+  ///
+  /// \param name Name of the column to update.
+  /// \param new_type Replacement type for the column (must be primitive).
+  /// \param new_doc Replacement documentation string for the column.
+  /// \return Reference to this for method chaining.
+  /// \throws InvalidArgument If name doesn't identify a column in the schema or if
+  ///         this change introduces a type incompatibility or if it conflicts with
+  ///         other additions, renames, or updates.
+  UpdateSchema& UpdateColumn(std::string_view name,
```
wgtmac: Same question for unifying functions with and without `doc`.

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+  UpdateSchema& UnionByNameWith(std::shared_ptr<Schema> new_schema);
+
+  /// \brief Set the identifier fields given a set of field names.
+  ///
+  /// Because identifier fields are unique, duplicated names will be ignored. See
+  /// Schema::identifier_field_ids() to learn more about Iceberg identifier.
+  ///
+  /// \param names Names of the columns to set as identifier fields.
+  /// \return Reference to this for method chaining.
+  UpdateSchema& SetIdentifierFields(const std::vector<std::string>& names);
```
wgtmac: Use `std::span<const std::string>` or `std::span<std::string_view>`?

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,451 @@
...
+#include "iceberg/type_fwd.h"
+#include "iceberg/update/pending_update.h"
+
+namespace iceberg {
+
+/// \brief API for schema evolution.
+///
+/// When committing, these changes will be applied to the current table metadata.
+/// Commit conflicts will not be resolved and will result in a CommitFailed error.
+class ICEBERG_EXPORT UpdateSchema : public PendingUpdate {
```
wgtmac, line 47: I think we need to add a TODO comment somewhere to indicate that V3 default values should be supported.

src/iceberg/update/update_schema.h
gty404: I removed the interface that updates both type and doc at the same time. In the Java implementation, update can only modify one field at a time.

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,402 @@
...
+  /// If type is a nested type, its field IDs are reassigned when added to the
+  /// existing schema.
+  ///
+  /// \param name Name for the new column.
+  /// \param type Type for the new column.
+  /// \param doc Documentation string for the new column.
+  /// \return Reference to this for method chaining.
+  /// \note InvalidArgument will be reported if name contains ".".
+  UpdateSchema& AddRequiredColumn(std::string_view name, std::shared_ptr<Type> type,
+                                  std::string_view doc);
```
wgtmac:
```suggestion
  /// \brief Add a new required top-level column with documentation.
  ///
  /// Adding a required column without a default is an incompatible change that can
  /// break reading older data. To suppress exceptions thrown when an incompatible
  /// change is detected, call AllowIncompatibleChanges().
  ///
  /// Because "." may be interpreted as a column path separator or may be used in
  /// field names, it is not allowed in names passed to this method. To add to nested
  /// structures or to add fields with names that contain ".", use
  /// AddRequiredColumn(parent, name, type, doc).
  ///
  /// If type is a nested type, its field IDs are reassigned when added to the
  /// existing schema.
  ///
  /// \param name Name for the new column.
  /// \param type Type for the new column.
  /// \param doc Documentation string for the new column.
  /// \return Reference to this for method chaining.
  /// \note InvalidArgument will be reported if name contains ".".
  UpdateSchema& AddRequiredColumn(std::string_view name, std::shared_ptr<Type> type,
                                  std::string_view doc = "");
```

src/iceberg/update/update_schema.h
```diff
@@ -0,0 +1,402 @@
...
+  ///
+  /// \param parent Name of the parent struct to which the column will be added.
+  /// \param name Name for the new column.
+  /// \param type Type for the new column.
+  /// \param doc Documentation string for the new column.
+  /// \return Reference to this for method chaining.
+  /// \note InvalidArgument will be reported if parent doesn't identify a struct.
+  UpdateSchema& AddRequiredColumn(std::optional<std::string_view> parent,
+                                  std::string_view name, std::shared_ptr<Type> type,
+                                  std::string_view doc);
```
wgtmac:
```suggestion
  /// \brief Add a new required column to a nested struct with documentation.
  ///
  /// Adding a required column without a default is an incompatible change that can
  /// break reading older data. To suppress exceptions thrown when an incompatible
  /// change is detected, call AllowIncompatibleChanges().
  ///
  /// The parent name is used to find the parent using Schema::FindFieldByName(). If
  /// the parent name is null or empty, the new column will be added to the root as a
  /// top-level column. If parent identifies a struct, a new column is added to that
  /// struct. If it identifies a list, the column is added to the list element struct,
  /// and if it identifies a map, the new column is added to the map's value struct.
  ///
  /// The given name is used to name the new column and names containing "." are not
  /// handled differently.
  ///
  /// If type is a nested type, its field IDs are reassigned when added to the
  /// existing schema.
  ///
  /// \param parent Name of the parent struct to which the column will be added.
  /// \param name Name for the new column.
  /// \param type Type for the new column.
  /// \param doc Documentation string for the new column.
  /// \return Reference to this for method chaining.
  /// \note InvalidArgument will be reported if parent doesn't identify a struct.
  UpdateSchema& AddRequiredColumn(std::optional<std::string_view> parent,
                                  std::string_view name, std::shared_ptr<Type> type,
                                  std::string_view doc = "");
```

src/iceberg/update/update_schema.cc
```diff
@@ -0,0 +1,214 @@
...
+                           doc);
+}
+
+UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> parent,
+                                              std::string_view name,
+                                              std::shared_ptr<Type> type) {
+  return AddColumnInternal(std::move(parent), name, /*is_optional=*/false,
+                           std::move(type), "");
+}
+
```
wgtmac:
```suggestion
```

wgtmac (changes requested):

wgtmac (approved):

PR #461, opened 2025-12-30
refactor: use template function for manifest reader

wgtmac:
- Removed PARSE_XXX macros as many as possible
- Refined error reporting

### DIFF
--- a/src/iceberg/manifest/manifest_reader.cc
+++ b/src/iceberg/manifest/manifest_reader.cc
@@ -46,24 +46,16 @@ namespace iceberg {
 
 namespace {
 
-template <typename... Args>
-[[nodiscard]] Status MakeError(ErrorKind error_kind, std::format_string<Args...> fmt,
-                               Args&&... args) {
-  return std::unexpected<Error>(
-      {error_kind, std::format(fmt, std::forward<Args>(args)...)});
-}
-
-[[nodiscard]] Status MakeNullError(ErrorKind error_kind, std::string_view field_name,
-                                   int64_t row_idx) {
-  return MakeError(error_kind, "Required field '{}' is null at row {}", field_name,
-                   row_idx);
+[[nodiscard]] Status MakeNullError(std::string_view field_name, int64_t row_idx) {
+  return InvalidArgument("Required field '{}' is null at row {}", field_name, row_idx);
 }
 
 [[nodiscard]] Status AssertViewType(const ArrowArrayView* view, ArrowType expected_type,
-                                    std::string_view field_name, ErrorKind error_kind) {
+                                    std::string_view field_name) {
   if (view->storage_type != expected_type) {
-    return MakeError(error_kind, "Field '{}' should be of type {}, got {}", field_name,
-                     ArrowTypeString(expected_type), ArrowTypeString(view->storage_type));
+    return InvalidArgument("Field '{}' should be of type {}, got {}", field_name,
+                           ArrowTypeString(expected_type),
+                           ArrowTypeString(view->storage_type));
   }
   return {};
 }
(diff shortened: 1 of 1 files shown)


src/iceberg/manifest/manifest_reader.cc
```diff
@@ -33,270 +33,357 @@
...
-  }                                                                              \
-  if (view->n_children != n_child) {                                             \
-    return InvalidManifest("Sub Field for:{} should have key&value:{} columns.", \
-                           field_name, n_child);                                 \
-  }
-
-std::vector<uint8_t> ArrowArrayViewGetInt8Vector(const ArrowArrayView* view,
-                                                 int32_t offset_idx) {
+template <typename... Args>
+[[nodiscard]] Status MakeError(ErrorKind error_kind, std::format_string<Args...> fmt,
```
HuaHuaY: This method is universal, and maybe it shouldn’t be placed here.

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -33,270 +33,357 @@
...
+                               Args&&... args) {
+  return std::unexpected<Error>(
+      {error_kind, std::format(fmt, std::forward<Args>(args)...)});
+}
+
+[[nodiscard]] Status MakeNullError(ErrorKind error_kind, std::string_view field_name,
```
HuaHuaY: Is it necessary to provide the first argument `error_kind`? Will any error kind except `ErrorKind::kInvalidManifestList` be used here?

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -33,270 +33,357 @@
...
+
+std::vector<uint8_t> ParseInt8VectorField(const ArrowArrayView* view,
+                                          int64_t offset_idx) {
   auto buffer = ArrowArrayViewGetBytesUnsafe(view, offset_idx);
   return {buffer.data.as_char, buffer.data.as_char + buffer.size_bytes};
 }
 
-Status ParsePartitionFieldSummaryList(ArrowArrayView* view_of_column,
+template <typename T, typename Container, typename Accessor>
+Status ParseIntegerField(const ArrowArrayView* array_view, Container& container,
```
HuaHuaY: I suggest to use a template to merge `ParseIntegerField`, `ParseStringField` and `ParseBinaryField` and simplify `Accessor` to just receive a projection.

```cpp
template <typename Container, typename Accessor, typename Transfer>
  requires std::ranges::forward_range<Container> &&
           requires(Accessor& a, Container& c, Transfer& t, const ArrowArrayView* v,
                    int64_t i) {
             std::invoke(a, *std::ranges::begin(c)) = std::invoke(t, v, i);
           }
Status ParseField(const ArrowArrayView* array_view, Container& container,
                  Accessor accessor, std::string_view field_name, bool required,
                  ErrorKind error_kind, Transfer transfer) {
  auto iter = std::ranges::begin(container);
  for (int64_t row_idx = 0; row_idx < array_view->length; row_idx++, iter++) {
    if (!ArrowArrayViewIsNull(array_view, row_idx)) {
      std::invoke(accessor, *iter) = std::invoke(transfer, array_view, row_idx);
    } else if (required) {
      return MakeNullError(error_kind, field_name, row_idx);
    }
  }
  return {};
}

template <typename T, typename Container, typename Accessor>
Status ParseIntegerField(const ArrowArrayView* array_view, Container& container,
                         Accessor accessor, std::string_view field_name, bool required,
                         ErrorKind error_kind) {
  return ParseField(array_view, container, accessor, field_name, required, error_kind,
                    [](const ArrowArrayView* view, int64_t row_idx) {
                      return static_cast<T>(ArrowArrayViewGetIntUnsafe(view, row_idx));
                    });
}

template <typename Container, typename Accessor>
Status ParseStringField(const ArrowArrayView* array_view, Container& container,
                        Accessor accessor, std::string_view field_name, bool required,
                        ErrorKind error_kind) {
  return ParseField(array_view, container, accessor, field_name, required, error_kind,
                    [](const ArrowArrayView* view, int64_t row_idx) {
                      auto value = ArrowArrayViewGetStringUnsafe(view, row_idx);
                      return std::string(value.data, value.size_bytes);
                    });
}

template <typename Container, typename Accessor>
Status ParseBinaryField(const ArrowArrayView* array_view, Container& container,
                        Accessor accessor, std::string_view field_name, bool required,
                        ErrorKind error_kind) {
  return ParseField(array_view, container, accessor, field_name, required, error_kind,
                    ParseInt8VectorField);
}
```

src/iceberg/manifest/manifest_reader.cc
```diff
@@ -33,270 +33,357 @@
...
+      auto key = ArrowArrayViewGetIntUnsafe(key_array_view, offset);
+      accessor(container, row_idx)[key] =
+          ArrowArrayViewGetIntUnsafe(value_array_view, offset);
+    }
+  }
+  return {};
+}
+
+template <typename Container, typename Accessor>
+Status ParseIntBinaryMapField(const ArrowArrayView* array_view, int64_t length,
```
HuaHuaY: ditto

src/iceberg/manifest/manifest_reader.cc
wgtmac: For now, `ErrorKind::kInvalidManifestList` is the only value passed. But this function is generic so perhaps it is ok to keep it to be consistent?

HuaHuaY: I just think that passing `ErrorKind::kInvalidManifestList` everywhere is a noise. Can we wrap this function and only use in this file?

src/iceberg/manifest/manifest_reader.cc
HuaHuaY: After offline discussion, I will open a new PR to modify this.

PR #465, opened 2025-12-30
fix: properly propagate errors in MetricsConfig validation

shangxinli:
The previous code had a logic error when checking the result of  FindFieldByName(). It used:                                                                                                                                                                                                                        
    if (!field.has_value() || !field.value().has_value())                                                                                                                                                                                            
                                                                                                                                                                                                                                                     
  This was incorrect because:                                                                                                                                                                                                                        
  1. If FindFieldByName() returns an error (Result contains an error), field.has_value() would be false, but the error was not propagated                                                                                                                                                                              
  2. field.value() returns std::optional<std::reference_wrapper<...>>,  so the second check was redundant and confusing                                                                                                                                                                                                 
                                                                                                                                                                                                                                                     
  Fixed by using ICEBERG_ASSIGN_OR_RAISE to properly propagate any errors from FindFieldByName(), then checking if the returned optional  contains a value. This matches the pattern used throughout the rest of the codebase and ensures errors are properly handled.

### DIFF
--- a/src/iceberg/metrics_config.cc
+++ b/src/iceberg/metrics_config.cc
@@ -36,8 +36,8 @@ Status MetricsConfig::VerifyReferencedColumns(
     }
     auto field_name =
         std::string_view(key).substr(TableProperties::kMetricModeColumnConfPrefix.size());
-    auto field = schema.FindFieldByName(field_name);
-    if (!field.has_value() || !field.value().has_value()) {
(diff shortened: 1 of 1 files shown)


wgtmac:
> `if (!field.has_value() || !field.value().has_value())`, the first is for `Result`, the second is for `optional`. I agree it should propagate any errors from FindFieldByName().

Agreed. I think they are different errors and should be handled differently.

PR #467, opened 2025-12-31
feat: implement location provider

WZhuo:


src/iceberg/location_provider.h
```diff
@@ -48,9 +50,57 @@ class ICEBERG_EXPORT LocationProvider {
   ///
   /// TODO(wgtmac): StructLike is not well thought yet, we may wrap an ArrowArray
```
wgtmac: Let's update the comment.

src/iceberg/location_provider.h
```diff
@@ -48,9 +50,57 @@ class ICEBERG_EXPORT LocationProvider {
...
+class ICEBERG_EXPORT LocationProviderFactory {
+ public:
+  virtual ~LocationProviderFactory() = default;
+
+  /// \brief Create a LocationProvider for the given table location and properties.
+  ///
+  /// \param input_location the table location
+  /// \param properties the table properties
+  /// \return a LocationProvider instance
+  static Result<std::unique_ptr<LocationProvider>> For(const std::string& input_location,
```
wgtmac: Use `Make` to be consistent with other APIs.

src/iceberg/expression/literal.cc
```diff
@@ -488,6 +488,7 @@ std::string Literal::ToString() const {
           .value_or("invalid literal of type decimal");
     }
     case TypeId::kString: {
+      // TODO(zhuo.wang): escape string?
```
wgtmac: IMO `ToString` is for human readable representation. So it is no need to escape it here?

src/iceberg/location_provider.h
```diff
@@ -48,9 +50,57 @@ class ICEBERG_EXPORT LocationProvider {
...
+                                                       const TableProperties& properties);
+};
+
+/// \brief DefaultLocationProvider privides default location provider for local file
+/// system.
+class ICEBERG_EXPORT DefaultLocationProvider : public LocationProvider {
+ public:
+  DefaultLocationProvider(const std::string& table_location,
```
wgtmac: Should we use `std::string_view` for table_location?

src/iceberg/location_provider.h
```diff
@@ -48,9 +50,57 @@ class ICEBERG_EXPORT LocationProvider {
...
+                          const TableProperties& properties);
+
+  std::string NewDataLocation(const std::string& filename) override;
+
+  Result<std::string> NewDataLocation(const PartitionSpec& spec,
+                                      const PartitionValues& partition_data,
+                                      const std::string& filename) override;
```
wgtmac: Same question for filename to use string_view in both `NewDataLocation` functions. We can change the base class function signature if appropriate.

src/iceberg/util/location_util.h
```diff
@@ -27,16 +27,15 @@ namespace iceberg {
...
     }
 
-    std::string_view result = path;
-    while (result.ends_with("/") && !result.ends_with("://")) {
-      result.remove_suffix(1);
+    while (path.ends_with("/") && !path.ends_with("://")) {
+      path.remove_suffix(1);
     }
-    return std::string(result);
+    return std::string(path);
```
wgtmac: It seems that we can just return a `std::string_view` and let the caller to decide which type to use to be more flexible.

src/iceberg/location_provider.cc
```diff
@@ -0,0 +1,176 @@
...
+  return DirsFromHash(hash_value);
+}
+
+}  // namespace
+
+Result<std::unique_ptr<LocationProvider>> LocationProviderFactory::For(
+    const std::string& input_location, const TableProperties& properties) {
+  std::string location = LocationUtil::StripTrailingSlash(input_location);
+
+  // Note: Not support dynamic constructor according to kWriteLocationProviderImpl
```
wgtmac: I think we need to allow users to register their own LocationProvider implementations, perhaps to a LocationProviderRegistry and we try to find it if `write.location-provider.impl` has been configured.

src/iceberg/location_provider.cc
```diff
@@ -0,0 +1,176 @@
...
+///
+/// If the low `kHashBits = 20` of `hash` is '10011001100110011001', then return
+/// '1001/1001/1001/10011001' with depth 3 and length 4.
+///
+/// \param hash The hash value to be divided.
+/// \return The path according to the `hash` value.
+std::string DirsFromHash(int32_t hash) {
+  std::string hash_with_dirs;
+
+  for (int i = 0; i < kEntropyDirDepth * kEntropyDirLength; i += kEntropyDirLength) {
```
wgtmac:
```suggestion
  for (int32_t i = 0; i < kEntropyDirDepth * kEntropyDirLength; i += kEntropyDirLength) {
```

src/iceberg/location_provider.cc
```diff
@@ -0,0 +1,176 @@
...
+                         const std::string& table_location) {
+  auto data_location = properties.Get(TableProperties::kWriteDataLocation);
+  if (data_location.empty()) {
+    data_location = std::format("{}/data", table_location);
+  }
+  return data_location;
+}
+
+std::string PathContext(std::string_view table_location) {
+  std::string path = LocationUtil::StripTrailingSlash(table_location);
```
wgtmac: If `StripTrailingSlash` returns a string_view, we can avoid creating a string object here.

src/iceberg/location_provider.cc
```diff
@@ -0,0 +1,176 @@
...
+
+  size_t last_slash = path.find_last_of('/');
+  if (last_slash != std::string::npos && last_slash < path.length() - 1) {
+    std::string_view data_path(path.data(), path.size() - last_slash - 1);
```
wgtmac:
```suggestion
    std::string_view data_path = std::string_view(path).substr(last_slash + 1);
```

src/iceberg/location_provider.h
```diff
@@ -48,9 +50,57 @@ class ICEBERG_EXPORT LocationProvider {
...
+  /// \param input_location the table location
+  /// \param properties the table properties
+  /// \return a LocationProvider instance
+  static Result<std::unique_ptr<LocationProvider>> For(const std::string& input_location,
+                                                       const TableProperties& properties);
+};
+
+/// \brief DefaultLocationProvider privides default location provider for local file
+/// system.
+class ICEBERG_EXPORT DefaultLocationProvider : public LocationProvider {
```
wgtmac: It seems that we can move these impl classes to location_provider.cc without declaring them here.

src/iceberg/partition_spec.h
```diff
@@ -64,6 +64,9 @@ class ICEBERG_EXPORT PartitionSpec : public util::Formattable {
   /// \brief Get the partition type binding to the input schema.
   Result<std::unique_ptr<StructType>> PartitionType(const Schema& schema) const;
 
+  /// \brief Get the partition path for the given partition data.
+  Result<std::string> PartitionPath(const PartitionValues& data) const;
```
wgtmac, line 68: Can we add some test cases for this? This function can also be in a separate PR if you see fit.

src/iceberg/partition_spec.cc
```diff
@@ -98,6 +100,22 @@ Result<std::unique_ptr<StructType>> PartitionSpec::PartitionType(
...
+  for (int32_t i = 0; i < fields_.size(); ++i) {
+    ICEBERG_ASSIGN_OR_RAISE(auto value, data.ValueAt(i));
+    if (i > 0) {
+      ss << "/";
+    }
+    // TODO(zhuo.wang): transform partition value to string
+    std::string partition_value = value.get().ToString();
+    // TODO(zhuo.wang): UrlEncoder::Encode for partition name and value
+    ss << fields_[i].name() << "="
+       << std::string_view(partition_value.data() + 1, partition_value.size() - 2);
```
wgtmac: Is this correct? Only string type carries quote in the `ToString()` output.

src/iceberg/test/location_provider_test.cc
```diff
@@ -0,0 +1,156 @@
...
+}
+}  // namespace
+
+// Test fixture for location provider tests
+class LocationProviderTest : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    // Setup a test table location and properties
+    table_location_ = "/test/table/location";
+    properties_ = TableProperties::FromMap({});
```
wgtmac: It seems that we don't need this line.

src/iceberg/test/location_provider_test.cc
```diff
@@ -0,0 +1,156 @@
...
+  ICEBERG_UNWRAP_OR_FAIL(auto provider,
+                         LocationProvider::Make(table_location_, properties_));
+
+  auto location = provider->NewDataLocation("test.parquet");
+  std::string relative_location = location;
+  if (relative_location.starts_with(table_location_)) {
+    relative_location = relative_location.substr(table_location_.size());
+  }
+
+  std::vector<std::string> parts = SplitString(relative_location, '/');
```
wgtmac, line 88: Why not directly comparing the `relative_location`? Is it non-deterministic?

src/iceberg/partition_spec.cc
WZhuo: The quotes added by Literal::ToString() for string types make it impossible to distinguish whether the quotes were originally part of the string or added by the method. I think we should not add extra quotes for StringType — in Java, they are not added either.

src/iceberg/test/location_provider_test.cc
WZhuo: This is a deterministic value, but since part of it is a hash value, the computed result is not known until after the calculation.

src/iceberg/partition_spec.h
WZhuo: Ok, will do

src/iceberg/partition_spec.cc
wgtmac: No, I don't think so. Java adds quote to `StringLiteral` as in https://github.com/apache/iceberg/blob/42cac92c4847407b8b56f976b83a0602d80dce0b/api/src/main/java/org/apache/iceberg/expressions/Literals.java#L607. Currently, `Literal.ToString()` is consistent with the Java impl.

wgtmac: BTW, I think we shouldn't rely on `Literal.ToString()` to create the partition path string. We might need to follow the Java impl to introduce a `Transform::ToHumanString` for this purpose.

PR #468, opened 2025-12-31
fix: return raw string for StringType literal

WZhuo:


src/iceberg/test/literal_test.cc
```diff
@@ -546,9 +546,9 @@ INSTANTIATE_TEST_SUITE_P(
                               .expected_type_id = TypeId::kDecimal,
                               .expected_string = "0"},
         BasicLiteralTestParam{.test_name = "String",
-                              .literal = Literal::String("hello world"),
+                              .literal = Literal::String("hello \" world"),
                               .expected_type_id = TypeId::kString,
-                              .expected_string = "\"hello world\""},
+                              .expected_string = "\"hello \\\" world\""},
```
wgtmac: Should it be escaped as `hello%20%22%20world`?

WZhuo: I think it's somehow like json literal. `true`, `false` for bool type, `1234` for integer and also json dump for string type, `"hello \" world"`

PR #478, opened 2026-01-02
feat: add support to read metadata columns

wgtmac:
- Support reading `_file` and `_pos`
- Support both avro and parquet files

### DIFF
--- a/src/iceberg/arrow/metadata_column_util_internal.h
+++ b/src/iceberg/arrow/metadata_column_util_internal.h
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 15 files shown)


WZhuo (approved):

src/iceberg/arrow/metadata_column_util_internal.h
```diff
@@ -0,0 +1,65 @@
...
+#include <string>
+
+#include <arrow/type_fwd.h>
+
+#include "iceberg/result.h"
+
+namespace iceberg::arrow {
+
+/// \brief Context for populating metadata columns during reading.
+struct MetadataColumnContext {
```
zhjwpku, line 36: Does it make sense to move this structure to metadata_columns.h?

zhjwpku (approved):

src/iceberg/arrow/metadata_column_util_internal.h
wgtmac: I've thought about this but finally decided not to do this. The reason is that there are more metadata columns in that file and we just need only two of them.

zhjwpku: Make sense, thanks for the explanation.

wgtmac:
Thanks all for the review!

PR #479, opened 2026-01-02
feat: add restcatalog authentication api

lishuxu:
feat: add restcatalog authentication api

### DIFF
--- a/src/iceberg/catalog/rest/auth/auth_manager.h
+++ b/src/iceberg/catalog/rest/auth/auth_manager.h
@@ -76,7 +76,7 @@ class ICEBERG_REST_EXPORT AuthManager {
   /// session is needed, or an error if session creation fails.
   virtual Result<std::shared_ptr<AuthSession>> ContextualSession(
       const std::unordered_map<std::string, std::string>& context,
-      const std::shared_ptr<AuthSession>& parent);
+      std::shared_ptr<AuthSession> parent);
 
   /// \brief Create or reuse a session scoped to a single table/view.
   ///
@@ -91,7 +91,7 @@ class ICEBERG_REST_EXPORT AuthManager {
   virtual Result<std::shared_ptr<AuthSession>> TableSession(
       const TableIdentifier& table,
       const std::unordered_map<std::string, std::string>& properties,
-      const std::shared_ptr<AuthSession>& parent);
+      std::shared_ptr<AuthSession> parent);
 
   /// \brief Release resources held by the manager.
   ///
--- a/src/iceberg/catalog/rest/auth/auth_managers.cc
+++ b/src/iceberg/catalog/rest/auth/auth_managers.cc
@@ -33,13 +33,7 @@ namespace {
 using AuthManagerRegistry =
     std::unordered_map<std::string, AuthManagerFactory, StringHash, StringEqual>;
 
-/// \brief Infer the authentication type from properties.
-///
-/// If no explicit auth type is set, this function tries to infer it from
-/// other properties. If "credential" or "token" is present, it implies
-/// OAuth2 authentication. Otherwise, defaults to no authentication.
-///
-/// This behavior is consistent with Java Iceberg's AuthManagers.
+// Infer the authentication type from properties.
 std::string InferAuthType(
     const std::unordered_map<std::string, std::string>& properties) {
   auto it = properties.find(AuthProperties::kAuthType);
@@ -54,11 +48,10 @@ std::string InferAuthType(
     return AuthProperties::kAuthTypeOAuth2;
   }
 
-  // Default to no authentication
   return AuthProperties::kAuthTypeNone;
 }
 
-/// \brief Get the global registry of auth manager factories.
+// Get the global registry of auth manager factories.
 AuthManagerRegistry& GetRegistry() {
   static AuthManagerRegistry registry;
   return registry;
--- a/src/iceberg/catalog/rest/auth/auth_managers.h
+++ b/src/iceberg/catalog/rest/auth/auth_managers.h
@@ -57,6 +57,9 @@ class ICEBERG_REST_EXPORT AuthManagers {
 
   /// \brief Register or override the factory for a given auth type.
   ///
+  /// This method is not thread-safe. All registrations should be done during
+  /// application startup before any concurrent access to Load().
+  ///
   /// \param auth_type Case-insensitive type identifier (e.g., "basic").
   /// \param factory Factory function that produces the manager.
   static void Register(std::string_view auth_type, AuthManagerFactory factory);
--- a/src/iceberg/catalog/rest/auth/auth_session.cc
+++ b/src/iceberg/catalog/rest/auth/auth_session.cc
@@ -44,7 +44,7 @@ class DefaultAuthSession : public AuthSession {
 
 }  // namespace
 
-std::shared_ptr<AuthSession> AuthSession::Make(
+std::shared_ptr<AuthSession> AuthSession::MakeDefault(
     std::unordered_map<std::string, std::string> headers) {
   return std::make_shared<DefaultAuthSession>(std::move(headers));
 }
--- a/src/iceberg/catalog/rest/auth/auth_session.h
+++ b/src/iceberg/catalog/rest/auth/auth_session.h
@@ -60,7 +60,7 @@ class ICEBERG_REST_EXPORT AuthSession {
   /// \return Status indicating success or failure of closing the session.
   virtual Status Close() { return {}; }
 
-  /// \brief Create a session with static headers.
+  /// \brief Create a default session with static headers.
   ///
   /// This factory method creates a session that adds a fixed set of headers to each
   /// request. It is suitable for authentication methods that use static credentials,
@@ -68,7 +68,7 @@ class ICEBERG_REST_EXPORT AuthSession {
   ///
   /// \param headers The headers to add to each request for authentication.
   /// \return A new session that adds the given headers to requests.
-  static std::shared_ptr<AuthSession> Make(
+  static std::shared_ptr<AuthSession> MakeDefault(
       std::unordered_map<std::string, std::string> headers);
 };
 
--- a/src/iceberg/catalog/rest/auth/auth_manager.cc
+++ b/src/iceberg/catalog/rest/auth/auth_manager.cc
@@ -32,7 +32,7 @@ Result<std::shared_ptr<AuthSession>> AuthManager::InitSession(
 
 Result<std::shared_ptr<AuthSession>> AuthManager::ContextualSession(
     [[maybe_unused]] const std::unordered_map<std::string, std::string>& context,
-    const std::shared_ptr<AuthSession>& parent) {
+    std::shared_ptr<AuthSession> parent) {
   // By default, return the parent session as-is
   return parent;
 }
@@ -40,7 +40,7 @@ Result<std::shared_ptr<AuthSession>> AuthManager::ContextualSession(
 Result<std::shared_ptr<AuthSession>> AuthManager::TableSession(
     [[maybe_unused]] const TableIdentifier& table,
     [[maybe_unused]] const std::unordered_map<std::string, std::string>& properties,
-    const std::shared_ptr<AuthSession>& parent) {
+    std::shared_ptr<AuthSession> parent) {
   // By default, return the parent session as-is
   return parent;
 }


src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+  /// The provided HTTP client is a short-lived client; it should only be used
+  /// to fetch initial credentials if required, and must be discarded after that.
+  ///
+  /// This method cannot return null. By default, it returns the catalog session.
+  ///
+  /// \param init_client A short-lived HTTP client for initialization.
+  /// \param properties Configuration properties.
+  /// \return A session for initialization, or the catalog session by default.
+  virtual std::shared_ptr<AuthSession> InitSession(
+      HttpClient* init_client,
```
HeartLinked: why use raw pointer？🤔 use std::shared_ptr instead?

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+  ///
+  /// This method cannot return null.
+  ///
+  /// It is not required to cache the returned session internally, as the catalog
+  /// will keep it alive for the lifetime of the catalog.
+  ///
+  /// \param shared_client A long-lived, shared HTTP client.
+  /// \param properties Configuration properties (merged with server config).
+  /// \return A session for catalog operations.
+  virtual std::shared_ptr<AuthSession> CatalogSession(
```
HeartLinked: I found these methods in java implementation may all throw exceptions, so return Result<...>?

src/iceberg/catalog/rest/auth/auth_manager.h
lishuxu: AuthManager simply borrows the HTTP client owned by the catalog, it isn’t supposed to manage that client’s lifetime, so here use raw pointer.

src/iceberg/catalog/rest/auth/auth_manager.h
lishuxu: Good catch! You're right that the Java implementations can throw exceptions (e.g., BasicAuthManager throws IllegalArgumentException when credentials are missing).

src/iceberg/catalog/rest/auth/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set(ICEBERG_REST_AUTH_SOURCES auth_session.cc auth_manager.cc auth_managers.cc)
+
+# Expose sources to parent scope for inclusion in iceberg_rest library
+set(ICEBERG_REST_AUTH_SOURCES
+    ${ICEBERG_REST_AUTH_SOURCES}
+    PARENT_SCOPE)
```
wgtmac:
```suggestion
iceberg_install_all_headers(iceberg/catalog/rest/auth)
```

The current pattern is inconsistent with other CMake config. We should only install headers for this subdirectory here.

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -26,6 +29,11 @@ set(ICEBERG_REST_SOURCES
     rest_util.cc
     types.cc)
 
+# Add auth sources with proper path prefix
+foreach(AUTH_SOURCE ${ICEBERG_REST_AUTH_SOURCES})
+  list(APPEND ICEBERG_REST_SOURCES "auth/${AUTH_SOURCE}")
+endforeach()
+
```
```suggestion
```

Let's remove these lines by directly adding those files in the above `set(ICEBERG_REST_SOURCES)`.

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -15,6 +15,9 @@
 # specific language governing permissions and limitations
 # under the License.
 
+# Include auth subdirectory
```
```suggestion
```

This comment is unnecessary.

src/iceberg/catalog/rest/type_fwd.h
```diff
@@ -34,3 +34,11 @@ class RestCatalog;
 class RestCatalogProperties;
 
 }  // namespace iceberg::rest
+
+namespace iceberg::rest::auth {
```
wgtmac, line 38: Do we really need the extra `auth` namespace?

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,91 @@
...
+
+namespace {
+
+/// \brief Convert a string to lowercase for case-insensitive comparison.
+std::string ToLower(std::string_view str) {
+  std::string result(str);
+  std::ranges::transform(result, result.begin(),
+                         [](unsigned char c) { return std::tolower(c); });
+  return result;
+}
```
WZhuo: There is a `StringUtils::ToLower` func in `util/string_util.h`

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -0,0 +1,65 @@
...
+class EmptyAuthSession : public AuthSession {
+ public:
+  void Authenticate(
+      [[maybe_unused]] std::unordered_map<std::string, std::string>& headers) override {
+    // No-op: empty session does not add any authentication
+  }
+
+  void Close() override {
+    // No resources to release
+  }
```
WZhuo: Can be removed if do nothing.

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+#include "iceberg/catalog/rest/auth/auth_session.h"
+#include "iceberg/catalog/rest/iceberg_rest_export.h"
+#include "iceberg/result.h"
+#include "iceberg/table_identifier.h"
+
+/// \file iceberg/catalog/rest/auth/auth_manager.h
+/// \brief Authentication manager interface for REST catalog.
+
+namespace iceberg::rest {
+class HttpClient;
```
wgtmac: Remove this by include `type_fwd.h`.

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "iceberg/catalog/rest/auth/auth_session.h"
```
wgtmac: Remove this to use forward declaration.

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+
+/// \brief Manager for authentication sessions.
+///
+/// This interface is used to create sessions for the catalog, tables/views,
+/// and any other context that requires authentication.
+///
+/// Managers are typically stateful and may require initialization and cleanup.
+/// The manager is created by the catalog and is closed when the catalog is closed.
+///
+/// This interface is modeled after Java Iceberg's AuthManager interface.
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/auth/auth_manager.h
wgtmac: Perhaps it is better to use `HttpClient&` to save the null check?

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+  /// close it when Close() is called.
+  ///
+  /// It is not required to cache the returned session internally, as the catalog
+  /// will keep it alive for the lifetime of the catalog.
+  ///
+  /// \param shared_client A long-lived, shared HTTP client.
+  /// \param properties Configuration properties (merged with server config).
+  /// \return A session for catalog operations, or an error if session creation fails
+  ///         (e.g., missing required credentials, network failure during token fetch).
+  virtual Result<std::shared_ptr<AuthSession>> CatalogSession(
```
wgtmac, line 64: ditto

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+  /// should manage their lifecycle and close them when they are no longer needed.
+  ///
+  /// \param table The table identifier.
+  /// \param properties Properties returned by the table/view endpoint.
+  /// \param parent The parent session (typically the catalog session).
+  /// \return A session for the table, or an error if session creation fails.
+  virtual Result<std::shared_ptr<AuthSession>> TableSession(
+      const TableIdentifier& table,
+      const std::unordered_map<std::string, std::string>& properties,
+      std::shared_ptr<AuthSession> parent);
```
wgtmac: Use `AuthSession*` for parent?

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+
+  /// \brief Close the manager and release any resources.
+  ///
+  /// This method is called when the owning catalog is closed. Implementations
+  /// should release any resources held by the manager, such as cached sessions
+  /// or background threads.
+  virtual void Close() {}
```
wgtmac:
```suggestion
  virtual Status Close() { return {}; }
```

Let's return Status just in case.

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,122 @@
...
+  ///
+  /// Implementors should cache table sessions internally, as the catalog will not
+  /// cache them. Also, the owning catalog never closes table sessions; implementations
+  /// should manage their lifecycle and close them when they are no longer needed.
+  ///
+  /// \param table The table identifier.
+  /// \param properties Properties returned by the table/view endpoint.
+  /// \param parent The parent session (typically the catalog session).
+  /// \return A session for the table, or an error if session creation fails.
+  virtual Result<std::shared_ptr<AuthSession>> TableSession(
```
wgtmac, line 91: Is it better to return `Result<std::unique_ptr<AuthSession>>` from all these functions? IMO, the created session is not supposed to be shared by different catalogs or tables. In most cases, they are short-lived. The main blocker might be here because by default parent session should be returned as is. Should we change the signature as:

```cpp
virtual Result<std::unique_ptr<AuthSession>> TableSession(
      const TableIdentifier& table_identifier,
      const std::unordered_map<std::string, std::string>& properties,
      std::unique_ptr<AuthSession>&& parent);
```

src/iceberg/catalog/rest/auth/auth_properties.h
```diff
@@ -0,0 +1,84 @@
...
+
+  /// \brief Property key for SigV4 region.
+  static constexpr std::string_view kSigV4Region = "rest.auth.sigv4.region";
+
+  /// \brief Property key for SigV4 service name.
+  static constexpr std::string_view kSigV4Service = "rest.auth.sigv4.service";
+
+  /// \brief Property key for SigV4 delegate auth type.
+  static constexpr std::string_view kSigV4DelegateAuthType =
+      "rest.auth.sigv4.delegate-auth-type";
```
wgtmac, line 70:
```suggestion
  /// \brief Property key for specifying the authentication type.
  static constexpr std::string_view kAuthType = "rest.auth.type";
  /// \brief Authentication type: no authentication.
  static constexpr std::string_view kAuthTypeNone = "none";
  /// \brief Authentication type: HTTP Basic authentication.
  static constexpr std::string_view kAuthTypeBasic = "basic";
  /// \brief Authentication type: OAuth2 authentication.
  static constexpr std::string_view kAuthTypeOAuth2 = "oauth2";
  /// \brief Authentication type: AWS SigV4 authentication.
  static constexpr std::string_view kAuthTypeSigV4 = "sigv4";

  /// \brief Property key for Basic auth username.
  static constexpr std::string_view kBasicUsername = "rest.auth.basic.username";
  /// \brief Property key for Basic auth password.
  static constexpr std::string_view kBasicPassword = "rest.auth.basic.password";

  /// \brief Property key for OAuth2 token (bearer token).
  static constexpr std::string_view kOAuth2Token = "token";
  /// \brief Property key for OAuth2 credential (client_id:client_secret).
  static constexpr std::string_view kOAuth2Credential = "credential";
  /// \brief Property key for OAuth2 scope.
  static constexpr std::string_view kOAuth2Scope = "scope";
  /// \brief Property key for OAuth2 server URI.
  static constexpr std::string_view kOAuth2ServerUri = "oauth2-server-uri";
  /// \brief Property key for enabling token refresh.
  static constexpr std::string_view kOAuth2TokenRefreshEnabled = "token-refresh-enabled";
  /// \brief Default OAuth2 scope for catalog operations.
  static constexpr std::string_view kOAuth2DefaultScope = "catalog";

  /// \brief Property key for SigV4 region.
  static constexpr std::string_view kSigV4Region = "rest.auth.sigv4.region";
  /// \brief Property key for SigV4 service name.
  static constexpr std::string_view kSigV4Service = "rest.auth.sigv4.service";
  /// \brief Property key for SigV4 delegate auth type.
  static constexpr std::string_view kSigV4DelegateAuthType =
      "rest.auth.sigv4.delegate-auth-type";
```

Let's put properties in the same catalog more compact?

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,88 @@
...
+
+namespace iceberg::rest::auth {
+
+/// \brief Factory function type for creating AuthManager instances.
+///
+/// \param name The name of the manager (used for logging).
+/// \param properties Configuration properties.
+/// \return A unique pointer to the created AuthManager.
+using AuthManagerFactory = std::function<std::unique_ptr<AuthManager>(
+    const std::string& name,
```
wgtmac:
```suggestion
    std::string_view name,
```

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,88 @@
...
+      const std::unordered_map<std::string, std::string>& properties);
+
+  /// \brief Register a custom authentication manager factory.
+  ///
+  /// This allows users to extend the supported authentication types by
+  /// registering their own AuthManager implementations.
+  ///
+  /// \param auth_type The authentication type name (e.g., "custom").
+  /// \param factory The factory function to create the AuthManager.
+  static void Register(const std::string& auth_type, AuthManagerFactory factory);
```
wgtmac:
```suggestion
  static void Register(std::string_view auth_type, AuthManagerFactory factory);
```

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,88 @@
...
+  ///
+  /// If no auth type is specified, the method will infer the type based on
+  /// other properties (e.g., presence of "credential" or "token" implies oauth2).
+  /// If no auth-related properties are found, it defaults to "none".
+  ///
+  /// \param name A name for the manager (used for logging).
+  /// \param properties Configuration properties.
+  /// \return A unique pointer to the created AuthManager, or an error.
+  static Result<std::unique_ptr<AuthManager>> Load(
+      const std::string& name,
```
wgtmac:
```suggestion
      std::string_view name,
```

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,88 @@
...
+    const std::string& name,
+    const std::unordered_map<std::string, std::string>& properties)>;
+
+/// \brief Factory class for loading authentication managers.
+///
+/// This class provides a registry-based approach to create AuthManager instances
+/// based on the configured authentication type. It supports built-in types
+/// (none, basic, oauth2) and allows registration of custom types.
+///
+/// This class is modeled after Java Iceberg's AuthManagers class.
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+
+namespace iceberg::rest::auth {
+
+/// \brief An authentication session that can authenticate outgoing HTTP requests.
+///
+/// Authentication sessions are typically immutable, but may hold resources that need
+/// to be released when the session is no longer needed (e.g., token refresh threads).
+/// Implementations should override Close() to release any such resources.
+///
+/// This interface is modeled after Java Iceberg's AuthSession interface.
```
wgtmac:
```suggestion
```

Let's remove comment like this.

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+class ICEBERG_REST_EXPORT AuthSession {
+ public:
+  virtual ~AuthSession() = default;
+
+  /// \brief Authenticate the given request headers.
+  ///
+  /// This method adds authentication information (e.g., Authorization header)
+  /// to the provided headers map. The implementation should be idempotent.
+  ///
+  /// \param headers The headers map to add authentication information to.
```
wgtmac:
```suggestion
  /// \param[in,out] headers The headers map to add authentication information to.
```

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+  virtual void Authenticate(std::unordered_map<std::string, std::string>& headers) = 0;
```
wgtmac:
```suggestion
  virtual Status Authenticate(std::unordered_map<std::string, std::string>& headers) = 0;
```

Let's return status just in case.

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+  /// \brief Close the session and release any resources.
+  ///
+  /// This method is called when the session is no longer needed. For stateful
+  /// sessions (e.g., OAuth2 with token refresh), this should stop any background
+  /// threads and release resources.
+  ///
+  /// Note: Since sessions may be cached, this method may not be called immediately
+  /// after the session is no longer needed, but rather when the session is evicted
+  /// from the cache or the cache itself is closed.
+  virtual void Close() {}
```
wgtmac: Return status

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+
+  /// \brief Get a shared pointer to an empty session that does nothing.
+  ///
+  /// The empty session is a singleton that simply returns the request unchanged.
+  /// It is useful as a default or placeholder session.
+  ///
+  /// \return A shared pointer to the empty session singleton.
+  static std::shared_ptr<AuthSession> Empty();
```
wgtmac: Return a singleton？ But if we return std::unique_ptr by AuthManager, here we can only return unique_ptr.

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+
+  /// \brief Create a DefaultAuthSession with the given headers.
+  ///
+  /// \param headers The headers to add to each request.
+  /// \return A shared pointer to the new session.
+  static std::shared_ptr<DefaultAuthSession> Of(
+      std::unordered_map<std::string, std::string> headers);
+
+ private:
+  std::unordered_map<std::string, std::string> headers_;
```
wgtmac: Why do we need this?

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+  /// (i.e., request headers take precedence).
+  ///
+  /// \param headers The headers map to add authentication information to.
+  void Authenticate(std::unordered_map<std::string, std::string>& headers) override;
+
+  /// \brief Create a DefaultAuthSession with the given headers.
+  ///
+  /// \param headers The headers to add to each request.
+  /// \return A shared pointer to the new session.
+  static std::shared_ptr<DefaultAuthSession> Of(
```
wgtmac: Let's be consistent to use `Make` as the function name.

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,105 @@
...
+
+  ~DefaultAuthSession() override = default;
+
+  /// \brief Add the configured headers to the request.
+  ///
+  /// Headers are added only if they don't already exist in the request
+  /// (i.e., request headers take precedence).
+  ///
+  /// \param headers The headers map to add authentication information to.
+  void Authenticate(std::unordered_map<std::string, std::string>& headers) override;
```
wgtmac: Do you think it is worth adding a class HttpRequest? We don't have it yet.

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,91 @@
...
+
+/// \brief Infer the authentication type from properties.
+///
+/// If no explicit auth type is set, this function tries to infer it from
+/// other properties. If "credential" or "token" is present, it implies
+/// OAuth2 authentication. Otherwise, defaults to no authentication.
+///
+/// This behavior is consistent with Java Iceberg's AuthManagers.
+std::string InferAuthType(
```
wgtmac, line 37: Should we add a enum class for the auth type?

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  // Default to no authentication
+  return std::string(AuthProperties::kAuthTypeNone);
+}
+
+}  // namespace
+
+std::unordered_map<std::string, AuthManagerFactory>& AuthManagers::GetRegistry() {
+  static std::unordered_map<std::string, AuthManagerFactory> registry;
```
wgtmac:
```suggestion
  static std::unordered_map<std::string, AuthManagerFactory, StringHash, StringEqual> registry;
```

Let's support heterogeneous lookup

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,88 @@
...
+  /// This allows users to extend the supported authentication types by
+  /// registering their own AuthManager implementations.
+  ///
+  /// \param auth_type The authentication type name (e.g., "custom").
+  /// \param factory The factory function to create the AuthManager.
+  static void Register(const std::string& auth_type, AuthManagerFactory factory);
+
+ private:
+  /// \brief Get the global registry of auth manager factories.
+  static std::unordered_map<std::string, AuthManagerFactory>& GetRegistry();
```
wgtmac: We don't need this function in the header file, right?

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,88 @@
...
+  /// - "none": NoopAuthManager (no authentication)
+  /// - "basic": BasicAuthManager (HTTP Basic authentication)
+  /// - "oauth2": OAuth2AuthManager (OAuth2 authentication)
+  /// - "sigv4": SigV4AuthManager (AWS Signature V4)
+  ///
+  /// If no auth type is specified, the method will infer the type based on
+  /// other properties (e.g., presence of "credential" or "token" implies oauth2).
+  /// If no auth-related properties are found, it defaults to "none".
+  ///
+  /// \param name A name for the manager (used for logging).
```
wgtmac:
```suggestion
  /// \param name A name for the manager.
```

src/iceberg/catalog/rest/type_fwd.h
lishuxu: Here mirrors the Java package structure (org.apache.iceberg.rest.auth), making it easier to navigate between implementations.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -0,0 +1,65 @@
...
+
+DefaultAuthSession::DefaultAuthSession(
+    std::unordered_map<std::string, std::string> headers)
+    : headers_(std::move(headers)) {}
+
+void DefaultAuthSession::Authenticate(
+    std::unordered_map<std::string, std::string>& headers) {
+  for (const auto& [key, value] : headers_) {
+    headers.try_emplace(key, value);
+  }
```
HuaHuaY:
```suggestion
  headers.merge(headers_);
```

src/iceberg/catalog/rest/auth/auth_session.h
lishuxu: The `headers_` member stores pre-computed authentication credentials (e.g., "Authorization: Basic <base64>") so they can be added to HTTP request when Authenticate() method is called

src/iceberg/catalog/rest/auth/auth_session.h
lishuxu: I don't think we need an HttpRequest class at this stage. The current design is sufficient for most auth schemes (Basic, Bearer, OAuth2, API keys), but we can add can add it later if needed.

src/iceberg/catalog/rest/auth/auth_managers.cc
lishuxu: Auth types come from external config files/properties as strings, so the string-based approach allows users to register custom auth types without modifying core code. It also matches the Iceberg REST spec.

src/iceberg/catalog/rest/auth/auth_session.cc
lishuxu: merge() is destructive - it moves nodes from headers_ to headers, leaving headers_ empty or partially empty.
so, I think the current implementation can be retained.

src/iceberg/catalog/rest/auth/auth_manager.h
lishuxu: Updated.

src/iceberg/catalog/rest/auth/auth_managers.cc
lishuxu: Good catch, fixed.

src/iceberg/catalog/rest/auth/auth_session.h
lishuxu: Currently we don't need this, I removed it.

src/iceberg/catalog/rest/CMakeLists.txt
```diff
@@ -24,7 +26,10 @@ set(ICEBERG_REST_SOURCES
     resource_paths.cc
     rest_catalog.cc
     rest_util.cc
-    types.cc)
+    types.cc
+    auth/auth_manager.cc
```
wgtmac: Sort them alphabetically

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,95 @@
...
+#pragma once
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "iceberg/catalog/rest/iceberg_rest_export.h"
+#include "iceberg/catalog/rest/type_fwd.h"
+#include "iceberg/result.h"
+#include "iceberg/table_identifier.h"
```
wgtmac:
```suggestion
```

We can remove this

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,95 @@
...
+namespace iceberg::rest::auth {
+
+/// \brief Produces authentication sessions for catalog and table requests.
+///
+/// AuthManager is responsible for creating authentication sessions at different scopes:
+/// - InitSession: Short-lived session for catalog initialization (optional)
+/// - CatalogSession: Long-lived session for catalog-level operations (required)
+/// - TableSession: Optional table-specific session or reuse of catalog session
+///
+/// Implementations are registered via AuthManagers::Register() and loaded by auth type.
```
wgtmac:
```suggestion
```

We don't need such verbose comments.

src/iceberg/catalog/rest/type_fwd.h
```diff
@@ -34,3 +34,11 @@ class RestCatalog;
...
+
+class AuthManager;
+class AuthSession;
+class DefaultAuthSession;
```
wgtmac: Why do we need to forward declare `DefaultAuthSession`? Isn't it used only by the implementation?

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,94 @@
...
+
+  /// \brief Authenticate the given request headers.
+  ///
+  /// This method adds authentication information (e.g., Authorization header)
+  /// to the provided headers map. The implementation should be idempotent.
+  ///
+  /// \param[in,out] headers The headers map to add authentication information to.
+  /// \return Status indicating success or failure of authentication.
+  ///         - Success: Returns Status::OK
+  ///         - Failure: Returns one of the following errors:
```
wgtmac:
```suggestion
  /// \return Status indicating success or one of the following errors:
```

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,94 @@
...
+/// \brief Authentication session interface for REST catalog.
+
+namespace iceberg::rest::auth {
+
+/// \brief An authentication session that can authenticate outgoing HTTP requests.
+///
+/// Authentication sessions are typically immutable, but may hold resources that need
+/// to be released when the session is no longer needed (e.g., token refresh threads).
+/// Implementations should override Close() to release any such resources.
+///
```
wgtmac:
```suggestion
```

They are duplicate in the comment for Close.

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,94 @@
...
+  /// \return Status indicating success or failure of closing the session.
+  virtual Status Close() { return {}; }
+};
+
+/// \brief A default authentication session that adds static headers to requests.
+///
+/// This implementation authenticates requests by adding a fixed set of headers.
+/// It is suitable for authentication methods that use static credentials,
+/// such as Basic auth or static bearer tokens.
+class ICEBERG_REST_EXPORT DefaultAuthSession : public AuthSession {
```
wgtmac: Is it better to just define it in the auth_session.cc and use a factory to create it?

src/iceberg/catalog/rest/auth/auth_properties.h
```diff
@@ -0,0 +1,69 @@
...
+  static constexpr std::string_view kAuthType = "rest.auth.type";
+  /// \brief Authentication type: no authentication.
+  static constexpr std::string_view kAuthTypeNone = "none";
+  /// \brief Authentication type: HTTP Basic authentication.
+  static constexpr std::string_view kAuthTypeBasic = "basic";
+  /// \brief Authentication type: OAuth2 authentication.
+  static constexpr std::string_view kAuthTypeOAuth2 = "oauth2";
+  /// \brief Authentication type: AWS SigV4 authentication.
+  static constexpr std::string_view kAuthTypeSigV4 = "sigv4";
+  /// \brief Property key for Basic auth username.
```
wgtmac, line 46:
```suggestion

  /// \brief Property key for Basic auth username.
```

src/iceberg/catalog/rest/auth/auth_properties.h
```diff
@@ -0,0 +1,69 @@
...
+  static constexpr std::string_view kBasicUsername = "rest.auth.basic.username";
+  /// \brief Property key for Basic auth password.
+  static constexpr std::string_view kBasicPassword = "rest.auth.basic.password";
+  /// \brief Property key for OAuth2 token (bearer token).
+  static constexpr std::string_view kOAuth2Token = "token";
```
wgtmac:
```suggestion

  static constexpr std::string_view kOAuth2Token = "token";
```

src/iceberg/catalog/rest/auth/auth_properties.h
```diff
@@ -0,0 +1,69 @@
...
+  static constexpr std::string_view kOAuth2Credential = "credential";
+  /// \brief Property key for OAuth2 scope.
+  static constexpr std::string_view kOAuth2Scope = "scope";
+  /// \brief Property key for OAuth2 server URI.
+  static constexpr std::string_view kOAuth2ServerUri = "oauth2-server-uri";
+  /// \brief Property key for enabling token refresh.
+  static constexpr std::string_view kOAuth2TokenRefreshEnabled = "token-refresh-enabled";
+  /// \brief Default OAuth2 scope for catalog operations.
+  static constexpr std::string_view kOAuth2DefaultScope = "catalog";
+  /// \brief Property key for SigV4 region.
```
wgtmac, line 64:
```suggestion

  /// \brief Property key for SigV4 region.
```

src/iceberg/catalog/rest/auth/auth_properties.h
wgtmac: Let's add blank line to separate different categories.

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,70 @@
...
+///
+/// \param name Catalog name passed to the manager.
+/// \param properties Consolidated catalog configuration.
+/// \return Newly created manager instance.
+using AuthManagerFactory = std::function<std::unique_ptr<AuthManager>(
+    std::string_view name,
+    const std::unordered_map<std::string, std::string>& properties)>;
+
+/// \brief Registry type for AuthManager factories with heterogeneous lookup support.
+using AuthManagerRegistry =
```
wgtmac: We can move this line to the cc file because users do not need to know this.

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,70 @@
...
+#include "iceberg/catalog/rest/iceberg_rest_export.h"
+#include "iceberg/result.h"
+#include "iceberg/util/string_util.h"
+
+/// \file iceberg/catalog/rest/auth/auth_managers.h
+/// \brief Factory for creating authentication managers.
+
+namespace iceberg::rest::auth {
+
+/// \brief Function that builds an AuthManager for a given catalog.
```
wgtmac:
```suggestion
/// \brief Function that creates an AuthManager from its name.
```

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,70 @@
...
+///
+/// \param name Catalog name passed to the manager.
```
wgtmac:
```suggestion
/// \param name Name of the auth manager.
```

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,70 @@
...
+/// \param properties Consolidated catalog configuration.
```
wgtmac:
```suggestion
/// \param properties Properties required by the auth manager.
```

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,70 @@
...
+/// \return Newly created manager instance.
+using AuthManagerFactory = std::function<std::unique_ptr<AuthManager>(
```
wgtmac: Should the factory return `Result<std::unique_ptr<AuthManager>>` just in case of any error or implementation is unavailable?

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,70 @@
...
+/// \brief Registry type for AuthManager factories with heterogeneous lookup support.
+using AuthManagerRegistry =
+    std::unordered_map<std::string, AuthManagerFactory, StringHash, StringEqual>;
+
+/// \brief Registry-backed factory for AuthManager implementations.
+class ICEBERG_REST_EXPORT AuthManagers {
+ public:
+  /// \brief Load a manager by consulting the "rest.auth.type" configuration.
+  ///
+  /// \param name Catalog name passed to the manager.
```
wgtmac: Is this correct? Why it is catalog name?

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,85 @@
...
+    const std::unordered_map<std::string, std::string>& properties) {
+  // Check for explicit auth type first
+  auto it = properties.find(std::string(AuthProperties::kAuthType));
+  if (it != properties.end() && !it->second.empty()) {
+    return StringUtils::ToLower(it->second);
+  }
+
+  // Infer from OAuth2 properties (credential or token)
+  bool has_credential =
+      properties.contains(std::string(AuthProperties::kOAuth2Credential));
```
wgtmac: Do we need to define them as const std::string? It creates a string every time it is used.

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,85 @@
...
+
+Result<std::unique_ptr<AuthManager>> AuthManagers::Load(
+    std::string_view name,
+    const std::unordered_map<std::string, std::string>& properties) {
+  std::string auth_type = InferAuthType(properties);
+
+  auto& registry = GetRegistry();
+  auto it = registry.find(auth_type);
+  if (it == registry.end()) {
+    return NotImplemented("Authentication type '{}' is not supported", auth_type);
```
wgtmac, line 75: Please add a TODO to say we need to fallback to our default auth manager implementations.

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,85 @@
...
+/// \brief Get the global registry of auth manager factories.
+AuthManagerRegistry& GetRegistry() {
+  static AuthManagerRegistry registry;
+  return registry;
+}
+
+}  // namespace
+
+void AuthManagers::Register(std::string_view auth_type, AuthManagerFactory factory) {
+  GetRegistry()[StringUtils::ToLower(std::string(auth_type))] = std::move(factory);
```
wgtmac: It seems that we have unnecessary string allocations here.

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,95 @@
...
+
+  /// \brief Create or reuse a session scoped to a single table/view.
+  ///
+  /// This method can return a new table-specific session or indicate that the parent
+  /// catalog session should be reused by returning nullptr.
+  ///
+  /// \param table Target table identifier.
+  /// \param properties Table-specific auth properties returned by the server.
+  /// \param parent Catalog session to inherit from or extract information from.
+  /// \return A new session for the table, nullptr to reuse parent session, or an error.
```
wgtmac: It looks weird to return nullptr to reuse parent session. I'm not sure if it is a good idea to use `std::unique_ptr<AuthSession>&& parent` to make the ownership explicit: either directly return `parent` or the created session owns its parent. I didn't any case to share the session, am I missing something?

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,95 @@
...
+  ///
+  /// This session is used for all catalog-level operations (list namespaces, list tables,
+  /// etc.) and serves as the parent session for table-specific operations. It is owned
+  /// by the catalog and reused throughout the catalog's lifetime.
+  ///
+  /// \param shared_client HTTP client owned by the catalog and reused for auth calls.
+  /// \param properties Catalog properties (client config + server defaults).
+  /// \return Session for catalog operations or an error if authentication cannot be set
+  /// up.
+  virtual Result<std::unique_ptr<AuthSession>> CatalogSession(
```
wgtmac: Why we don't have contextual session as the Java impl? I think this is used by each rest endpoint before sending the http request.

src/iceberg/catalog/rest/auth/auth_manager.h
lishuxu: Good point! I will change the return value to Result<std::optional<std::unique_ptr<AuthSession>>>. Now std::nullopt explicitly means "reuse parent session", which is clearer than returning nullptr.
The catalog session can't be transferred via unique_ptr&& because it's shared across multiple table operations throughout the catalog's lifetime.

src/iceberg/catalog/rest/auth/auth_manager.h
lishuxu: Good catch! Added ContextualSession method. In Java, this is used by SessionCatalog to create sessions for different contexts (users/tenants).

src/iceberg/catalog/rest/auth/auth_manager.h
wgtmac: After double checking, I figured out that parent session can indeed be the catalog session. Do you think it will be much simpler to use `std::shared_ptr<AuthSession>` for the return and input types?

src/iceberg/catalog/rest/auth/auth_managers.h
lishuxu: I meant the auth type, not the catalog itself.
To avoid ambiguity, I have changed it to “Name of the auth manager.”

src/iceberg/catalog/rest/auth/auth_manager.h
lishuxu: Changed the return value to shared_ptr so that ContextualSession/TableSession can directly return the parent session when no specific session is needed, the trade-off is slightly higher overhead from atomic reference counting.

src/iceberg/catalog/rest/auth/auth_managers.cc
lishuxu: Agreed, updated.

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,102 @@
...
+  /// This method is used by SessionCatalog to create sessions for different contexts
+  /// (e.g., different users or tenants).
+  ///
+  /// \param context Context properties (e.g., user credentials, tenant info).
+  /// \param parent Catalog session to inherit from or return as-is.
+  /// \return A context-specific session, or the parent session if no context-specific
+  /// session is needed, or an error if session creation fails.
+  virtual Result<std::shared_ptr<AuthSession>> ContextualSession(
+      const std::unordered_map<std::string, std::string>& context,
+      const std::shared_ptr<AuthSession>& parent);
```
wgtmac:
```suggestion
      std::shared_ptr<AuthSession> parent);
```

src/iceberg/catalog/rest/auth/auth_manager.h
```diff
@@ -0,0 +1,102 @@
...
+  ///
+  /// \param table Target table identifier.
+  /// \param properties Table-specific auth properties returned by the server.
+  /// \param parent Catalog or contextual session to inherit from or return as-is.
+  /// \return A table-specific session, or the parent session if no table-specific
+  /// session is needed, or an error if session creation fails.
+  virtual Result<std::shared_ptr<AuthSession>> TableSession(
+      const TableIdentifier& table,
+      const std::unordered_map<std::string, std::string>& properties,
+      const std::shared_ptr<AuthSession>& parent);
```
wgtmac:
```suggestion
      std::shared_ptr<AuthSession> parent);
```

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,88 @@
...
+using AuthManagerRegistry =
+    std::unordered_map<std::string, AuthManagerFactory, StringHash, StringEqual>;
+
+/// \brief Infer the authentication type from properties.
+///
+/// If no explicit auth type is set, this function tries to infer it from
+/// other properties. If "credential" or "token" is present, it implies
+/// OAuth2 authentication. Otherwise, defaults to no authentication.
+///
+/// This behavior is consistent with Java Iceberg's AuthManagers.
```
wgtmac:
```suggestion
// Infer the authentication type from properties.
```

wgtmac: Let's avoid verbose comments.

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  // Infer from OAuth2 properties (credential or token)
+  bool has_credential = properties.contains(AuthProperties::kOAuth2Credential);
+  bool has_token = properties.contains(AuthProperties::kOAuth2Token);
+  if (has_credential || has_token) {
+    return AuthProperties::kAuthTypeOAuth2;
+  }
+
+  // Default to no authentication
```
wgtmac:
```suggestion
```

src/iceberg/catalog/rest/auth/auth_managers.cc
```diff
@@ -0,0 +1,88 @@
...
+  return AuthProperties::kAuthTypeNone;
+}
+
+/// \brief Get the global registry of auth manager factories.
```
wgtmac:
```suggestion
// Get the global registry of auth manager factories.
```

For comments that are not in the header, we don't need to use the docxygen style `///`.

src/iceberg/catalog/rest/auth/auth_managers.h
```diff
@@ -0,0 +1,65 @@
...
+  /// \return Manager instance or an error if no factory matches.
+  static Result<std::unique_ptr<AuthManager>> Load(
+      std::string_view name,
+      const std::unordered_map<std::string, std::string>& properties);
+
+  /// \brief Register or override the factory for a given auth type.
+  ///
+  /// \param auth_type Case-insensitive type identifier (e.g., "basic").
+  /// \param factory Factory function that produces the manager.
+  static void Register(std::string_view auth_type, AuthManagerFactory factory);
```
wgtmac, line 65: nit: add some comment to say the thread safety implication.

src/iceberg/catalog/rest/auth/auth_session.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /// \brief Create a session with static headers.
+  ///
+  /// This factory method creates a session that adds a fixed set of headers to each
+  /// request. It is suitable for authentication methods that use static credentials,
+  /// such as Basic auth or static bearer tokens.
+  ///
+  /// \param headers The headers to add to each request for authentication.
+  /// \return A new session that adds the given headers to requests.
+  static std::shared_ptr<AuthSession> Make(
```
wgtmac:
```suggestion
  static std::shared_ptr<AuthSession> MakeDefault(
```

I'm not sure if we need to extend it to accept more types. Let's make it explicit for now that it creates a default session.

wgtmac (approved):
Mostly looks good! I've left some nits.

PR #481, opened 2026-01-03
fix: use /v1/config endpoint for catalog config

plusplusjiajia:
base on the define in https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L65


wgtmac (approved):

PR #482, opened 2026-01-03
rest: build resource paths from final config

plusplusjiajia:
Build ResourcePaths using final URI + prefix after fetching server config: ref:https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L233

### DIFF
--- a/src/iceberg/catalog/rest/resource_paths.h
+++ b/src/iceberg/catalog/rest/resource_paths.h
@@ -44,7 +44,7 @@ class ICEBERG_REST_EXPORT ResourcePaths {
                                                      const std::string& prefix);
 
   /// \brief Get the /v1/config endpoint path.
-  static Result<std::string> Config(const std::string& base_uri);
+  Result<std::string> Config() const;
 
   /// \brief Get the /v1/{prefix}/oauth/tokens endpoint path.
   Result<std::string> OAuth2Tokens() const;
--- a/src/iceberg/catalog/rest/rest_catalog.cc
+++ b/src/iceberg/catalog/rest/rest_catalog.cc
@@ -63,8 +63,9 @@ std::unordered_set<Endpoint> GetDefaultEndpoints() {
 }
 
 /// \brief Fetch server config and merge it with client config
-Result<CatalogConfig> FetchServerConfig(const std::string& config_path,
+Result<CatalogConfig> FetchServerConfig(const ResourcePaths& paths,
                                         const RestCatalogProperties& current_config) {
+  ICEBERG_ASSIGN_OR_RAISE(auto config_path, paths.Config());
   HttpClient client(current_config.ExtractHeaders());
   ICEBERG_ASSIGN_OR_RAISE(const auto response,
                           client.Get(config_path, /*params=*/{}, /*headers=*/{},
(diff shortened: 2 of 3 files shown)


src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -117,7 +117,7 @@ Result<std::shared_ptr<RestCatalog>> RestCatalog::Make(
   ICEBERG_ASSIGN_OR_RAISE(auto server_config, FetchServerConfig(*paths, config));
 
   std::unique_ptr<RestCatalogProperties> final_config = RestCatalogProperties::FromMap(
-      MergeConfigs(server_config.overrides, config.configs(), server_config.defaults));
+      MergeConfigs(server_config.defaults, config.configs(), server_config.overrides));
```
wgtmac, line 120: Thanks for fixing this! Let's not complicate the PR. I think this change can be added without the test.

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -132,7 +132,11 @@ Result<std::shared_ptr<RestCatalog>> RestCatalog::Make(
 
   // Update resource paths based on the final config
   ICEBERG_ASSIGN_OR_RAISE(auto final_uri, final_config->Uri());
-  ICEBERG_RETURN_UNEXPECTED(paths->SetBaseUri(std::string(TrimTrailingSlash(final_uri))));
+  ICEBERG_ASSIGN_OR_RAISE(
```
wgtmac, line 135: I'm not sure if it is worth changing this. In theory the `prefix` should not change.

src/iceberg/catalog/rest/rest_catalog.cc
plusplusjiajia: > Thanks for fixing this! Let's not complicate the PR. I think this change can be added without the test.

Yes, i create a new pull request for it.https://github.com/apache/iceberg-cpp/pull/483/files

src/iceberg/catalog/rest/rest_catalog.cc
plusplusjiajia: > I'm not sure if it is worth changing this. In theory the `prefix` should not change.
maybe, we can build after getting config?

HeartLinked:
> > I think it makes sense to build resource paths from the final config, and I agree to create a new `ResourcePaths` instance. In this case, the `SetBaseUri` will no longer be necessary. Could you help remove it? Btw, I don't think it's worth changing to the signature of the `ResourcePaths::Config`.
> 
> SetBaseUri removed. Do you mean build a ResourcePaths to get config?

Well, sorry—I might not have put it clearly. I just wanted to say that I agree with your changes to `src/iceberg/catalog/rest/rest_catalog.cc`. :)

src/iceberg/catalog/rest/resource_paths.h
```diff
@@ -44,11 +43,8 @@ class ICEBERG_REST_EXPORT ResourcePaths {
   static Result<std::unique_ptr<ResourcePaths>> Make(std::string base_uri,
                                                      const std::string& prefix);
 
-  /// \brief Set the base URI of the REST catalog server.
-  Status SetBaseUri(const std::string& base_uri);
-
   /// \brief Get the /v1/config endpoint path.
-  Result<std::string> Config() const;
+  static Result<std::string> Config(const std::string& base_uri);
```
wgtmac: It sounds good to remove `SetBaseUri`. However, it leads to confusion to make this a static function. I'd suggest to revert this function as you have created a new ResourcePaths based on the server config.

src/iceberg/catalog/rest/rest_catalog.cc
wgtmac: I think it is cheap to create a new ResourcePaths. I'm fine with this change and please revert changes from 113 to 115 above.

src/iceberg/catalog/rest/resource_paths.h
plusplusjiajia: > It sounds good to remove `SetBaseUri`. However, it leads to confusion to make this a static function. I'd suggest to revert this function as you have created a new ResourcePaths based on the server config.

it's reasonable, i've reverted.

HeartLinked (approved):

PR #488, opened 2026-01-05
feat: impl metrics config

WZhuo:


src/iceberg/util/type_util.h
```diff
@@ -122,6 +122,20 @@ class PruneColumnVisitor {
   const bool select_full_types_;
 };
 
+/// \brief Visitor for get field IDs which could be used for projection.
+class GetProjectedIdsVisitor {
+ public:
+  explicit GetProjectedIdsVisitor(bool include_struct_ids = false);
+
+  Status Visit(const std::shared_ptr<Type>& type);
```
wgtmac:
```suggestion
  Status Visit(const Type& type);
```

It looks weird that one is shared_ptr but the other is not. Is it better to use `PrimitiveType` as the type?

src/iceberg/sort_order.cc
```diff
@@ -132,4 +132,18 @@ Result<std::unique_ptr<SortOrder>> SortOrder::Make(int32_t sort_id,
...
 }
 
+std::unordered_set<std::string_view> SortOrder::OrderPreservingSortedColumns(
+    const Schema& schema, const SortOrder& order) {
+  return order.fields() | std::views::filter([&schema](const SortField& field) {
+           return field.transform()->PreservesOrder();
+         }) |
+         std::views::transform([&schema](const SortField& field) {
+           return schema.FindColumnNameById(field.source_id())
+               .value_or(std::nullopt)
```
wgtmac, line 142: Why there is two `value_or`? Should we return error if `FindColumnNameById` returns error instead of swallowing it? (This is not a strong opinion)

src/iceberg/metrics_config.h
```diff
@@ -22,24 +22,120 @@
...
+class ICEBERG_EXPORT MetricsMode : public util::Formattable {
+ public:
+  enum class Kind : uint8_t {
+    kNone,
+    kCounts,
+    kTruncate,
+    kFull,
+  };
+
+  static Result<std::shared_ptr<MetricsMode>> FromString(const std::string& mode);
```
wgtmac:
```suggestion
  static Result<std::shared_ptr<MetricsMode>> FromString(std::string_view mode);
```

src/iceberg/metrics_config.h
```diff
@@ -22,24 +22,120 @@
...
+
+  static const std::shared_ptr<MetricsMode>& None();
+  static const std::shared_ptr<MetricsMode>& Counts();
+  static const std::shared_ptr<MetricsMode>& Truncate();
```
wgtmac: Is it useful to keep this parameterless `Truncate`?

src/iceberg/metrics_config.h
```diff
@@ -22,24 +22,120 @@
...
+
+  std::string ToString() const override;
+};
+
 /// \brief Configuration utilities for table metrics
 class ICEBERG_EXPORT MetricsConfig {
  public:
+  MetricsConfig(
+      std::unordered_map<std::string, std::shared_ptr<MetricsMode>> column_modes,
+      std::shared_ptr<MetricsMode> default_mode);
```
wgtmac:
```suggestion
      std::unordered_map<std::string, std::string>> properties);
```

Should we use this as the function signature? I think MetricsConfig is created directly from the table property map.

src/iceberg/metrics_config.h
```diff
@@ -22,24 +22,120 @@
...
+#include <unordered_set>
 
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
 #include "iceberg/type_fwd.h"
+#include "iceberg/util/formattable.h"
 
 namespace iceberg {
 
+class ICEBERG_EXPORT MetricsMode : public util::Formattable {
```
wgtmac: Is it an overkill to use class hierarchy for it? Can we do this instead?

```cpp
struct MetricsMode {
  enum class Kind : uint8_t { kNone, ... };

  Kind kind;
  std::variant<std::monostate, int32_t> param;
};
```

src/iceberg/metrics_config.h
```diff
@@ -22,24 +22,120 @@
...
  public:
+  MetricsConfig(
+      std::unordered_map<std::string, std::shared_ptr<MetricsMode>> column_modes,
+      std::shared_ptr<MetricsMode> default_mode);
+
+  /// \brief Get the default metrics config.
+  static const std::shared_ptr<MetricsConfig>& Default();
+
+  /// \brief Creates a metrics config from a table.
+  static Result<std::shared_ptr<MetricsConfig>> Make(std::shared_ptr<Table> table);
```
wgtmac:
```suggestion
  static Result<std::unique_ptr<MetricsConfig>> Make(const Table& table);
```

src/iceberg/metrics_config.h
wgtmac: BTW, should we make it a private ctor and add a Make function to create one by validating all inputs?

src/iceberg/metrics_config.cc
```diff
@@ -19,15 +19,235 @@
...
+Result<std::shared_ptr<MetricsMode>> MetricsMode::FromString(const std::string& mode) {
+  if (StringUtils::EqualsIgnoreCase(mode, kNoneName)) {
+    return MetricsMode::None();
+  } else if (StringUtils::EqualsIgnoreCase(mode, kCountsName)) {
+    return MetricsMode::Counts();
+  } else if (StringUtils::EqualsIgnoreCase(mode, kFullName)) {
+    return MetricsMode::Full();
+  }
+
+  if (mode.starts_with(kTruncatePrefix) && mode.ends_with(")")) {
```
wgtmac: It should also usse case insensitive comparison.

src/iceberg/metrics_config.cc
```diff
@@ -19,15 +19,235 @@
...
+                                         std::move(default_mode));
+}
+
+Result<std::unordered_set<int32_t>> MetricsConfig::LimitFieldIds(const Schema& schema,
+                                                                 int32_t limit) {
+  class Visitor {
+   public:
+    explicit Visitor(int32_t limit) : limit_(limit) {}
+
+    Status Visit(const std::shared_ptr<Type>& type) {
```
wgtmac: Similarly, it looks weird to use `const std::shared_ptr<Type>&` and `const NestedType&` inconsistently. Can you fix this? You may find `VisitTypeInline` useful.

src/iceberg/metrics_config.cc
```diff
@@ -19,15 +19,235 @@
...
+      }
+      return {};
+    }
+
+    Status Visit(const NestedType& type) {
+      for (auto& field : type.fields()) {
+        if (!ShouldContinue()) {
+          break;
+        }
+        if (field.type()->is_primitive()) {
```
wgtmac, line 187: Add a TODO comment to support `variant` type in the future.

wgtmac (changes requested):

src/iceberg/metrics_config.h
WZhuo: The constructor will not be called directly, defined as private

src/iceberg/metrics_config.h
WZhuo: looks good

src/iceberg/metrics_config.h
WZhuo: no used, remove it

src/iceberg/sort_order.cc
WZhuo: I think `FindColumnNameById` is not expected to return an error, and doing so would break the entire pipeline.

wgtmac (approved):

PR #489, opened 2026-01-05
feat: extend table scan to support v2 deletes

wgtmac:


src/iceberg/table_scan.h
```diff
@@ -50,175 +63,293 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
+  TableScanBuilder& Select(const std::vector<std::string>& column_names);
+
+  /// \brief Set the expression to filter data.
+  /// \param filter a filter expression
+  TableScanBuilder& Filter(std::shared_ptr<Expression> filter);
+
+  /// \brief Request data filtering to files but not to rows in those files.
+  TableScanBuilder& IgnoreResiduals();
+
+  /// \brief Request this can to return at least the given number of rows.
```
gty404:
```suggestion
  /// \brief Request this scan to return at least the given number of rows.
```

src/iceberg/table_scan.h
```diff
@@ -50,175 +63,293 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
+  /// \brief Instructs this scan to look for changes starting from a particular snapshot
+  /// (inclusive).
+  ///
+  /// If the start snapshot is not configured, it defaults to the oldest ancestor of the
+  /// end snapshot (inclusive).
+  ///
+  /// \param from_snapshot_id the start snapshot ID (inclusive)
+  /// \note InvalidArgument will be returned if the start snapshot is not an ancestor of
+  /// the end snapshot
+  TableScanBuilder& FromSnapshotInclusive(int64_t from_snapshot_id);
```
gty404: TableScanBuilder& FromSnapshot(int64_t from_snapshot_id, bool inclusive);
TableScanBuilder& FromSnapshot(const std::string& ref,  bool inclusive);

How about merging these interfaces of FromSnapshot into this one? Can ref also be changed to tag? SnapshotRef might be Branch or Tag. Using ref here would be a bit strange

src/iceberg/table_scan.h
```diff
@@ -50,175 +63,293 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
+  bool return_column_stats{false};
+  std::unordered_set<int32_t> columns_to_keep_stats;
+  std::vector<std::string> selected_columns;
+  std::shared_ptr<Schema> projected_schema;
   std::unordered_map<std::string, std::string> options;
-  /// \brief Optional limit on the number of rows to scan.
-  std::optional<int64_t> limit;
+  bool from_snapshot_id_inclusive{false};
+  std::optional<int64_t> from_snapshot_id;
+  std::optional<int64_t> to_snapshot_id;
```
gty404, line 118: How about defining a VersionRange or DataRange to express a specific version or version range?

src/iceberg/table_scan.cc
```diff
@@ -135,186 +153,346 @@ Result<ArrowArrayStream> MakeArrowArrayStream(std::unique_ptr<Reader> reader) {
 
 }  // namespace
 
+namespace internal {
+
+Status TableScanContext::Validate() const {
+  if (!columns_to_keep_stats.empty() && !return_column_stats) {
```
gty404: use  ICEBERG_PRECHECK ?

src/iceberg/table_scan.cc
```diff
@@ -135,186 +153,346 @@ Result<ArrowArrayStream> MakeArrowArrayStream(std::unique_ptr<Reader> reader) {
...
+                        context_.snapshot_id.value());
+  auto iter = metadata_->refs.find(ref);
+  if (iter != metadata_->refs.end()) {
+    ICEBERG_BUILDER_CHECK(iter->second != nullptr, "Ref {} is null", ref);
+    int32_t snapshot_id = iter->second->snapshot_id;
+    ICEBERG_BUILDER_ASSIGN_OR_RETURN(std::ignore, metadata_->SnapshotById(snapshot_id));
+    context_.snapshot_id = snapshot_id;
+  } else {
+    return AddError(InvalidArgument("Cannot find ref {}", ref));
   }
```
gty404:
```suggestion
  ICEBERG_BUILDER_CHECK(iter != metadata_->refs.end(), "Cannot find ref {}", ref);
  ICEBERG_BUILDER_CHECK(iter->second != nullptr, "Ref {} is null", ref);
  int32_t snapshot_id = iter->second->snapshot_id;
  ICEBERG_BUILDER_ASSIGN_OR_RETURN(std::ignore, metadata_->SnapshotById(snapshot_id));
  context_.snapshot_id = snapshot_id;
```

src/iceberg/table_scan.h
wgtmac: I think that may be a little bit complicated and `from/to` do not need to appear at the same time.

src/iceberg/table_scan.h
wgtmac: It makes sense to combine them. I don't think the ref applies to tag.

src/iceberg/table_scan.cc
```diff
@@ -135,186 +153,327 @@ Result<ArrowArrayStream> MakeArrowArrayStream(std::unique_ptr<Reader> reader) {
...
+  ICEBERG_ASSIGN_OR_RAISE(auto data_manifests, snapshot_cache.DataManifests(io_));
+  ICEBERG_ASSIGN_OR_RAISE(auto delete_manifests, snapshot_cache.DeleteManifests(io_));
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto manifest_group,
+      ManifestGroup::Make(io_, schema_, specs_by_id,
+                          {data_manifests.begin(), data_manifests.end()},
+                          {delete_manifests.begin(), delete_manifests.end()}));
+  (*manifest_group)
+      .CaseSensitive(context_.case_sensitive)
```
HuaHuaY:
```suggestion
  manifest_group->CaseSensitive(context_.case_sensitive)
```

wgtmac: I wrote this by purpose to use `.` in this chain.

HuaHuaY: You can use `->` for first call and `.` for others.

src/iceberg/table_scan.h
gty404: snapshot_id can also be expressed as [-1, snapshot_id], It can avoid passing multiple parameters when used internally.

wgtmac: Let's revisit this once we add incremental scan support.

wgtmac:
Thanks all for the review!

PR #491, opened 2026-01-06
feat: implement reference visitor

wgtmac:


src/iceberg/expression/binder.cc
```diff
@@ -113,4 +115,54 @@ Result<bool> IsBoundVisitor::Aggregate(
...
+
+Result<FieldIdsSetRef> ReferenceVisitor::Predicate(
+    const std::shared_ptr<BoundPredicate>& pred) {
+  referenced_field_ids_.insert(pred->reference()->field_id());
+  return referenced_field_ids_;
+}
+
+Result<FieldIdsSetRef> ReferenceVisitor::Predicate(
+    [[maybe_unused]] const std::shared_ptr<UnboundPredicate>& pred) {
+  return referenced_field_ids_;
```
zhjwpku: Should we error on unbound predicates and unbound aggregates? ISTM that an expression should be either fully bound or fully unbound, and that mixing the two is not possible, and visiting an unbound expression would be meaningless.

zhjwpku: I just took a look at the Java's impl of ReferenceVisitor and IsBoundVisitor, it appears that mixed bound and unbound predicate should error, I created PR #503 to fix our IsBoundVisitor logic, please correct me if f I'm wrong.

PR #495, opened 2026-01-07
fix: YearTransform return years since 1970

WZhuo:


WZhuo:
``` 
   Types.TimestampType type = Types.TimestampType.withoutZone();
    Literal<Long> ts = Literal.of("2017-12-01T10:12:55.038194").to(type);
    Literal<Long> pts = Literal.of("1970-01-01T00:00:01.000001").to(type);
    Literal<Long> nts = Literal.of("1969-12-31T23:59:59.999999").to(type);

    Transform<Long, Integer> years = Transforms.year();
    assertThat((int) years.bind(type).apply(ts.value()))
        .as("Should produce 2017 - 1970 = 47")
        .isEqualTo(47);
    assertThat((int) years.bind(type).apply(pts.value()))
        .as("Should produce 1970 - 1970 = 0")
        .isZero();
    assertThat((int) years.bind(type).apply(nts.value()))
        .as("Should produce 1969 - 1970 = -1")
        .isEqualTo(-1);
```

Above is the YearTransform's test case in Java, it should return delta years between the time and 1970.

zhjwpku:
Good catch, you may also want to change the comments of Transform::Year() and YearTransform.

src/iceberg/transform_function.h
```diff
@@ -100,13 +100,14 @@ class ICEBERG_EXPORT TruncateTransform : public TransformFunction {
...
 
-/// \brief Year transform that extracts the year component from timestamp inputs.
+/// \brief Year transform that extracts the number of years from timestamp inputs since
+/// the epoch.
 class ICEBERG_EXPORT YearTransform : public TransformFunction {
  public:
   /// \param source_type Must be a timestamp type.
   explicit YearTransform(std::shared_ptr<Type> const& source_type);
 
-  /// \brief Extract a date or timestamp year, as years from 1970.
```
zhjwpku, line 109: This is the same as the spec[1], maybe we should keep this.

ditto for all the Transform methods comments.

[1] https://iceberg.apache.org/spec/#partition-transforms

wgtmac (approved):

wgtmac:
Thanks @WZhuo for fixing this and @zhjwpku for the review!

PR #498, opened 2026-01-08
feat: implement update column

gty404:


src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
...
-  AddError(NotImplemented("UpdateSchema::UpdateColumn not implemented"));
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto field_opt, FindFieldForUpdate(name));
+
+  // If not found, check if there's a deleted field with this name
+  if (!field_opt.has_value()) {
+    ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto original_field, FindField(name));
+    ICEBERG_BUILDER_CHECK(!original_field.has_value() ||
+                              !deletes_.contains(original_field->get().field_id()),
+                          "Cannot update a column that will be deleted: {}", name);
+    // No field found at all
```
wgtmac:
```suggestion
```

Let's remove duplicate comments like this and below. LLMs are notorious at generating too verbose comments. Let's keep functions short and compact as much as possible.

src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
...
+  // If the type is already the same, this is a noop
+  if (*field.type() == *new_type) {
+    return *this;
+  }
+
+  int32_t field_id = field.field_id();
+
+  // Check if field is being deleted
+  // (This shouldn't happen given FindFieldForUpdate logic, but check for safety)
+  ICEBERG_BUILDER_CHECK(!deletes_.contains(field_id),
```
wgtmac, line 301: This should be checked before the type equality check at line 347. Otherwise we end up with updating a deleted column.

src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
 
 UpdateSchema& UpdateSchema::UpdateColumn(std::string_view name,
                                          std::shared_ptr<PrimitiveType> new_type) {
-  // TODO(Guotao Yu): Implement UpdateColumn
-  AddError(NotImplemented("UpdateSchema::UpdateColumn not implemented"));
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto field_opt, FindFieldForUpdate(name));
+
+  // If not found, check if there's a deleted field with this name
+  if (!field_opt.has_value()) {
```
wgtmac: Why not directly error out? The Java impl does not check original field.

src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
...
 }
 
 UpdateSchema& UpdateSchema::UpdateColumnDoc(std::string_view name,
                                             std::string_view new_doc) {
-  // TODO(Guotao Yu): Implement UpdateColumnDoc
-  AddError(NotImplemented("UpdateSchema::UpdateColumnDoc not implemented"));
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto field_opt, FindFieldForUpdate(name));
+
+  // If not found, check if there's a deleted field with this name
+  if (!field_opt.has_value()) {
```
wgtmac: Same question as in UpdateColumn.

src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
...
+  // If the doc is already the same, this is a noop
+  if (field.doc() == new_doc) {
+    return *this;
+  }
+
+  int32_t field_id = field.field_id();
+
+  // Check if field is being deleted
+  // (This shouldn't happen given FindFieldForUpdate logic, but check for safety)
+  ICEBERG_BUILDER_CHECK(!deletes_.contains(field_id),
```
wgtmac, line 326: ditto

src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
...
 }
 
 UpdateSchema& UpdateSchema::RenameColumn(std::string_view name,
                                          std::string_view new_name) {
-  // TODO(Guotao Yu): Implement RenameColumn
-  AddError(NotImplemented("UpdateSchema::RenameColumn not implemented"));
+  ICEBERG_BUILDER_CHECK(!new_name.empty(), "Cannot rename a column to null");
+
+  // Find the field for update (considers pending changes)
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto field_opt, FindFieldForUpdate(name));
```
wgtmac: Java calls `findField` not `findForUpdate`. Is this correct? Can we rename a updated or added column?

src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
...
+
+  // If not found, check if there's a deleted field with this name
+  if (!field_opt.has_value()) {
+    ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto original_field, FindField(name));
```
wgtmac: Same question as in UpdateColumn.

src/iceberg/update/update_schema.cc
```diff
@@ -327,34 +328,179 @@ UpdateSchema& UpdateSchema::AddRequiredColumn(std::optional<std::string_view> pa
...
+  }
+
+  int32_t field_id = field.field_id();
+
+  // TODO(GuotaoYu): support added column with default value
+  // Check for incompatible change: making a column required
+  // This is only allowed if:
+  // 1. allowIncompatibleChanges is true, OR
+  // 2. The field is newly added with a default value (not yet supported in C++)
+  // For now, we only check allowIncompatibleChanges
```
wgtmac:
```suggestion
```

These are not necessary.

src/iceberg/update/update_schema.cc
```diff
@@ -509,12 +655,26 @@ UpdateSchema& UpdateSchema::AddColumnInternal(std::optional<std::string_view> pa
...
                           "Cannot add to a column that will be deleted: {}", *parent);
 
-    // Check field doesn't already exist (unless it's being deleted)
+    // Check field doesn't already exist in original schema (unless being deleted/renamed)
     std::string nested_name = std::format("{}.{}", *parent, name);
-    ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto current_field, FindField(nested_name));
-    ICEBERG_BUILDER_CHECK(
-        !current_field.has_value() || deletes_.contains(current_field->get().field_id()),
-        "Cannot add column, name already exists: {}.{}", *parent, name);
+    ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto original_field, FindField(nested_name));
```
wgtmac: The java impl does not have this logic, why do we introduce these?

src/iceberg/update/update_schema.cc
```diff
@@ -525,10 +685,24 @@ UpdateSchema& UpdateSchema::AddColumnInternal(std::optional<std::string_view> pa
     full_name = std::format("{}.{}", *parent_name_opt, name);
   } else {
     // Top-level field
-    ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto current_field, FindField(name));
-    ICEBERG_BUILDER_CHECK(
-        !current_field.has_value() || deletes_.contains(current_field->get().field_id()),
-        "Cannot add column, name already exists: {}", name);
+    // Check field doesn't already exist in original schema (unless being deleted/renamed)
```
wgtmac: ditto

src/iceberg/update/update_schema.cc
```diff
@@ -572,4 +746,43 @@ Result<std::optional<std::reference_wrapper<const SchemaField>>> UpdateSchema::F
...
+        return std::optional<std::reference_wrapper<const SchemaField>>(
+            std::cref(*update_it->second));
+      }
+      // Return the original field
+      return existing_field_opt;
+    }
+  }
+
+  // Field not in original schema (or is being deleted), check if it's a newly added field
+  auto added_it = added_name_to_id_.find(std::string(name));
```
wgtmac: Forget to call caseSensitivityAwareName?

src/iceberg/update/update_schema.cc
```diff
@@ -572,4 +746,43 @@ Result<std::optional<std::reference_wrapper<const SchemaField>>> UpdateSchema::F
   return schema_->FindFieldByName(name, case_sensitive_);
 }
 
+Result<std::optional<std::reference_wrapper<const SchemaField>>>
+UpdateSchema::FindFieldForUpdate(std::string_view name) const {
+  // First, check the original schema
```
wgtmac: Can we remove these comments to be shorter?

src/iceberg/update/update_schema.cc
gty404: This is to distinguish between conflicting types, whether columns do not exist at all, or whether deleted columns are updated. I will modify the logic of FindFieldForUpdate to no longer check the delete list, but check and return the corresponding conflict information in UpdateColumns, which is consistent with java

PR #501, opened 2026-01-08
feat: add utility to print transformed value as human string

WZhuo:


src/iceberg/util/transform_util.h
```diff
@@ -0,0 +1,105 @@
...
+  /// \return a string representation of this time
+  static std::string HumanTime(int64_t micros_from_midnight);
+
+  /// \brief Returns a string representation of a timestamp in microseconds.
+  ///
+  /// The output will be one of the following forms, according to the precision of the
+  /// timestamp:
+  ///  - yyyy-MM-dd HH:mm:ss
+  ///  - yyyy-MM-dd HH:mm:ss.SSS
+  ///  - yyyy-MM-dd HH:mm:ss.SSSSSS
```
wgtmac:
```suggestion
  ///  - yyyy-MM-ddTHH:mm:ss
  ///  - yyyy-MM-ddTHH:mm:ss.SSS
  ///  - yyyy-MM-ddTHH:mm:ss.SSSSSS
```

The java formatter is as below:

```java
        ISO_LOCAL_DATE_TIME = new DateTimeFormatterBuilder()
                .parseCaseInsensitive()
                .append(ISO_LOCAL_DATE)
                .appendLiteral('T')
                .append(ISO_LOCAL_TIME)
                .toFormatter(ResolverStyle.STRICT, IsoChronology.INSTANCE);
```

src/iceberg/util/transform_util.h
```diff
@@ -0,0 +1,105 @@
...
+  /// \param timestamp_micros the timestamp in microseconds.
+  /// \return a string representation of this timestamp.
+  static std::string HumanTimestamp(int64_t timestamp_micros);
+
+  /// \brief Returns a human-readable string representation of a timestamp with a time
+  /// zone.
+  ///
+  /// The output will be one of the following forms, according to the precision of the
+  /// timestamp:
+  ///  - yyyy-MM-dd HH:mm:ss+00:00
```
wgtmac: ditto

src/iceberg/test/transform_util_test.cc
```diff
@@ -0,0 +1,149 @@
...
+ */
+
+#include "iceberg/util/transform_util.h"
+
+#include <gtest/gtest.h>
+
+namespace iceberg {
+
+TEST(TransformUtilTest, HumanYear) {
+  EXPECT_EQ("1970", TransformUtil::HumanYear(0));
```
wgtmac, line 27: Let's add large negative number to test year smaller than 1000. Same for other cases.

wgtmac (approved):

src/iceberg/util/transform_util.cc
```diff
@@ -0,0 +1,144 @@
...
+    return std::format("{:%FT%T}.{:06d}+00:00", tp, micros);
+  }
+}
+
+std::string TransformUtil::Base64Encode(std::string_view str_to_encode) {
+  static constexpr std::string_view kBase64Chars =
+      "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+  int32_t i = 0;
+  int32_t j = 0;
+  std::array<unsigned char, 3> char_array_3;
```
zhjwpku, line 100: nit: should we use uint8_t as elsewhere? Not a strong opinion.

wgtmac: That may require extra cast when appending to the string.

PR #503, opened 2026-01-10
chore: fix IsBoundVisitor to error on AlwaysTrue & AlwaysFalse

zhjwpku:


src/iceberg/expression/binder.cc
```diff
@@ -79,36 +96,47 @@ Result<std::shared_ptr<Expression>> Binder::Aggregate(
...
-  return Visit<bool, IsBoundVisitor>(expr, visitor);
+  auto visit_result = Visit<std::optional<bool>, IsBoundVisitor>(expr, visitor);
+  ICEBERG_RETURN_UNEXPECTED(visit_result);
+  auto result = std::move(visit_result.value());
+  // If the result is null, return true
+  return result.value_or(true);
 }
 
-Result<bool> IsBoundVisitor::AlwaysTrue() { return true; }
+Result<std::optional<bool>> IsBoundVisitor::AlwaysTrue() { return std::nullopt; }
```
wgtmac: I think the Java impl is overly complicated to return `null` by default and then a lot of places should handle `null` with different meanings.

wgtmac: What about letting each visitors to handle this instead of complicating the return type?

zhjwpku: Fixed.

src/iceberg/expression/binder.cc
```diff
@@ -92,11 +92,13 @@ Result<bool> IsBoundVisitor::AlwaysFalse() { return true; }
 Result<bool> IsBoundVisitor::Not(bool child_result) { return child_result; }
 
 Result<bool> IsBoundVisitor::And(bool left_result, bool right_result) {
-  return left_result && right_result;
+  ICEBERG_PRECHECK(left_result == right_result, "Found partially bound expression");
```
wgtmac: I don't think this is correct. If we want to keep the same behavior as the Java impl, we might need to return error for `AlwaysTrue()` and `AlwaysFalse()` and then revert changes to `And` and `Or`.

zhjwpku: But if we do that, mixed bound and unbound will be evaluated as false, is that expected?

wgtmac: That is expected. `IsBoundVisitor` checks if all expressions are bound and returns false if any is unbound. This is used to check if the expression has called bind function along the tree.

zhjwpku: Ok, I get it.

PR #505, opened 2026-01-12
feat: impl Transform::ToHumanString

WZhuo:


src/iceberg/transform.h
```diff
@@ -194,6 +194,12 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
   Result<std::unique_ptr<UnboundPredicate>> ProjectStrict(
       std::string_view name, const std::shared_ptr<BoundPredicate>& predicate);
 
+  /// \brief Returns a human-readable String representation of a transformed value.
+  ///
+  /// \param value The literal value to be transformed.
+  /// @return a human-readable String representation of the value
```
zhjwpku:
```suggestion
  /// \return a human-readable String representation of the value
```

src/iceberg/transform.h
```diff
@@ -194,6 +194,12 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
   Result<std::unique_ptr<UnboundPredicate>> ProjectStrict(
       std::string_view name, const std::shared_ptr<BoundPredicate>& predicate);
 
+  /// \brief Returns a human-readable String representation of a transformed value.
```
```suggestion
  /// \brief Returns a human-readable string representation of a transformed value.
```

src/iceberg/transform.h
```diff
@@ -194,6 +194,12 @@ class ICEBERG_EXPORT Transform : public util::Formattable {
...
+  ///
+  /// \param value The literal value to be transformed.
+  /// \return A human-readable String representation of the value
```
```suggestion
  /// \return A human-readable string representation of the value
```

src/iceberg/transform.cc
```diff
@@ -366,6 +367,50 @@ Result<std::unique_ptr<UnboundPredicate>> Transform::ProjectStrict(
   std::unreachable();
 }
 
+Result<std::string> Transform::ToHumanString(const Literal& value) {
+  if (value.IsNull()) {
```
wgtmac, line 371: Let's return error when `IsAboveMax()` or `IsBelowMin()` is true.

src/iceberg/transform.cc
```diff
@@ -366,6 +367,50 @@ Result<std::unique_ptr<UnboundPredicate>> Transform::ProjectStrict(
...
+    return "null";
+  }
+
+  switch (transform_type_) {
+    case TransformType::kYear:
+      return TransformUtil::HumanYear(std::get<int32_t>(value.value()));
```
wgtmac, line 385: We might need to check the literal type before calling `std::get` on the variant value to avoid exception.

src/iceberg/transform.cc
```diff
@@ -366,6 +367,50 @@ Result<std::unique_ptr<UnboundPredicate>> Transform::ProjectStrict(
...
+    case TransformType::kMonth:
+      return TransformUtil::HumanMonth(std::get<int32_t>(value.value()));
+    case TransformType::kDay:
+      return TransformUtil::HumanDay(std::get<int32_t>(value.value()));
+    case TransformType::kHour:
+      return TransformUtil::HumanHour(std::get<int32_t>(value.value()));
+    default: {
```
wgtmac: Can we remove this `default` and add all other branches explicitly to avoid forgetting to update here in the future once we add a new transform type?

PR #506, opened 2026-01-12
feat: expose  `location_provider` interface for `Table`

WZhuo:


src/iceberg/table.h
```diff
@@ -116,6 +116,9 @@ class ICEBERG_EXPORT Table : public std::enable_shared_from_this<Table> {
   /// \brief Returns the catalog that this table belongs to
   const std::shared_ptr<Catalog>& catalog() const;
 
+  /// \brief Return a LocationProvider for this table
```
zhjwpku: nit: %s/Return/Returns as the above.

PR #548, opened 2026-01-29
feat: add s3 file io integration

MisterRaindrop:
I have implemented Arrow FileSystem to access S3, but I'm still not sure if it meets the requirements.

There are still task or question to complete for the current PR, and it is not ready for merging yet.

Question:   
Currently, the object storage options include Azure, AWS, and GCS. I have chosen AWS as the implementation for now is ok?
  
Task:  
I need to deploy MinIO to facilitate testing access to S3, but I'm not sure where it would be best to set it up?


wgtmac:
I think it is fine to use minio at this moment to unblock us. Let me know what you think on my proposed approach above. We might also need to add a FileIO registry to provide default implementation on us and enable users to override their own implementations of s3 and others. The key in the FileIO registry can be associated with table property `io-impl`.

wgtmac:
We may also need to add top-level CMake options like `ICEBERG_S3`, `ICEBERG_AZURE`, `ICEBERG_GCS` to mirror Arrow equivalents and vendor Arrow with `ARROW_S3=ON`. This might be tricky in the CI because we need to deal with Arrow's dependency on AWS SDK and others.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -87,6 +87,7 @@ function(resolve_arrow_dependency)
   # Work around undefined symbol: arrow::ipc::ReadSchema(arrow::io::InputStream*, arrow::ipc::DictionaryMemo*)
   set(ARROW_IPC ON)
   set(ARROW_FILESYSTEM ON)
+  set(ARROW_S3 ON)
```
wgtmac: Can we add a cmake option `ICEBERG_S3` and only toggle on `ARROW_S3` when `ICEBERG_S3` is on?

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,236 @@
...
+ * under the License.
+ */
+
+#include <cstdlib>
+#include <mutex>
+#include <string_view>
+
+#include <arrow/filesystem/filesystem.h>
+#include <arrow/filesystem/localfs.h>
+#if __has_include(<arrow/filesystem/s3fs.h>)
```
wgtmac: If we add `ICEBERG_S3` option, we don't need to deal with this check.

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -144,6 +145,40 @@ Result<std::shared_ptr<RestCatalog>> RestCatalog::Make(
...
+  // Check for user-specified io-impl property
+  auto io_impl = config.configs().find(FileIOProperties::kImpl);
+  std::string impl_name;
+
+  if (io_impl != config.configs().end() && !io_impl->second.empty()) {
+    // User specified a custom io-impl
+    impl_name = io_impl->second;
+  } else {
+    // Use default based on warehouse URI scheme
+    if (warehouse.rfind("s3://", 0) == 0) {
```
wgtmac: Why using `rfind`?

wgtmac: BTW, shouldn't we use `uri` instead of `warehouse` property?

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,236 @@
...
+  ICEBERG_ARROW_ASSIGN_OR_RETURN(auto fs, ::arrow::fs::FileSystemFromUri(uri, out_path));
+  return fs;
+}
+
+/// \brief ArrowUriFileIO resolves FileSystem from URI for each operation.
+///
+/// This implementation is thread-safe as it creates a new FileSystem instance
+/// for each operation. However, it may be less efficient than caching the
+/// FileSystem. S3 initialization is done once per process.
+class ArrowUriFileIO : public FileIO {
```
wgtmac: Why do we need this instead of reusing `ArrowFIleSystemFileIO`?

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,236 @@
...
+  return NotImplemented("Arrow S3 support is not enabled");
+#endif
+}
+
+#if ICEBERG_ARROW_HAS_S3
+/// \brief Configure S3Options from a properties map.
+///
+/// \param properties The configuration properties map.
+/// \return Configured S3Options.
+::arrow::fs::S3Options ConfigureS3Options(
```
wgtmac: I agree this is something that we need.

src/iceberg/catalog/rest/rest_catalog.h
```diff
@@ -53,6 +53,30 @@ class ICEBERG_REST_EXPORT RestCatalog : public Catalog,
   static Result<std::shared_ptr<RestCatalog>> Make(const RestCatalogProperties& config,
                                                    std::shared_ptr<FileIO> file_io);
 
+  /// \brief Create a RestCatalog instance with auto-detected FileIO.
+  ///
+  /// This overload automatically creates an appropriate FileIO based on the "io-impl"
+  /// property or the warehouse location URI scheme.
+  ///
+  /// FileIO selection logic:
```
wgtmac: It is better to add a `iceberg/util/file_io_util.h` to handle this logic and support reusing. Please note that Arrow Filesystem support is only available in the iceberg-bundle library, so we can only talk to the FileIO registry to create an FileIO instance.

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,236 @@
...
+    const std::unordered_map<std::string, std::string>& properties) {
+  ::arrow::fs::S3Options options;
+
+  // Configure credentials
+  auto access_key_it = properties.find(S3Properties::kAccessKeyId);
```
wgtmac: What is `S3Properties` defined?

wgtmac (commented):
I just took a quick pass on the latest commit. Can we simplify the implementation like this:

1. Use a CMake option to enable S3.
2. Define reserved iceberg properties for S3 and add functions to convert them to Arrow S3 options.
3. To create a concrete S3FileIO, using Arrow API to create a S3FileSystem and wrap it by ArrowFileSystemFileIO.
4. Register the factory to create FileIO of S3 to the registry before use.
5. Add a file io utility to create the FileIO instance based on various condition.

src/iceberg/file_io_registry.h
```diff
@@ -0,0 +1,102 @@
...
+
+/// \brief Registry for FileIO implementations.
+///
+/// Provides a mechanism to register and load FileIO implementations by name.
+/// This allows the REST catalog (and others) to resolve FileIO implementations
+/// at runtime based on configuration properties like "io-impl".
+class ICEBERG_EXPORT FileIORegistry {
+ public:
+  /// Well-known implementation names
+  static constexpr const char* kArrowLocalFileIO = "org.apache.iceberg.arrow.ArrowFileIO";
```
wgtmac: Let's use `std::string_view` instead of C-style string.

src/iceberg/file_io_registry.h
```diff
@@ -0,0 +1,102 @@
...
+  static constexpr const char* kArrowS3FileIO = "org.apache.iceberg.arrow.ArrowS3FileIO";
```
wgtmac: It looks odd to use the Java classpath style here. I haven't looked at other impls yet, perhaps it is worth investigating their conventions as well? My initial idea is just to use default keys like `"local"` and `"s3"` to locate the FileIO implementations.

src/iceberg/file_io_registry.h
```diff
@@ -0,0 +1,102 @@
...
+
+  /// Factory function type for creating FileIO instances.
+  using Factory = std::function<Result<std::shared_ptr<FileIO>>(
```
wgtmac: It is better to return unique_ptr by default.

src/iceberg/file_io_registry.h
```diff
@@ -0,0 +1,102 @@
...
+      const std::string& warehouse,
```
wgtmac:
```suggestion
      const std::string& name,
```

src/iceberg/arrow/file_io_register.cc
```diff
@@ -0,0 +1,53 @@
...
+
+namespace iceberg::arrow {
+
+void RegisterFileIO() {
+  static std::once_flag flag;
+  std::call_once(flag, []() {
+    // Register Arrow local filesystem FileIO
+    FileIORegistry::Register(
+        FileIORegistry::kArrowLocalFileIO,
+        [](const std::string& /*warehouse*/,
```
wgtmac:
```suggestion
        [](const std::string& /*name*/,
```

Same for below.

src/iceberg/arrow/s3/s3_properties.h
```diff
@@ -0,0 +1,52 @@
...
+#include <string>
+
+namespace iceberg::arrow {
+
+/// \brief S3 configuration property keys for ArrowS3FileIO.
+///
+/// These constants define the property keys used to configure S3 access
+/// via the Arrow filesystem integration, following the Iceberg spec for
+/// S3 configuration properties.
+struct S3Properties {
```
wgtmac, line 31: Should we add a separate subdirectory for s3 and move all s3-related files  there?

src/iceberg/arrow/s3_properties.h
```diff
@@ -0,0 +1,52 @@
...
+namespace iceberg::arrow {
+
+/// \brief S3 configuration property keys for ArrowS3FileIO.
+///
+/// These constants define the property keys used to configure S3 access
+/// via the Arrow filesystem integration, following the Iceberg spec for
+/// S3 configuration properties.
+struct S3Properties {
+  /// AWS access key ID
+  static constexpr const char* kAccessKeyId = "s3.access-key-id";
```
wgtmac: Use std::string_view. Same for other places in this PR.

src/iceberg/arrow/file_io_register.h
```diff
@@ -0,0 +1,40 @@
...
+/// \brief Register Arrow FileIO implementations (local and S3) into the
+/// FileIORegistry.
+///
+/// This function is idempotent and thread-safe. It registers:
+///   - ArrowFileIO (local filesystem)
+///   - ArrowS3FileIO (S3 filesystem)
+///
+/// Must be called before using FileIORegistry::Load() with the built-in
+/// implementation names (e.g., from RestCatalog::Make(config)).
+ICEBERG_BUNDLE_EXPORT void RegisterFileIO();
```
wgtmac: Should this be a separate register per FileIO implementation?

src/iceberg/test/CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ if(ICEBERG_BUILD_BUNDLE)
                    USE_BUNDLE
                    SOURCES
                    arrow_fs_file_io_test.cc
+                   arrow_s3_file_io_test.cc
```
wgtmac: Let's create a new `file_io_test` executable?

src/iceberg/file_io_registry.h
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /// \brief Load a FileIO implementation by name.
+  ///
+  /// \param name The implementation name to look up.
+  /// \param warehouse The warehouse location URI.
+  /// \param properties Configuration properties to pass to the factory.
+  /// \return A shared_ptr to the FileIO instance, or an error if not found.
+  static Result<std::shared_ptr<FileIO>> Load(
+      const std::string& name, const std::string& warehouse,
```
wgtmac: Why do we need a two-jump design for registry? Specifically I don't quite understand why it has anything to do with `warehouse`. Is it much simpler to locate the FileIO by the `name`?

src/iceberg/file_io_registry.h
```diff
@@ -0,0 +1,102 @@
...
+  static std::mutex& Mutex() {
+    static std::mutex mutex;
+    return mutex;
+  }
+};
+
+/// \brief Property keys for FileIO configuration.
+struct FileIOProperties {
+  /// The FileIO implementation class name (e.g., "org.apache.iceberg.arrow.ArrowFileIO")
+  static constexpr const char* kImpl = "io-impl";
```
wgtmac: A relevant followup is to map and register known `"io-impl"` names from other impls (especially Java).

src/iceberg/file_io_registry.h
```diff
@@ -0,0 +1,102 @@
...
+      }
+      factory = it->second;
+    }
+    // Invoke factory outside the lock to avoid blocking other Register/Load
+    // calls and to prevent deadlocks if the factory calls back into the registry.
+    return factory(warehouse, properties);
+  }
+
+ private:
+  static std::unordered_map<std::string, Factory>& Registry() {
```
wgtmac: Do we really need to add two wrappers for `registry` and `mutex`? We don't benefit too much from their lazy initialization.

src/iceberg/catalog/rest/rest_catalog.h
```diff
@@ -54,6 +54,31 @@ class ICEBERG_REST_EXPORT RestCatalog : public Catalog,
...
+  /// FileIORegistry::Register("com.mycompany.MyFileIO",
+  ///     [](const std::string& warehouse, const auto& props) {
+  ///         return std::make_shared<MyFileIO>(warehouse, props);
+  ///     });
+  /// \endcode
+  ///
+  /// \param config the configuration for the RestCatalog, including warehouse location
+  ///        and optional "io-impl" property
+  /// \return a shared_ptr to RestCatalog instance, or an error if FileIO creation fails
+  static Result<std::shared_ptr<RestCatalog>> Make(const RestCatalogProperties& config);
```
wgtmac, line 50: A cleaner approach is to make file_io a member of `RestCatalogProperties` and remove the above 2-arg Make function.

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -174,6 +175,40 @@ Result<std::shared_ptr<RestCatalog>> RestCatalog::Make(
...
+    return InvalidArgument(
+        "Warehouse location is required when FileIO is not explicitly provided. "
+        "Set the 'warehouse' property to an S3 URI (s3://...) or local path.");
+  }
+
+  // Check for user-specified io-impl property
+  auto io_impl = config.configs().find(FileIOProperties::kImpl);
+  std::string impl_name;
+
+  if (io_impl != config.configs().end() && !io_impl->second.empty()) {
```
wgtmac: nit: make the url scheme detection a separate function and move it to be an utility function of FileIO

src/iceberg/test/arrow_s3_file_io_test.cc
```diff
@@ -0,0 +1,238 @@
...
+///
+/// Arrow's S3 initialization creates global state that must be cleaned up via
+/// FinalizeS3() before the process exits.  Without this, Arrow's static destructor
+/// detects the missing finalization and causes a non-zero exit (which fails under
+/// sanitizers).  GTest Environment::TearDown() runs after all tests but before
+/// static destructors, making it the safe place to finalize.
+class ArrowS3TestEnvironment : public ::testing::Environment {
+ public:
+  void TearDown() override {
+    auto status = ::arrow::fs::FinalizeS3();
```
wgtmac: Should we add an equivalent function like `::iceberg::arrow::FinalizeS3()`? In this case, users do not need to write `include <arrow/filesystem/s3fs.h>` and the test is closer to the real world use case.

CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ option(ICEBERG_BUILD_TESTS "Build tests" ON)
 option(ICEBERG_BUILD_BUNDLE "Build the battery included library" ON)
 option(ICEBERG_BUILD_REST "Build rest catalog client" ON)
 option(ICEBERG_BUILD_REST_INTEGRATION_TESTS "Build rest catalog integration tests" OFF)
+option(ICEBERG_S3 "Build with S3 support" ON)
```
wgtmac: It is worth noting that `ICEBERG_S3` should be disabled if `ICEBERG_BUILD_BUNDLE` is `OFF`.

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,166 @@
...
+    ::arrow::fs::S3GlobalOptions options;
+    init_status = ::arrow::fs::InitializeS3(options);
+  });
+  if (!init_status.ok()) {
+    return std::unexpected(Error{.kind = ::iceberg::arrow::ToErrorKind(init_status),
+                                 .message = init_status.ToString()});
+  }
+  return {};
+#else
+  return NotImplemented("Arrow S3 support is not enabled");
```
wgtmac:
```suggestion
  return NotSupported("Arrow S3 support is not enabled");
```

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,166 @@
...
+namespace iceberg::arrow {
+
+namespace {
+
+Status EnsureS3Initialized() {
+#if ICEBERG_ARROW_HAS_S3
+  static std::once_flag init_flag;
+  static ::arrow::Status init_status = ::arrow::Status::OK();
+  std::call_once(init_flag, []() {
+    ::arrow::fs::S3GlobalOptions options;
```
wgtmac: nit: add a TODO comment to support options supported by `::arrow::fs::S3GlobalOptions`.

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,166 @@
...
+  auto ssl_it = properties.find(S3Properties::kSslEnabled);
+  if (ssl_it != properties.end() && ssl_it->second == "false") {
+    options.scheme = "http";
+  }
+
+  // Configure timeouts
+  auto connect_timeout_it = properties.find(S3Properties::kConnectTimeoutMs);
+  if (connect_timeout_it != properties.end()) {
+    try {
+      options.connect_timeout = std::stod(connect_timeout_it->second) / 1000.0;
```
wgtmac: Can we use from_chars just like `ParseInteger` does?

src/iceberg/arrow/arrow_s3_file_io.cc
```diff
@@ -0,0 +1,166 @@
...
+  return options;
+}
+#endif
+
+}  // namespace
+
+Result<std::unique_ptr<FileIO>> MakeS3FileIO(
+    const std::string& uri,
+    const std::unordered_map<std::string, std::string>& properties) {
+  if (!uri.starts_with("s3://")) {
```
wgtmac: Define a constant for the magic `s3://`. `s3a` should be supported as well.

CMakeLists.txt
wgtmac: Should we disable it by default?

wgtmac (changes requested):
Thanks for updating this! I haven't looked at the test yet and have some initial comments.

src/iceberg/arrow/arrow_s3_file_io.cc
MisterRaindrop: s3a, as far as I know, is a protocol header for Hadoop. Don't we support the standard S3?

CMakeLists.txt
MisterRaindrop: Currently, it appears that the entire project will prioritize development around the REST catalog, which primarily interacts with S3. Are we certain we want to disable ICEBERG_S3 by default?

wgtmac: Yes, but it will incur unnecessary burden to developers who do not care about S3. For example, Arrow will need to download and bundle AWS SDK and use docker to run the test.

MisterRaindrop: 👌

MisterRaindrop:
@wgtmac  I update, please review it, thanks

wgtmac (approved):
Sorry for the delay. I've polished its structure a little bit and now it looks good to me. Let me merge this. Thanks for working on this!

PR #576, opened 2026-02-26
feat: retry failed transaction commit

linguoxuan:
This commit implements the retry for transaction commits. It introduces a generic RetryRunner utility with exponential backoff and error-kind filtering, and integrates it into Transaction::Commit() to automatically refresh table metadata and retry on commit conflicts.


wgtmac (commented):
## Review Report: PR #576
*(Generated by gemini-cli)*

### 📄 File: `src/iceberg/table.cc`
**Java Counterpart:** `Table.java` / `BaseTable.java`

*   **Parity Check:** ✅ Correctly matches the Java behavior where the cached metadata location is updated upon `Refresh()`.
*   **Style & Comments:** ✅ Clean and straightforward.
*   **Logic Check:** ✅ Fixes a bug where `metadata_location_` could be stale.
*   **Design & Conciseness:** ✅ Good.
*   **Test Quality:** ✅ Covered by general refresh logic in existing tests.

### 📄 File: `src/iceberg/transaction.cc` & `src/iceberg/transaction.h`
**Java Counterpart:** `BaseTransaction.java`

*   **Parity Check:** ✅ The retry execution cleanly mimics `BaseTransaction.commitSimpleTransaction()` behavior utilizing `RetryRunner`. The use of `applying_updates_` correctly prevents recursive commits for single-operation transactions.
*   **Style & Comments:** ✅ Well documented.
*   **Logic Check:** ✅ Mapping `ErrorKind::kCommitFailed` inside `CommitOnce()` during a re-apply to `ValidationFailed` is a clever and correct way to abort the retry loop when a logical conflict occurs with the newly refreshed state.
*   **Design & Conciseness:** ✅ Elegant restructuring of the commit phase.
*   **Test Quality:** ✅ Good.

### 📄 File: `src/iceberg/update/snapshot_update.cc`
**Java Counterpart:** `SnapshotProducer.java`

*   **Parity Check:** ✅ Clearing `manifest_lists_` and cleaning uncommitted files before reapplying correctly mirrors the cleanup loop in Java's `SnapshotProducer.apply()`. Snapshot ID generation also properly handles collisions, achieving parity.
*   **Style & Comments:** ✅ Good.
*   **Logic Check:** ✅ State resets cleanly before each retry attempt.
*   **Design & Conciseness:** ✅ Well integrated into the existing `Apply()` flow.
*   **Test Quality:** ✅ Assumed covered by integration tests simulating conflicts.

### 📄 File: `src/iceberg/update/update_snapshot_reference.cc`
**Java Counterpart:** `UpdateSnapshotReferencesOperation.java`

*   **Parity Check:** ✅ Matches and arguably improves upon Java's `internalApply()`.
*   **Style & Comments:** ✅ Good.
*   **Logic Check:** ✅ The introduction of `initial_refs_` to compute the user's *intended delta* (additions/removals) and applying that delta to the *refreshed* `current_refs` is a very robust way to handle transaction rebasing. It properly avoids inadvertently dropping concurrent reference changes during a retry.
*   **Design & Conciseness:** ✅ Concise and logical.
*   **Test Quality:** ✅ Good.

### 📄 File: `src/iceberg/util/retry_util.h` & `src/iceberg/test/retry_util_test.cc`
**Java Counterpart:** `Tasks.java` (specifically the retry builder)

*   **Parity Check:** ✅ Config defaults (e.g., 4 retries, 100ms min wait, 60s max wait, 30m total timeout) directly match Iceberg Java's `COMMIT_NUM_RETRIES_DEFAULT`, etc.
*   **Style & Comments:** ⚠️ **Minor Style Point:** In `RetryRunner::CalculateDelay`, `std::random_device rd;` and `std::mt19937 gen(rd());` are instantiated on every single retry attempt. While the overhead is mostly negligible due to the surrounding `sleep`, `std::random_device` can block on `/dev/urandom` or throw exceptions in constrained environments. It is standard C++ practice to use a `thread_local` generator to avoid repeatedly initializing the PRNG. For example:
  ```cpp
  static thread_local std::mt19937 gen(std::random_device{}());
  ```
*   **Logic Check:** ✅ Exponential backoff and jitter math is sound. Timeout and max attempts logic is flawless.
*   **Design & Conciseness:** ✅ Excellent standalone utility. Does not reinvent the wheel as STL/Arrow lack an equivalent drop-in.
*   **Test Quality:** ✅ `retry_util_test.cc` is exhaustive and tests all major edge cases (exhaustion, filters, zero retries).

---

## Summary & Recommendation
*   **Comment**
Overall, a beautifully constructed PR. The implementation of transaction retries and rebasing (especially calculating the intent delta in `UpdateSnapshotReference`) is very robust and accurately achieves parity with the Java spec. 

I have left a minor suggestion regarding the PRNG initialization in `RetryRunner::CalculateDelay` to adhere to modern C++ best practices, but otherwise, the code is solid.

wgtmac (changes requested):
I've carefully reviewed the retry mechanism and found a few parity issues and a structural data-loss concern regarding how pending updates are held during retries. I have left inline comments on the specific lines.

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,180 @@
...
+
+      auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
+                         std::chrono::steady_clock::now() - start_time)
+                         .count();
+
+      // total_timeout_ms <= 0 means no total timeout limit
+      bool timed_out = config_.total_timeout_ms > 0 &&
+                       elapsed > config_.total_timeout_ms && attempt > 1;
+      if (attempt >= max_attempts || timed_out) {
+        return result;
```
wgtmac: The `timed_out` check requires `attempt > 1`. If the *first* execution takes longer than `total_timeout_ms` to fail, `timed_out` will be falsely evaluated as `false`, and the runner will erroneously proceed to sleep and execute a second attempt. Java's `Tasks.java` strictly validates `durationMs > maxDurationMs` unconditionally and aborts immediately without attempting a retry. Remove the `&& attempt > 1` condition.

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,180 @@
...
+        std::min(base_delay, static_cast<double>(config_.max_wait_ms)));
+
+    static thread_local std::mt19937 gen(std::random_device{}());
+    int32_t jitter_range = std::max(1, delay_ms / 10);
+    std::uniform_int_distribution<> dis(-jitter_range, jitter_range);
+    delay_ms += dis(gen);
+    return std::max(1, delay_ms);
+  }
+
+  /// \brief Sleep for the specified duration
```
wgtmac: The C++ jitter calculation uses a bidirectional spread `[-jitter_range, jitter_range]`. Java's `Tasks.java` specifically adds a strictly positive jitter: `[0, delayMs * 0.1)`. Consider generating a strictly positive random value `[0, jitter_range]` to align precisely with Java.

src/iceberg/update/update_snapshot_reference.h
```diff
@@ -134,6 +134,8 @@ class ICEBERG_EXPORT UpdateSnapshotReference : public PendingUpdate {
 
   Kind kind() const final { return Kind::kUpdateSnapshotReference; }
 
+  bool IsRetryable() const override { return false; }
```
wgtmac: Overriding `IsRetryable()` to explicitly return `false` causes `Transaction::CanRetry()` to fail any transaction containing branch or tag updates on conflicts. In Java, `SnapshotManager.commit()` utilizes `transaction.commitTransaction()`, which safely retries `UpdateSnapshotReferencesOperation`. Branch and tag creations should be retryable. Consider removing this override or returning `true`.

wgtmac (changes requested):
I've carefully reviewed the retry mechanism and found a few parity issues and a structural data-loss concern regarding how pending updates are held during retries. Please see the inline comments.

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,180 @@
...
+  void Sleep(int32_t ms) const {
+    std::this_thread::sleep_for(std::chrono::milliseconds(ms));
+  }
+
+  RetryConfig config_;
+  std::optional<std::vector<ErrorKind>> only_retry_on_;
+  std::optional<std::vector<ErrorKind>> stop_retry_on_;
```
zhjwpku: Can we just use vector here?

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  RetryRunner& WithExponentialBackoff(int32_t min_wait_ms, int32_t max_wait_ms,
+                                      int32_t total_timeout_ms, double scale_factor) {
+    config_.min_wait_ms = min_wait_ms;
+    config_.max_wait_ms = max_wait_ms;
+    config_.total_timeout_ms = total_timeout_ms;
+    config_.scale_factor = scale_factor;
+    return *this;
+  }
```
WZhuo: I think it can be simply use a constructor with `RetryConfig`

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,180 @@
...
+
+  /// \brief Specify error types that should trigger a retry
+  RetryRunner& OnlyRetryOn(std::initializer_list<ErrorKind> error_kinds) {
```
WZhuo: What if some error exist in `only_retry_on_` and also in `stop_retry_on_`? Should we do some check in setting the retry errors?

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,180 @@
...
+ private:
+  /// \brief Check if the given error kind should trigger a retry
+  bool ShouldRetry(ErrorKind kind) const {
+    if (only_retry_on_.has_value()) {
+      for (const auto& retry_kind : only_retry_on_.value()) {
+        if (kind == retry_kind) {
+          return true;
+        }
+      }
+      return false;
```
WZhuo: Can instead of using `std::range::any_of`
```suggestion
      return std::ranges::any_of(only_retry_on_.value(), [kind](ErrorKind retry_kind) {
        return kind == retry_kind;
      });
```

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    if (stop_retry_on_.has_value()) {
+      for (const auto& stop_kind : stop_retry_on_.value()) {
+        if (kind == stop_kind) {
+          return false;
+        }
+      }
```
WZhuo:
```suggestion
      return !std::ranges::any_of(stop_retry_on_.value(), [kind](ErrorKind stop_kind) {
        return kind == stop_kind;
      });
```

src/iceberg/util/retry_util.h
linguoxuan: The current implementation handles this by giving OnlyRetryOn precedence over StopRetryOn, which aligns with the Java upstream behavior in Tasks.java (no runtime check either). I think it's enough since the API is unlikely to be called with both set simultaneously. WDYT？

src/iceberg/update/update_snapshot_reference.h
linguoxuan: When used within an external transaction, SnapshotManager.commit() does not call transaction.commitTransaction():
```
// SnapshotManager.java
public void commit() {
    commitIfRefUpdatesExist();
    if (!isExternalTransaction) { 
        transaction.commitTransaction();
    }
} 
```
The retry then happens in the outer BaseTransaction.commitSimpleTransaction() → applyUpdates(). During retry, applyUpdates() resets this.current = underlyingOps.current() (refreshed metadata), then re-calls update.commit(). But UpdateSnapshotReferencesOperation holds this.base captured at construction time (the old metadata). When commit() calls ops.commit(base, updated), TransactionTableOperations.commit() checks:
```
// BaseTransaction.java
public void commit(TableMetadata underlyingBase, TableMetadata metadata) {
    if (underlyingBase != current) {
        throw new CommitFailedException("Table metadata refresh is required");
    }
    ...
} 
```
CommitFailedException is caught by applyUpdates() and wrapped as PendingUpdateFailedException (which is not CommitFailedException), terminating the retry loop immediately.
```
private void applyUpdates(TableOperations underlyingOps) {
    if (base != underlyingOps.refresh()) {
      // use refreshed the metadata
      this.base = underlyingOps.current();
      this.current = underlyingOps.current();
      for (PendingUpdate update : updates) {
        // re-commit each update in the chain to apply it and update current
        try {
          update.commit();
        } catch (CommitFailedException e) {
          // Cannot pass even with retry due to conflicting metadata changes. So, break the
          // retry-loop.
          throw new PendingUpdateFailedException(e);
        }
      }
    }
  }
```

So Java's UpdateSnapshotReferencesOperation also cannot be retried within an external transaction. Our IsRetryable() = false is consistent with Java's behavior. 
@wgtmac The architectures of C++ and Java differ somewhat; consequently, the modifications required when `IsRetryable` returns `true` could be quite substantial.

src/iceberg/table.cc
```diff
@@ -87,6 +87,7 @@ Status Table::Refresh() {
   ICEBERG_ASSIGN_OR_RAISE(auto refreshed_table, catalog_->LoadTable(identifier_));
   if (metadata_location_ != refreshed_table->metadata_file_location()) {
     metadata_ = std::move(refreshed_table->metadata_);
+    metadata_location_ = std::string(refreshed_table->metadata_file_location());
```
wgtmac, line 90: This looks like a bug that can be fixed separately and merged quickly

src/iceberg/transaction.cc
```diff
@@ -378,6 +389,43 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
   return ctx_->table;
 }
 
+Result<std::shared_ptr<Table>> Transaction::CommitOnce() {
+  auto refresh_result = ctx_->table->Refresh();
```
wgtmac: It seems unnecessary to issue a `Refresh()` call on the first attempt since it may prohibit fast commit. Can we change to only call Refresh during retry?

src/iceberg/transaction.cc
```diff
@@ -378,6 +389,43 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
+  if (!refresh_result.has_value()) {
```
wgtmac: Let's reuse macros like `ICEBERG_RETURN_UNEXPECTED` and `ICEBERG_ASSIGN_OR_RAISE` to write less lines.

src/iceberg/transaction.cc
```diff
@@ -346,23 +347,33 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
-      ICEBERG_ASSIGN_OR_RAISE(requirements, TableRequirements::ForCreateTable(updates));
-    } break;
-    case TransactionKind::kUpdate: {
-      ICEBERG_ASSIGN_OR_RAISE(
-          requirements,
-          TableRequirements::ForUpdateTable(*ctx_->metadata_builder->base(), updates));
-
-    } break;
+  Result<std::shared_ptr<Table>> commit_result;
+  if (!CanRetry()) {
```
wgtmac: The non-retry path and CommitOnce() both build requirements and call UpdateTable, but through different code paths. Consider unifying: even the non-retry case could go through CommitOnce() with num_retries=0, which would eliminate the branching and reduce maintenance burden.

src/iceberg/transaction.cc
```diff
@@ -378,6 +389,43 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
+  auto refresh_result = ctx_->table->Refresh();
+  if (!refresh_result.has_value()) {
+    return std::unexpected(refresh_result.error());
+  }
+
+  if (ctx_->metadata_builder->base() != ctx_->table->metadata().get()) {
+    ctx_->metadata_builder =
+        TableMetadataBuilder::BuildFrom(ctx_->table->metadata().get());
+    for (const auto& update : pending_updates_) {
+      auto commit_status = update->Commit();
```
wgtmac: Should we directly call `this->Apply(*update)` to avoid an indirection?

src/iceberg/update/snapshot_update.cc
```diff
@@ -218,14 +218,27 @@ Result<std::vector<ManifestFile>> SnapshotUpdate::WriteDeleteManifests(
 }
 
 int64_t SnapshotUpdate::SnapshotId() {
-  if (!snapshot_id_.has_value()) {
+  while (!snapshot_id_.has_value() ||
+         base().SnapshotById(snapshot_id_.value()).has_value()) {
```
wgtmac: Doesn't `SnapshotUtil::GenerateSnapshotId` below have the same loop?

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,160 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <algorithm>
+#include <chrono>
+#include <functional>
```
wgtmac: Unused?

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  /// \brief Specify a single error type that should trigger a retry
+  RetryRunner& OnlyRetryOn(ErrorKind error_kind) {
+    only_retry_on_ = std::vector<ErrorKind>{error_kind};
+    return *this;
+  }
+
+  /// \brief Specify error types that should stop retries immediately
+  RetryRunner& StopRetryOn(std::initializer_list<ErrorKind> error_kinds) {
```
wgtmac: It would be good to document the priority between OnlyRetryOn and StopRetryOn.

src/iceberg/update/pending_update.h
```diff
@@ -58,6 +58,9 @@ class ICEBERG_EXPORT PendingUpdate : public ErrorCollector {
   /// \brief Return the kind of this pending update.
   virtual Kind kind() const = 0;
 
+  /// \brief Whether this update can be retried after a commit conflict.
+  virtual bool IsRetryable() const { return true; }
```
wgtmac: I think the Java impl has other types of updates that disallow retry, e.g. UpdateSchema. We might need to evaluate if we want to keep the same behavior.

src/iceberg/test/retry_util_test.cc
```diff
@@ -0,0 +1,312 @@
...
+
+#include "iceberg/result.h"
+#include "iceberg/test/matchers.h"
+
+namespace iceberg {
+
+// --------------------------------------------------------------------------
+// Test: Successful on first attempt — no retries
+// --------------------------------------------------------------------------
+TEST(RetryRunnerTest, SuccessOnFirstAttempt) {
```
wgtmac: We only have test on RetryRunner. Can we add some integration test directly using Table or Transaction?

wgtmac (changes requested):
Thanks for the update and fixing the CI! I've just reviewed it for a quick pass.

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,160 @@
...
+    stop_retry_on_ = std::vector<ErrorKind>(error_kinds);
+    return *this;
+  }
+
+  /// \brief Run a task that returns a Result<T>
+  template <typename F, typename T = typename std::invoke_result_t<F>::value_type>
+  Result<T> Run(F&& task, int32_t* attempt_counter = nullptr) {
+    auto start_time = std::chrono::steady_clock::now();
+    int32_t attempt = 0;
+    int32_t max_attempts = config_.num_retries + 1;
```
wgtmac: Do we need to validate `config_.num_retries`?

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,160 @@
...
+
+  /// \brief Specify error types that should stop retries immediately
+  RetryRunner& StopRetryOn(std::initializer_list<ErrorKind> error_kinds) {
+    stop_retry_on_ = std::vector<ErrorKind>(error_kinds);
+    return *this;
+  }
+
+  /// \brief Run a task that returns a Result<T>
+  template <typename F, typename T = typename std::invoke_result_t<F>::value_type>
+  Result<T> Run(F&& task, int32_t* attempt_counter = nullptr) {
```
wgtmac: Instead of a raw pointer `attempt_counter`, is it better to incorporate metrics reporter that is being added by @evindj?

src/iceberg/util/retry_util.h
```diff
@@ -0,0 +1,160 @@
...
+
+    static thread_local std::mt19937 gen(std::random_device{}());
+    int32_t jitter_range = std::max(1, delay_ms / 10);
+    std::uniform_int_distribution<> dis(0, jitter_range - 1);
+    delay_ms += dis(gen);
+    return std::max(1, delay_ms);
+  }
+
+  /// \brief Sleep for the specified duration
+  void Sleep(int32_t ms) const {
```
wgtmac: This function seems unnecessary

src/iceberg/table.cc
linguoxuan: @wgtmac move to here: https://github.com/apache/iceberg-cpp/pull/612

wgtmac:
Is this ready to review again?

src/iceberg/util/retry_util.h
linguoxuan: @wgtmac  It seems that metrics reporter is not yet available in the codebase. I've added a TODO comment for this.

linguoxuan:
> Is this ready to review again?

Yes. PTAL.

src/iceberg/update/pending_update.h
wgtmac: nit: make it pure virtual to make sure new implementation (especially those cannot be retried) doesn't forget to set it correctly.

wgtmac:
Sorry I've accidentally pushed my local main (which is up to sync) against your main branch and resulted in this PR being closed due to empty diff. Could you either force push your local main or merge https://github.com/linguoxuan/iceberg-cpp/pull/1 to proceed? I've polished the design a little bit and fixed some issues so it is ready to merge. Thanks!

linguoxuan:
> Sorry I've accidentally pushed my local main (which is up to sync) against your main branch and resulted in this PR being closed due to empty diff. Could you either force push your local main or merge [linguoxuan#1](https://github.com/linguoxuan/iceberg-cpp/pull/1) to proceed? I've polished the design a little bit and fixed some issues so it is ready to merge. Thanks!

It's ok. I make a new pr: https://github.com/apache/iceberg-cpp/pull/626

PR #585, opened 2026-03-02
feat: implement truncate max for literals

WZhuo:


src/iceberg/util/truncate_util.cc
```diff
@@ -29,6 +29,93 @@
 namespace iceberg {
 
 namespace {
+constexpr uint32_t kUtf8MaxCodePoint = 0x10FFFF;
+constexpr uint32_t kUtf8MinSurrogate = 0xD800;
+constexpr uint32_t kUtf8MaxSurrogate = 0xDFFF;
+
+bool DecodeUtf8CodePoint(std::string_view source, uint32_t& code_point) {
```
wgtmac:
```suggestion
std::optional<uint32_t> DecodeUtf8CodePoint(std::string_view source) {
```

nit: this might look simpler

src/iceberg/util/truncate_util.cc
```diff
@@ -72,8 +159,85 @@ Literal TruncateLiteralImpl<TypeId::kBinary>(const Literal& literal, int32_t wid
   return Literal::Binary(std::vector<uint8_t>(data.begin(), data.begin() + width));
 }
 
+template <TypeId type_id>
+Result<Literal> TruncateLiteralMaxImpl(const Literal& literal, int32_t width) {
+  std::unreachable();
```
wgtmac: nit: let's return a `NotSupported` error instead of an exception since you already return Result here?

src/iceberg/util/truncate_util.cc
```diff
@@ -72,8 +159,85 @@ Literal TruncateLiteralImpl<TypeId::kBinary>(const Literal& literal, int32_t wid
...
+
+template <>
+Result<Literal> TruncateLiteralMaxImpl<TypeId::kString>(const Literal& literal,
+                                                        int32_t width) {
+  const auto& str = std::get<std::string>(literal.value());
+  ICEBERG_ASSIGN_OR_RAISE(std::string truncated,
+                          TruncateUtils::TruncateUTF8Max(str, width));
+  if (truncated == str) {
+    return literal;
+  }
```
wgtmac:
```suggestion
```

Should we remove this block? It does not run faster than line 176.

wgtmac (approved):

src/iceberg/util/truncate_util.cc
WZhuo: Indeed it is.

src/iceberg/util/truncate_util.cc
WZhuo: Use `= delete` for unsupported types, give out errors in compile time.

wgtmac (approved):

Issue #322, opened 2025-11-15
Fix CI build warnings


wgtmac:
There are a lot of CI warnings that are worth fixing. For example: https://github.com/apache/iceberg-cpp/actions/runs/19386197700/job/55473420291?pr=320

PR #587, opened 2026-03-08
build: Fix various compiler warnings and uninitialized members

Joe-Abraham:
This PR fixes various build warnings reported during CI as part of resolving [#322](https://github.com/apache/iceberg-cpp/issues/322).

To make the review process easier, the fixes have been broken down into the following logical commits:

1. Fix unused parameter warnings: Added [[maybe_unused]] to unused parameters across abstract interface implementations and visitors, suppressing compiler warnings.
2. Fix uninitialized member warnings: Added default initializers (e.g., =, = nullptr, or {}) to struct and class members to ensure deterministic initialization.
3. Fix integer sign comparison and loop counter warnings: Transitioned to std::cmp_equal, std::cmp_less, std::cmp_not_equal, etc., from <utility>, and ensured loop counters appropriately use size_t where necessary.
4. Fix mixed build warnings: Addressed specific operator precedence warnings (e.g., adding parentheses around || operations in std::isnan checks in literal.cc) and other mixed build issues.
5. Fix redundant std::move warnings: Removed unnecessary std::move calls that prevent copy elision (e.g., on projected->term()) in projection utilities.
6. Fix shadowed function and class/struct friend declaration warnings: Renamed shadowed Equals methods to EqualsSchema, and converted friend class errors to friend struct where applicable.
7. Remove unused functions: Dropped unused local/helper functions from schema_util_test.cc that were triggering unused-function warnings.


src/iceberg/catalog/rest/types.h
```diff
@@ -39,9 +39,9 @@ namespace iceberg::rest {
 
 /// \brief Server-provided configuration for the catalog.
 struct ICEBERG_REST_EXPORT CatalogConfig {
-  std::unordered_map<std::string, std::string> defaults;   // required
-  std::unordered_map<std::string, std::string> overrides;  // required
-  std::vector<Endpoint> endpoints;
+  std::unordered_map<std::string, std::string> defaults{};   // required
```
wgtmac: Do we really need to explicitly initialize STL containers?

src/iceberg/catalog/rest/types.h
```diff
@@ -142,13 +142,13 @@ struct ICEBERG_REST_EXPORT RenameTableRequest {
...
 /// \brief Request to create a table.
 struct ICEBERG_REST_EXPORT CreateTableRequest {
-  std::string name;  // required
-  std::string location;
-  std::shared_ptr<Schema> schema;  // required
-  std::shared_ptr<PartitionSpec> partition_spec;
-  std::shared_ptr<SortOrder> write_order;
+  std::string name = "";  // required
+  std::string location = "";
+  std::shared_ptr<Schema> schema = nullptr;  // required
```
wgtmac: Same question here, do we actually need this?

wgtmac (commented):
Thanks for taking this! Before proceeding, I think we need to sort out what kind of fixes are required. In addition, it would be good to enforce these warnings in the ci so we don't need to fix similar issues again.

PR #590, opened 2026-03-12
feat: Impl IncrementalAppendScan

WZhuo:


src/iceberg/table_scan.h
```diff
@@ -29,6 +29,7 @@
 
 #include "iceberg/arrow_c_data.h"
 #include "iceberg/result.h"
+#include "iceberg/table_metadata.h"
```
wgtmac, line 32: Can we remove this? Is forward declaration enough?

src/iceberg/table_scan.cc
```diff
@@ -539,17 +641,75 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
...
+  std::unordered_set<int64_t> snapshot_ids;
+  std::ranges::transform(append_snapshots,
+                         std::inserter(snapshot_ids, snapshot_ids.end()),
+                         [](const auto& snapshot) { return snapshot->snapshot_id; });
+
+  std::vector<ManifestFile> data_manifests;
+  for (const auto& snapshot : append_snapshots) {
+    SnapshotCache snapshot_cache(snapshot.get());
+    ICEBERG_ASSIGN_OR_RAISE(auto manifests, snapshot_cache.DataManifests(io_));
+    std::ranges::copy_if(manifests, std::back_inserter(data_manifests),
```
wgtmac: In Java's `BaseIncrementalAppendScan.java`, manifests from append snapshots are collected into a `Set<ManifestFile>` to prevent duplicate processing if multiple snapshots reference the same manifest. The C++ implementation pushes directly into `std::vector<ManifestFile> data_manifests`, which will cause duplicate processing if a manifest is retained across multiple append snapshots since `ManifestGroup::Make` does not deduplicate them internally.

**Suggestion:** Deduplicate `data_manifests` (e.g., by tracking seen `manifest_path`s using a `std::unordered_set<std::string>`) before passing them to `ManifestGroup::Make`.

wgtmac (commented):
> *Note: This review was generated by Gemini.*

src/iceberg/table_scan.h
```diff
@@ -132,6 +133,17 @@ struct TableScanContext {
 
   // Validate the context parameters to see if they have conflicts.
   [[nodiscard]] Status Validate() const;
+
+  /// \brief Returns true if this scan is a current lineage scan, which means it does not
+  /// specify from/to snapshot IDs.
+  bool IsScanCurrentLineage() const;
```
wgtmac: Can we just move these three new functions to the anonymous namespace in the table_scan.cc?

src/iceberg/table_scan.h
```diff
@@ -373,6 +383,26 @@ class ICEBERG_EXPORT IncrementalScan : public TableScan {
   using TableScan::TableScan;
 };
 
+// Template method implementation (must be in header for MSVC)
+template <typename ScanTaskType>
+Result<std::vector<std::shared_ptr<ScanTaskType>>>
+IncrementalScan<ScanTaskType>::PlanFiles() const {
+  if (context_.IsScanCurrentLineage()) {
+    ICEBERG_ASSIGN_OR_RAISE(auto current_snapshot, metadata_->Snapshot());
```
wgtmac: Do we need to consider if branch has been specified?

src/iceberg/table_scan.h
WZhuo: `IncrementalScan<ScanTaskType>::PlanFiles()` method must implement in header file, it compiles error in msvc env if implement it in source file. Of course, It's better not include it, will see how to fix it.

src/iceberg/table_scan.h
WZhuo: I made these functions members because I don't want to keep writing context.xxx inside the method bodies

src/iceberg/table_scan.h
WZhuo: If `current_snapshot` is nullptr, it indicates that the current table is empty, so we can simply return an empty task list. No need care about the `branch`.

src/iceberg/table_scan.cc
```diff
@@ -536,20 +639,110 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
   return manifest_group->PlanFiles();
 }
 
+// Friend function template for IncrementalScan that implements the shared PlanFiles
+// logic. It resolves the from/to snapshot range from the scan context and delegates
+// to the two-arg virtual PlanFiles() override in the concrete subclass.
+// Defined as a friend to access the protected two-arg PlanFiles().
+template <typename ScanTaskType>
+Result<std::vector<std::shared_ptr<ScanTaskType>>> ResolvePlanFiles(
```
wgtmac, line 647: **Logic / Parity Issue:** `scan.metadata()->Snapshot()` returns a `NotFound` error if the table has no current snapshot (e.g., table is empty), rather than a successful result with `nullptr`. `ICEBERG_ASSIGN_OR_RAISE` will immediately propagate this `NotFound` error, making the `if (current_snapshot == nullptr)` check unreachable. This breaks parity with Java, which handles empty tables gracefully by returning an empty iterable.

*Suggested fix:* Check the ID directly: `if (scan.metadata()->current_snapshot_id == kInvalidSnapshotId) { return ...; }` or use `scan.snapshot()` which properly resolves to `nullptr`.

*(Note: The same issue exists on line 214 in `ToSnapshotIdInclusive`, where `ICEBERG_ASSIGN_OR_RAISE` will propagate `NotFound` before the custom error message can be triggered).*

src/iceberg/manifest/manifest_list.h
```diff
@@ -272,3 +272,12 @@ ICEBERG_EXPORT inline constexpr Result<ManifestContent> ManifestContentFromStrin
 }
 
 }  // namespace iceberg
+
+namespace std {
+template <>
+struct hash<iceberg::ManifestFile> {
```
wgtmac, line 280: **Style / Logic Issue:** The added `std::hash<iceberg::ManifestFile>` specialization only hashes `manifest_path`. However, `ManifestFile::operator==` defaults to member-wise equality. If an `unordered_set` (like the one used in `IncrementalAppendScan::PlanFiles`) encounters two `ManifestFile` objects with the same path but slightly different metadata fields, they will have the same hash but fail equality, resulting in duplicates being inserted. In Java, `GenericManifestFile` uses *only* `manifestPath` for both `hashCode()` and `equals()`.

*Suggested fix:* Modify `ManifestFile::operator==` to only compare `manifest_path` to align with Java, or provide a custom equality functor specifically for the `unordered_set<ManifestFile>`.

wgtmac (changes requested):
> *Note: This review was generated by Gemini.*

## Summary & Recommendation
*   **Request Changes:** There are logic bugs in handling empty table snapshots and a potential hashing/equality discrepancy for manifest files that need to be addressed.

wgtmac (approved):

PR #490, opened 2026-01-06
feat: support expire snapshots

dongxiao1198:
This PR is part1 implements to support expire snapshots described in issue https://github.com/apache/iceberg-cpp/issues/364.

1 Add ExpireSnapshot in TableMetadataBuilder,Table,Transaction...
2 Add basic UT for  TableMetadataBuilder and E2E with in memory catalog

TODO: FileCleanupStrategy and file recycle need another independent PR to do，and need more UT after all part finished.

PR #592, opened 2026-03-16
feat: add ReachableFileCleanup to expire snapshots

shangxinli:
## Summary

- Implement the file cleanup logic missing from expire snapshots (#490 noted "TODO: File recycling will be added in a followup PR")
- Port the "reachable file cleanup" strategy from Java's `ReachableFileCleanup`
- Single-threaded implementation; multi-threaded and incremental cleanup as TODOs

### DIFF
--- a/src/iceberg/update/expire_snapshots.cc
+++ b/src/iceberg/update/expire_snapshots.cc
@@ -49,7 +49,8 @@ Result<std::shared_ptr<ManifestReader>> MakeManifestReader(
     const ManifestFile& manifest, const std::shared_ptr<FileIO>& file_io,
     const TableMetadata& metadata) {
   ICEBERG_ASSIGN_OR_RAISE(auto schema, metadata.Schema());
-  ICEBERG_ASSIGN_OR_RAISE(auto spec, metadata.PartitionSpecById(manifest.partition_spec_id));
+  ICEBERG_ASSIGN_OR_RAISE(auto spec,
+                          metadata.PartitionSpecById(manifest.partition_spec_id));
   return ManifestReader::Make(manifest, file_io, std::move(schema), std::move(spec));
 }
 
@@ -98,8 +99,7 @@ class FileCleanupStrategy {
   /// if the same file path is shared across snapshots, it is only deleted when
   /// no retained snapshot references it.
   std::unordered_set<std::string> ExpiredStatisticsFilePaths(
-      const TableMetadata& metadata,
-      const std::unordered_set<int64_t>& expired_ids) {
+      const TableMetadata& metadata, const std::unordered_set<int64_t>& expired_ids) {
     std::unordered_set<std::string> retained_paths;
     for (const auto& stat : metadata.statistics) {
       if (stat && !expired_ids.contains(stat->snapshot_id)) {
@@ -568,9 +568,8 @@ Status ExpireSnapshots::Finalize(std::optional<Error> commit_error) {
     return {};
   }
 
-  std::unordered_set<int64_t> expired_ids(
-      apply_result_->snapshot_ids_to_remove.begin(),
-      apply_result_->snapshot_ids_to_remove.end());
+  std::unordered_set<int64_t> expired_ids(apply_result_->snapshot_ids_to_remove.begin(),
+                                          apply_result_->snapshot_ids_to_remove.end());
   apply_result_.reset();
 
   // File cleanup is best-effort: log and continue on individual file deletion failures
--- a/src/iceberg/update/expire_snapshots.cc
+++ b/src/iceberg/update/expire_snapshots.cc
@@ -33,6 +33,7 @@
 #include "iceberg/manifest/manifest_reader.h"
 #include "iceberg/schema.h"
 #include "iceberg/snapshot.h"
+#include "iceberg/statistics_file.h"
 #include "iceberg/table.h"
 #include "iceberg/table_metadata.h"
 #include "iceberg/transaction.h"
@@ -54,9 +55,6 @@ Result<std::shared_ptr<ManifestReader>> MakeManifestReader(
 }
 
 /// \brief Abstract strategy for cleaning up files after snapshot expiration.
-///
-/// Mirrors Java's FileCleanupStrategy: provides shared delete utilities while
-/// allowing different cleanup algorithms (ReachableFileCleanup, IncrementalFileCleanup).
 class FileCleanupStrategy {
  public:
   FileCleanupStrategy(std::shared_ptr<FileIO> file_io,
@@ -67,18 +65,17 @@ class FileCleanupStrategy {
 
   /// \brief Clean up files that are only reachable by expired snapshots.
   ///
-  /// \param metadata Table metadata before expiration (contains all snapshots).
+  /// \param metadata_before_expiration Table metadata before expiration.
+  /// \param metadata_after_expiration Table metadata after expiration.
   /// \param expired_snapshot_ids Snapshot IDs that were expired during this operation.
   /// \param level Controls which types of files are eligible for deletion.
-  virtual Status CleanFiles(const TableMetadata& metadata,
+  virtual Status CleanFiles(const TableMetadata& metadata_before_expiration,
+                            const TableMetadata& metadata_after_expiration,
                             const std::unordered_set<int64_t>& expired_snapshot_ids,
                             CleanupLevel level) = 0;
 
  protected:
-  /// \brief Delete a file, suppressing errors (best-effort).
-  ///
-  /// Uses the custom delete function if set, otherwise FileIO::DeleteFile.
-  /// Matches Java's suppressFailureWhenFinished behavior.
+  /// \brief Delete a single file
   void DeleteFile(const std::string& path) {
     try {
       if (delete_func_) {
@@ -87,9 +84,52 @@ class FileCleanupStrategy {
         std::ignore = file_io_->DeleteFile(path);
       }
     } catch (...) {
-      // Suppress all exceptions during file cleanup to match Java's
-      // suppressFailureWhenFinished behavior.
+      /// TODO(shangxinli): add retry
     }
+  }
+
+  /// TODO(shangxinli): Add bulk deletion
+  void DeleteFiles(const std::unordered_set<std::string>& paths) {
+    for (const auto& path : paths) {
+      DeleteFile(path);
+    }
+  }
+
+  bool HasAnyStatisticsFiles(const TableMetadata& metadata) const {
+    return !metadata.statistics.empty() || !metadata.partition_statistics.empty();
+  }
+
+  std::unordered_set<std::string> StatisticsFilesToDelete(
+      const TableMetadata& metadata_before_expiration,
+      const TableMetadata& metadata_after_expiration) const {
+    std::unordered_set<std::string> stats_files_to_delete;
+    std::unordered_set<std::string> live_stats_paths;
+
+    for (const auto& stats_file : metadata_after_expiration.statistics) {
+      if (stats_file) {
+        live_stats_paths.insert(stats_file->path);
+      }
+    }
+
+    for (const auto& part_stats_file : metadata_after_expiration.partition_statistics) {
+      if (part_stats_file) {
+        live_stats_paths.insert(part_stats_file->path);
+      }
+    }
+
+    for (const auto& stats_file : metadata_before_expiration.statistics) {
+      if (stats_file && !live_stats_paths.contains(stats_file->path)) {
+        stats_files_to_delete.insert(stats_file->path);
+      }
+    }
+
+    for (const auto& part_stats_file : metadata_before_expiration.partition_statistics) {
+      if (part_stats_file && !live_stats_paths.contains(part_stats_file->path)) {
+        stats_files_to_delete.insert(part_stats_file->path);
+      }
+    }
+
+    return stats_files_to_delete;
   }
 
   std::shared_ptr<FileIO> file_io_;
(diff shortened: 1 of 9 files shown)


src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+    std::unordered_set<std::string>& data_files_to_delete) {
+  const TableMetadata& metadata = base();
+  auto file_io = ctx_->table->io();
+
+  // Step 1: Collect all file paths from manifests being deleted
+  for (const auto& manifest_path : manifests_to_delete) {
+    // Find the ManifestFile for this path by scanning expired snapshots
+    for (const auto& snapshot : metadata.snapshots) {
+      if (!snapshot) continue;
+      SnapshotCache snapshot_cache(snapshot.get());
```
wgtmac: **Performance / Logic Issue:** Instantiating `SnapshotCache` locally inside the nested loop causes O(M*S) I/O complexity. The cache is not preserved across loop iterations, meaning for *every* manifest path being processed, the manifest list of *every* snapshot is repeatedly downloaded and parsed from storage.

*Suggestion:* Consider collecting `ManifestFile` objects in Phase 1 (similar to Java's `Set<ManifestFile> deletionCandidates`) or pre-caching them in a map to avoid redundant I/O.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+        auto schema_result = metadata.Schema();
+        if (!schema_result.has_value()) continue;
+        auto spec_result = metadata.PartitionSpecById(manifest.partition_spec_id);
+        if (!spec_result.has_value()) continue;
+
+        auto reader_result = ManifestReader::Make(
+            manifest, file_io, schema_result.value(), spec_result.value());
+        if (!reader_result.has_value()) continue;
+
+        auto entries_result = reader_result.value()->Entries();
```
wgtmac: **Parity / Logic Issue:** C++ uses `Entries()` which includes `DELETED` entries, whereas Java uses `ManifestFiles.readPaths` which delegates to `liveEntries()` (only `ADDED` and `EXISTING`).

*Details:* If a data file is added in an expired snapshot and deleted in a retained snapshot, using `Entries()` will read the DELETED entry in Phase 2 and subtract the data file from `data_files_to_delete`. This prevents the physical file from being deleted, resulting in a storage leak.

*Suggestion:* Change `reader_result.value()->Entries()` to `reader_result.value()->LiveEntries()` to strictly match Java's behavior.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+  }
+
+  // Step 2: Remove any files that are still referenced by retained manifests.
+  // This ensures we don't delete files that are shared across manifests.
+  for (const auto& manifest_path : retained_manifests) {
+    if (data_files_to_delete.empty()) break;
+
+    for (const auto& snapshot : metadata.snapshots) {
+      if (!snapshot) continue;
+      SnapshotCache snapshot_cache(snapshot.get());
```
wgtmac: **Performance / Logic Issue:** Similar to line 371, instantiating `SnapshotCache` here inside the loop repeats the O(M*S) I/O complexity.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+        auto schema_result = metadata.Schema();
+        if (!schema_result.has_value()) continue;
+        auto spec_result = metadata.PartitionSpecById(manifest.partition_spec_id);
+        if (!spec_result.has_value()) continue;
+
+        auto reader_result = ManifestReader::Make(
+            manifest, file_io, schema_result.value(), spec_result.value());
+        if (!reader_result.has_value()) continue;
+
+        auto entries_result = reader_result.value()->Entries();
```
wgtmac: **Parity / Logic Issue:** Similar to line 387, use `LiveEntries()` here instead of `Entries()` to avoid a potential storage leak.

wgtmac (changes requested):
> *Note: This review was generated by Gemini.*

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+      DeleteFilePath(snapshot->manifest_list);
+    }
+  }
+
+  // Phase 6: Delete expired statistics files.
+  // Use set difference between before and after states (matching Java behavior).
+  // Since Finalize runs before table_ is updated, "after" is base() minus expired.
+  std::unordered_set<int64_t> retained_stats_snapshots(retained_snapshot_ids);
+  for (const auto& stat_file : metadata.statistics) {
+    if (stat_file && !retained_stats_snapshots.contains(stat_file->snapshot_id)) {
```
wgtmac: **Inconsistent Behavior:** C++ deletes statistics files by checking if the `snapshot_id` of the `StatisticsFile` is in the `retained_stats_snapshots` set. Java computes the set difference of actual **file paths**. If a statistics file path is referenced by multiple snapshots, deleting purely based on the expiration of a specific `snapshot_id` could erroneously delete a physical file that is still referenced by a newer, retained snapshot. Diff paths, not just snapshot IDs.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+        if (manifest.manifest_path != manifest_path) continue;
+
+        auto schema_result = metadata.Schema();
+        if (!schema_result.has_value()) continue;
+        auto spec_result = metadata.PartitionSpecById(manifest.partition_spec_id);
+        if (!spec_result.has_value()) continue;
+
+        auto reader_result = ManifestReader::Make(
+            manifest, file_io, schema_result.value(), spec_result.value());
+        if (!reader_result.has_value()) continue;
```
wgtmac: **Inconsistent Behavior (Data Loss Risk):** If reading a **retained** manifest fails, the C++ implementation silently ignores it (`continue`). Java uses `.retry(3)` and `.throwFailureWhenFinished()`. If we silently ignore a read failure here, we will fail to subtract its live data files from `data_files_to_delete`, resulting in **accidental data loss** (deleting a physical file that is still actively used). Failures here should abort the deletion of those specific data files.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+
+        auto entries_result = reader_result.value()->Entries();
+        if (!entries_result.has_value()) continue;
+
+        for (const auto& entry : entries_result.value()) {
+          if (entry.data_file) {
+            data_files_to_delete.insert(entry.data_file->file_path);
+          }
+        }
+        goto next_manifest;  // Found and processed this manifest, move to next
```
wgtmac: **C++ Style Issue:** The use of `goto next_manifest;` to break out of nested loops is a non-idiomatic C++ anti-pattern. Consider moving this manifest lookup logic into a helper function (e.g., `std::optional<ManifestFile> GetManifestByPath(path)`).

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,247 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+    manifest_paths.insert(manifest.manifest_path);
+  }
+
+  return {};
+}
+
+Status ExpireSnapshots::FindDataFilesToDelete(
+    const std::unordered_set<std::string>& manifests_to_delete,
+    const std::unordered_set<std::string>& retained_manifests,
+    std::unordered_set<std::string>& data_files_to_delete) {
```
wgtmac: **C++ Style Issue:** Using mutable out-parameters (`std::unordered_set<std::string>& data_files_to_delete`) over `Result` is a less modern C++ pattern. Consider returning `Result<std::unordered_set<std::string>>`.

wgtmac (commented):
> *Note: This review was generated by Gemini.*

src/iceberg/update/expire_snapshots.cc
shangxinli: Good catch.

src/iceberg/update/expire_snapshots.cc
shangxinli: changed to `LiveEntries()` to match Java's `ManifestFiles.readPaths` which delegates to `liveEntries()` (only ADDED and EXISTING). Using `Entries()` would include DELETED entries and cause storage leaks.

src/iceberg/update/expire_snapshots.cc
shangxinli: Same as #1

src/iceberg/update/expire_snapshots.cc
shangxinli: Fixed — changed to `LiveEntries()` here as well, consistent with Step 1.

src/iceberg/update/expire_snapshots.cc
shangxinli: Good point. Changed to path-based set difference: collect paths from retained snapshots first, then only delete expired statistics files whose paths are not in the retained set. This prevents erroneous deletion when a statistics file path is shared across snapshots

src/iceberg/update/expire_snapshots.cc
shangxinli: Fixed

src/iceberg/update/expire_snapshots.cc
shangxinli: Removed. The nested loop + goto was eliminated by introducing `manifest_cache_` (populated in Phase 1) and a `MakeManifestReader()` helper. `FindDataFilesToDelete()` now iterates over the cache directly with simple map lookups.

src/iceberg/update/expire_snapshots.cc
shangxinli: Changed `FindDataFilesToDelete` to return `Result<std::unordered_set<std::string>>` instead of using a mutable out-parameter.

src/iceberg/update/expire_snapshots.h
```diff
@@ -22,10 +22,14 @@
 #include <cstdint>
 #include <functional>
 #include <memory>
+#include <string>
+#include <unordered_map>
 #include <unordered_set>
 #include <vector>
 
 #include "iceberg/iceberg_export.h"
+#include "iceberg/manifest/manifest_list.h"
```
wgtmac: Can we remove these includes by using forward declaration? `iceberg/type_fwd.h` has already been included.

src/iceberg/update/expire_snapshots.h
```diff
@@ -169,6 +224,14 @@ class ICEBERG_EXPORT ExpireSnapshots : public PendingUpdate {
...
   bool clean_expired_metadata_{false};
   bool specified_snapshot_id_{false};
+
+  /// Cached result from Apply(), used during Finalize() for file cleanup
+  std::optional<ApplyResult> apply_result_;
+
+  /// Cache of manifest path -> ManifestFile, built during ReadManifestsForSnapshot
+  /// to avoid O(M*S) repeated I/O from re-reading manifest lists in
+  /// FindDataFilesToDelete.
+  std::unordered_map<std::string, ManifestFile> manifest_cache_;
```
wgtmac: This is the internal and ephemeral state used when cleaning up files. Can we remove it from here and move to the source file (or concrete FileCleanupStrategy class)?

src/iceberg/update/expire_snapshots.h
```diff
@@ -159,6 +173,47 @@ class ICEBERG_EXPORT ExpireSnapshots : public PendingUpdate {
...
+  /// by any branch or tag, as computed by Apply(). This prevents deleting files
+  /// that are still reachable from any ref.
+  ///
+  /// TODO(shangxinli): Add multi-threaded file deletion support.
+  /// TODO(shangxinli): Add IncrementalFileCleanup strategy for linear ancestry.
+  Status CleanExpiredFiles(const std::vector<int64_t>& expired_snapshot_ids);
+
+  /// \brief Read manifest paths from a single snapshot.
+  /// Best-effort: returns OK even if the snapshot or its manifests can't be read.
+  Status ReadManifestsForSnapshot(int64_t snapshot_id,
```
wgtmac: It seems worth adding a similar abstraction like the Java `FileCleanupStrategy` if we will add `IncrementalFileCleanup` or more strategies in the future. We can add them only in the expire_snapshots.cc for now as it is unlikely to be used elsewhere.

I'm also thinking if it is valuable to enable users to customize any strategy by registering their own implementations but that's a separate topic.

src/iceberg/update/expire_snapshots.h
```diff
@@ -159,6 +173,47 @@ class ICEBERG_EXPORT ExpireSnapshots : public PendingUpdate {
...
+                                  std::unordered_set<std::string>& manifest_paths);
+
+  /// \brief Find data files to delete by reading live entries from manifests being
+  /// deleted, then subtracting files still reachable from retained manifests.
+  /// If a retained manifest cannot be read, returns an empty set to prevent
+  /// accidental data loss.
+  Result<std::unordered_set<std::string>> FindDataFilesToDelete(
```
wgtmac: This function seems to be related to ReachableFileCleanup only.

src/iceberg/update/expire_snapshots.h
```diff
@@ -159,6 +173,47 @@ class ICEBERG_EXPORT ExpireSnapshots : public PendingUpdate {
...
+      const std::unordered_set<std::string>& manifests_to_delete,
+      const std::unordered_set<std::string>& retained_manifests);
+
+  /// \brief Create a ManifestReader for the given ManifestFile.
+  Result<std::shared_ptr<ManifestReader>> MakeManifestReader(
```
wgtmac: This function doesn't justify to be a member function.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,250 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+    return {};
+  }
+
+  if (!apply_result_.has_value() || apply_result_->snapshot_ids_to_remove.empty()) {
+    return {};
+  }
+
+  // File cleanup is best-effort: log and continue on individual file deletion failures
+  // to avoid blocking metadata updates (matching Java behavior).
+  return CleanExpiredFiles(apply_result_->snapshot_ids_to_remove);
```
wgtmac: Should we clear `apply_result_` if `CleanExpiredFiles` succeeds?

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,250 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+}
+
+void ExpireSnapshots::DeleteFilePath(const std::string& path) {
```
wgtmac:
```suggestion
void ExpireSnapshots::DeleteFile(const std::string& path) {
```

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,250 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+  try {
+    if (delete_func_) {
+      delete_func_(path);
+    } else {
+      auto status = ctx_->table->io()->DeleteFile(path);
+      // Best-effort: ignore NotFound (file already deleted) and other errors.
+      // Java uses suppressFailureWhenFinished + onFailure logging.
+      std::ignore = status;
```
wgtmac:
```suggestion
      std::ignore = ctx_->table->io()->DeleteFile(path);
```

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,250 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+    }
+  } catch (...) {
+    // Suppress all exceptions during file cleanup to match Java's
+    // suppressFailureWhenFinished behavior.
+  }
+}
+
+Result<std::shared_ptr<ManifestReader>> ExpireSnapshots::MakeManifestReader(
```
wgtmac: Make it a local function in the anonymous namespace?

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,250 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+    const ManifestFile& manifest, const std::shared_ptr<FileIO>& file_io) {
+  const TableMetadata& metadata = base();
+  auto schema_result = metadata.Schema();
+  if (!schema_result.has_value()) return std::unexpected<Error>(schema_result.error());
+  auto spec_result = metadata.PartitionSpecById(manifest.partition_spec_id);
+  if (!spec_result.has_value()) return std::unexpected<Error>(spec_result.error());
+  return ManifestReader::Make(manifest, file_io, schema_result.value(),
+                              spec_result.value());
```
wgtmac:
```suggestion
  ICEBERG_ASSIGN_OR_RAISE(auto schema, metadata.Schema());
  ICEBERG_ASSIGN_OR_RAISE(auto spec, metadata.PartitionSpecById(manifest.partition_spec_id));
  return ManifestReader::Make(manifest, file_io, std::move(schema), std::move(spec));
```

src/iceberg/update/expire_snapshots.cc
```diff
@@ -285,7 +291,250 @@ Result<ExpireSnapshots::ApplyResult> ExpireSnapshots::Apply() {
...
+    if (snapshot && !expired_id_set.contains(snapshot->snapshot_id)) {
+      retained_snapshot_ids.insert(snapshot->snapshot_id);
+    }
+  }
+
+  // Phase 1: Collect manifest paths from expired and retained snapshots.
+  // TODO(shangxinli): Parallelize manifest collection with a thread pool.
+  std::unordered_set<std::string> expired_manifest_paths;
+  for (int64_t snapshot_id : expired_snapshot_ids) {
+    std::ignore = ReadManifestsForSnapshot(snapshot_id, expired_manifest_paths);
```
wgtmac: Why we can ignore the error? Isn't it safe to abort entire operation?

src/iceberg/update/expire_snapshots.h
shangxinli: Done — removed manifest_list.h and manifest_reader.h from the header.

src/iceberg/update/expire_snapshots.h
shangxinli: Done — introduced FileCleanupStrategy abstract base class and ReachableFileCleanup concrete subclass in the anonymous namespace of expire_snapshots.cc, mirroring Java's class hierarchy. They live in the .cc for now since they're unlikely to be used elsewhere; we can promote to a header when IncrementalFileCleanup is added.

src/iceberg/update/expire_snapshots.h
shangxinli: Done — ReadManifestsForSnapshot is now a private method of ReachableFileCleanup only.

src/iceberg/update/expire_snapshots.h
shangxinli: Done — MakeManifestReader is now a free function in the anonymous namespace

src/iceberg/update/expire_snapshots.cc
shangxinli: ReadManifestsForSnapshot now returns void since it is intentionally best-effort. For expired snapshots, a read failure means we might miss some file deletions, but orphaned files can be cleaned up by GC later. This is the conservative direction (never delete what we can't confirm is unreachable). For retained manifests in FindDataFilesToDelete, failures still abort data file deletion entirely to prevent accidental data loss, matching Java's throwFailureWhenFinished behavior there.

src/iceberg/update/expire_snapshots.cc
shangxinli: Done — MakeManifestReader is now a free function in the anonymous namespace.

src/iceberg/update/expire_snapshots.h
shangxinli: Done — manifest_cache_ is now a private member of ReachableFileCleanup in expire_snapshots.cc. It no longer appears in the header at all.

shangxinli:
@wgtmac do you have time to have another look?

src/iceberg/update/expire_snapshots.cc
```diff
@@ -37,6 +43,264 @@
...
+  }
+
+ private:
+  /// Cache of manifest path -> ManifestFile, populated during Phase 1 to avoid
+  /// re-reading manifest lists in FindDataFilesToDelete.
+  std::unordered_map<std::string, ManifestFile> manifest_cache_;
+
+  /// \brief Collect manifest paths for a snapshot into manifest_paths.
+  ///
+  /// Best-effort: if the snapshot or its manifest list cannot be read, the error
```
wgtmac: [P0] Retained-manifest read failures can delete live manifests. `ReadManifestsForSnapshot` silently returns when a retained snapshot's manifest list cannot be read, but phase 2 still computes `manifests_to_delete` from that incomplete retained set and phase 4 deletes those manifests. That means a transient read failure while enumerating the current table state can delete manifest files that are still referenced by live snapshots. Java's `ReachableFileCleanup` aborts cleanup on this path (`throwFailureWhenFinished`) specifically to avoid deleting live metadata, so we should fail or skip manifest deletion entirely when retained manifests cannot be enumerated.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -37,6 +43,264 @@
...
+    for (int64_t snapshot_id : expired_snapshot_ids) {
+      auto snapshot_result = metadata.SnapshotById(snapshot_id);
+      if (!snapshot_result.has_value()) continue;
+      const auto& snapshot = snapshot_result.value();
+      if (!snapshot->manifest_list.empty()) {
+        DeleteFile(snapshot->manifest_list);
+      }
+    }
+
+    // Phase 6: Delete expired statistics files using path-based set difference.
```
wgtmac: [P1] Stats files are deleted even though snapshot removal still leaves their metadata entries behind. This PR now deletes expired statistics and partition-statistics files in `Finalize`, but C++ snapshot removal still does not remove the corresponding `statistics` / `partition_statistics` metadata entries (`src/iceberg/table_metadata.cc` still has a FIXME instead of calling `RemoveStatistics` / `RemovePartitionStatistics`). That means a table can commit metadata that still references a stats file and then immediately delete the file on disk. Java removes those metadata entries as part of snapshot removal and has regression coverage for expiration and reuse of stats files. Please either wire snapshot removal to drop the metadata entries first or defer physical stats-file deletion until that is implemented.

wgtmac (commented):
Codex review comments for the two issues found in expire snapshots cleanup.

src/iceberg/update/expire_snapshots.cc
shangxinli: Fixed. ReadManifestsForSnapshot now returns bool. In CleanFiles, I track retained_manifests_complete; if any retained snapshot's manifest list can't be read the entire manifest/data-file deletion block (phases 2–4) is skipped, matching Java's throwFailureWhenFinished guard. Manifest-list deletion (phase 5) is unaffected since it depends only on expired snapshots.

src/iceberg/update/expire_snapshots.cc
shangxinli: Fixed

wgtmac (approved):
Thanks @shangxinli for adding this!

PR #600, opened 2026-03-24
feat(rest): Parse JWT exp claim from token in AuthProperties

wgtmac:
- Implement Base64UrlDecode in TransformUtil.
- Update AuthProperties::FromProperties to parse the JWT exp claim from the token and set expires_at_millis_.
- Add expires_at_millis() getter to AuthProperties.
- Add unit tests for Base64UrlDecode and JWT expiration parsing.

### DIFF
--- a/src/iceberg/catalog/rest/auth/auth_properties.cc
+++ b/src/iceberg/catalog/rest/auth/auth_properties.cc
@@ -21,7 +21,11 @@
 
 #include <utility>
 
+#include <nlohmann/json.hpp>
+
 #include "iceberg/catalog/rest/catalog_properties.h"
+#include "iceberg/util/macros.h"
+#include "iceberg/util/transform_util.h"
 
 namespace iceberg::rest::auth {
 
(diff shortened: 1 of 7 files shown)


src/iceberg/util/transform_util.cc
```diff
@@ -283,4 +283,40 @@ std::string TransformUtil::Base64Encode(std::string_view str_to_encode) {
...
+    if (c >= 'A' && c <= 'Z')
+      v = static_cast<int8_t>(c - 'A');
+    else if (c >= 'a' && c <= 'z')
+      v = static_cast<int8_t>(c - 'a' + 26);
+    else if (c >= '0' && c <= '9')
+      v = static_cast<int8_t>(c - '0' + 52);
+    else if (c == '-' || c == '+')
+      v = 62;
+    else if (c == '_' || c == '/')
+      v = 63;
```
lishuxu, line 304: `Base64UrlDecode` accepts `+` and `/` which are standard Base64 characters, not Base64Url. This means inputs that should be rejected are silently accepted, which contradicts the function name and the error message.

  Base64Url spec only uses `-` (62) and `_` (63). `+` and `/` should fall through to the invalid character check.

  Suggested fix:

      else if (c == '-')
        v = 62;
      else if (c == '_')
        v = 63;

src/iceberg/catalog/rest/auth/auth_properties.cc
```diff
@@ -75,7 +79,25 @@ Result<AuthProperties> AuthProperties::FromProperties(
...
+    auto first_dot = token.find('.');
+    auto last_dot = token.find('.', first_dot + 1);
+    if (first_dot != std::string::npos && last_dot != std::string::npos) {
+      auto payload_encoded = token.substr(first_dot + 1, last_dot - first_dot - 1);
+      auto payload_decoded = TransformUtil::Base64UrlDecode(payload_encoded);
+      if (payload_decoded.has_value()) {
+        try {
+          auto payload_json = nlohmann::json::parse(payload_decoded.value());
+          if (payload_json.contains("exp") && payload_json["exp"].is_number()) {
+            config.expires_at_millis_ = payload_json["exp"].get<int64_t>() * 1000;
```
lishuxu, line 93: `is_number()` returns true for both integers and floats, but `get<int64_t>()`
  throws `nlohmann::json::type_error` when the value is a float (e.g. `"exp": 1735689600.0`).
  This exception is not caught by the `parse_error` handler, so it would propagate
  unexpectedly.

src/iceberg/catalog/rest/auth/auth_properties.cc
```diff
@@ -75,7 +79,25 @@ Result<AuthProperties> AuthProperties::FromProperties(
     }
   }
 
-  // TODO(lishuxu): Parse JWT exp claim from token to set expires_at_millis_.
+  // Parse JWT exp claim from token to set expires_at_millis_.
+  if (auto token = config.token(); !token.empty()) {
+    auto first_dot = token.find('.');
+    auto last_dot = token.find('.', first_dot + 1);
```
lishuxu, line 85: nit: `last_dot` is a bit misleading — it could imply `rfind` was used.
  Since JWT has exactly three parts, `second_dot` is more precise.

wgtmac:
I'm closing this. @lishuxu feel free to open a PR when you are ready.

PR #603, opened 2026-03-24
feat(puffin): add format constants, utilities, and JSON serialization

zhaoxuan1994:
Second PR in the Puffin series (following #588).

puffin_format.h/.cc: format constants, footer flag operations, little-endian I/O, compress/decompress dispatch (LZ4/Zstd stubbed as NotSupported)
puffin_json_internal.h/.cc: JSON serialization for BlobMetadata and FileMetadata

### DIFF
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -66,6 +66,8 @@ set(ICEBERG_SOURCES
     partition_spec.cc
     partition_summary.cc
     puffin/file_metadata.cc
+    puffin/puffin_format.cc
+    puffin/json_serde.cc
     row/arrow_array_wrapper.cc
     row/manifest_wrapper.cc
     row/partition_values.cc
--- a/src/iceberg/puffin/json_serde_internal.h
+++ b/src/iceberg/puffin/json_serde_internal.h
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/puffin/json_serde_internal.h
+/// JSON serialization/deserialization for Puffin file metadata.
+
+#include <string>
+#include <string_view>
+
+#include <nlohmann/json_fwd.hpp>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/puffin/type_fwd.h"
+#include "iceberg/result.h"
+
+namespace iceberg::puffin {
+
+/// \brief Serialize a BlobMetadata to JSON.
+ICEBERG_EXPORT nlohmann::json ToJson(const BlobMetadata& blob_metadata);
+
+/// \brief Deserialize a BlobMetadata from JSON.
+ICEBERG_EXPORT Result<BlobMetadata> BlobMetadataFromJson(const nlohmann::json& json);
+
+/// \brief Serialize a FileMetadata to JSON.
+ICEBERG_EXPORT nlohmann::json ToJson(const FileMetadata& file_metadata);
+
+/// \brief Deserialize a FileMetadata from JSON.
+ICEBERG_EXPORT Result<FileMetadata> FileMetadataFromJson(const nlohmann::json& json);
+
+/// \brief Serialize a FileMetadata to a JSON string.
+ICEBERG_EXPORT std::string ToJsonString(const FileMetadata& file_metadata,
+                                        bool pretty = false);
+
+/// \brief Deserialize a FileMetadata from a JSON string.
+ICEBERG_EXPORT Result<FileMetadata> FileMetadataFromJsonString(
+    std::string_view json_string);
+
+}  // namespace iceberg::puffin
--- a/src/iceberg/puffin/puffin_format.cc
+++ b/src/iceberg/puffin/puffin_format.cc
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/puffin/puffin_format.h"
+
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+namespace iceberg::puffin {
+
+namespace {
+
+// Returns (byte_index, bit_index) for a given flag within the 4-byte flags field.
+constexpr std::pair<int, int> GetFlagPosition(PuffinFlag flag) {
+  switch (flag) {
+    case PuffinFlag::kFooterPayloadCompressed:
+      return {0, 0};
+  }
+  std::unreachable();
+}
+
+// TODO(zhaoxuan1994): Move compression logic to a unified codec interface.
+Result<std::vector<std::byte>> Compress(PuffinCompressionCodec codec,
+                                        std::span<const std::byte> input) {
+  switch (codec) {
+    case PuffinCompressionCodec::kNone:
+      return std::vector<std::byte>(input.begin(), input.end());
+    case PuffinCompressionCodec::kLz4:
+      return NotSupported("LZ4 compression is not yet supported");
+    case PuffinCompressionCodec::kZstd:
+      return NotSupported("Zstd compression is not yet supported");
+  }
+  std::unreachable();
+}
+
+Result<std::vector<std::byte>> Decompress(PuffinCompressionCodec codec,
+                                          std::span<const std::byte> input) {
+  switch (codec) {
+    case PuffinCompressionCodec::kNone:
+      return std::vector<std::byte>(input.begin(), input.end());
+    case PuffinCompressionCodec::kLz4:
+      return NotSupported("LZ4 decompression is not yet supported");
+    case PuffinCompressionCodec::kZstd:
+      return NotSupported("Zstd decompression is not yet supported");
+  }
+  std::unreachable();
+}
+
+}  // namespace
+
+bool IsFlagSet(std::span<const uint8_t, 4> flags, PuffinFlag flag) {
+  auto [byte_num, bit_num] = GetFlagPosition(flag);
+  return (flags[byte_num] & (1 << bit_num)) != 0;
+}
+
+void SetFlag(std::span<uint8_t, 4> flags, PuffinFlag flag) {
+  auto [byte_num, bit_num] = GetFlagPosition(flag);
+  flags[byte_num] |= (1 << bit_num);
+}
+
+}  // namespace iceberg::puffin
--- a/src/iceberg/puffin/puffin_format.h
+++ b/src/iceberg/puffin/puffin_format.h
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/puffin/puffin_format.h
+/// Puffin file format constants and utilities.
+
+#include <array>
+#include <cstdint>
+#include <span>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/puffin/file_metadata.h"
+#include "iceberg/result.h"
+
+namespace iceberg::puffin {
+
+/// \brief Puffin file format constants.
+struct ICEBERG_EXPORT PuffinFormat {
+  /// Magic bytes: "PFA1" (Puffin Fratercula arctica, version 1)
+  static constexpr std::array<uint8_t, 4> kMagicV1 = {0x50, 0x46, 0x41, 0x31};
+
+  static constexpr int32_t kMagicLength = 4;
+  static constexpr int32_t kFooterStartMagicOffset = 0;
+  static constexpr int32_t kFooterStartMagicLength = kMagicLength;
+  static constexpr int32_t kFooterStructPayloadSizeOffset = 0;
+  static constexpr int32_t kFooterStructFlagsOffset = kFooterStructPayloadSizeOffset + 4;
+  static constexpr int32_t kFooterStructFlagsLength = 4;
+  static constexpr int32_t kFooterStructMagicOffset =
+      kFooterStructFlagsOffset + kFooterStructFlagsLength;
+
+  /// Total length of the footer struct: payload_size(4) + flags(4) + magic(4)
+  static constexpr int32_t kFooterStructLength = kFooterStructMagicOffset + kMagicLength;
+
+  /// Default compression codec for footer payload.
+  static constexpr PuffinCompressionCodec kDefaultFooterCompressionCodec =
+      PuffinCompressionCodec::kLz4;
+};
+
+/// \brief Footer flags for Puffin files.
+enum class PuffinFlag : uint8_t {
+  /// Whether the footer payload is compressed.
+  kFooterPayloadCompressed = 0,
+};
+
+/// \brief Check if a flag is set in the flags bytes.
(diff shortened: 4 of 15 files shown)


src/iceberg/puffin/puffin_format.h
```diff
@@ -0,0 +1,89 @@
...
+ICEBERG_EXPORT bool IsFlagSet(std::span<const uint8_t, 4> flags, PuffinFlag flag);
+
+/// \brief Set a flag in the flags bytes.
+ICEBERG_EXPORT void SetFlag(std::span<uint8_t, 4> flags, PuffinFlag flag);
+
+/// \brief Write a 32-bit integer in little-endian format.
+ICEBERG_EXPORT void WriteInt32LittleEndian(int32_t value, std::span<uint8_t, 4> output);
+
+/// \brief Read a 32-bit integer from a fixed-size span in little-endian format.
+ICEBERG_EXPORT int32_t ReadInt32LittleEndian(std::span<const uint8_t, 4> input);
```
emkornfield: There is very code for this in the roaring bitmap code it should probably standardized in a common place?

src/iceberg/puffin/puffin_format.h
```diff
@@ -0,0 +1,89 @@
...
+ICEBERG_EXPORT int32_t ReadInt32LittleEndian(std::span<const uint8_t> data,
+                                             int32_t offset);
+
+/// \brief Compress data using the specified codec.
+ICEBERG_EXPORT Result<std::vector<uint8_t>> Compress(PuffinCompressionCodec codec,
+                                                     std::span<const uint8_t> input);
+
+/// \brief Decompress data using the specified codec.
+ICEBERG_EXPORT Result<std::vector<uint8_t>> Decompress(PuffinCompressionCodec codec,
+                                                       std::span<const uint8_t> input);
```
emkornfield: Roaring bitmap is using string_view as collection of bytes, we should probably standardize on one of them.

src/iceberg/puffin/puffin_json_internal.h
```diff
@@ -0,0 +1,55 @@
...
+/// \brief Deserialize a FileMetadata from JSON.
+ICEBERG_EXPORT Result<FileMetadata> FileMetadataFromJson(const nlohmann::json& json);
+
+/// \brief Serialize a FileMetadata to a JSON string.
+ICEBERG_EXPORT std::string ToJsonString(const FileMetadata& file_metadata,
+                                        bool pretty = false);
+
+/// \brief Deserialize a FileMetadata from a JSON string.
+ICEBERG_EXPORT Result<FileMetadata> FileMetadataFromJsonString(
+    const std::string& json_string);
```
emkornfield: nit: std::string_view?

src/iceberg/test/puffin_json_test.cc
emkornfield, line 1: seems like tests should be seperated per .cc?

src/iceberg/puffin/puffin_format.h
```diff
@@ -0,0 +1,89 @@
...
+  static constexpr int32_t kFooterStructFlagsLength = 4;
+  static constexpr int32_t kFooterStructMagicOffset =
+      kFooterStructFlagsOffset + kFooterStructFlagsLength;
+
+  /// Total length of the footer struct: payload_size(4) + flags(4) + magic(4)
+  static constexpr int32_t kFooterStructLength = kFooterStructMagicOffset + kMagicLength;
+
+  /// Default compression codec for footer payload.
+  static constexpr PuffinCompressionCodec kFooterCompressionCodec =
+      PuffinCompressionCodec::kLz4;
```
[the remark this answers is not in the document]
zhaoxuan1994: Keeping kLz4 as it reflects the Puffin spec's default footer compression codec, consistent with the Java implementation. This constant is a spec-level declaration — actual usage in the writer will check the Result from Compress(), so unsupported codecs surface as explicit errors.

src/iceberg/test/puffin_json_test.cc
zhaoxuan1994: Done

src/iceberg/puffin/puffin_json_internal.h
zhaoxuan1994: Done

src/iceberg/puffin/puffin_format.h
zhaoxuan1994: Done. Added WriteLittleEndian/ReadLittleEndian templates to endian.h as public API. Updated both puffin_format.cc and roaring_position_bitmap.cc to use them, removing the duplicated local helpers.

src/iceberg/puffin/puffin_format.h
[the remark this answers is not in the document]
emkornfield: I think this is actually a real compatibility concern where java puffin readers would likely break if a compressed footer was ever properly written.  Java has never implemented lz4 compression: https://github.com/apache/iceberg/blob/8a51a685894eace51474f512b46a187566f27202/core/src/main/java/org/apache/iceberg/puffin/PuffinFormat.java#L112 which IIUC means anything written with a compressed footer would not be readable from Java.  If you agree with this assessment it might be misreading the code but might be worth a discussion on the mailing list.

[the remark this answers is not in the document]
emkornfield: Looking more closely it looks like java would just throw unsupported codec for compressed footers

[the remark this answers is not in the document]
emkornfield: I agree the spec calls for lz4, so there is maybe nothing to be done here

src/iceberg/puffin/puffin_format.h
zhaoxuan1994: I used span<const uint8_t> here since it feels more natural for binary data. But I can see the argument for string_view given that CRoaring already uses std::string for serialization. Do you have a preference on which one the project should standardize on?

emkornfield: I don't.  Another things to decide is if we are using C++17 or above if [std::byte](https://en.cppreference.com/w/cpp/types/byte.html) should be used instead.  This is probably something that should be discussed on the mailing list I would guess, or at least get @wgtmac and other committers opinion on.

src/iceberg/CMakeLists.txt
```diff
@@ -64,6 +64,8 @@ set(ICEBERG_SOURCES
     partition_spec.cc
     partition_summary.cc
     puffin/file_metadata.cc
+    puffin/puffin_format.cc
+    puffin/puffin_json_internal.cc
```
wgtmac: Let's rename to `puffin/json_serde_internal.h` and `puffin/json_serde.cc` respectively for consistency.

src/iceberg/puffin/puffin_format.h
```diff
@@ -0,0 +1,89 @@
...
+#include "iceberg/iceberg_export.h"
+#include "iceberg/puffin/file_metadata.h"
+#include "iceberg/result.h"
+
+namespace iceberg::puffin {
+
+/// \brief Puffin file format constants.
+struct ICEBERG_EXPORT PuffinFormat {
+  /// Magic bytes: "PFA1" (Puffin Fratercula arctica, version 1)
+  static constexpr std::array<uint8_t, 4> kMagic = {0x50, 0x46, 0x41, 0x31};
```
wgtmac:
```suggestion
  static constexpr std::array<uint8_t, 4> kMagicV1 = {0x50, 0x46, 0x41, 0x31};
```

Let's make V1 explicit.

src/iceberg/puffin/puffin_format.h
```diff
@@ -0,0 +1,89 @@
...
+  static constexpr int32_t kFooterStructFlagsOffset = kFooterStructPayloadSizeOffset + 4;
+  static constexpr int32_t kFooterStructFlagsLength = 4;
+  static constexpr int32_t kFooterStructMagicOffset =
+      kFooterStructFlagsOffset + kFooterStructFlagsLength;
+
+  /// Total length of the footer struct: payload_size(4) + flags(4) + magic(4)
+  static constexpr int32_t kFooterStructLength = kFooterStructMagicOffset + kMagicLength;
+
+  /// Default compression codec for footer payload.
+  static constexpr PuffinCompressionCodec kFooterCompressionCodec =
```
wgtmac:
```suggestion
  static constexpr PuffinCompressionCodec kDefaultFooterCompressionCodec =
```

src/iceberg/test/CMakeLists.txt
```diff
@@ -125,6 +125,10 @@ add_iceberg_test(util_test
 
 add_iceberg_test(roaring_test SOURCES roaring_test.cc)
 
+add_iceberg_test(puffin_format_test SOURCES puffin_format_test.cc)
+
+add_iceberg_test(puffin_json_test SOURCES puffin_json_test.cc)
```
wgtmac: Can we merge them into a single puffin_test executable?

src/iceberg/puffin/puffin_format.h
```diff
@@ -0,0 +1,89 @@
...
+/// \brief Read a 32-bit integer from a fixed-size span in little-endian format.
+ICEBERG_EXPORT int32_t ReadInt32LittleEndian(std::span<const uint8_t, 4> input);
+
+/// \brief Read a 32-bit integer from a buffer at the given offset in little-endian
+/// format.
+ICEBERG_EXPORT int32_t ReadInt32LittleEndian(std::span<const uint8_t> data,
+                                             int32_t offset);
+
+/// \brief Compress data using the specified codec.
+ICEBERG_EXPORT Result<std::vector<uint8_t>> Compress(PuffinCompressionCodec codec,
```
wgtmac: It is not the right place to put compression logic here as this can be shared by TableMetadata compression as well. We can consider a single place to unify the codec implementation (as well as the codec availability).

src/iceberg/puffin/puffin_format.h
wgtmac: We're on C++23 so `std::byte` is definitely available. Together with my previous comment for a generic codec interface/implementation, we can use `std::span<const std::byte>` as input and `std::vector<std::byte>` as output but with overloads that accepting `std::string_view` and returning `std::string`, etc.

src/iceberg/puffin/puffin_format.cc
```diff
@@ -0,0 +1,91 @@
...
+#include <utility>
+
+#include "iceberg/util/endian.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg::puffin {
+
+namespace {
+
+constexpr std::pair<int, int> GetFlagPosition(PuffinFlag flag) {
```
wgtmac, line 31: Add a comment to improve the readability?

src/iceberg/puffin/puffin_format.cc
```diff
@@ -0,0 +1,91 @@
...
+void WriteInt32LittleEndian(int32_t value, std::span<uint8_t, 4> output) {
+  WriteLittleEndian(value, output.data());
+}
+
+int32_t ReadInt32LittleEndian(std::span<const uint8_t, 4> input) {
+  return ReadLittleEndian<int32_t>(input.data());
+}
+
+int32_t ReadInt32LittleEndian(std::span<const uint8_t> data, int32_t offset) {
+  ICEBERG_DCHECK(offset >= 0, "Offset must be non-negative");
```
wgtmac: ICEBERG_DCHECK is a debug check and should only be used in a rare case. If this is a generic function and cannot guarantee the input, we should return `Result<int32_t>` and replace it by `ICEBERG_PRECHECK` here.

src/iceberg/test/puffin_format_test.cc
```diff
@@ -0,0 +1,69 @@
...
+#include <cstdint>
+#include <vector>
+
+#include <gtest/gtest.h>
+
+namespace iceberg::puffin {
+
+TEST(PuffinFormatTest, ByteOrderRoundTrip) {
+  std::array<uint8_t, 4> buf{};
+  WriteInt32LittleEndian(0x12345678, buf);
```
wgtmac: Should these functions be moved to endian.h as they are not puffin specific?

src/iceberg/test/puffin_json_test.cc
```diff
@@ -0,0 +1,167 @@
...
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
+
+#include "iceberg/puffin/file_metadata.h"
+#include "iceberg/puffin/puffin_json_internal.h"
+
+namespace iceberg::puffin {
+
+TEST(PuffinJsonTest, BlobMetadataRoundTrip) {
+  BlobMetadata blob;
```
wgtmac: nit: use aggregate initialization.

src/iceberg/test/puffin_json_test.cc
```diff
@@ -0,0 +1,167 @@
...
+    "length": 200,
+    "compression-codec": "zstd",
+    "properties": {"key": "value"}
+  })"_json;
+
+  EXPECT_EQ(ToJson(blob), expected_json);
+
+  auto result = BlobMetadataFromJson(expected_json);
+  ASSERT_TRUE(result.has_value());
+  EXPECT_EQ(result.value(), blob);
```
wgtmac: nit: merge these two checks into `ASSERT_THAT`

src/iceberg/test/puffin_json_test.cc
```diff
@@ -0,0 +1,167 @@
...
+
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
+
+#include "iceberg/puffin/file_metadata.h"
+#include "iceberg/puffin/puffin_json_internal.h"
+
+namespace iceberg::puffin {
+
+TEST(PuffinJsonTest, BlobMetadataRoundTrip) {
```
wgtmac: We can improve these test cases by refactoring them into parameterized test with more valid and invalid cases. A lot of other test files have done similar jobs.

wgtmac (changes requested):

src/iceberg/puffin/puffin_format.h
emkornfield: @wgtmac I guess my main concern here would be to try to maintain internal consistency within all the code in the project.  For consuming use-cases the overloads might be worth it.  I wonder if it would be too premature to consider having a Buffer object?

wgtmac: I'm hesitant to introduce a Buffer (and corresponding MemoryPool) at this moment. They are only used as ephemeral input/output buffers for CompressionCodec and FileIO.

src/iceberg/util/endian.h
```diff
@@ -94,4 +95,19 @@ constexpr T FromBigEndian(T value) {
   }
 }
 
+/// \brief Write a value in little-endian format to a buffer.
```
wgtmac, line 98: It would be good to add a comment to say that it is caller's responsibility to guarantee that `output` has sufficient length, otherwise it is a UB.

src/iceberg/util/endian.h
```diff
@@ -94,4 +95,19 @@ constexpr T FromBigEndian(T value) {
...
+template <EndianConvertible T>
+void WriteLittleEndian(T value, void* output) {
+  auto le = ToLittleEndian(value);
+  std::memcpy(output, &le, sizeof(le));
+}
+
+/// \brief Read a value in little-endian format from a buffer.
+template <EndianConvertible T>
+T ReadLittleEndian(const void* input) {
```
wgtmac, line 109: ditto

src/iceberg/puffin/file_metadata.h
```diff
@@ -107,6 +108,10 @@ struct ICEBERG_EXPORT BlobMetadata {
 
 ICEBERG_EXPORT std::string ToString(const BlobMetadata& blob_metadata);
 
+inline std::ostream& operator<<(std::ostream& os, const BlobMetadata& b) {
```
wgtmac: We don't have any precedence to add this. It is preferable to use std::format wherever possible.

src/iceberg/puffin/json_serde_internal.h
```diff
@@ -0,0 +1,56 @@
...
+/// \file iceberg/puffin/json_serde_internal.h
+/// JSON serialization/deserialization for Puffin file metadata.
+
+#include <string>
+#include <string_view>
+
+#include <nlohmann/json_fwd.hpp>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/puffin/file_metadata.h"
```
wgtmac: Can we remove this and add a `iceberg/puffin/type_fwd.h` for forward declarations?

src/iceberg/puffin/puffin_format.h
```diff
@@ -0,0 +1,74 @@
...
+
+/// \brief Check if a flag is set in the flags bytes.
+ICEBERG_EXPORT bool IsFlagSet(std::span<const uint8_t, 4> flags, PuffinFlag flag);
+
+/// \brief Set a flag in the flags bytes.
+ICEBERG_EXPORT void SetFlag(std::span<uint8_t, 4> flags, PuffinFlag flag);
+
+/// \brief Read a 32-bit integer from a buffer at the given offset in little-endian
+/// format.
+ICEBERG_EXPORT Result<int32_t> ReadInt32LittleEndian(std::span<const uint8_t> data,
```
wgtmac: Should we move this to `endian.h`? If not, I'd be inclined to remove this for now and then add it back as an internal function in the source file where it is called.

src/iceberg/puffin/json_serde_internal.h
```diff
@@ -0,0 +1,56 @@
...
+ICEBERG_EXPORT Result<BlobMetadata> BlobMetadataFromJson(const nlohmann::json& json);
+
+/// \brief Serialize a FileMetadata to JSON.
+ICEBERG_EXPORT nlohmann::json ToJson(const FileMetadata& file_metadata);
+
+/// \brief Deserialize a FileMetadata from JSON.
+ICEBERG_EXPORT Result<FileMetadata> FileMetadataFromJson(const nlohmann::json& json);
+
+/// \brief Serialize a FileMetadata to a JSON string.
+ICEBERG_EXPORT std::string ToJsonString(const FileMetadata& file_metadata,
```
wgtmac, line 49: Do we really need to specialize this?

src/iceberg/test/puffin_format_test.cc
```diff
@@ -0,0 +1,51 @@
...
+#include <cstdint>
+
+#include <gtest/gtest.h>
+
+#include "iceberg/util/endian.h"
+
+namespace iceberg::puffin {
+
+TEST(PuffinFormatTest, ByteOrderRoundTrip) {
+  std::array<uint8_t, 4> buf{};
```
wgtmac: This test seems unrelated to puffin. Move it to `endian_test.cc`?

src/iceberg/test/puffin_json_test.cc
```diff
@@ -0,0 +1,193 @@
...
+
+TEST_P(InvalidBlobMetadataJsonTest, DeserializeFails) {
+  auto json = nlohmann::json::parse(GetParam().json);
+  EXPECT_THAT(BlobMetadataFromJson(json), IsError(ErrorKind::kJsonParseError));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    PuffinJson, InvalidBlobMetadataJsonTest,
+    ::testing::Values(
+        InvalidBlobMetadataJsonParam{.name = "MissingType",
```
wgtmac, line 110: Add a MissingSequenceNumber test case for completeness.

src/iceberg/puffin/json_serde_internal.h
zhaoxuan1994: ToJsonString/FileMetadataFromJsonString are convenience wrappers for the reader/writer — they handle the json::parse/json::dump + error wrapping that would otherwise be duplicated at each call site. Happy to remove if you prefer callers to use ToJson().dump() directly.

wgtmac (approved):

PR #607, opened 2026-03-29
feat(inspect): add base metadata table interface

All-less:
This PR kicks off the implementation of table inspection support.

- Add `BaseMetadataTable` class (without scan support for now).
- Add `MetadataTableFactory` as an entry point for creating metadata tables.
- Add `SnapshotsTable` and `HistoryTable` as example impl.


src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,180 @@
...
+namespace iceberg {
+
+/// Forward declarations
+class FileIO;
+
+/// \brief Base class for Iceberg metadata tables
+///
+/// Metadata tables expose table metadata as queryable tables with schemas and scan
+/// support. They provide read-only access to metadata.
+class ICEBERG_EXPORT BaseMetadataTable : public Table {
```
wgtmac: Let's inherit `StaticTable` to make it read-only. What do you think if we simply name it as `MetadataTable`?


BTW, it seems that `StaticTable` misses some overrides of update functions (this can be a separate PR to fix).

src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,180 @@
...
+ public:
+  /// \brief Returns the identifier of this table
+  const TableIdentifier& name() const { return identifier_; }
```
wgtmac: Functions like these shadow those defined by base class. We may need to make them virtual to support overrides.

src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,180 @@
...
+  Status Refresh() override;
+
+  /// \brief Create a new table scan builder for this table
+  ///
+  /// Once a table scan builder is created, it can be refined to project columns and
+  /// filter data.
+  Result<std::unique_ptr<TableScanBuilder>> NewScan() const;
+
+  /// \brief Creating transactions is not supported in metadata tables.
+  Result<std::shared_ptr<Transaction>> NewTransaction() override;
```
wgtmac: Same as above comment. Let's remove these functions by inheriting `StaticTable`.

src/iceberg/inspect/metadata_table.cc
```diff
@@ -0,0 +1,87 @@
...
+      source_table_(std::move(source_table)),
+      schema_(schema) {
+  uuid_ = Uuid::GenerateV4().ToString();
+  schemas_[schema->schema_id()] = schema;
+}
+
+BaseMetadataTable::~BaseMetadataTable() = default;
+
+Status BaseMetadataTable::Refresh() {
+  return NotSupported("Cannot refresh a metadata table");
```
wgtmac: This can be supported by refreshing the base table.

src/iceberg/CMakeLists.txt
```diff
@@ -59,12 +60,12 @@ set(ICEBERG_SOURCES
     manifest/v2_metadata.cc
     manifest/v3_metadata.cc
     metadata_columns.cc
+    inspect/metadata_table.cc
```
wgtmac: Let's sort them alphabetically.

src/iceberg/inspect/snapshots_table.h
```diff
@@ -0,0 +1,57 @@
...
+/// - parent_id (long)
+/// - manifest_list (string)
+/// - summary (map<string, string>)
+class ICEBERG_EXPORT SnapshotsTable : public BaseMetadataTable {
+ public:
+  /// \brief Create a SnapshotsTable from table metadata
+  ///
+  /// \param[in] table The source table
+  /// \return A SnapshotsTable instance or error status
+  static Result<std::shared_ptr<SnapshotsTable>> Make(std::shared_ptr<Table> table);
```
wgtmac:
```suggestion
  static Result<std::unique_ptr<SnapshotsTable>> Make(std::shared_ptr<Table> table);
```

wgtmac: It is better to return unique_ptr unless unrealistic.

src/iceberg/inspect/snapshots_table.h
```diff
@@ -0,0 +1,57 @@
...
+
+  ~SnapshotsTable() override;
+
+ private:
+  SnapshotsTable(std::shared_ptr<Table> table);
```
wgtmac:
```suggestion
  explicit SnapshotsTable(std::shared_ptr<Table> table);
```

src/iceberg/inspect/snapshots_table.h
```diff
@@ -0,0 +1,57 @@
...
+
+  std::shared_ptr<Schema> CreateSchema();
```
wgtmac: This does not have to be a member function.

src/iceberg/inspect/snapshots_table.cc
```diff
@@ -0,0 +1,61 @@
...
+
+SnapshotsTable::SnapshotsTable(std::shared_ptr<Table> table)
+    : BaseMetadataTable(table, CreateName(table->name()), CreateSchema()) {}
+
+SnapshotsTable::~SnapshotsTable() = default;
+
+std::shared_ptr<Schema> SnapshotsTable::CreateSchema() {
+  return std::make_shared<Schema>(
+      std::vector<SchemaField>{SchemaField::MakeRequired(1, "committed_at", int64()),
+                               SchemaField::MakeOptional(2, "snapshot_id", int64()),
```
wgtmac: Let's make sure the schema definition is consistent with the Java implementation.

src/iceberg/inspect/history_table.cc
```diff
@@ -0,0 +1,56 @@
...
+
+HistoryTable::HistoryTable(std::shared_ptr<Table> table)
+    : BaseMetadataTable(table, CreateName(table->name()), CreateSchema()) {}
+
+HistoryTable::~HistoryTable() = default;
+
+std::shared_ptr<Schema> HistoryTable::CreateSchema() {
+  return std::make_shared<Schema>(
+      std::vector<SchemaField>{
+          SchemaField::MakeRequired(1, "made_current_at", int64()),
```
wgtmac:
```suggestion
          SchemaField::MakeRequired(1, "made_current_at", timestamptz()),
```

src/iceberg/inspect/snapshots_table.h
```diff
@@ -0,0 +1,57 @@
...
+
+  TableIdentifier CreateName(const TableIdentifier& source_name);
```
wgtmac: ditto

src/iceberg/CMakeLists.txt
```diff
@@ -44,6 +42,9 @@ set(ICEBERG_SOURCES
     expression/term.cc
     file_reader.cc
     file_writer.cc
+    inspect/history_table.cc
```
wgtmac, line 44: Add them to meson.build as well.

src/iceberg/inspect/metadata_table_factory.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <memory>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+#include "iceberg/table.h"
+
+namespace iceberg {
+
+class HistoryTable;
```
wgtmac: Add them to `iceberg/type_fwd.h` and include it here.

src/iceberg/inspect/metadata_table_factory.h
```diff
@@ -0,0 +1,61 @@
...
+///
+/// MetadataTable provides factory methods to create specific metadata tables for
+/// inspecting table metadata. Each metadata table exposes a different aspect of the
+/// table's metadata as a scannable Iceberg table.
+///
+/// Usage:
+///   auto snapshots = ICEBERG_TRY(MetadataTable::GetSnapshotsTable(table));
+///   auto scan = ICEBERG_TRY(snapshots->NewScan());
+///   // ... scan and read snapshot data
+class ICEBERG_EXPORT MetadataTableFactory {
```
wgtmac: Let's introduce an enum MetadataTableType as well. Each MetadataTable instance should also add a function to return this type.

wgtmac: By convention, we don't add a new file for the factory. Consider moving them to metadata_table.h/cc.

wgtmac (commented):
Thanks for working on this! I've left some preliminary comments.

src/iceberg/type_fwd.h
```diff
@@ -221,6 +204,11 @@ class UpdateSnapshotReference;
 class UpdateSortOrder;
 class UpdateStatistics;
 
+/// \brief Metadata tables.
```
wgtmac, line 243: Please sort them alphabetically

src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /// \brief Refreshing is not supported in metadata tables.
+  Status Refresh() override;
+
+  /// \brief Create a new table scan builder for this table
+  ///
+  /// Once a table scan builder is created, it can be refined to project columns and
+  /// filter data.
+  Result<std::unique_ptr<TableScanBuilder>> NewScan() const;
```
wgtmac: Forget to mark `override`?

src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,191 @@
...
+
+ protected:
+  explicit MetadataTable(std::shared_ptr<Table> source_table, TableIdentifier identifier);
+
+  ~MetadataTable();
+
+  std::shared_ptr<Table> source_table_;
+  std::string uuid_;
+  std::shared_ptr<Schema> schema_;
```
wgtmac, line 58: Member variables here and below are not necessary

src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,191 @@
...
+  /// \brief Returns the identifier of this table
+  const TableIdentifier& name() const override { return identifier_; }
+
+  virtual MetadataTableType type() const noexcept = 0;
+
+  /// \brief Returns the UUID of the table
+  const std::string& uuid() const { return uuid_; }
+
+  /// \brief Returns the schema for this table, return NotFoundError if not found
+  Result<std::shared_ptr<Schema>> schema() const { return schema_; }
```
wgtmac: Let's remove these redefinitions here and below

src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,191 @@
...
+///   auto snapshots = ICEBERG_TRY(MetadataTable::GetSnapshotsTable(table));
+///   auto scan = ICEBERG_TRY(snapshots->NewScan());
+///   // ... scan and read snapshot data
+class ICEBERG_EXPORT MetadataTableFactory {
+ public:
+  /// \brief Create a SnapshotsTable from a table
+  ///
+  /// \param table The source table
+  /// \return A SnapshotsTable exposing all snapshots or error status
+  static Result<std::unique_ptr<SnapshotsTable>> GetSnapshotsTable(
```
wgtmac: It looks cumbersome to use a function per type. Can we use enum metadata type or canonical metadata table name suffix instead to return a `std::unique_ptr<Table>` instead?

wgtmac: BTW, do we want to support the case when users directly create a metadata table without having a table object? For example, users may directly create a metadata table from catalog by calling LoadTable and providing full table identifier with metadata table suffix.

wgtmac:
Please rebase to resolve conflicts. Let me know when it is ready to review again.

src/iceberg/inspect/metadata_table.h
All-less: The purpose of those variables is that the MetadataTable won't have the same uuid/schema/etc as the source table. If we remove them, the MetadataTable will return those properties directly from the source table's TableMetadata.

src/iceberg/inspect/metadata_table.h
All-less: > BTW, do we want to support the case when users directly create a metadata table without having a table object? For example, users may directly create a metadata table from catalog by calling LoadTable and providing full table identifier with metadata table suffix.

This breaks the assumption that MetadataTables are companion objects to an existing table. Allowing this might call for unnecessary consideration (for example, how to read the history/snapshots of a non-existent table).

src/iceberg/inspect/metadata_table.h
wgtmac: I agree that `uuid_` and `schema_` are required. Other variables below do not seem necessary to me.

src/iceberg/inspect/metadata_table.h
wgtmac: This is a valid use case where users directly query against the metadata table. You may refer to https://github.com/apache/iceberg/blob/29ba0a14dc6667db0683fdfcd520639b3da77774/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L504 and see Spark use case here: https://iceberg.apache.org/docs/latest/spark-queries/#inspecting-tables

All-less: I see what you mean. I think we will support it on the Catalog level, i.e., let the Catalog load the base table and then create the metadata table. The MetadataTable APIs won't directly support identifier-only creation.

wgtmac:
Could you please make ci happy (and possibly rebase on the latest main branch)?

src/iceberg/inspect/metadata_table.cc
```diff
@@ -0,0 +1,95 @@
...
+
+namespace iceberg {
+
+MetadataTable::MetadataTable(std::shared_ptr<Table> source_table,
+                             TableIdentifier identifier)
+    : StaticTable(identifier, source_table->metadata(),
+                  std::string(source_table->metadata_file_location()), source_table->io(),
+                  source_table->catalog()),
+      source_table_(std::move(source_table)) {
+  auto schema = GetSchema();
```
wgtmac: Calling `GetSchema()` here does not call the derived override. During base construction, C++ dispatches to `MetadataTable::GetSchema()`, so snapshots/history tables get the empty fallback schema. Please pass the schema into the base constructor or set it after the derived object is built.

src/iceberg/inspect/snapshots_table.cc
```diff
@@ -0,0 +1,63 @@
...
+
+SnapshotsTable::~SnapshotsTable() = default;
+
+std::shared_ptr<Schema> SnapshotsTable::GetSchema() const {
+  return std::make_shared<Schema>(
+      std::vector<SchemaField>{
+          SchemaField::MakeRequired(1, "committed_at", timestamp_tz()),
+          SchemaField::MakeRequired(2, "snapshot_id", int64()),
+          SchemaField::MakeRequired(3, "parent_id", int64()),
+          SchemaField::MakeOptional(4, "operation", int64()),
```
wgtmac: This schema does not match Iceberg. `parent_id` must be optional and `operation` must be a string, not int64. Valid snapshots can have no parent and operations such as `append`.

src/iceberg/inspect/metadata_table.h
```diff
@@ -0,0 +1,141 @@
...
+///   auto scan = ICEBERG_TRY(metadata_table->NewScan());
+///   // ... scan and read metadata table data
+class ICEBERG_EXPORT MetadataTableFactory {
+ public:
+  /// \brief Create a metadata table from a table
+  ///
+  /// \param table The source table
+  /// \param type The metadata table type to create
+  /// \return A MetadataTable instance or error status
+  static Result<std::unique_ptr<MetadataTable>> CreateMetadataTable(
```
wgtmac: Returning `unique_ptr` is unsafe for a `Table` subclass. `Table` uses `enable_shared_from_this()`, and inherited methods like `NewFastAppend()` can throw `bad_weak_ptr`. Metadata tables also need to override all mutating APIs as read-only.

src/iceberg/inspect/metadata_table.cc
```diff
@@ -0,0 +1,95 @@
...
+MetadataTable::~MetadataTable() = default;
+
+Status MetadataTable::Refresh() { return source_table_->Refresh(); }
+
+Result<std::unique_ptr<DataTableScanBuilder>> MetadataTable::NewScan() const {
+  return NotSupported("TODO: Scanning metadata tables is not yet supported");
+};
+
+Result<std::unique_ptr<MetadataTable>> MetadataTableFactory::CreateMetadataTable(
+    std::shared_ptr<Table> table, MetadataTableType type) {
```
wgtmac: Please validate `table` here. Passing `nullptr` reaches the concrete constructor and crashes on `table->name()` instead of returning `InvalidArgument`.

src/iceberg/CMakeLists.txt
```diff
@@ -41,6 +41,9 @@ set(ICEBERG_SOURCES
     file_io_registry.cc
     file_reader.cc
     file_writer.cc
+    inspect/history_table.cc
+    inspect/metadata_table.cc
+    inspect/snapshots_table.cc
```
wgtmac, line 46: These sources are added, but the new `inspect/CMakeLists.txt` is not included below. Please add `add_subdirectory(inspect)`, otherwise CMake install will miss the public headers.

wgtmac (commented):
Please also fix the current CI failure before merging.

wgtmac (approved):
I thought the design could be simplified if `MetadataTable` doesn't inherit `Table` so I just made this change to reflect this. I've also simplified the factory and public functions. Since this PR has been opened too long, I'll  merge it to move forward. Thanks @All-less for initiating this!

PR #608, opened 2026-03-29
feat/perf: optimize AddRange in RoaringPositionBitmap

Baunsgaard:
Replace the per-position loop in `RoaringPositionBitmap::AddRange` with direct calls to `roaring::Roaring::addRange`, matching the optimization in the Java implementation (https://github.com/apache/iceberg/pull/15791).

Added tests for single-position ranges, large contiguous ranges, three-key spanning ranges, and invalid inputs.

### DIFF
--- a/src/iceberg/deletes/roaring_position_bitmap.cc
+++ b/src/iceberg/deletes/roaring_position_bitmap.cc
@@ -128,7 +128,12 @@ void RoaringPositionBitmap::Add(int64_t pos) {
 }
 
 void RoaringPositionBitmap::AddRange(int64_t pos_start, int64_t pos_end) {
-  if (pos_start >= pos_end || pos_start < 0 || pos_end - 1 > kMaxPosition) {
+  if (pos_start >= pos_end) {
+    return;
+  }
+  pos_start = std::max(pos_start, int64_t{0});
+  pos_end = std::min(pos_end, kMaxPosition + 1);
+  if (pos_start >= pos_end) {
     return;
   }
 
--- a/src/iceberg/deletes/roaring_position_bitmap.cc
+++ b/src/iceberg/deletes/roaring_position_bitmap.cc
@@ -141,20 +141,12 @@ void RoaringPositionBitmap::AddRange(int64_t pos_start, int64_t pos_end) {
   int32_t end_key = Key(pos_end - 1);
   impl_->AllocateBitmapsIfNeeded(end_key + 1);
 
-  if (start_key == end_key) {
-    uint64_t low_start = Pos32Bits(pos_start);
-    uint64_t low_end = static_cast<uint64_t>(Pos32Bits(pos_end - 1)) + 1;
-    impl_->bitmaps[start_key].addRange(low_start, low_end);
-  } else {
-    uint64_t first_low_start = Pos32Bits(pos_start);
-    impl_->bitmaps[start_key].addRange(first_low_start, uint64_t{1} << 32);
-
-    for (int32_t key = start_key + 1; key < end_key; ++key) {
-      impl_->bitmaps[key].addRange(uint64_t{0}, uint64_t{1} << 32);
-    }
-
-    uint64_t last_low_end = static_cast<uint64_t>(Pos32Bits(pos_end - 1)) + 1;
-    impl_->bitmaps[end_key].addRange(uint64_t{0}, last_low_end);
+  for (int32_t key = start_key; key <= end_key; ++key) {
+    uint64_t low_start = (key == start_key) ? Pos32Bits(pos_start) : uint64_t{0};
+    uint64_t low_end = (key == end_key)
+                            ? static_cast<uint64_t>(Pos32Bits(pos_end - 1)) + 1
+                            : (uint64_t{1} << 32);
+    impl_->bitmaps[key].addRange(low_start, low_end);
   }
 }
 
--- a/src/iceberg/deletes/roaring_position_bitmap.cc
+++ b/src/iceberg/deletes/roaring_position_bitmap.cc
@@ -22,7 +22,6 @@
 #include <cstring>
 #include <exception>
 #include <limits>
-#include <stdexcept>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


src/iceberg/deletes/roaring_position_bitmap.h
```diff
@@ -65,6 +65,7 @@ class ICEBERG_EXPORT RoaringPositionBitmap {
   void Add(int64_t pos);
 
   /// \brief Sets a range of positions [pos_start, pos_end).
+  /// If pos_start >= pos_end, this method does nothing.
   /// \note Invalid positions are silently ignored
```
evindj: rephrase may be state pos_start >= pos_end is considered invalid.

src/iceberg/deletes/roaring_position_bitmap.cc
```diff
@@ -128,8 +128,31 @@ void RoaringPositionBitmap::Add(int64_t pos) {
...
 
 void RoaringPositionBitmap::AddRange(int64_t pos_start, int64_t pos_end) {
-  for (int64_t pos = pos_start; pos < pos_end; ++pos) {
-    Add(pos);
+  if (pos_start >= pos_end) {
+    return;
+  }
+  if (pos_start < 0 || pos_end - 1 > kMaxPosition) {
+    return;
+  }
```
evindj: Nit: keep only one ```if``` block?

src/iceberg/test/roaring_position_bitmap_test.cc
```diff
@@ -128,9 +128,72 @@ INSTANTIATE_TEST_SUITE_P(
...
+  ASSERT_EQ(bitmap.Cardinality(), 0u);
+}
+
+TEST(RoaringPositionBitmapTest, TestAddRangeEqualStartEndIsNoOp) {
+  RoaringPositionBitmap bitmap;
+  bitmap.AddRange(100, 100);
+  ASSERT_TRUE(bitmap.IsEmpty());
+  ASSERT_EQ(bitmap.Cardinality(), 0u);
+  ASSERT_FALSE(bitmap.Contains(100));
+}
```
evindj: could parameterize these noop tests.

src/iceberg/deletes/roaring_position_bitmap.cc
```diff
@@ -128,8 +128,31 @@ void RoaringPositionBitmap::Add(int64_t pos) {
 }
 
 void RoaringPositionBitmap::AddRange(int64_t pos_start, int64_t pos_end) {
-  for (int64_t pos = pos_start; pos < pos_end; ++pos) {
-    Add(pos);
+  if (pos_start >= pos_end) {
+    return;
+  }
+  if (pos_start < 0 || pos_end - 1 > kMaxPosition) {
```
emkornfield: its not clear to me this is correct, don't we need to fill in valid values?

src/iceberg/deletes/roaring_position_bitmap.cc
```diff
@@ -128,8 +128,31 @@ void RoaringPositionBitmap::Add(int64_t pos) {
...
+    return;
+  }
+
+  int32_t start_key = Key(pos_start);
+  int32_t end_key = Key(pos_end - 1);
+  impl_->AllocateBitmapsIfNeeded(end_key + 1);
+
+  if (start_key == end_key) {
```
emkornfield: I think it might end up being more maintainable to have a single loop that handles both cases?

src/iceberg/deletes/roaring_position_bitmap.cc
wgtmac: +1

We need to ignore invalid positions but set values for valid ones.

src/iceberg/deletes/roaring_position_bitmap.h
```diff
@@ -65,6 +65,7 @@ class ICEBERG_EXPORT RoaringPositionBitmap {
   void Add(int64_t pos);
 
   /// \brief Sets a range of positions [pos_start, pos_end).
+  /// If pos_start >= pos_end, this method does nothing.
```
wgtmac: This contradicts the note sentence below.

src/iceberg/test/roaring_position_bitmap_test.cc
Baunsgaard: Okay, i have changed it to parameterized tests.

src/iceberg/deletes/roaring_position_bitmap.cc
Baunsgaard: sure, done

src/iceberg/deletes/roaring_position_bitmap.h
Baunsgaard: I have made it invalid if  pos_start > pos_end, technically I can imagine cases where pos_start == pos_end, therefore i keep that as a no-op. 

Let me know what you think.

src/iceberg/deletes/roaring_position_bitmap.cc
Baunsgaard: Okay, i follow the style of the normal operation, to ignore  invalid positions, and clamp inputs to correct values. 

However, this is inconsistent with the behaviour of core Java, where positions gets validated, and errors out if they are out of range.

src/iceberg/deletes/roaring_position_bitmap.cc
Baunsgaard: Okay, changed to single control flow..

src/iceberg/deletes/roaring_position_bitmap.cc
```diff
@@ -128,8 +129,27 @@ void RoaringPositionBitmap::Add(int64_t pos) {
 }
 
 void RoaringPositionBitmap::AddRange(int64_t pos_start, int64_t pos_end) {
-  for (int64_t pos = pos_start; pos < pos_end; ++pos) {
-    Add(pos);
+  if (pos_start > pos_end) {
+    throw std::invalid_argument("AddRange requires pos_start <= pos_end, got [" +
```
wgtmac: Let's return `Status` instead of throwing exception if you really think this is an invalid case. IMHO, we need to be consistent with line 123 and line 140 to return on invalid inputs.

Baunsgaard: Well, it is an invalid input, however for consistency with other methods in the c++ implementation, then both versions are valid. 

Before my changes in the PR to the Java base, it also would not throw any exceptions. since it used the same for loop definition. To keep the existing behaviour, I have removed the exception throwing.

wgtmac (approved):
LGTM. Thanks @Baunsgaard!

PR #610, opened 2026-03-31
feat(data): add DeleteLoader for position and equality delete files

wgtmac:
- LoadPositionDeletes dispatches to LoadPositionDelete (file-based) and LoadDV (deletion vectors, not yet implemented).
- LoadEqualityDeletes reads equality delete files into a StructLikeSet.

### DIFF
--- a/src/iceberg/data/delete_loader.cc
+++ b/src/iceberg/data/delete_loader.cc
@@ -110,13 +110,19 @@ Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
   PositionDeleteIndex index;
 
   for (const auto& file : delete_files) {
+    if (file->referenced_data_file.has_value() &&
+        file->referenced_data_file.value() != data_file_path) {
+      continue;
+    }
+
     if (file->IsDeletionVector()) {
       ICEBERG_RETURN_UNEXPECTED(LoadDV(*file, index));
+      continue;
     }
 
     ICEBERG_PRECHECK(file->content == DataFile::Content::kPositionDeletes,
                      "Expected position delete file but got content type {}",
-                     static_cast<int>(file->content));
+                     ToString(file->content));
 
     ICEBERG_RETURN_UNEXPECTED(LoadPositionDelete(*file, index, data_file_path));
   }
--- a/src/iceberg/data/delete_loader.cc
+++ b/src/iceberg/data/delete_loader.cc
@@ -1,0 +1,167 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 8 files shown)


src/iceberg/data/delete_loader.h
```diff
@@ -0,0 +1,80 @@
...
+      std::span<const std::shared_ptr<DataFile>> delete_files,
+      std::string_view data_file_path) const;
+
+  /// \brief Load equality deletes into an in-memory set.
+  ///
+  /// \param delete_files Equality delete files to load (must have
+  ///   content == Content::kEqualityDeletes)
+  /// \param equality_type The struct type describing the equality columns
+  /// \return A StructLikeSet containing the deleted rows
+  Result<std::unique_ptr<UncheckedStructLikeSet>> LoadEqualityDeletes(
```
zhjwpku, line 64: Why LoadEqualityDeletes returns a unique ptr while LoadPositionDeletes returns a struct?

zhjwpku (approved):

src/iceberg/data/delete_loader.cc
```diff
@@ -0,0 +1,167 @@
...
+
+Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
+    std::span<const std::shared_ptr<DataFile>> delete_files,
+    std::string_view data_file_path) const {
+  PositionDeleteIndex index;
+
+  for (const auto& file : delete_files) {
+    if (file->IsDeletionVector()) {
+      ICEBERG_RETURN_UNEXPECTED(LoadDV(*file, index));
+    }
```
gty404, line 121:
```suggestion
    if (file->IsDeletionVector()) {
      ICEBERG_RETURN_UNEXPECTED(LoadDV(*file, index));
      continue;
    }
```

src/iceberg/data/delete_loader.h
```diff
@@ -0,0 +1,80 @@
...
+/// Loads position and equality delete files into in-memory indexes.
+
+#include <memory>
+#include <span>
+#include <string_view>
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+#include "iceberg/util/struct_like_set.h"
```
gty404: Use forward declaration?

src/iceberg/data/delete_loader.cc
```diff
@@ -0,0 +1,167 @@
...
+Status DeleteLoader::LoadDV(const DataFile& file, PositionDeleteIndex& index) const {
+  return NotSupported("Loading deletion vectors is not yet supported");
+}
+
+Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
+    std::span<const std::shared_ptr<DataFile>> delete_files,
+    std::string_view data_file_path) const {
+  PositionDeleteIndex index;
+
+  for (const auto& file : delete_files) {
```
gty404, line 112: Can we skip opening the file directly if `DataFile::referenced_data_file` exists and is not equal to `data_file_path`?

src/iceberg/data/delete_loader.h
wgtmac: PositionDeleteIndex is a movable value type, so it can be returned directly by value. UncheckedStructLikeSet contains a `std::pmr::monotonic_buffer_resource`, which is neither copyable nor movable, so it must be heap-allocated and returned via unique_ptr.

wgtmac:
I've addressed the comments. Thanks for the review, @zhjwpku and @gty404!

blackmwk (approved):

PR #611, opened 2026-03-31
feat: implement incremental changelog scan

WZhuo:


src/iceberg/table_scan.h
```diff
@@ -102,14 +102,87 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
-  int64_t estimated_row_count() const override { return 0; }
+
+  int64_t size_bytes() const override;
+  int32_t files_count() const override;
+  int64_t estimated_row_count() const override;
+
+  virtual ChangelogOperation operation() const = 0;
+
+  /// \brief The position of this change in the changelog order (0-based).
+  virtual int32_t change_ordinal() const { return change_ordinal_; }
```
wgtmac: This doesn't need to be virtual.

src/iceberg/table_scan.h
```diff
@@ -102,14 +102,87 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
+
+  /// \brief The snapshot ID that committed this change.
+  virtual int64_t commit_snapshot_id() const { return commit_snapshot_id_; }
```
wgtmac: Ditto

src/iceberg/table_scan.h
```diff
@@ -102,14 +102,87 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
+
+  /// \brief The data file containing the added rows.
+  const std::shared_ptr<DataFile>& data_file() const { return data_file_; }
```
wgtmac: The meaning of `data_file()` and `delete_files()` vary for different `ChangelogOperation`. Let's move their definitions to subclasses to pair with the Java implementation. The evidence is that comment above does not apply to the Java `BaseDeletedRowsScanTask` class.

src/iceberg/table_scan.h
```diff
@@ -102,14 +102,87 @@ class ICEBERG_EXPORT FileScanTask : public ScanTask {
...
+  std::shared_ptr<DataFile> data_file_;
+  std::vector<std::shared_ptr<DataFile>> delete_files_;
+  std::shared_ptr<Expression> residual_filter_;
+};
+
+/// \brief A scan task for reading rows that were added between snapshots.
+///
+/// This task represents data files that were added to the table, along with any
+/// delete files that should be applied when reading the data.
+class ICEBERG_EXPORT AddedRowsScanTask : public ChangelogScanTask {
```
wgtmac, line 175: Let's add the comment from Java which helps understand what is for.

```
/**
 * A scan task for inserts generated by adding a data file to the table.
 *
 * <p>Note that added data files may have matching delete files. This may happen if a matching
 * position delete file is committed in the same snapshot or if changes for multiple snapshots are
 * squashed together.
 *
 * <p>Suppose snapshot S1 adds data files F1, F2, F3 and a position delete file, D1, that marks
 * particular records in F1 as deleted. A scan for changes generated by S1 should include the
 * following tasks:
 *
 * <ul>
 *   <li>AddedRowsScanTask(file=F1, deletes=[D1], snapshot=S1)
 *   <li>AddedRowsScanTask(file=F2, deletes=[], snapshot=S1)
 *   <li>AddedRowsScanTask(file=F3, deletes=[], snapshot=S1)
 * </ul>
 *
 * <p>Readers consuming these tasks should produce added records with metadata like change ordinal
 * and commit snapshot ID.
 */
```

src/iceberg/table_scan.h
```diff
@@ -133,8 +206,24 @@ struct TableScanContext {
...
+
+  /// \brief Get the snapshot ID to scan up to (inclusive) based on the context.
+  Result<int64_t> ToSnapshotIdInclusive(const TableMetadata& metadata) const;
+
+  /// \brief Get the snapshot ID to scan from (exclusive) based on the context.
+  Result<std::optional<int64_t>> FromSnapshotIdExclusive(
+      const TableMetadata& metadata, int64_t to_snapshot_id_inclusive) const;
 };
 
+// Internal validation functions for IncrementalScanBuilder
```
wgtmac: Is it better to move them to snapshot_util.h/cc?

src/iceberg/table_scan.cc
```diff
@@ -294,6 +296,24 @@ Result<ArrowArrayStream> FileScanTask::ToArrow(
   return MakeArrowArrayStream(std::move(reader));
 }
 
+// ChangelogScanTask implementation
+
+int64_t ChangelogScanTask::size_bytes() const {
+  int64_t total_size = data_file_->file_size_in_bytes;
+  for (const auto& delete_file : delete_files_) {
+    total_size +=
+        (delete_file->IsDeletionVector() ? delete_file->content_size_in_bytes.value_or(0)
```
wgtmac: nit: add a `ICEBERG_DCHECK(delete_file->content_size_in_bytes.has_value())` as it is unlikely to be unset but just in case

src/iceberg/table_scan.cc
```diff
@@ -762,7 +784,126 @@ IncrementalChangelogScan::PlanFiles() const {
...
+      changelog_snapshots.emplace_back(snapshot, std::move(snapshot_cache));
+    }
+  }
+  if (changelog_snapshots.empty()) {
+    return std::vector<std::shared_ptr<ChangelogScanTask>>{};
+  }
+
+  std::unordered_set<int64_t> snapshot_ids;
+  std::unordered_map<int64_t, int32_t> snapshot_ordinals;
+  int32_t ordinal = 0;
```
wgtmac: nit: we can use `snapshot_ordinals.size()` to replace `ordinal`

src/iceberg/table_scan.cc
```diff
@@ -762,7 +784,126 @@ IncrementalChangelogScan::PlanFiles() const {
...
+  if (data_manifests.empty()) {
+    return std::vector<std::shared_ptr<ChangelogScanTask>>{};
+  }
+
+  TableMetadataCache metadata_cache(metadata_.get());
+  ICEBERG_ASSIGN_OR_RAISE(auto specs_by_id, metadata_cache.GetPartitionSpecsById());
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto manifest_group,
+      ManifestGroup::Make(io_, schema_, specs_by_id, std::move(data_manifests), {}));
```
```suggestion
      ManifestGroup::Make(io_, schema_, specs_by_id, std::move(data_manifests), /*delete_manifests=*/{}));
```

src/iceberg/table_scan.cc
```diff
@@ -762,7 +784,126 @@ IncrementalChangelogScan::PlanFiles() const {
...
+  auto create_tasks_func =
+      [&snapshot_ordinals](
+          std::vector<ManifestEntry>&& entries,
+          const TaskContext& ctx) -> Result<std::vector<std::shared_ptr<ScanTask>>> {
+    std::vector<std::shared_ptr<ScanTask>> tasks;
+    tasks.reserve(entries.size());
+
+    for (auto& entry : entries) {
+      if (!entry.snapshot_id.has_value() || entry.data_file == nullptr) {
+        continue;
```
wgtmac: This should not happen. Use ICEBERG_PRECHECK to return an error instead?

src/iceberg/table_scan.cc
```diff
@@ -762,7 +784,126 @@ IncrementalChangelogScan::PlanFiles() const {
...
+      }
+
+      int64_t commit_snapshot_id = entry.snapshot_id.value();
+      auto ordinal_it = snapshot_ordinals.find(commit_snapshot_id);
+      if (ordinal_it == snapshot_ordinals.end()) {
+        continue;
```
wgtmac: Same question here.

src/iceberg/table_scan.cc
```diff
@@ -762,7 +784,126 @@ IncrementalChangelogScan::PlanFiles() const {
...
+      int32_t change_ordinal = ordinal_it->second;
+
+      if (ctx.drop_stats) {
+        ContentFileUtil::DropAllStats(*entry.data_file);
+      } else if (!ctx.columns_to_keep_stats.empty()) {
+        ContentFileUtil::DropUnselectedStats(*entry.data_file, ctx.columns_to_keep_stats);
+      }
+
+      ICEBERG_ASSIGN_OR_RAISE(auto residual,
+                              ctx.residuals->ResidualFor(entry.data_file->partition));
```
wgtmac, line 889: Why do we need this specifically for changelog scan?

src/iceberg/test/incremental_changelog_scan_test.cc
```diff
@@ -0,0 +1,478 @@
...
+    }
+    return t1->data_file()->file_path < t2->data_file()->file_path;
+  });
+}
+
+}  // namespace
+
+class IncrementalChangelogScanTest : public ScanTestBase {};
+
+TEST_P(IncrementalChangelogScanTest, DataFilters) {
```
wgtmac, line 70: Java's `testDataFilters` uses `withUnavailableLocations` to verify that snap1's manifest is actually pruned and never read. This test only checks result correctness but doesn't verify manifest pruning. Should we add a similar one?

src/iceberg/test/scan_test_base.h
```diff
@@ -261,6 +261,87 @@ class ScanTestBase : public testing::TestWithParam<int8_t> {
     });
   }
 
+  /// \brief Create a delete snapshot with the given files.
+  std::shared_ptr<Snapshot> MakeDeleteSnapshot(
```
wgtmac, line 265: Should we add partition_spec as an input to `MakeDeleteSnapshot` and `MakeOverwriteSnapshot`?

wgtmac (commented):
Mostly looks good! I have some inline comments.

src/iceberg/table_scan.h
WZhuo: It's a dead code perhaps be left over rebase, remove it.

src/iceberg/table_scan.cc
WZhuo: It's also set for `FileScanTask`.

src/iceberg/test/incremental_changelog_scan_test.cc
WZhuo: The manifest pruning is tested in the ManifestGroupTest.

src/iceberg/test/scan_test_base.h
WZhuo: The `partition_spec` is not used currently. We can add it later if needed.

wgtmac (approved):
+1

PR #614, opened 2026-04-05
feat(rest): add scan plan endpoints to REST catalog client

gsandeep1241:
* Introduced 4 endpoints: planning table scan, fetching a scan result, cancelling a scan plan and fetching scan planning tasks (endpoint.h)
* Introduced corresponding request and response objects (types.h)
* Implemented handlers (catalog.h). The handlers construct a request object and return a response object
    - Handlers use newly introduced helper methods to serialize and deserialize request response json objects
* Added unit tests for all the above helper methods and handler methods
* Added integration tests for the handlers


src/iceberg/catalog/rest/error_handlers.cc
```diff
@@ -183,4 +186,32 @@ Status ViewCommitErrorHandler::Accept(const ErrorResponse& error) const {
...
+      if (error.type == kNoSuchNamespaceException) {
+        return NoSuchNamespace(error.message);
+      }
+      if (error.type == kNoSuchTableException) {
+        return NoSuchTable(error.message);
+      }
+      if (error.type == kNoSuchPlanIdException) {
+        return NoSuchPlanId(error.message);
+      }
+      if (error.type == kNoSuchPlanTaskException) {
```
zhjwpku, line 228: Should we add a PlanTask level error handler like java implementation? see https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/ErrorHandlers.java#L197

src/iceberg/catalog/rest/json_serde.cc
```diff
@@ -506,6 +523,114 @@ Result<OAuthTokenResponse> OAuthTokenResponseFromJson(const nlohmann::json& json
...
+  return json;
+}
+
+nlohmann::json ToJson(const FetchScanTasksRequest& request) {
+  nlohmann::json json;
+  json[kPlanTask] = request.planTask;
+  return json;
+}
+
+Status BaseScanTaskResponseFromJson(
```
zhjwpku: This looks like a helper function that isn's exposed in the header; perhaps it should be placed in an anonymous namespace?

src/iceberg/catalog/rest/resource_paths.cc
```diff
@@ -102,4 +102,26 @@ Result<std::string> ResourcePaths::CommitTransaction() const {
...
+  return std::format("{}/v1/{}namespaces/{}/tables/{}/plan", base_uri_, prefix_,
+                     encoded_namespace, encoded_table_name);
+}
+
+Result<std::string> ResourcePaths::ScanPlan(const TableIdentifier& ident,
+                                            const std::string& plan_id) const {
+  ICEBERG_ASSIGN_OR_RAISE(std::string encoded_namespace, EncodeNamespace(ident.ns));
+  ICEBERG_ASSIGN_OR_RAISE(std::string encoded_table_name, EncodeString(ident.name));
+  return std::format("{}/v1/{}namespaces/{}/tables/{}/plan/{}", base_uri_, prefix_,
+                     encoded_namespace, encoded_table_name, plan_id);
```
zhjwpku: Should we also encode plan_id?

src/iceberg/catalog/rest/json_serde.cc
```diff
@@ -78,6 +80,21 @@ constexpr std::string_view kExpiresIn = "expires_in";
...
+constexpr std::string_view kDeleteFiles = "delete-files";
+constexpr std::string_view kSnapshotId = "snapshot-id";
+constexpr std::string_view kSelect = "select";
+constexpr std::string_view kFilter = "filter";
+constexpr std::string_view kCaseSensitive = "case-sensitive";
+constexpr std::string_view kUseSnapshotSchema = "use-snapshot-schema";
+constexpr std::string_view kStartSnapshotId = "start-snapshot-id";
+constexpr std::string_view kEndSnapshotId = "end-snapshot-id";
+constexpr std::string_view kStatsFields = "stats-fields";
+constexpr std::string_view kMinRowsRequired = "min-rows-required";
```
zhjwpku:
```suggestion
constexpr std::string_view kMinRowsRequested = "min-rows-requested";
```

gsandeep1241: Thank you!

src/iceberg/catalog/rest/json_serde.cc
gsandeep1241: Agreed, done.

src/iceberg/catalog/rest/resource_paths.cc
gsandeep1241: Yes - done.

src/iceberg/catalog/rest/error_handlers.cc
gsandeep1241: Done!

src/iceberg/catalog/rest/error_handlers.cc
```diff
@@ -183,4 +186,52 @@ Status ViewCommitErrorHandler::Accept(const ErrorResponse& error) const {
   return DefaultErrorHandler::Accept(error);
 }
 
+const std::shared_ptr<ScanPlanErrorHandler>& ScanPlanErrorHandler::Instance() {
```
wgtmac: What about renaming it to `PlanErrorHandler` to be consistent with the Java parity?

src/iceberg/catalog/rest/error_handlers.cc
```diff
@@ -183,4 +186,52 @@ Status ViewCommitErrorHandler::Accept(const ErrorResponse& error) const {
...
+      }
+      if (error.type == kNoSuchPlanIdException) {
+        return NoSuchPlanId(error.message);
+      }
+      if (error.type == kNoSuchPlanTaskException) {
+        return NoSuchPlanTask(error.message);
+      }
+      return NotFound(error.message);
+    case 406:
+      return NotSupported(error.message);
```
wgtmac:
```suggestion
```

I don't see these lines from Java. Why do we need them?

src/iceberg/catalog/rest/resource_paths.h
```diff
@@ -81,6 +81,19 @@ class ICEBERG_REST_EXPORT ResourcePaths {
   /// \brief Get the /v1/{prefix}/transactions/commit endpoint path.
   Result<std::string> CommitTransaction() const;
 
+  /// \brief Get the /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan endpoint
+  /// path.
+  Result<std::string> ScanPlan(const TableIdentifier& ident) const;
```
wgtmac: Can we follow the same name of Java impl to use `Plan` and `FetchScanTasks`? This will make it easier to keep them in sync in the future.

src/iceberg/catalog/rest/resource_paths.cc
```diff
@@ -102,4 +102,27 @@ Result<std::string> ResourcePaths::CommitTransaction() const {
   return std::format("{}/v1/{}transactions/commit", base_uri_, prefix_);
 }
 
+Result<std::string> ResourcePaths::ScanPlan(const TableIdentifier& ident) const {
+  ICEBERG_ASSIGN_OR_RAISE(std::string encoded_namespace, EncodeNamespace(ident.ns));
```
wgtmac: We need to respect `namespace_separator_` just like other functions.

src/iceberg/catalog/rest/resource_paths.cc
```diff
@@ -102,4 +102,27 @@ Result<std::string> ResourcePaths::CommitTransaction() const {
...
+Result<std::string> ResourcePaths::ScanPlan(const TableIdentifier& ident,
+                                            const std::string& plan_id) const {
+  ICEBERG_ASSIGN_OR_RAISE(std::string encoded_namespace, EncodeNamespace(ident.ns));
+  ICEBERG_ASSIGN_OR_RAISE(std::string encoded_table_name, EncodeString(ident.name));
+  ICEBERG_ASSIGN_OR_RAISE(std::string encoded_plan_id, EncodeString(plan_id));
+  return std::format("{}/v1/{}namespaces/{}/tables/{}/plan/{}", base_uri_, prefix_,
+                     encoded_namespace, encoded_table_name, encoded_plan_id);
+}
+
+Result<std::string> ResourcePaths::ScanTask(const TableIdentifier& ident) const {
```
wgtmac: Should we merge this with the above one by using `std::optional<int64_t> plan_id`?

src/iceberg/catalog.h
```diff
@@ -26,8 +26,10 @@
 #include <unordered_set>
 #include <vector>
 
+#include "iceberg/catalog/rest/types.h"
 #include "iceberg/result.h"
 #include "iceberg/table_identifier.h"
+#include "iceberg/table_scan.h"
```
wgtmac: Let's use forward declaration as much as possible

src/iceberg/catalog.h
```diff
@@ -26,8 +26,10 @@
 #include <unordered_set>
 #include <vector>
 
+#include "iceberg/catalog/rest/types.h"
```
wgtmac: This is a generic catalog interface so we need abstract away rest catalog specific concepts.

src/iceberg/catalog.h
```diff
@@ -188,6 +190,44 @@ class ICEBERG_EXPORT Catalog {
   /// \return a Table instance or ErrorKind::kAlreadyExists if the table already exists
   virtual Result<std::shared_ptr<Table>> RegisterTable(
       const TableIdentifier& identifier, const std::string& metadata_file_location) = 0;
+
+  /// \brief Initiate a scan planning operation for the given table.
+  ///
+  /// \param table The table to scan.
+  /// \param context The scan context containing snapshot, filter, and other options.
+  /// \return A PlanTableScanResponse with the plan status and initial scan tasks.
+  virtual Result<rest::PlanTableScanResponse> PlanTableScan(
```
wgtmac: Same as above, we should use better abstraction. Server side scan planning needs to be handled inside the table object returned by `LoadTable` above. Perhaps we need to introduce the same `RestTable` class hierachy just like the Java impl.

src/iceberg/catalog/rest/types.h
```diff
@@ -295,4 +298,73 @@ struct ICEBERG_REST_EXPORT OAuthTokenResponse {
...
+struct ICEBERG_REST_EXPORT PlanTableScanRequest {
+  std::optional<int64_t> snapshot_id;
+  std::vector<std::string> select;
+  std::shared_ptr<Expression> filter;
+  bool case_sensitive = true;
+  bool use_snapshot_schema = false;
+  std::optional<int64_t> start_snapshot_id;
+  std::optional<int64_t> end_snapshot_id;
+  std::vector<std::string> statsFields;
+  std::optional<int64_t> min_rows_required;
```
```suggestion
  std::optional<int64_t> min_rows_requested;
```

src/iceberg/json_serde_internal.h
```diff
@@ -413,4 +417,39 @@ ICEBERG_EXPORT nlohmann::json ToJson(const TableRequirement& requirement);
...
+/// value-counts, etc.) use the CountMap/BinaryMap parallel arrays format.
+/// Binary fields (lower-bounds, upper-bounds, key-metadata) are base64-encoded.
+///
+/// \param json The JSON object representing a `DataFile`.
+/// \param partitionSpecById Map from spec ID to PartitionSpec for type-aware partition
+/// parsing.
+/// \param schema The table schema, used with partitionSpecById to resolve partition
+/// types.
+/// \return A `DataFile` object or an error if the conversion fails.
+ICEBERG_EXPORT Result<DataFile> DataFileFromJson(const nlohmann::json& json);
```
wgtmac: I would highly recommend adding `ToJson` for `DataFile` and `FileScanTask` so we can have roundtrip test coverage. In addition, although this library currently does not need to implement rest catalog server, it should be a building block in case downstream users need to do that.

src/iceberg/catalog/rest/json_serde_internal.h
```diff
@@ -62,4 +66,26 @@ ICEBERG_DECLARE_JSON_SERDE(OAuthTokenResponse)
 
 #undef ICEBERG_DECLARE_JSON_SERDE
 
+ICEBERG_REST_EXPORT Result<PlanTableScanResponse> PlanTableScanResponseFromJson(
```
wgtmac, line 70: Same as my other comments, it would be good to add both `ToJson` and `FromJson` for these new request and response types just like others.

src/iceberg/catalog.h
wgtmac: It this requires larger changes, we can split them into smaller PRs. This PR can just focus on adding rest models with ser/de support and error handlers.

src/iceberg/catalog/rest/types.h
```diff
@@ -295,4 +298,73 @@ struct ICEBERG_REST_EXPORT OAuthTokenResponse {
...
+/// \brief Request to initiate a server-side scan planning operation.
+struct ICEBERG_REST_EXPORT PlanTableScanRequest {
+  std::optional<int64_t> snapshot_id;
+  std::vector<std::string> select;
+  std::shared_ptr<Expression> filter;
+  bool case_sensitive = true;
+  bool use_snapshot_schema = false;
+  std::optional<int64_t> start_snapshot_id;
+  std::optional<int64_t> end_snapshot_id;
+  std::vector<std::string> statsFields;
```
```suggestion
  std::vector<std::string> stats_fields;
```

src/iceberg/catalog/rest/types.h
```diff
@@ -295,4 +298,73 @@ struct ICEBERG_REST_EXPORT OAuthTokenResponse {
...
+  std::optional<int64_t> min_rows_required;
+
+  Status Validate() const;
+
+  bool operator==(const PlanTableScanRequest&) const;
+};
+
+/// \brief Base response containing scan tasks and delete files returned by scan plan
+/// endpoints.
+struct ICEBERG_REST_EXPORT BaseScanTaskResponse {
```
wgtmac: There is a trade-off in designing the `BaseScanTaskResponse` class hierachy. All other rest models defined in this file are trivial C++ structs that support aggregate initialization. That's why we didn't declare the `Validate()` function to be virtual. Using class hierachy like the Java equivalents disables aggregate initialization. That works perfect before adding `BaseScanTaskResponse`.

If we want to keep the current `BaseScanTaskResponse` class hierachy, declaring `BaseScanTaskResponse::Validate` as non-virtual looks error-prone and downstreams have to figure out which specific subclass of BaseScanTaskResponse is.

Instead, should we define it as something like below?

```cpp
struct ICEBERG_REST_EXPORT ScanTaskResponse {
  // common fields
  std::vector<std::string> plan_tasks;
  ...

  struct PlanTableScanResponse {
    std::string plan_status;
    std::string plan_id;
  };

  struct FetchPlanningResultResponse {
    std::string plan_status;
  };

  ...

  std::variant<PlanTableScanResponse, FetchPlanningResultResponse, ...>

  Status Validate() const;
}'
```

src/iceberg/catalog/rest/types.h
```diff
@@ -295,4 +298,73 @@ struct ICEBERG_REST_EXPORT OAuthTokenResponse {
...
+  std::vector<std::string> plan_tasks;
+  std::vector<FileScanTask> file_scan_tasks;
```
wgtmac: It seems that using `std::shared_ptr<FileScanTask>` and `std::shared_ptr<DataFile>` is more flexible and consistent with the repo.

src/iceberg/test/rest_json_serde_test.cc
```diff
@@ -1380,4 +1381,139 @@ INSTANTIATE_TEST_SUITE_P(
...
+// Helper: empty schema and specs for scan response tests that don't need partition
+// parsing.
+static Schema EmptySchema() { return Schema({}, 0); }
+static std::unordered_map<int32_t, std::shared_ptr<PartitionSpec>> EmptySpecs() {
+  return {};
+}
+
+// --- PlanTableScanResponse ---
+
+TEST(PlanTableScanResponseFromJsonTest, SubmittedStatusMissingOptionalFields) {
```
wgtmac, line 1419: Can we switch to paramterized test here and port as many cases as possible from the Jave tests like `TestFetchScanTasksResponseParser.java`, `TestPlanTableScanResponseParser.java` and `TestPlanTableScanRequestParser.java`?

src/iceberg/json_serde_internal.h
```diff
@@ -413,4 +417,39 @@ ICEBERG_EXPORT nlohmann::json ToJson(const TableRequirement& requirement);
...
+
+ICEBERG_EXPORT Result<DataFile> DataFileFromJson(
```
wgtmac: BTW, let's revert changes in this file. These json serde impls on DataFile and FileScanTask are specific to the rest catalog spec. `src/iceberg/catalog/rest/json_serde.cc` is the home. Java `TableScanResponseParser.java` is doing the same thing.

src/iceberg/json_serde_internal.h
```diff
@@ -413,4 +417,39 @@ ICEBERG_EXPORT nlohmann::json ToJson(const TableRequirement& requirement);
...
+    const nlohmann::json& json,
+    const std::unordered_map<int32_t, std::shared_ptr<PartitionSpec>>& partitionSpecById,
```
wgtmac:
```suggestion
    const std::unordered_map<int32_t, std::shared_ptr<PartitionSpec>>& partition_spec_by_id,
```

wgtmac:
I've just finished my first round of review. Thanks @gsandeep1241 for working on this and everyone for the review! Let me know what you think.

src/iceberg/catalog/rest/types.h
```diff
@@ -295,4 +298,73 @@ struct ICEBERG_REST_EXPORT OAuthTokenResponse {
...
+
+  Status Validate() const { return {}; };
+
+  bool operator==(const BaseScanTaskResponse&) const;
+};
+
+/// \brief Response from initiating a scan planning operation, including plan status and
+/// initial scan tasks.
+struct ICEBERG_REST_EXPORT PlanTableScanResponse : BaseScanTaskResponse {
+  std::string plan_status;
```
wgtmac: Is it better to use an enum to define the plan status?

src/iceberg/catalog/rest/error_handlers.cc
gsandeep1241: Done!

src/iceberg/catalog/rest/error_handlers.cc
gsandeep1241: NoSuchPlanTask is definitely not required. After @zhjwpku 's comment, I created a new PlanTaskHandler and handled it there. This is a leftover. Removing it now. 

As for the 406, I got it from here: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L705

This is not handled in the DefaultErrorHandler. Wdyt?

src/iceberg/catalog/rest/resource_paths.cc
gsandeep1241: Done!

src/iceberg/catalog/rest/resource_paths.h
gsandeep1241: Makes sense, done.

src/iceberg/catalog/rest/types.h
gsandeep1241: Done.

src/iceberg/catalog/rest/types.h
gsandeep1241: Done

src/iceberg/catalog/rest/types.h
gsandeep1241: I think you mean vector<shared_ptr>, right?

src/iceberg/catalog.h
gsandeep1241: I debated this and I felt that a catalog handler should really honor a Request-Response contract. (That said, I'm not doing the "Request" part as such). The java implementation I felt handled core logic directly coupled with the handlers.  

But if we want to keep it similar, it makes sense to split into two PRs to prevent this from getting complicated. I'll remove the stuff in catalog.h and rest_catalog.h. 

Agree?

gsandeep1241 (commented):
Thanks @wgtmac  for the detailed review. I am part way into addressing your comments. I'm posting the partial comments early because there was one open question I wanted to check - changes to catalog.h because if we go that route, we can simplify the PR. 

The rest of your comments I  don't have a conflict with and will address them.

src/iceberg/catalog/rest/error_handlers.cc
wgtmac: Ah, that's a good call. Now I'm convinced to keep it here. Thanks!

src/iceberg/catalog.h
gsandeep1241: Sounds good, I removed the changes to this file and will handle it in a different PR as discussed below.

src/iceberg/catalog.h
gsandeep1241: Sounds good, I removed the changes to this file and will handle it in a different PR as discussed below.

src/iceberg/catalog.h
gsandeep1241: Removed the changes. from this PR.

src/iceberg/catalog/rest/types.h
gsandeep1241: Makes sense, done.

src/iceberg/catalog/rest/types.h
gsandeep1241: Done!

src/iceberg/json_serde_internal.h
gsandeep1241: Done!

src/iceberg/json_serde_internal.h
gsandeep1241: Done!

src/iceberg/json_serde_internal.h
gsandeep1241: Done.

src/iceberg/catalog/rest/json_serde_internal.h
gsandeep1241: Done!

wgtmac:
Is this ready to review? I see there are CI failures and unaddressed comments. Please also rebase it to the latest main branch to make sure namespace separator is respected.

src/iceberg/catalog/rest/resource_paths.cc
gsandeep1241: Done!

src/iceberg/catalog/rest/types.h
gsandeep1241: Good point on the virtual/non-virtual issue! But instead I'd propose flattening into three independent structs rather than have the variant. 

The variant pushes complexity onto every caller. Each response type has its own dedicated fromJson method, so we already know the type there and variant would force a runtime check.  

Wdyt?

src/iceberg/test/rest_json_serde_test.cc
gsandeep1241: Done!

gsandeep1241:
@wgtmac Addressed all comments (except one, where I left a question for you). Please take a look when you get the chance. Thank you!

src/iceberg/catalog/rest/types.h
wgtmac: I'm open to this. Thanks!

src/iceberg/catalog/rest/json_serde.cc
```diff
@@ -78,6 +83,406 @@ constexpr std::string_view kExpiresIn = "expires_in";
...
+  for (const auto& task : response.file_scan_tasks) {
+    if (!task) continue;
+    nlohmann::json task_json;
+    if (task->data_file()) {
+      task_json[kDataFile] = ToJson(*task->data_file());
+    }
+    if (!task->delete_files().empty()) {
+      std::vector<int32_t> refs;
+      for (const auto& df : task->delete_files()) {
+        auto it = delete_file_index.find(df.get());
```
wgtmac: Using pointer identity here can silently drop delete references when `response.delete_files` and `task->delete_files()` hold equivalent `DataFile`s through different `shared_ptr`s. Please key by stable file identity, as Java does with delete file location, or derive the response delete list from task deletes.

src/iceberg/catalog/rest/types.h
```diff
@@ -295,4 +299,73 @@ struct ICEBERG_REST_EXPORT OAuthTokenResponse {
...
+
+  Status Validate() const { return {}; };
+
+  bool operator==(const BaseScanTaskResponse&) const;
+};
+
+/// \brief Response from initiating a scan planning operation, including plan status and
+/// initial scan tasks.
+struct ICEBERG_REST_EXPORT PlanTableScanResponse : BaseScanTaskResponse {
+  PlanStatus plan_status = PlanStatus::kCompleted;
```
wgtmac, line 332: Failed planning responses also carry the REST `error` payload. Please model and parse/serialize it for both `PlanTableScanResponse` and `FetchPlanningResultResponse`; otherwise callers lose the server error type/message/code.

src/iceberg/catalog/rest/json_serde.cc
```diff
@@ -78,6 +83,406 @@ constexpr std::string_view kExpiresIn = "expires_in";
...
+    }
+
+    file_scan_tasks.push_back(std::make_shared<FileScanTask>(
+        std::make_shared<DataFile>(std::move(data_file)), std::move(task_delete_files),
+        std::move(residual_filter)));
+  }
+  return file_scan_tasks;
+}
+
+nlohmann::json ToJson(const DataFile& df) {
```
wgtmac: `partition` is a required ContentFile field in the REST schema, and Java's `ContentFileParser` writes it for scan tasks. This serializer never emits it, so partitioned scan tasks are not spec-compatible and lose partition data on round-trip.

wgtmac:
I've pushed a commit to tighten the REST scan JSON serde to match Java/spec behavior more closely. This includes preserving residual filters, distinguishing missing vs empty scan task fields, requiring contextual DataFile serialization/deserialization with partition spec/schema validation, adding request-side FromJson support for scan requests, and simplifying delete-file deduplication by reusing DeleteFileSet. I also adjusted the tests to cover these parity cases in the existing style.

Thanks @gsandeep1241 for working on this!

wgtmac (approved):

PR #624, opened 2026-04-23
feat(puffin): add puffin file reader and writer

zhaoxuan1994:
- PuffinWriter: in-memory writer that builds complete Puffin files
  - Add() writes blobs with optional compression
  - Finish() serializes footer with JSON metadata
  - Tracks BlobMetadata for all written blobs
- PuffinReader: in-memory reader that parses Puffin files
  - ReadFileMetadata() parses footer and validates magic bytes
  - ReadBlob() reads and decompresses individual blobs
  - ReadAll() reads all blobs from metadata
- Expose Compress/Decompress as public API in puffin_format.h
- Register new sources in CMake and Meson build systems
- Add comprehensive tests

### DIFF
--- a/src/iceberg/puffin/puffin_format.cc
+++ b/src/iceberg/puffin/puffin_format.cc
@@ -36,6 +36,18 @@ constexpr std::pair<int, int> GetFlagPosition(PuffinFlag flag) {
   std::unreachable();
 }
 
+}  // namespace
+
+bool IsFlagSet(std::span<const uint8_t, 4> flags, PuffinFlag flag) {
+  auto [byte_num, bit_num] = GetFlagPosition(flag);
+  return (flags[byte_num] & (1 << bit_num)) != 0;
+}
+
+void SetFlag(std::span<uint8_t, 4> flags, PuffinFlag flag) {
+  auto [byte_num, bit_num] = GetFlagPosition(flag);
+  flags[byte_num] |= (1 << bit_num);
+}
+
 // TODO(zhaoxuan1994): Move compression logic to a unified codec interface.
 Result<std::vector<std::byte>> Compress(PuffinCompressionCodec codec,
                                         std::span<const std::byte> input) {
@@ -61,18 +73,6 @@ Result<std::vector<std::byte>> Decompress(PuffinCompressionCodec codec,
       return NotSupported("Zstd decompression is not yet supported");
   }
   std::unreachable();
-}
-
-}  // namespace
-
-bool IsFlagSet(std::span<const uint8_t, 4> flags, PuffinFlag flag) {
-  auto [byte_num, bit_num] = GetFlagPosition(flag);
-  return (flags[byte_num] & (1 << bit_num)) != 0;
-}
-
-void SetFlag(std::span<uint8_t, 4> flags, PuffinFlag flag) {
-  auto [byte_num, bit_num] = GetFlagPosition(flag);
-  flags[byte_num] |= (1 << bit_num);
 }
 
 }  // namespace iceberg::puffin
--- a/src/iceberg/puffin/puffin_reader.cc
+++ b/src/iceberg/puffin/puffin_reader.cc
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/puffin/puffin_reader.h"
+
+#include <algorithm>
+#include <array>
+#include <cstring>
+#include <string_view>
+
+#include "iceberg/puffin/json_serde_internal.h"
+#include "iceberg/puffin/puffin_format.h"
+#include "iceberg/util/endian.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg::puffin {
+
+namespace {
+
+// Validate magic bytes at the given offset.
+Status CheckMagic(std::span<const std::byte> data, int64_t offset) {
+  if (offset < 0 ||
+      offset + PuffinFormat::kMagicLength > static_cast<int64_t>(data.size())) {
+    return Invalid("Invalid file: cannot read magic at offset {}", offset);
+  }
+  auto* begin = reinterpret_cast<const uint8_t*>(data.data() + offset);
+  if (!std::equal(PuffinFormat::kMagicV1.begin(), PuffinFormat::kMagicV1.end(), begin)) {
+    return Invalid("Invalid file: expected magic at offset {}", offset);
+  }
+  return {};
+}
+
+}  // namespace
+
+PuffinReader::PuffinReader(std::span<const std::byte> data) : data_(data) {}
+
+Result<FileMetadata> PuffinReader::ReadFileMetadata() {
+  auto file_size = static_cast<int64_t>(data_.size());
+
+  if (file_size < PuffinFormat::kFooterStructLength) {
+    return Invalid("Invalid file: file length {} is less than minimal footer size {}",
+                   file_size, PuffinFormat::kFooterStructLength);
+  }
+
+  // Read footer struct from end of file
+  auto footer_struct_offset = file_size - PuffinFormat::kFooterStructLength;
+
+  // Validate footer end magic
+  ICEBERG_RETURN_UNEXPECTED(
+      CheckMagic(data_, footer_struct_offset + PuffinFormat::kFooterStructMagicOffset));
+
+  // Read payload size from footer struct
+  auto payload_size = ReadLittleEndian<int32_t>(
+      data_.data() + footer_struct_offset + PuffinFormat::kFooterStructPayloadSizeOffset);
+
+  if (payload_size < 0) {
+    return Invalid("Invalid file: negative payload size {}", payload_size);
+  }
+
+  // Calculate total footer size and validate
+  int64_t footer_size = PuffinFormat::kFooterStartMagicLength +
+                        static_cast<int64_t>(payload_size) +
+                        PuffinFormat::kFooterStructLength;
+  auto footer_offset = file_size - footer_size;
+  if (footer_offset < 0) {
+    return Invalid("Invalid file: footer size {} exceeds file size {}", footer_size,
+                   file_size);
+  }
+
+  // Validate footer start magic
+  ICEBERG_RETURN_UNEXPECTED(CheckMagic(data_, footer_offset));
+
+  // Check flags for footer compression
+  std::array<uint8_t, 4> flags{};
+  std::memcpy(
+      flags.data(),
+      data_.data() + footer_struct_offset + PuffinFormat::kFooterStructFlagsOffset, 4);
+
+  PuffinCompressionCodec footer_compression = PuffinCompressionCodec::kNone;
+  if (IsFlagSet(flags, PuffinFlag::kFooterPayloadCompressed)) {
+    footer_compression = PuffinFormat::kDefaultFooterCompressionCodec;
+  }
+
+  // Extract footer payload
+  auto payload_offset = footer_offset + PuffinFormat::kFooterStartMagicLength;
+  std::span<const std::byte> payload_span(data_.data() + payload_offset, payload_size);
+  ICEBERG_ASSIGN_OR_RAISE(auto payload_bytes,
+                          Decompress(footer_compression, payload_span));
+
+  // Parse JSON
+  std::string_view json_str(reinterpret_cast<const char*>(payload_bytes.data()),
+                            payload_bytes.size());
+  ICEBERG_ASSIGN_OR_RAISE(auto file_metadata, FileMetadataFromJsonString(json_str));
+
+  // Validate header magic
+  ICEBERG_RETURN_UNEXPECTED(CheckMagic(data_, 0));
+
+  return file_metadata;
+}
+
+Result<std::pair<BlobMetadata, std::vector<std::byte>>> PuffinReader::ReadBlob(
+    const BlobMetadata& blob_metadata) {
+  auto file_size = static_cast<int64_t>(data_.size());
+
+  if (blob_metadata.offset < 0 || blob_metadata.length < 0 ||
+      blob_metadata.offset > file_size ||
+      blob_metadata.length > file_size - blob_metadata.offset) {
+    return Invalid("Invalid blob: offset {} + length {} exceeds file size {}",
+                   blob_metadata.offset, blob_metadata.length, file_size);
+  }
+
+  std::span<const std::byte> raw_data(data_.data() + blob_metadata.offset,
+                                      blob_metadata.length);
+
+  // Determine compression codec
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto codec, PuffinCompressionCodecFromName(blob_metadata.compression_codec));
+  ICEBERG_ASSIGN_OR_RAISE(auto decompressed, Decompress(codec, raw_data));
+
+  return std::pair{blob_metadata, std::move(decompressed)};
+}
+
+Result<std::vector<std::pair<BlobMetadata, std::vector<std::byte>>>>
+PuffinReader::ReadAll(const std::vector<BlobMetadata>& blobs) {
+  std::vector<std::pair<BlobMetadata, std::vector<std::byte>>> results;
+  results.reserve(blobs.size());
+  for (const auto& blob : blobs) {
+    ICEBERG_ASSIGN_OR_RAISE(auto blob_pair, ReadBlob(blob));
+    results.push_back(std::move(blob_pair));
+  }
+  return results;
+}
+
+}  // namespace iceberg::puffin
--- a/src/iceberg/puffin/puffin_reader.cc
+++ b/src/iceberg/puffin/puffin_reader.cc
@@ -1,0 +1,202 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/puffin/puffin_reader.h"
+
+#include <algorithm>
+#include <array>
+#include <cstdint>
+#include <cstring>
+#include <string_view>
+
+#include "iceberg/file_io.h"
+#include "iceberg/puffin/json_serde_internal.h"
+#include "iceberg/puffin/puffin_format.h"
+#include "iceberg/util/endian.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg::puffin {
+
+namespace {
+
+// Validate magic bytes in a buffer at offset 0.
+Status CheckMagic(std::span<const std::byte> data) {
+  if (static_cast<int64_t>(data.size()) < PuffinFormat::kMagicLength) {
+    return Invalid("Invalid file: buffer too small for magic");
+  }
+  auto* begin = reinterpret_cast<const uint8_t*>(data.data());
+  if (!std::equal(PuffinFormat::kMagicV1.begin(), PuffinFormat::kMagicV1.end(), begin)) {
+    return Invalid(
+        "Invalid file: expected magic, got [{:#04x}, {:#04x}, {:#04x}, {:#04x}]",
+        begin[0], begin[1], begin[2], begin[3]);
+  }
+  return {};
+}
+
+// Validate that no unknown flag bits are set.
+Status CheckUnknownFlags(std::span<const uint8_t, 4> flags) {
+  constexpr uint8_t kKnownBitsMask = 0x01;
+  if ((flags[0] & ~kKnownBitsMask) != 0 || flags[1] != 0 || flags[2] != 0 ||
+      flags[3] != 0) {
+    return Invalid(
+        "Invalid file: unknown footer flags set [{:#04x}, {:#04x}, {:#04x}, {:#04x}]",
+        flags[0], flags[1], flags[2], flags[3]);
+  }
+  return {};
+}
+
+}  // namespace
+
+PuffinReader::PuffinReader(std::unique_ptr<SeekableInputStream> stream, int64_t file_size,
+                           std::optional<int64_t> known_footer_size)
+    : stream_(std::move(stream)),
+      file_size_(file_size),
+      known_footer_size_(known_footer_size) {}
+
+PuffinReader::~PuffinReader() = default;
+
+Result<std::unique_ptr<PuffinReader>> PuffinReader::Make(
+    std::unique_ptr<InputFile> input_file, std::optional<int64_t> footer_size) {
+  if (!input_file) {
+    return InvalidArgument("input_file must not be null");
+  }
+  ICEBERG_ASSIGN_OR_RAISE(auto file_size, input_file->Size());
+  ICEBERG_ASSIGN_OR_RAISE(auto stream, input_file->Open());
+  return std::unique_ptr<PuffinReader>(
+      new PuffinReader(std::move(stream), file_size, footer_size));
+}
+
+Result<std::vector<std::byte>> PuffinReader::ReadBytes(int64_t offset, int64_t length) {
+  if (offset < 0 || length < 0 || offset > file_size_ || length > file_size_ - offset) {
+    return Invalid("Read out of bounds: offset {} + length {} exceeds file size {}",
+                   offset, length, file_size_);
+  }
+  std::vector<std::byte> buf(length);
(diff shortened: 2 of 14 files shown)


src/iceberg/puffin/puffin_format.cc
```diff
@@ -36,6 +36,18 @@ constexpr std::pair<int, int> GetFlagPosition(PuffinFlag flag) {
   std::unreachable();
 }
 
+}  // namespace
```
kamcheungting-db, line 39: is there any reason to move both `Result<std::vector<std::byte>> Compress` and `Result<std::vector<std::byte>> Decompress` out of the namespace?

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,150 @@
...
+
+// Validate magic bytes at the given offset.
+Status CheckMagic(std::span<const std::byte> data, int64_t offset) {
+  if (offset < 0 ||
+      offset + PuffinFormat::kMagicLength > static_cast<int64_t>(data.size())) {
+    return Invalid("Invalid file: cannot read magic at offset {}", offset);
+  }
+  auto* begin = reinterpret_cast<const uint8_t*>(data.data() + offset);
+  if (!std::equal(PuffinFormat::kMagicV1.begin(), PuffinFormat::kMagicV1.end(), begin)) {
+    return Invalid("Invalid file: expected magic at offset {}", offset);
```
kamcheungting-db: please also save the value of incorrect magic number into `Invalid` for easy debugging.

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,150 @@
...
+
+  // Validate footer end magic
+  ICEBERG_RETURN_UNEXPECTED(
+      CheckMagic(data_, footer_struct_offset + PuffinFormat::kFooterStructMagicOffset));
+
+  // Read payload size from footer struct
+  auto payload_size = ReadLittleEndian<int32_t>(
+      data_.data() + footer_struct_offset + PuffinFormat::kFooterStructPayloadSizeOffset);
+
+  if (payload_size < 0) {
```
kamcheungting-db: should `payload_size == 0` also be considered as an error.

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // Validate footer start magic
+  ICEBERG_RETURN_UNEXPECTED(CheckMagic(data_, footer_offset));
+
+  // Check flags for footer compression
+  std::array<uint8_t, 4> flags{};
+  std::memcpy(
+      flags.data(),
+      data_.data() + footer_struct_offset + PuffinFormat::kFooterStructFlagsOffset, 4);
```
kamcheungting-db:
```suggestion
  std::memcpy(
      flags.data(),
      data_.data() + footer_struct_offset + PuffinFormat::kFooterStructFlagsOffset,
      4);
```

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,145 @@
...
+  PuffinCompressionCodec footer_compression = PuffinCompressionCodec::kNone;
+  if (IsFlagSet(flags, PuffinFlag::kFooterPayloadCompressed)) {
+    footer_compression = PuffinFormat::kDefaultFooterCompressionCodec;
+  }
+
+  // Extract footer payload
+  auto payload_offset = footer_offset + PuffinFormat::kFooterStartMagicLength;
+  std::span<const std::byte> payload_span(data_.data() + payload_offset, payload_size);
+  ICEBERG_ASSIGN_OR_RAISE(auto payload_bytes,
+                          Decompress(footer_compression, payload_span));
```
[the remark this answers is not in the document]
kamcheungting-db: payload_size is already validated above:
```
if (payload_size < 0) {
  return Invalid("Invalid file: negative payload size {}", payload_size);
}
```

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,150 @@
...
+      data_.data() + footer_struct_offset + PuffinFormat::kFooterStructFlagsOffset, 4);
+
+  PuffinCompressionCodec footer_compression = PuffinCompressionCodec::kNone;
+  if (IsFlagSet(flags, PuffinFlag::kFooterPayloadCompressed)) {
+    footer_compression = PuffinFormat::kDefaultFooterCompressionCodec;
+  }
+
+  // Extract footer payload
+  auto payload_offset = footer_offset + PuffinFormat::kFooterStartMagicLength;
+  std::span<const std::byte> payload_span(data_.data() + payload_offset, payload_size);
```
kamcheungting-db:
```suggestion
  const std::span<const std::byte> payload_span(data_.data() + payload_offset, payload_size);
```

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,150 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto payload_bytes,
+                          Decompress(footer_compression, payload_span));
+
+  // Parse JSON
+  std::string_view json_str(reinterpret_cast<const char*>(payload_bytes.data()),
+                            payload_bytes.size());
+  ICEBERG_ASSIGN_OR_RAISE(auto file_metadata, FileMetadataFromJsonString(json_str));
+
+  // Validate header magic
+  ICEBERG_RETURN_UNEXPECTED(CheckMagic(data_, 0));
```
kamcheungting-db: should we move this validation to the beginning of this method?

src/iceberg/puffin/puffin_reader.h
```diff
@@ -0,0 +1,66 @@
...
+  Result<std::pair<BlobMetadata, std::vector<std::byte>>> ReadBlob(
+      const BlobMetadata& blob_metadata);
+
+  /// \brief Read all blobs described in the file metadata.
+  /// \return A vector of (BlobMetadata, decompressed data) pairs, or an error.
+  Result<std::vector<std::pair<BlobMetadata, std::vector<std::byte>>>> ReadAll(
+      const std::vector<BlobMetadata>& blobs);
+
+ private:
+  std::span<const std::byte> data_;
```
kamcheungting-db:
```suggestion
  const std::span<const std::byte> data_;
```

src/iceberg/puffin/puffin_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+namespace iceberg::puffin {
+
+PuffinWriter::PuffinWriter(PuffinCompressionCodec default_codec)
+    : default_codec_(default_codec) {}
+
+void PuffinWriter::WriteHeader() {
+  if (header_written_) return;
+  const auto& magic = PuffinFormat::kMagicV1;
+  buffer_.insert(buffer_.end(), reinterpret_cast<const std::byte*>(magic.data()),
+                 reinterpret_cast<const std::byte*>(magic.data() + magic.size()));
```
kamcheungting-db:
```suggestion
  const auto start_byte = magic.data();
  buffer_.insert(buffer_.end(),
                 reinterpret_cast<const std::byte*>(start_byte),
                 reinterpret_cast<const std::byte*>(start_byte + magic.size()));
```

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,80 @@
...
+  /// \return The footer size, or std::nullopt if Finish() has not been called.
+  std::optional<int64_t> footer_size() const;
+
+ private:
+  PuffinCompressionCodec default_codec_;
+  std::vector<std::byte> buffer_;
+  std::vector<BlobMetadata> written_blobs_metadata_;
+  bool header_written_ = false;
+  bool finished_ = false;
+  std::optional<int64_t> footer_size_;
```
kamcheungting-db: please add one line comment for these varaibles

src/iceberg/puffin/puffin_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+#include "iceberg/util/endian.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg::puffin {
+
+PuffinWriter::PuffinWriter(PuffinCompressionCodec default_codec)
+    : default_codec_(default_codec) {}
+
+void PuffinWriter::WriteHeader() {
+  if (header_written_) return;
```
kamcheungting-db: shouldn't we error out here?
It looks like dangerous to call WriteHeader() multiple time.

kamcheungting-db:
```suggestion
  if (header_written_) {  
      return Invalid("Header has already been written.");
  }
```

src/iceberg/puffin/puffin_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+
+  WriteHeader();
+
+  FileMetadata file_metadata{
+      .blobs = written_blobs_metadata_,
+      .properties = std::move(properties),
+  };
+
+  auto footer_json = ToJsonString(file_metadata);
+  auto footer_payload = std::span<const std::byte>(
```
kamcheungting-db:
```suggestion
  const auto footer_payload = std::span<const std::byte>(
```

src/iceberg/puffin/puffin_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+  WriteLittleEndian(payload_size, size_buf.data());
+  buffer_.insert(buffer_.end(), size_buf.begin(), size_buf.end());
+
+  // Flags (no compression for now)
+  std::array<std::byte, 4> flags{};
+  buffer_.insert(buffer_.end(), flags.begin(), flags.end());
+
+  // Footer end magic
+  buffer_.insert(buffer_.end(), reinterpret_cast<const std::byte*>(magic.data()),
+                 reinterpret_cast<const std::byte*>(magic.data() + magic.size()));
```
kamcheungting-db:
```suggestion
  buffer_.insert(buffer_.end(),
                 reinterpret_cast<const std::byte*>(magic.data()),
                 reinterpret_cast<const std::byte*>(magic.data() + magic.size()));
```

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,80 @@
...
+
+  /// \brief Add a blob to be written.
+  /// \return The BlobMetadata for the written blob, or an error.
+  Result<BlobMetadata> Add(const Blob& blob);
+
+  /// \brief Finalize the file and return the serialized bytes.
+  /// \param properties File-level properties to include in the footer.
+  /// \return The complete Puffin file as a byte vector, or an error.
+  Result<std::vector<std::byte>> Finish(
+      std::unordered_map<std::string, std::string> properties = {});
```
kamcheungting-db: do we really need to copy the whole `std::unordered_map<std::string, std::string>`

kamcheungting-db (commented):
Some minor comments.
Nice code

src/iceberg/puffin/puffin_reader.h
```diff
@@ -0,0 +1,66 @@
...
+#include "iceberg/puffin/file_metadata.h"
+#include "iceberg/result.h"
+
+namespace iceberg::puffin {
+
+/// \brief Reader for Puffin files.
+///
+/// Parses a Puffin file from an in-memory buffer. Usage:
+///   PuffinReader reader(file_data);
+///   auto metadata = reader.ReadFileMetadata();
```
wgtmac: This might not be the recommended approach unless error handling has been added. Otherwise users may silently ignore the error and call `metadata.value()` unconsciously.

src/iceberg/puffin/puffin_reader.h
```diff
@@ -0,0 +1,66 @@
...
+///   auto blob = reader.ReadBlob(metadata.value().blobs[0]);
+class ICEBERG_EXPORT PuffinReader {
+ public:
+  /// \brief Construct a reader from file data.
+  explicit PuffinReader(std::span<const std::byte> data);
```
wgtmac: The main issue of this ctor is that we lose the capability of seeking into a segment of a puffin file, which is the main use case of v3 deletion vector.

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,150 @@
...
+  ICEBERG_RETURN_UNEXPECTED(CheckMagic(data_, footer_offset));
+
+  // Check flags for footer compression
+  std::array<uint8_t, 4> flags{};
+  std::memcpy(
+      flags.data(),
+      data_.data() + footer_struct_offset + PuffinFormat::kFooterStructFlagsOffset, 4);
+
+  PuffinCompressionCodec footer_compression = PuffinCompressionCodec::kNone;
+  if (IsFlagSet(flags, PuffinFlag::kFooterPayloadCompressed)) {
```
wgtmac: Please validate that all unknown/reserved footer flag bits are unset. Java's PuffinReader rejects unknown flags, and the spec says reserved bits should be 0; silently ignoring them may cause C++ to accept future or invalid files and interpret the footer incorrectly.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/puffin/file_metadata.h"
+#include "iceberg/result.h"
+
+namespace iceberg::puffin {
+
+/// \brief Writer for Puffin files.
+///
+/// Builds a complete Puffin file in memory. Usage:
```
wgtmac: Same as my comment in the reader, I think we should remove comment like this.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,80 @@
...
+///   // result.value() contains the serialized file bytes
+class ICEBERG_EXPORT PuffinWriter {
+ public:
+  /// \brief Construct a writer with the given default compression codec.
+  explicit PuffinWriter(
+      PuffinCompressionCodec default_codec = PuffinCompressionCodec::kNone);
+
+  /// \brief Add a blob to be written.
+  /// \return The BlobMetadata for the written blob, or an error.
+  Result<BlobMetadata> Add(const Blob& blob);
```
wgtmac: Do we really need to return a copy of `BlobMetadata` here? IMO, users usually do not care about it and it pays to do so.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,80 @@
...
+///   PuffinWriter writer;
+///   writer.Add(blob1);
+///   writer.Add(blob2);
+///   auto result = writer.Finish({{"created-by", "iceberg-cpp"}});
+///   // result.value() contains the serialized file bytes
+class ICEBERG_EXPORT PuffinWriter {
+ public:
+  /// \brief Construct a writer with the given default compression codec.
+  explicit PuffinWriter(
+      PuffinCompressionCodec default_codec = PuffinCompressionCodec::kNone);
```
wgtmac: Do we need to introduce a `PuffinWriterOptions` to hold all possible (but optional) configs in case of expanding the ctor signature in the future?

src/iceberg/puffin/puffin_writer.h
wgtmac: Why do not set the properties while creating the writer?

wgtmac: BTW, it seems that we need to extend our `FileIO` abstraction to support puffin I/O. It looks inefficient to write all blobs into a buffer like this.

wgtmac (commented):
Sorry for the long delay. I have some concerns about its API design which would be inefficient in the real world use case. I believe the main issue arises from the lack of streaming support of FileIO. I will put up a PR to address this.

wgtmac:
I've merged https://github.com/apache/iceberg-cpp/pull/641 so we can rebase on it to design better API for puffin.

src/iceberg/puffin/puffin_reader.cc
zhaoxuan1994: payload_size == 0 is valid — it represents an empty footer payload (e.g. {"blobs":[]}). Java's PuffinReader also does not reject zero payload size. Only negative values indicate corruption.

src/iceberg/puffin/puffin_reader.cc
zhaoxuan1994: Yeah I thought about this — span is basically a pointer+size pair like string_view, so const on the member itself doesn't buy us much. The span is set once in the constructor and never reassigned. I'll leave it as-is to stay consistent with how the rest of the project uses span/string_view members.

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,204 @@
...
+    return Invalid("Read out of bounds: offset {} + length {} exceeds file size {}",
+                   offset, length, file_size_);
+  }
+  return std::vector<std::byte>(data_.data() + offset, data_.data() + offset + length);
+}
+
+Result<FileMetadata> PuffinReader::ReadFileMetadata() {
+  // Get file size
+  if (IsFileMode()) {
+    ICEBERG_ASSIGN_OR_RAISE(file_size_, input_file_->Size());
```
wgtmac: Please fail on unknown footer flags here, matching Java. Ignoring extra bits can make newer or corrupted Puffin files look valid.

src/iceberg/puffin/puffin_reader.h
```diff
@@ -0,0 +1,86 @@
...
+class InputFile;
+class SeekableInputStream;
+}  // namespace iceberg
+
+namespace iceberg::puffin {
+
+/// \brief Reader for Puffin files.
+///
+/// Supports two modes:
+/// - Memory mode: parses from an in-memory buffer.
```
wgtmac: Can we avoid making the public Puffin API in-memory only? Java and the C++ IO layer are file/stream based, which fits object stores and range reads better.

src/iceberg/puffin/puffin_reader.h
```diff
@@ -0,0 +1,86 @@
...
+/// - File mode: reads from an InputFile with seek support (for DV use case).
+class ICEBERG_DATA_EXPORT PuffinReader {
+ public:
+  /// \brief Construct a memory-mode reader from file data.
+  explicit PuffinReader(std::span<const std::byte> data);
```
wgtmac: Let's remove the api for pure in-memory data which distracts and complicates its implementation.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,115 @@
...
+namespace iceberg::puffin {
+
+/// \brief Writer for Puffin files.
+///
+/// Supports two modes:
+/// - Stream mode: writes directly to an OutputFile/PositionOutputStream.
+/// - Memory mode: builds the file in an internal buffer and returns it via Finish().
+class ICEBERG_DATA_EXPORT PuffinWriter {
+ public:
+  /// \brief Construct a memory-mode writer.
```
wgtmac: Let's remove this use case not to complicate the implementation.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,115 @@
...
+      PuffinCompressionCodec default_codec = PuffinCompressionCodec::kNone);
+
+  /// \brief Construct a stream-mode writer from an OutputFile.
+  /// \param output_file The output file to write to.
+  /// \param properties File-level properties to include in the footer.
+  /// \param default_codec Default compression codec for blobs.
+  explicit PuffinWriter(
+      std::unique_ptr<OutputFile> output_file,
+      std::unordered_map<std::string, std::string> properties = {},
+      PuffinCompressionCodec default_codec = PuffinCompressionCodec::kNone);
```
wgtmac: Why there is no `bool compressFooter` parameter?

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,115 @@
...
+
+  ~PuffinWriter();
+
+  /// \brief Add a blob to be written.
+  Status Add(const Blob& blob);
```
wgtmac: Why there is no `Result<BlobMetadata> Write(const Blob& blob)`?

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,115 @@
...
+  ///
+  /// In memory mode, returns the complete serialized file bytes.
+  /// In stream mode, flushes and closes the stream, returns empty vector.
+  Result<std::vector<std::byte>> Finish();
+
+  /// \brief Get metadata for all blobs written so far.
+  const std::vector<BlobMetadata>& written_blobs_metadata() const;
+
+  /// \brief Get the footer size after Finish() has been called.
+  std::optional<int64_t> footer_size() const;
```
wgtmac: Return `Result<int64_t>` to emit unclosed error.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,115 @@
...
+
+  /// \brief Get the total file size after Finish() has been called.
+  std::optional<int64_t> file_size() const;
```
wgtmac: ditto

src/iceberg/puffin/puffin_reader.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /// \brief Construct a file-mode reader from an InputFile.
+  explicit PuffinReader(std::unique_ptr<InputFile> input_file);
```
wgtmac: We cannot throw exception in practice, so we may need a static function like `Result<std::unique_ptr<PuffinReader>> Make(std::unique_ptr<InputFile> input_file)` to handle any error, like null input. Same for writer.

wgtmac (changes requested):
Thanks for incorporating the new FileIO abstraction! My main concern is the API inconsistency compared to Java. Let me know what you think.

src/iceberg/puffin/puffin_reader.h
wgtmac: BTW, do we want to provide footerSize as the hint? This saves a seek btw.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,106 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "iceberg/iceberg_data_export.h"
+#include "iceberg/puffin/file_metadata.h"
+#include "iceberg/result.h"
+
+namespace iceberg {
+class OutputFile;
```
wgtmac: nit: we should put these forward declarations in `type_fwd.h` and include it here.

src/iceberg/puffin/puffin_writer.cc
```diff
@@ -0,0 +1,176 @@
...
+PuffinWriter::~PuffinWriter() = default;
+
+Result<std::unique_ptr<PuffinWriter>> PuffinWriter::Make(
+    std::unique_ptr<OutputFile> output_file,
+    std::unordered_map<std::string, std::string> properties,
+    PuffinCompressionCodec default_codec, bool compress_footer) {
+  if (!output_file) {
+    return InvalidArgument("output_file must not be null");
+  }
+  ICEBERG_ASSIGN_OR_RAISE(auto stream, output_file->CreateOrOverwrite());
```
wgtmac: Java PuffinWriter opens the output with OutputFile.create(), which fails when the target already exists. This path uses CreateOrOverwrite(), so a caller can silently truncate an existing Puffin file instead of getting the Java-aligned failure.

src/iceberg/puffin/puffin_reader.cc
```diff
@@ -0,0 +1,202 @@
...
+      known_footer_size_(known_footer_size) {}
+
+PuffinReader::~PuffinReader() = default;
+
+Result<std::unique_ptr<PuffinReader>> PuffinReader::Make(
+    std::unique_ptr<InputFile> input_file, std::optional<int64_t> footer_size) {
+  if (!input_file) {
+    return InvalidArgument("input_file must not be null");
+  }
+  ICEBERG_ASSIGN_OR_RAISE(auto file_size, input_file->Size());
```
wgtmac: The reader accepts a known footer_size, but Make still always calls InputFile::Size() and ReadFileMetadata recomputes the footer size from the file tail. Java's ReadBuilder supports both withFileSize and withFooterSize and uses/validates the hints, so this API is not strictly aligned and loses the intended object-store optimization.

src/iceberg/puffin/puffin_writer.cc
```diff
@@ -0,0 +1,176 @@
...
+  ICEBERG_RETURN_UNEXPECTED(WriteBytes(size_buf));
+
+  // Flags
+  ICEBERG_RETURN_UNEXPECTED(WriteBytes(std::span<const std::byte>(
+      reinterpret_cast<const std::byte*>(flags.data()), flags.size())));
+
+  // Footer end magic
+  ICEBERG_RETURN_UNEXPECTED(WriteMagic());
+
+  ICEBERG_ASSIGN_OR_RAISE(auto end_pos, stream_->Position());
```
wgtmac, line 153: Finish records file_size/footer_size and sets finished_ before Flush/Close returns. Java only marks the writer finished and records storedLength after close, because some streams emit final bytes on close. If Close fails or finalizes extra bytes, this writer reports a finished state and wrong size.

src/iceberg/puffin/puffin_writer.h
```diff
@@ -0,0 +1,106 @@
...
+  /// \param properties File-level properties to include in the footer.
+  /// \param default_codec Default compression codec for blobs.
+  /// \param compress_footer Whether to compress the footer payload.
+  static Result<std::unique_ptr<PuffinWriter>> Make(
+      std::unique_ptr<OutputFile> output_file,
+      std::unordered_map<std::string, std::string> properties = {},
+      PuffinCompressionCodec default_codec = PuffinCompressionCodec::kNone,
+      bool compress_footer = false);
+
+  ~PuffinWriter();
```
wgtmac, line 56: Java PuffinWriter implements close(), and close() implicitly finishes an unfinished file. This API exposes only Finish() and the destructor is a no-op, so ported code has no Java-equivalent close path and can leave an invalid Puffin file if Finish() is not called explicitly.

src/iceberg/puffin/puffin_reader.h
```diff
@@ -0,0 +1,85 @@
...
+class ICEBERG_DATA_EXPORT PuffinReader {
+ public:
+  /// \brief Create a PuffinReader for the given input file.
+  /// \param input_file The input file to read from.
+  /// \param footer_size Optional known footer size hint to avoid an extra seek.
+  static Result<std::unique_ptr<PuffinReader>> Make(
+      std::unique_ptr<InputFile> input_file,
+      std::optional<int64_t> footer_size = std::nullopt);
+
+  ~PuffinReader();
```
wgtmac, line 53: Java PuffinReader.close() closes the input stream. This reader owns a SeekableInputStream whose interface has an explicit Close() contract, but PuffinReader exposes no Close() and its destructor is defaulted, so readers can leak or delay releasing file handles/object-store resources.

wgtmac:
I'll directly push some changes to resolve my comments from previous round of review to save some time.

wgtmac (approved):

wgtmac:
Thank you for working on this, @zhaoxuan1994!

PR #625, opened 2026-04-24
fix(ci): restore auth for docs gh-pages deploy

wgtmac:


wgtmac:
@kevinjqliu I'm not sure if this is the right approach to fix https://github.com/apache/iceberg-cpp/actions/runs/24883794037/job/72858619932. I can only verify it by merging this PR.

.github/workflows/docs.yml
```diff
@@ -63,9 +63,12 @@ jobs:
 
       - name: Deploy to gh-pages
         if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
kevinjqliu:
```suggestion
          GITHUB_TOKEN: ${{ github.token }}
```

kevinjqliu (approved):
great catch! this should work, similar to https://github.com/apache/iceberg/blob/main/.github/workflows/site-ci.yml

one nit: we can make the permissions more granular by moving `contents: write` to the job level, like so https://github.com/apache/iceberg/blob/f475ccb5c579b8a1d4842507b12e854161dea7af/.github/workflows/site-ci.yml#L30-L37

wgtmac:
Thanks @kevinjqliu! I will explore that later. Let me fix this the breaking site first.

PR #633, opened 2026-04-29
refactor(util): make retry policy explicit and deterministic

wgtmac:


src/iceberg/test/retry_util_test.cc
```diff
@@ -140,6 +201,32 @@ TEST(RetryRunnerTest, OnlyRetryOnMatchingError) {
...
+TEST(RetryRunnerTest, OnlyRetryOnTakesPrecedenceOverStopRetryOn) {
+  int call_count = 0;
+  int32_t attempts = 0;
+
+  auto result = RetryRunner(RetryConfig{.num_retries = 2,
+                                        .min_wait_ms = 1,
+                                        .max_wait_ms = 10,
+                                        .total_timeout_ms = 5000})
+                    .OnlyRetryOn(ErrorKind::kCommitFailed)
+                    .StopRetryOn({ErrorKind::kCommitFailed})
```
zhjwpku: I see `StopRetryOn` doesn't have a single error type signature like `OnlyRetryOn`, is that by design?

zhjwpku (approved):
LGTM

src/iceberg/util/retry_util.h
```diff
@@ -121,89 +162,34 @@ class ICEBERG_EXPORT RetryRunner {
   }
 
  private:
+  enum class RetryPolicyMode {
+    kUnset,
+    kOnlyRetryOn,
+    kStopRetryOn,
+  };
```
kamcheungting-db, line 172: could you add one line comments for these RetryPolicyModes' behavior?

src/iceberg/util/retry_util.h
kamcheungting-db, line 160: should we record the result detail for each failed run?

src/iceberg/util/retry_util.h
```diff
@@ -121,89 +162,34 @@ class ICEBERG_EXPORT RetryRunner {
...
-    if (config_.total_timeout_ms <= 0) {
-      return std::nullopt;
-    }
-    return Clock::now() + Duration(config_.total_timeout_ms);
-  }
-
-  bool HasTimedOut(const std::optional<TimePoint>& deadline) const {
-    return deadline.has_value() && Clock::now() >= *deadline;
-  }
+  Status ValidateConfig() const;
```
kamcheungting-db, line 179: please add simple comments for this method. Its implementation is not trivial

src/iceberg/util/retry_util.cc
```diff
@@ -0,0 +1,186 @@
...
+  ActiveRetryTestHooks() = hooks;
+}
+
+Status RetryRunner::ValidateConfig() const {
+  if (config_.num_retries < 0) {
+    return InvalidArgument("num_retries must be non-negative, got {}",
+                           config_.num_retries);
+  }
+  if (config_.num_retries == 0) {
+    return {};
```
kamcheungting-db, line 88: nitpick: could we explicitly call out?
```suggestion
    return RetryExhausted(<".... list of failed run reason">);
```

src/iceberg/util/retry_util_internal.h
```diff
@@ -0,0 +1,61 @@
...
+#include <functional>
+
+#include "iceberg/iceberg_export.h"
+
+namespace iceberg {
+
+struct RetryTestHooks {
+  using Clock = std::chrono::steady_clock;
+  using Duration = std::chrono::milliseconds;
+  using TimePoint = Clock::time_point;
```
kamcheungting-db, line 33: these are duplicated with these [few line code](https://github.com/apache/iceberg-cpp/pull/633/changes#diff-e67d5eeb935f4299db9ad0d2c4115fff174f89f73c4f41d740335338293c3d7bL124-L126)

Have we considered about deduplicating them?

kamcheungting-db (commented):
Nice factor.
I have some non-blocker comments for understanding the underlying intention.

src/iceberg/test/retry_util_test.cc
wgtmac: Added `StopRetryOn(ErrorKind)` and switched this precedence test to use the single-kind overload.

src/iceberg/util/retry_util.h
wgtmac: Added comments for each `RetryPolicyMode` value to spell out unset, only-retry-on, and stop-retry-on behavior.

src/iceberg/util/retry_util.h
wgtmac: Added a declaration comment summarizing that `ValidateConfig` checks retry counts, timing bounds, and the selected retry policy.

src/iceberg/util/retry_util.h
wgtmac: Keeping this as-is for now. `RetryRunner` currently preserves the existing contract of returning the error from the attempt that caused retrying to stop. Recording every failed attempt would require separate error aggregation or metrics reporting; the existing TODO on `Run` points toward using a metrics reporter once available.

src/iceberg/util/retry_util.cc
wgtmac: Keeping the current return behavior here. There is no `RetryExhausted` wrapper today; retry exhaustion returns the last task error so callers keep the original `ErrorKind` and message. Collecting all failed-run reasons would be a separate error aggregation or metrics change.

src/iceberg/util/retry_util_internal.h
wgtmac: I considered this and kept them separate. `RetryTestHooks` lives in the internal test hook header, while `RetryRunner` keeps its aliases private in the public retry header. Deduplicating these three simple aliases would either expose internals or add a public dependency on test hooks.

kamcheungting-db (approved):
LGTM

zhjwpku (approved):

blackmwk (approved):

wgtmac:
Thanks all for the review!

PR #635, opened 2026-05-02
fix: repair shared package linkage and export scan builder symbols

zhjwpku:
1. Update the shared installation interface to depend on the static nanoarrow target, since vendored libraries are built only as static artifacts(see prepare_fetchcontent).

2. Switch the example build to use shared libraries to surface linkage issues. This revealed that `TableScanBuilder` symbols are not exported. Explicitly export the `TableScanBuilder` template instantiations so that shared-library consumers can successfully link against `Build()`.


wgtmac (approved):

example/CMakeLists.txt
```diff
@@ -26,5 +26,5 @@ find_package(iceberg CONFIG REQUIRED)
 
 add_executable(demo_example demo_example.cc)
 
-target_link_libraries(demo_example PRIVATE iceberg::iceberg_bundle_static
-                                           iceberg::iceberg_rest_static)
+target_link_libraries(demo_example PRIVATE iceberg::iceberg_bundle_shared
```
wgtmac, line 29: The example now unconditionally links `iceberg::iceberg_bundle_shared` and `iceberg::iceberg_rest_shared`, but the project default is ICEBERG_BUILD_SHARED=OFF and ICEBERG_BUILD_STATIC=ON. A default static-only install exports only the static targets, so configuring the example against that package fails at CMake generate with missing iceberg::iceberg_bundle_shared.

zhjwpku: I hadn't considered that. Should we set `ICEBERG_BUILD_SHARED=ON` and `ICEBERG_BUILD_STATIC=OFF`, or set both `ON`?

PR #641, opened 2026-05-06
feat(io): add streaming FileIO support

wgtmac:
Introduce InputFile/OutputFile stream APIs and Arrow IO adapters so bundled Avro/Parquet readers and writers can work with generic FileIO implementations.


src/iceberg/arrow/arrow_io.cc
```diff
@@ -0,0 +1,617 @@
...
+  }
+
+  ::arrow::Result<std::shared_ptr<::arrow::Buffer>> Read(int64_t nbytes) override {
+    if (nbytes < 0) {
+      return ::arrow::Status::Invalid("Cannot read a negative number of bytes");
+    }
+    {
+      std::lock_guard lock(mutex_);
+      ARROW_RETURN_NOT_OK(CheckOpenLocked());
+    }
```
WZhuo: It checked in the inner `Read` function, no need to check here.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -0,0 +1,617 @@
...
+  ::arrow::Result<std::shared_ptr<::arrow::Buffer>> ReadAt(int64_t position,
+                                                           int64_t nbytes) override {
+    if (position < 0 || nbytes < 0) {
+      return ::arrow::Status::Invalid("ReadAt position and length must be non-negative");
+    }
+    {
+      std::lock_guard lock(mutex_);
+      ARROW_RETURN_NOT_OK(CheckOpenLocked());
+    }
+    auto bytes_to_read = position >= size_ ? 0 : std::min(nbytes, size_ - position);
```
WZhuo: Should it return error if `position > size_`?

src/iceberg/arrow/arrow_io.cc
```diff
@@ -0,0 +1,617 @@
...
+  mutable std::mutex mutex_;
+};
+
+class ArrowSeekableInputStream : public SeekableInputStream {
+ public:
+  explicit ArrowSeekableInputStream(std::shared_ptr<::arrow::io::RandomAccessFile> input)
+      : input_(std::move(input)) {}
+
+  Result<int64_t> Position() const override {
+    ICEBERG_RETURN_UNEXPECTED(CheckOpen());
```
WZhuo: We do the open check in `InputStream/OutputStream`, I think the open check in `InputStreamAdapter/OutputStreamAdapter` can be removed and call the inner stream directly.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -0,0 +1,617 @@
...
+
+  int64_t size;
+  std::unique_ptr<InputFile> input_file;
+  if (length.has_value()) {
+    ICEBERG_ASSIGN_OR_RAISE(size, ToInt64Length(*length));
+    ICEBERG_ASSIGN_OR_RAISE(input_file, io->NewInputFile(path, *length));
+  } else {
+    ICEBERG_ASSIGN_OR_RAISE(input_file, io->NewInputFile(path));
+    ICEBERG_ASSIGN_OR_RAISE(size, input_file->Size());
+  }
```
WZhuo, line 505:
```suggestion
  if (length.has_value()) {
    ICEBERG_ASSIGN_OR_RAISE(input_file, io->NewInputFile(path, *length));
  } else {
    ICEBERG_ASSIGN_OR_RAISE(input_file, io->NewInputFile(path));
  }
  ICEBERG_ASSIGN_OR_RAISE(size, input_file->Size());
```

src/iceberg/file_io.h
```diff
@@ -50,21 +138,14 @@ class ICEBERG_EXPORT FileIO {
   /// \return The content of the file if the read succeeded, an error code if the read
   /// failed.
   virtual Result<std::string> ReadFile(const std::string& file_location,
-                                       std::optional<size_t> length) {
-    // We provide a default implementation to avoid Windows linker error LNK2019.
-    return NotImplemented("ReadFile not implemented");
-  }
+                                       std::optional<size_t> length);
```
zhjwpku, line 141: Can we just use int64_t for the length? Not a strong opinion, just bring this on table.

src/iceberg/test/std_io.h
```diff
@@ -19,78 +19,297 @@
...
+      return IOError("Failed to write to file: {}", location_);
     }
+    return {};
   }
 
-  Status WriteFile(const std::string& file_location, std::string_view content) override {
-    // Create parent directories if they don't exist
-    std::filesystem::path path(file_location);
+  Status Flush() override {
+    file_.flush();
```
zhjwpku, line 186: Do we need is_open() here?

zhjwpku (approved):
Generally LGTM except some trivial comments.

blackmwk (approved):

wgtmac:
Thanks @WZhuo @zhjwpku @blackmwk for the review! I'll address comments in a separate PR.

src/iceberg/test/std_io.h
wgtmac: Yes, it's needed. `close()` on an already-closed stream sets `failbit`, which would cause the `if (!file_)` check below to return a spurious error.

src/iceberg/file_io.h
wgtmac: That would need to check negative values.

PR #642, opened 2026-05-06
ci: split S3/MinIO into a Linux and macOS-only workflow

### LINKED ISSUE #640: [DISCUSSION] Docker MinIO image not available for Windows platform
We encountered a CI failure on `main` and in a recent PR [1]:

```
docker run -d --name iceberg-minio \
  -p 9000:9000 -p 9001:9001 \
  -e MINIO_ROOT_USER=minio \
  -e MINIO_ROOT_PASSWORD=minio123 \
  minio/minio:latest server /data --console-address :9001

Unable to find image 'minio/minio:latest' locally
latest: Pulling from minio/minio
docker: no matching manifest for windows(10.0.26100)/amd64 in the manifest list entries
```

It seems that the `minio/minio:latest` image no longer provide a Windows-compatible manifest, which causes failures on Windows runners.

There has been some discussion in the Apache Iceberg project about replacing MinIO with RustFS due to maintenance and security concerns [2]. However, even if we adopt RustFS, we would still face the same limitation, as it also does not provide Windows container images.

One suggestion (from @wgtmac) is to decouple S3-related testing from the Windows workflow: run S3-backed tests (e.g., against AWS or a compatible service) in a dedicated CI workflow, and disable that workflow on Windows. This would avoid relying on Windows-compatible Docker images for S3 emulation.

As a short-term approach, I propose that we continue using MinIO for Linux-based CI, while excluding S3-related tests from Windows jobs. In parallel, we can evaluate alternative S3-compatible solutions or a potential migration strategy in the future.

Thoughts?

[1] https://github.com/apache/iceberg-cpp/actions/runs/25410582555/job/74535575933?pr=639
[2] https://github.com/apache/iceberg/issues/14638

zhjwpku:
Closes #640

### DIFF
--- a/.github/workflows/s3_test.yml
+++ b/.github/workflows/s3_test.yml
@@ -1,0 +1,81 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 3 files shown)


wgtmac (approved):

wgtmac:
Currently the new split workflows still do the same thing. In the future we may use a dedicated script to build and run components relevant to AWS only.

PR #645, opened 2026-05-08
feat(data): coalesce position deletes into range inserts

Baunsgaard:
Add ForEachPositionDelete (the C++ equivalent of Java's PositionDeleteRangeConsumer) and route DeleteLoader through it, replacing the per-position PositionDeleteIndex::Delete(pos) call. The function sniffs a 1024-position prefix and dispatches to either run coalescing (CRoaring addRange) or bulk addMany grouped by high-32-bit key.

Also rework DeleteLoader::LoadPositionDelete to read Arrow batches via nanoarrow's ArrowArrayView directly. When the delete file's referenced_data_file matches the target (V2 writer hint), positions are passed as a zero-copy span; otherwise a per-batch staging vector filters by path.

Local microbenchmarks: 2.2x-10.6x for ForEachPositionDelete and 2.1x-2.5x end-to-end through LoadPositionDeletes. Equivalent of apache/iceberg#16052.

### DIFF
--- a/src/iceberg/deletes/position_delete_range_consumer.cc
+++ b/src/iceberg/deletes/position_delete_range_consumer.cc
@@ -1,0 +1,147 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/deletes/position_delete_range_consumer.h"
+
+#include <algorithm>
+#include <cstdint>
+#include <span>
+#include <vector>
+
+#include "iceberg/deletes/position_delete_index.h"
+#include "iceberg/deletes/roaring_position_bitmap.h"
+
+namespace iceberg {
+
+namespace {
+
+bool IsValidPosition(int64_t pos) {
+  return pos >= 0 && pos <= RoaringPositionBitmap::kMaxPosition;
+}
+
+// Unsigned subtraction so negative or wrap-around input can't
+// false-positive via signed overflow.
+bool IsAdjacent(int64_t prev, int64_t next) {
+  return (static_cast<uint64_t>(next) - static_cast<uint64_t>(prev)) == 1;
+}
+
+// `RoaringPositionBitmap` shards positions by their high 32 bits; the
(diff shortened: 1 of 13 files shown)


src/iceberg/deletes/roaring_position_bitmap.h
```diff
@@ -110,6 +112,12 @@ class ICEBERG_DATA_EXPORT RoaringPositionBitmap {
   std::unique_ptr<Impl> impl_;
 
   explicit RoaringPositionBitmap(std::unique_ptr<Impl> impl);
+
+  // Bulk-add positions sharing high-32-bit `key`. Internal hook for
+  // `PositionDeleteIndex::BulkAddForKey`; per-key grouping is the caller's
+  // job, keeping this a thin wrapper around CRoaring's `addMany`.
+  void AddManyForKey(int32_t key, const uint32_t* positions, size_t n);
```
wgtmac:
```suggestion
  void AddManyForKey(int32_t key, std::span<const uint32_t> positions);
```

I'd prefer `span` here

src/iceberg/deletes/position_delete_range_consumer.cc
```diff
@@ -0,0 +1,147 @@
...
+    boundaries += static_cast<size_t>(!IsAdjacent(positions[i - 1], positions[i]));
+  }
+
+  // boundaries / (sniff - 1) > kBulkThresholdPercent / 100, without FP.
+  if (boundaries * 100 > (sniff - 1) * kBulkThresholdPercent) {
+    // Bulk path: group by high-32-bit key, flush each group via CRoaring's
+    // `addMany` (through `BulkAddForKey`). The thread-local buffer is
+    // reused across calls; nested invocations on the same thread would
+    // corrupt it -- see `\warning` on `ForEachPositionDelete`.
+    thread_local std::vector<uint32_t> bulk_key_positions;
```
wgtmac: This looks hacky and limits its use. How about passing `bulk_key_positions` as a parameter so that caller can take full control over it?

src/iceberg/deletes/roaring_position_bitmap.h
Baunsgaard: Okay, i changed it to use std:span

src/iceberg/deletes/position_delete_range_consumer.cc
Baunsgaard: Sure, i changed it to do that as well, It is also more in line with what I did in Java.

Baunsgaard:
Thanks for the Look @wgtmac, I have addressed your comments!

It extended the implementation quite a bit across more files, I am not sure i like it, but let me know if you would prefer something less invasive, and if you find something else!

src/iceberg/meson.build
```diff
@@ -152,6 +152,7 @@ iceberg_data_sources = files(
     'data/position_delete_writer.cc',
     'data/writer.cc',
     'deletes/position_delete_index.cc',
+    'deletes/position_delete_range_consumer.cc',
```
wgtmac, line 155: We need to install `deletes/position_delete_range_consumer.h` in the meson.build.

wgtmac (approved):
Nice improvement! Thanks @Baunsgaard!

src/iceberg/meson.build
wgtmac: Fixed in https://github.com/apache/iceberg-cpp/pull/684

Baunsgaard: Thanks, I will keep this in mind in the future!

PR #646, opened 2026-05-09
feat(rest): implement OAuth2 token auto-refresh for REST catalog

lishuxu:
Replace the MakeOAuth2 stub with a full OAuth2AuthSession that automatically refreshes tokens before expiration using the client_credentials grant.

### DIFF
--- a/src/iceberg/catalog/rest/auth/auth_manager.cc
+++ b/src/iceberg/catalog/rest/auth/auth_manager.cc
@@ -130,7 +130,8 @@ class OAuth2Manager : public AuthManager {
       init_token_response_.reset();
       return AuthSession::MakeOAuth2(token_response, config.oauth2_server_uri(),
                                      config.client_id(), config.client_secret(),
-                                     config.scope(), client);
+                                     config.scope(), config.keep_refreshed(),
+                                     config.optional_oauth_params(), client);
     }
 
     // If token is provided, use it directly.
@@ -143,10 +144,10 @@ class OAuth2Manager : public AuthManager {
       auto base_session = AuthSession::MakeDefault(AuthHeaders(config.token()));
       OAuthTokenResponse token_response;
       ICEBERG_ASSIGN_OR_RAISE(token_response, FetchToken(client, *base_session, config));
-      // TODO(lishuxu): should we directly pass config to the MakeOAuth2 call?
       return AuthSession::MakeOAuth2(token_response, config.oauth2_server_uri(),
                                      config.client_id(), config.client_secret(),
-                                     config.scope(), client);
+                                     config.scope(), config.keep_refreshed(),
+                                     config.optional_oauth_params(), client);
     }
 
     return AuthSession::MakeDefault({});
--- a/src/iceberg/catalog/rest/auth/auth_manager.cc
+++ b/src/iceberg/catalog/rest/auth/auth_manager.cc
@@ -25,8 +25,8 @@
 #include "iceberg/catalog/rest/auth/auth_properties.h"
 #include "iceberg/catalog/rest/auth/auth_session.h"
 #include "iceberg/catalog/rest/auth/oauth2_util.h"
+#include "iceberg/util/base64.h"
 #include "iceberg/util/macros.h"
-#include "iceberg/util/transform_util.h"
 
 namespace iceberg::rest::auth {
 
@@ -83,8 +83,7 @@ class BasicAuthManager : public AuthManager {
                      "Missing required property '{}'", AuthProperties::kBasicPassword);
     std::string credential = username_it->second + ":" + password_it->second;
     return AuthSession::MakeDefault(
-        {{std::string(kAuthorizationHeader),
-          "Basic " + TransformUtil::Base64Encode(credential)}});
+        {{std::string(kAuthorizationHeader), "Basic " + Base64::Encode(credential)}});
   }
 };
 
--- a/src/iceberg/catalog/rest/auth/auth_session.cc
+++ b/src/iceberg/catalog/rest/auth/auth_session.cc
@@ -19,9 +19,17 @@
 
 #include "iceberg/catalog/rest/auth/auth_session.h"
 
+#include <algorithm>
+#include <chrono>
+#include <memory>
+#include <shared_mutex>
+#include <thread>
 #include <utility>
 
+#include "iceberg/catalog/rest/auth/auth_properties.h"
 #include "iceberg/catalog/rest/auth/oauth2_util.h"
+#include "iceberg/catalog/rest/auth/token_refresh_scheduler.h"
+#include "iceberg/catalog/rest/http_client.h"
 
 namespace iceberg::rest::auth {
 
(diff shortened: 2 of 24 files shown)


src/iceberg/test/meson.build
```diff
@@ -136,23 +135,23 @@ if get_option('rest').enabled()
                     ),
                     'dependencies': [iceberg_rest_dep],
                 },
+                'polaris_oauth2_integration_test': {
+                    'sources': files('polaris_oauth2_test.cc'),
```
zhjwpku: It seems you may have missed adding this file to the commit.

src/iceberg/util/transform_util.cc
```diff
@@ -283,4 +283,76 @@ std::string TransformUtil::Base64Encode(std::string_view str_to_encode) {
   return encoded;
 }
 
+namespace {
+
+// Shared base64 decode logic. The decode table maps ASCII char → 6-bit value.
```
zhjwpku: nit
```suggestion
// Shared base64 decode logic. The decode table maps ASCII char -> 6-bit value.
```

src/iceberg/util/transform_util.cc
```diff
@@ -283,4 +283,76 @@ std::string TransformUtil::Base64Encode(std::string_view str_to_encode) {
...
+  std::string output;
+  output.reserve((input.size() * 3) / 4);
+
+  uint32_t buffer = 0;
+  int bits_collected = 0;
+
+  for (char c : input) {
+    uint8_t val = table[static_cast<uint8_t>(c)];
+    if (val == 0xFF) {
+      return {};  // Invalid character
```
zhjwpku: Should we return Result<std::string> so that we know input is just empty or invalid?

src/iceberg/test/meson.build
lishuxu: Good catch. This integration test (polaris_oauth2_integration_test) was included by mistake — it's not ready to be part of this PR.

src/iceberg/util/transform_util.cc
lishuxu: Good suggestion. Updated to return Result<std::string> — empty input now returns Ok("") and invalid characters return InvalidArgument. This makes the semantics unambiguous for callers.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -44,6 +51,183 @@ class DefaultAuthSession : public AuthSession {
...
+    auto session = std::shared_ptr<OAuth2AuthSession>(
+        new OAuth2AuthSession(std::move(config), client));
+    session->SetInitialToken(initial_token);
+    return session;
+  }
+
+  Status Authenticate(std::unordered_map<std::string, std::string>& headers) override {
+    std::shared_lock lock(mutex_);
+    for (const auto& [key, value] : headers_) {
+      headers.insert_or_assign(key, value);
```
wgtmac: Use `try_emplace` here so a request-specific `Authorization` header can win, matching `DefaultAuthSession` and Java’s `putIfAbsent` behavior.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -44,6 +51,183 @@ class DefaultAuthSession : public AuthSession {
...
+    // Build credential and properties once (invariant across retries)
+    std::string credential = config_.client_id.empty()
+                                 ? config_.client_secret
+                                 : config_.client_id + ":" + config_.client_secret;
+
+    // Use an empty session for the refresh request (no auth headers —
+    // avoids circular dependency of using an expired token to refresh itself)
+    auto empty_session = AuthSession::MakeDefault({});
+
+    AuthProperties props;
```
wgtmac: Carry `audience` and `resource` through to refresh requests too. Some OAuth servers require the same optional params on every token request.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -44,6 +51,183 @@ class DefaultAuthSession : public AuthSession {
...
+    if (result.has_value()) {
+      auto& response = result.value();
+      {
+        std::unique_lock lock(mutex_);
+        token_ = response.access_token;
+        headers_ = {
+            {std::string(kAuthorizationHeader), std::string(kBearerPrefix) + token_}};
+
+        // Update expiration
+        if (response.expires_in_secs.has_value()) {
```
wgtmac, line 194: Reset `expires_at_` before deriving the new expiry, and skip scheduling when the refreshed token has neither `expires_in` nor JWT `exp`.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -52,12 +236,17 @@ std::shared_ptr<AuthSession> AuthSession::MakeDefault(
...
-                       std::string(kBearerPrefix) + initial_token.access_token}});
+    const OAuthTokenResponse& initial_token, const std::string& token_endpoint,
+    const std::string& client_id, const std::string& client_secret,
+    const std::string& scope, HttpClient& client) {
+  OAuth2AuthSession::Config config{
+      .token_endpoint = token_endpoint,
+      .client_id = client_id,
+      .client_secret = client_secret,
+      .scope = scope,
+      .keep_refreshed = true,
```
wgtmac: Pass the parsed `token-refresh-enabled` value into this config instead of forcing refresh on. Users who disable refresh should not get background token requests.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -44,6 +51,183 @@ class DefaultAuthSession : public AuthSession {
...
+  struct Config {
+    std::string token_endpoint;
+    std::string client_id;
+    std::string client_secret;
+    std::string scope;
+    bool keep_refreshed;
+  };
+
+  /// \brief Create an OAuth2 session and optionally schedule refresh.
+  static std::shared_ptr<OAuth2AuthSession> Create(
```
wgtmac:
```suggestion
  static std::shared_ptr<OAuth2AuthSession> Make(
```

nit: rename it to Make to be consistent with this repo.

src/iceberg/catalog/rest/auth/auth_session.cc
lishuxu: Fixed. Changed to try_emplace to match DefaultAuthSession and Java's putIfAbsent semantics.

src/iceberg/catalog/rest/auth/auth_session.cc
lishuxu: Fixed. Now resetting expires_at_ to epoch before computing the new value.

src/iceberg/catalog/rest/auth/auth_session.cc
lishuxu: Good catch. Added optional_oauth_params to the session config.

src/iceberg/catalog/rest/auth/auth_session.cc
lishuxu: Fixed. Added keep_refreshed parameter to MakeOAuth2, passed from config.keep_refreshed().

src/iceberg/catalog/rest/auth/token_refresh_scheduler.h
```diff
@@ -0,0 +1,109 @@
...
+namespace iceberg::rest::auth {
+
+/// \brief A process-global scheduler for delayed token refresh tasks.
+///
+/// Uses a single background thread that sleeps until the next task is due.
+/// All OAuth2AuthSession instances share this scheduler. Tasks are lightweight
+/// (a single HTTP POST to refresh a token), so one thread is sufficient.
+///
+/// Thread safety: All public methods are thread-safe.
+class ICEBERG_REST_EXPORT TokenRefreshScheduler {
```
wgtmac, line 47: @HuaHuaY is working on the adding a thread pool abstraction and it would be good to reuse that once available. The main issue of current design is that only a single thread is working on token refresh and a slow request would starve all other tasks.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -44,6 +51,189 @@ class DefaultAuthSession : public AuthSession {
...
+
+    AuthProperties props;
+    props.Set(AuthProperties::kCredential, credential);
+    props.Set(AuthProperties::kScope, config_.scope);
+    props.Set(AuthProperties::kOAuth2ServerUri, config_.token_endpoint);
+    for (const auto& [key, value] : config_.optional_oauth_params) {
+      props.mutable_configs().insert_or_assign(key, value);
+    }
+
+    auto result = FetchToken(client_, *empty_session, props);
```
wgtmac: Java refreshes through token exchange by default and only falls back to client credentials in specific cases. This path always uses `client_credentials` and ignores `token-exchange-enabled`, which breaks parity for exchange-based refresh.

src/iceberg/catalog/rest/auth/auth_manager.cc
```diff
@@ -130,7 +130,8 @@ class OAuth2Manager : public AuthManager {
       init_token_response_.reset();
       return AuthSession::MakeOAuth2(token_response, config.oauth2_server_uri(),
                                      config.client_id(), config.client_secret(),
-                                     config.scope(), client);
+                                     config.scope(), config.keep_refreshed(),
+                                     config.optional_oauth_params(), client);
     }
 
     // If token is provided, use it directly.
```
wgtmac, line 136: Java wraps configured access tokens in `fromAccessToken()` so expiring tokens can be refreshed. Returning a static session here means JWT or `token-expires-in-ms` based tokens never refresh.

src/iceberg/util/transform_util.h
```diff
@@ -139,6 +140,19 @@ class ICEBERG_EXPORT TransformUtil {
 
   /// \brief Base64 encode a string
   static std::string Base64Encode(std::string_view str_to_encode);
+
+  /// \brief Base64 decode a string (standard alphabet: +/).
+  ///
+  /// Handles optional padding ('=').
+  /// \return Decoded string, or an error if the input contains invalid characters.
+  static Result<std::string> Base64Decode(std::string_view encoded);
```
wgtmac: nit: we can refactor a separate `iceberg/util/base64.h` for these `Base64XXX` functions.

src/iceberg/util/transform_util.h
```diff
@@ -139,6 +140,19 @@ class ICEBERG_EXPORT TransformUtil {
...
+
+  /// \brief Base64url decode a string (URL-safe alphabet: -_).
+  ///
+  /// Handles optional padding ('='). This variant uses '-' and '_' instead of
+  /// '+' and '/' per RFC 4648 §5.
+  /// \return Decoded string, or an error if the input contains invalid characters.
+  static Result<std::string> Base64UrlDecode(std::string_view encoded);
```
wgtmac: Do we need to add a `Base64UrlEncode` for parity?

src/iceberg/catalog/rest/auth/token_refresh_scheduler.h
lishuxu: Makes sense. Added a TODO referencing this discussion to migrate to the shared thread pool once it lands. Agreed the single-thread design can let a slow request starve other refreshes; will switch over in a follow-up.

src/iceberg/catalog/rest/auth/auth_session.cc
lishuxu: Good catch. This path always uses client_credentials and ignores token-exchange-enabled, unlike Java which refreshes via token exchange by default. I'll address this in a dedicated follow-up PR (exchange request builder + refresh flow rework), and link it here. Keeping this PR scoped to client_credentials auto-refresh.

src/iceberg/catalog/rest/auth/auth_manager.cc
lishuxu: Agreed. This is tightly coupled to the token-exchange work (a configured token has no credential, so Java refreshes it via exchange), so I'll bundle it into the same follow-up PR: add token-expires-in-ms, populate expires_at_millis_, and wrap configured tokens in a refreshable session. Will link the PR here.

src/iceberg/util/transform_util.h
lishuxu: Done. Moved the Base64* functions into a new base64.h/base64.cc (Base64 class), and migrated the tests to base64_test.cc.

src/iceberg/util/transform_util.h
lishuxu: Added Base64::UrlEncode for parity. It also replaces the hand-rolled base64url encoder in auth_manager_test.cc.

wgtmac (approved):
I've rebased with some minor fixes. Thanks @lishuxu for adding this!

PR #648, opened 2026-05-10
feat: add IncrementalFileCleanup strategy and dispatch in ExpireSnapshots::Finalize

shangxinli:
Mirrors Java's IncrementalFileCleanup for the linear-ancestry case: each manifest is attributed to its writer snapshot, so two passes are enough instead of the full reachability scan. Cherry-pick protection via SnapshotSummaryFields::kSourceSnapshotId is preserved.

Finalize() now picks IncrementalFileCleanup when the expiration is "simple" (no explicit snapshot IDs, no removed snapshots outside the current main ancestry, and no retained snapshots outside the current main ancestry), and falls back to ReachableFileCleanup otherwise. The dispatch matches Java RemoveSnapshots.cleanExpiredSnapshots.

Two existing cleanup tests (DeletesExpiredFiles, IgnoresExpiredDeleteManifestReadFailures) used an empty current manifest list, which is an unreachable-orphan scenario that only ReachableFileCleanup can resolve. They now call ExpireSnapshotId() to force the reachable path, which keeps their original intent and matches Java behavior. New tests cover both dispatch branches.


src/iceberg/update/expire_snapshots.cc
```diff
@@ -331,6 +331,303 @@ class ReachableFileCleanup : public FileCleanupStrategy {
...
+    std::unordered_set<int64_t> picked_ancestor_snapshot_ids;
+    for (const auto& ancestor : ancestors_result.value()) {
+      if (!ancestor) continue;
+      const auto& summary = ancestor->summary;
+      auto it = summary.find(SnapshotSummaryFields::kSourceSnapshotId);
+      if (it == summary.end()) continue;
+      try {
+        picked_ancestor_snapshot_ids.insert(std::stoll(it->second));
+      } catch (...) {
+        // Malformed source-snapshot-id; skip rather than fail cleanup.
```
wgtmac: Java fails cleanup when `source-snapshot-id` is malformed. Please fail closed here (return/propagate an error) instead of skipping, so cherry-pick protection cannot be bypassed.

shangxinli: Good catch — fixed in 7184725. Both parse sites now return `InvalidArgument` (matching Java propagating `NumberFormatException`), so a malformed `source-snapshot-id` aborts cleanup instead of silently bypassing cherry-pick protection.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -331,6 +332,309 @@ class ReachableFileCleanup : public FileCleanupStrategy {
...
+    // Iterate the ancestor pointers we already have rather than re-looking-up
+    // each snapshot by id.
+    std::unordered_set<int64_t> picked_ancestor_snapshot_ids;
+    for (const auto& ancestor : ancestors_result.value()) {
+      if (!ancestor) continue;
+      const auto& summary = ancestor->summary;
+      auto it = summary.find(SnapshotSummaryFields::kSourceSnapshotId);
+      if (it == summary.end()) continue;
+      try {
+        picked_ancestor_snapshot_ids.insert(std::stoll(it->second));
```
wgtmac: Please use the repo integer parser instead of `std::stoll` (e.g. `StringUtils::ParseNumber<int64_t>`, with `+` handling if needed). It returns `Result` and rejects trailing junk like Java.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -613,11 +917,23 @@ Status ExpireSnapshots::Finalize(Result<const TableMetadata*> commit_result) {
...
+                                                         metadata_after_expiration) &&
+                             !HasNonMainSnapshots(metadata_after_expiration);
+
+  std::unique_ptr<FileCleanupStrategy> strategy;
+  if (can_use_incremental) {
+    strategy = std::make_unique<IncrementalFileCleanup>(ctx_->table->io(), delete_func_);
+  } else {
+    strategy = std::make_unique<ReachableFileCleanup>(ctx_->table->io(), delete_func_);
+  }
+  return strategy->CleanFiles(metadata_before_expiration, metadata_after_expiration,
```
wgtmac: This status is ignored by the normal `Commit()` paths today. Please add a commit-path test and propagate parsing/planning errors, otherwise malformed `source-snapshot-id` still reports success.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -331,6 +332,309 @@ class ReachableFileCleanup : public FileCleanupStrategy {
   }
 };
 
+/// \brief Incremental file cleanup strategy for simple linear-ancestry expirations.
+///
+/// Mirrors Java's IncrementalFileCleanup. Only safe when:
```
wgtmac: Please avoid Java-specific wording in production comments. Keep this to the invariant and dispatch conditions, and drop the “Mirrors Java” phrasing here and below.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -613,11 +917,23 @@ Status ExpireSnapshots::Finalize(Result<const TableMetadata*> commit_result) {
...
+  // Pick incremental cleanup when the expiration is a simple linear-ancestry walk:
+  // no explicit snapshot IDs, no removed snapshots outside main ancestry, and no
+  // retained snapshots outside main ancestry. Mirrors Java RemoveSnapshots's
+  // dispatch in cleanExpiredSnapshots().
+  bool can_use_incremental = !specified_snapshot_id_ &&
+                             !HasRemovedNonMainAncestors(metadata_before_expiration,
+                                                         metadata_after_expiration) &&
+                             !HasNonMainSnapshots(metadata_after_expiration);
+
+  std::unique_ptr<FileCleanupStrategy> strategy;
```
wgtmac: This does not need heap allocation or a virtual dispatch object. Please instantiate the selected strategy in each branch and return `CleanFiles(...)` directly.

src/iceberg/test/expire_snapshots_test.cc
```diff
@@ -573,4 +580,75 @@ TEST_F(ExpireSnapshotsCleanupTest, KeepsReusedPartitionStats) {
   EXPECT_THAT(deleted_files, testing::Not(testing::Contains(reused_statistics_path)));
 }
 
+// Linear-ancestry, no specified ID: dispatch must pick IncrementalFileCleanup.
```
wgtmac: This comment is doing too much and calls out Java again. Please trim it to the setup expectation: no explicit ID should take the incremental path and preserve the added data file.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -331,6 +332,309 @@ class ReachableFileCleanup : public FileCleanupStrategy {
...
+/// current state under a different id.
+///
+/// TODO(shangxinli): Add multi-threaded manifest reading and file deletion support.
+class IncrementalFileCleanup : public FileCleanupStrategy {
+ public:
+  using FileCleanupStrategy::FileCleanupStrategy;
+
+  Status CleanFiles(const TableMetadata& metadata_before_expiration,
+                    const TableMetadata& metadata_after_expiration,
+                    const std::unordered_set<int64_t>& expired_snapshot_ids,
```
wgtmac: Java derives expired IDs from `beforeExpiration` and `afterExpiration` inside cleanup. Please do the same here instead of passing a separate set that can drift from the two metadata states.

src/iceberg/update/expire_snapshots.cc
shangxinli: Fixed in b57207a — both parse sites now propagate via ICEBERG_ASSIGN_OR_RAISE.

src/iceberg/update/expire_snapshots.cc
shangxinli: Done in b57207a — both strategies now derive expired IDs from before/after metadata.

src/iceberg/test/expire_snapshots_test.cc
shangxinli: Trimmed in b57207a.

src/iceberg/update/expire_snapshots.cc
shangxinli: Dropped in b57207a.

src/iceberg/update/expire_snapshots.cc
shangxinli: Done in b57207a — PendingUpdate::Commit() now returns Finalize's status. Added CommitPropagatesMalformedSourceSnapshotId.

src/iceberg/update/expire_snapshots.cc
shangxinli: Done in b57207a.

src/iceberg/update/expire_snapshots.cc
shangxinli: Done in b57207a.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -608,16 +917,24 @@ Status ExpireSnapshots::Finalize(Result<const TableMetadata*> commit_result) {
...
+      !HasRemovedNonMainAncestors(metadata_before_expiration,
+                                  metadata_after_expiration) &&
+      !HasNonMainSnapshots(metadata_after_expiration);
+
+  if (can_use_incremental) {
+    return IncrementalFileCleanup(ctx_->table->io(), delete_func_)
+        .CleanFiles(metadata_before_expiration, metadata_after_expiration,
+                    cleanup_level_);
+  }
+  return ReachableFileCleanup(ctx_->table->io(), delete_func_)
```
wgtmac, line 945: Reachable still reads manifests-to-delete with the after metadata. Please use the pre-expiration metadata for that read so `CleanExpiredMetadata(true)` cannot prune the spec before cleanup.

src/iceberg/update/pending_update.cc
```diff
@@ -47,8 +47,7 @@ Status PendingUpdate::Commit() {
       return std::unexpected(commit_result.error());
     }
 
-    std::ignore = Finalize(commit_result.value()->metadata().get());
-    return {};
+    return Finalize(commit_result.value()->metadata().get());
```
wgtmac: Please keep `Finalize()` best-effort here. Commit status should reflect the metadata commit, so cleanup/parsing failures should not be returned from `Commit()`.

src/iceberg/test/expire_snapshots_test.cc
```diff
@@ -350,6 +351,11 @@ TEST_F(ExpireSnapshotsCleanupTest, IgnoresExpiredDeleteManifestReadFailures) {
 
   std::vector<std::string> deleted_files;
   ICEBERG_UNWRAP_OR_FAIL(auto update, table_->NewExpireSnapshots());
+  // Force ReachableFileCleanup: the empty current manifest list makes this an
+  // unreachable-orphan scenario, which Incremental cannot detect (added-in-ancestor
+  // files are preserved unless an explicit DELETED entry exists). Specifying the
+  // snapshot id forces the dispatch to Reachable.
+  update->ExpireSnapshotId(kExpiredSnapshotId);
```
wgtmac, line 397: This now forces the reachable path, but the assertion still omits `expired_delete_file_path`. Please make reachable cleanup delete orphaned delete files under `kAll` and assert it here.

wgtmac (approved):
Thanks for adding this! It is a very complex algorithm that I spent some time to understand it and make sure it aligns with the Java parity. I think now it looks good to me and it still has some followups to make it work as expected.

PR #649, opened 2026-05-10
refactor: harden FileCleanupStrategy with retry and parallel deletes

shangxinli:
Rebased onto #687's executor pool support per @wgtmac's review.

### What

* `FileCleanupStrategy` now takes an `OptionalExecutor` in its constructor. When `DeleteWith()` is configured, per-path deletes fan out through `TaskGroup<>` -- uses the supplied executor when set, otherwise runs the callbacks serially on the calling thread via `RunTasksSingleThreaded`. Preserves prior single-threaded behavior by default.
* The `FileIO::DeleteFiles` bulk path is wrapped in `RetryRunner<retry::StopRetryOn<ErrorKind::kNotFound>>` with a tight budget (2 retries, 100ms-1s backoff, 5s total). Mirrors Java's `Tasks.foreach(...).stopRetryOn(NotFoundException.class).retry(N)`. The retry primarily helps atomic-bulk FileIO impls (e.g. an S3 `DeleteObjects`-backed FileIO) ride out transient throttles -- best-effort, see code comment for the fail-fast-iterative caveat.
* New public builder `ExpireSnapshots::ExecuteDeleteWith(OptionalExecutor)` (named after Java's `executeDeleteWith(ExecutorService)`) so callers opt in to parallel deletion; threaded through both `IncrementalFileCleanup` and `ReachableFileCleanup`.
* `DeleteWith()` doc note clarifies that the user-supplied callback may be invoked concurrently from worker threads and must be thread-safe.
* Drops the `std::async` / `std::thread` / `std::span` machinery and the ad-hoc retry loop -- replaced by `util/task_group.h` and `util/retry_util.h` from #687.

### Test coverage

Existing 25 `ExpireSnapshots*` tests continue to pass unchanged (they hit the default no-executor path).

Adds `ExpireSnapshotsCleanupTest.ExecutorDispatchesDeletesConcurrently` -- wires a `test::ThreadExecutor` through `ExecuteDeleteWith()`, runs an `ExpireSnapshotId` cleanup with a custom `DeleteWith` callback, and asserts the executor received one submission per file (data + manifest + manifest-list = 3) and that all paths show up in the mutex-guarded sink.


src/iceberg/update/expire_snapshots.cc
```diff
@@ -54,6 +58,64 @@ Result<std::shared_ptr<ManifestReader>> MakeManifestReader(
...
+
+  std::vector<std::future<void>> futures;
+  futures.reserve(n);
+  std::size_t per = (items.size() + n - 1) / n;
+  for (std::size_t i = 0; i < n; ++i) {
+    std::size_t begin = i * per;
+    if (begin >= items.size()) break;
+    std::size_t end = std::min(begin + per, items.size());
+    auto slice = items.subspan(begin, end - begin);
+    futures.emplace_back(std::async(std::launch::async, [slice, &work]() {
```
wgtmac: I don't think it is a good idea to use `std::async` here. Instead, we may want to introduce a thread pool to fully control the resource.

shangxinli: Good call -- replaced `std::async` with a dedicated `ThreadPool` in ff3a254. The pool is owned by `FileCleanupStrategy`, sized once at construction (`min(8, hardware_concurrency)`), and drained on destruction, so `CleanFiles` no longer spins up fresh threads on each `DeleteFiles` call. Added `util/thread_pool_internal.{h,cc}` plus `ThreadPoolTest` coverage (submit, fan-out, exception isolation, drain). PTAL.

src/iceberg/util/thread_pool_internal.h
```diff
@@ -0,0 +1,103 @@
...
+
+/// \brief Fixed-size worker pool for fire-and-wait task batches.
+///
+/// Internal API: not part of the public iceberg surface. Used today by
+/// FileCleanupStrategy in update/expire_snapshots.cc to parallelize bulk deletes
+/// without spawning a fresh std::thread per call. Workers are created eagerly in
+/// the constructor and joined in the destructor.
+///
+/// All public methods are thread-safe.
+class ICEBERG_EXPORT ThreadPool {
```
wgtmac: @HuaHuaY is already working on a design of thread pool that with pluggability. I'd prefer to pause this PR until we finalize the design.

HuaHuaY: I have submitted a new PR https://github.com/apache/iceberg-cpp/pull/687, which allows users to use their own thread pools through our specified interface.

wgtmac: Thanks to @HuaHuaY, https://github.com/apache/iceberg-cpp/pull/687 is merged. Do you want to rebase on it, @shangxinli?

shangxinli: Rebased onto #687 — force-pushed as `7a329ad`. Single commit on top of current `main`.

Changes:
- `FileCleanupStrategy` now takes an `OptionalExecutor`. When a custom `DeleteWith()` callback is set, per-path deletes fan out through `TaskGroup<>` (uses the executor if provided, otherwise runs serially via `RunTasksSingleThreaded`).
- The `FileIO::DeleteFiles` bulk path is wrapped in `RetryRunner<retry::StopRetryOn<ErrorKind::kNotFound>>` with a tight budget (2 retries, 5s total). Mirrors Java's `Tasks.foreach(...).stopRetryOn(NotFoundException.class).retry(N)`.
- Added `ExpireSnapshots::Executor(OptionalExecutor)` so callers can opt in to parallel deletion; threaded through both `IncrementalFileCleanup` and `ReachableFileCleanup`.
- Dropped the `std::async` / `std::thread` / `std::span` machinery and the ad-hoc retry loop in favor of `util/task_group.h` and `util/retry_util.h` from #687.
- New `ExecutorDispatchesDeletesConcurrently` test wires a `test::ThreadExecutor` through the new API and asserts one submission per file.

PTAL @wgtmac, and thanks @HuaHuaY for #687.

shangxinli: Heads up @wgtmac — quick self-review pass after the rebase, force-pushed as `70d507d` (amends `7a329ad`):

- Renamed the new setter `Executor()` → `ExecuteDeleteWith()` to match Java's `executeDeleteWith(ExecutorService)` and remove the name shadow over `iceberg::Executor`.
- Sharpened the comment above the bulk-path `RetryRunner` to be honest about its limits: it helps atomic-bulk FileIO impls (e.g. S3 `DeleteObjects`); for the default fail-fast iterative impl a retried attempt re-submits the full path list and may short-circuit early on `kNotFound`. Still best-effort, no worse than the prior un-retried call.
- Refreshed the PR description body — it had been describing the old `std::async`-based approach.

No behavioural changes vs `7a329ad`; same 26 tests passing locally.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -99,19 +105,48 @@ class FileCleanupStrategy {
...
+      // Status to match Java's suppressFailureWhenFinished semantics.
+      RetryRunner<retry::StopRetryOn<ErrorKind::kNotFound>> runner(kDeleteRetryConfig);
+      std::ignore = runner.Run([&]() { return file_io_->DeleteFiles(path_list); });
+      return;
+    }
+
+    // Custom callback path: invoke one path at a time, optionally on a worker thread
+    // pulled from the configured executor. Without an executor TaskGroup runs the
+    // callbacks synchronously on the calling thread.
+    TaskGroup<> group;
```
HuaHuaY: You can use `TaskGroup<retry::StopRetryOn<ErrorKind::kNotFound>> group(kDeleteRetryConfig)` to simplify code.

shangxinli: sure

HuaHuaY: My mistakes. I confused this with line 128 and didn’t notice that there is no retry here. If you need to retry in the future, you can do this.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -99,19 +105,52 @@ class FileCleanupStrategy {
...
-    try {
-      if (delete_func_) {
-        for (const auto& path : paths) {
-          delete_func_(path);
-        }
-      } else {
-        std::vector<std::string> path_list(paths.begin(), paths.end());
-        std::ignore = file_io_->DeleteFiles(path_list);
+    if (paths.empty()) return;
+    std::vector<std::string> path_list(paths.begin(), paths.end());
```
HuaHuaY: We only need to create the vector when there is not `delete_func_`.

HuaHuaY (approved):

wgtmac (approved):
Thank you, @shangxinli and @HuaHuaY!

PR #651, opened 2026-05-11
feat: metrics for parquet writer

WZhuo:


src/iceberg/parquet/parquet_metrics.cc
```diff
@@ -0,0 +1,370 @@
...
+        .field_id = field_id,
+        .value_count = value_count,
+        .null_value_count = null_count,
+    };
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto truncated_lower,
+                          TruncateUtils::TruncateLowerBound(
+                              *iceberg_type, lower_bound.value(), truncate_length));
+  ICEBERG_ASSIGN_OR_RAISE(auto truncated_upper,
```
wgtmac, line 157: This turns a missing representable upper bound into a hard metrics failure. For `truncate(N)`, Java returns null from `BinaryUtil.truncateBinaryMax` / `UnicodeUtil.truncateStringMax` when values like `0xff...` cannot produce a safe upper bound, and `ParquetMetrics` just omits the upper bound. Here `TruncateUpperBound` bubbles an `InvalidArgument` out of `writer->metrics()`, so a valid file can be written but fail while building DataFile metrics. Can we treat this case as no upper bound instead?

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -152,15 +157,34 @@ class ParquetWriter::Impl {
...
 
+  Result<Metrics> metrics() {
+    if (writer_) {
+      return Invalid("Cannot return metrics for unclosed writer");
+    }
+    if (!metadata_) {
+      return Metrics();
+    }
+    return ParquetMetrics::GetMetrics(*schema_, *parquet_schema_, *metrics_config_,
+                                      *metadata_, {});
```
wgtmac, line 169: This drops the write-side field metrics. Java passes `model.metrics()` into `ParquetMetrics.metrics(...)`, which is where float/double NaN counts come from. With `{}` here, `nan_value_counts` stays empty even when the file has NaNs, and the tests currently skip that assertion. We should either collect/pass `FieldMetrics` here or leave NaN metrics unsupported explicitly.

wgtmac (commented):
A couple of metrics edge cases to fix before this lands.

src/iceberg/parquet/parquet_writer.cc
WZhuo: NaN value support will be implemented in a separate PR. This change keeps the current scope limited to Parquet metrics/docs and does not add write-side NaN counting yet.

src/iceberg/parquet/parquet_metrics.cc
WZhuo: Updated this path so upper-bound truncation now returns an optional value. When no safe upper bound can be represented, Parquet metrics omit upper_bounds for that field instead of failing metrics generation; real errors still propagate via Result.

src/iceberg/metrics.h
```diff
@@ -30,6 +30,35 @@
 
 namespace iceberg {
 
+/// \brief Field-level metrics for a single column.
+///
+/// This structure captures value counts, null counts, NaN counts, and optional
+/// lower/upper bounds for a specific field identified by its field_id.
+struct ICEBERG_EXPORT FieldMetrics {
+  /// \brief The field ID this metrics belongs to.
+  int32_t field_id;
```
wgtmac, line 39: Do we need to initialize it by an invalid value just in case undefined behavior?

src/iceberg/parquet/parquet_metrics.h
```diff
@@ -0,0 +1,64 @@
...
+  ///
+  /// \param schema The Iceberg schema for the table.
+  /// \param parquet_schema The Parquet schema descriptor.
+  /// \param metrics_config The configuration specifying how to collect metrics.
+  /// \param metadata The Parquet file metadata containing row group statistics.
+  /// \param field_metrics Optional per-field metrics computed during write.
+  ///        If provided, these take precedence over footer statistics.
+  /// \return Result containing the computed Metrics or an error.
+  static Result<Metrics> GetMetrics(
+      const Schema& schema, const ::parquet::SchemaDescriptor& parquet_schema,
```
wgtmac, line 59: In principle, we cannot expose symbols like `::parquet::` which belongs to thirdparty dependencies. So we need to rename this header file to `parquet_metrics_internal.h` to not install it. (note that the source file does not need to rename)

src/iceberg/parquet/parquet_metrics.cc
```diff
@@ -0,0 +1,370 @@
...
+
+    null_count += stats->null_count();
+    value_count += column_chunk->num_values();
+
+    if (stats->HasMinMax()) {
+      auto min_bytes = stats->EncodeMin();
+      auto min_span = std::span<const uint8_t>(
+          reinterpret_cast<const uint8_t*>(min_bytes.data()), min_bytes.size());
+      ICEBERG_ASSIGN_OR_RAISE(auto min_value,
+                              Conversions::FromBytes(iceberg_type, min_span));
```
wgtmac, line 129: `stats->EncodeMin()` is encoded by Parquet plain encoding but `Conversions::FromBytes` expects Iceberg single value serialization. This conversion may work coincidently for some primitive types. We need to fix it by creating a mapping between Parquet and Iceberg types and then use explicit cast to obtain typed Parquet values to create correct literals.

src/iceberg/parquet/parquet_metrics.cc
```diff
@@ -0,0 +1,370 @@
...
+      ICEBERG_ASSIGN_OR_RAISE(auto max_value,
+                              Conversions::FromBytes(iceberg_type, max_span));
+      if (!upper_bound.has_value() || max_value > upper_bound.value()) {
+        upper_bound = std::move(max_value);
+      }
+    }
+  }
+
+  if (!lower_bound.has_value() || !upper_bound.has_value() || lower_bound->IsNaN() ||
+      upper_bound->IsNaN()) {
```
wgtmac, line 146: If the above conversion has been fixed, we can move NaN check only for floating types.

src/iceberg/parquet/parquet_metrics.cc
```diff
@@ -0,0 +1,370 @@
...
+    return FieldMetrics{
+        .field_id = field_id,
+        .value_count = value_count,
+        .null_value_count = null_count,
+    };
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto truncated_lower,
+                          TruncateUtils::TruncateLowerBound(
```
wgtmac, line 155: Ditto, truncation can only be done on string types.

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -165,15 +157,35 @@ class ParquetWriter::Impl {
 
   std::vector<int64_t> split_offsets() const { return split_offsets_; }
 
+  Result<Metrics> metrics() {
+    if (writer_) {
+      return Invalid("Cannot return metrics for unclosed writer");
+    }
+    if (!metadata_) {
```
wgtmac, line 164: Isn't this an error when `metadata_` is not available if writer is closed?

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -130,6 +134,7 @@ class ParquetWriter::Impl {
     for (int i = 0; i < metadata->num_row_groups(); ++i) {
       split_offsets_.push_back(metadata->RowGroup(i)->file_offset());
     }
+    metadata_ = writer_->metadata();
```
wgtmac, line 137: We can consolidate `metadata` with `metadata_`

src/iceberg/metrics_config.cc
```diff
@@ -116,6 +130,14 @@ Result<std::shared_ptr<MetricsConfig>> MetricsConfig::Make(const Table& table) {
                       *sort_order.value_or(SortOrder::Unsorted()));
 }
 
+Result<std::shared_ptr<MetricsConfig>> MetricsConfig::Make(
+    std::unordered_map<std::string, std::string> properties) {
+  // Create a minimal TableProperties wrapper for the properties
+  TableProperties props = TableProperties::FromMap(std::move(properties));
+
+  return MakeInternal(props, Schema({}), *SortOrder::Unsorted());
```
wgtmac, line 138: This looks inconsistent with the Java impl. Here we pass an empty schema to `MakeInternal` which is consumed later but the Java impl passes a null schema and directly assign the metrics mode to default later to bypass the schema checking. Perhaps we should change signature of MakeInternal to use `const Schema*` to indicate a missing schema.

src/iceberg/util/truncate_util.cc
```diff
@@ -167,42 +168,46 @@ Literal TruncateLiteralImpl<TypeId::kBinary>(const Literal& literal, int32_t wid
...
   for (auto it = truncated.rbegin(); it != truncated.rend(); ++it) {
     if (*it < 0xFF) {
       ++(*it);
       truncated.resize(truncated.size() - std::distance(truncated.rbegin(), it));
-      return Literal::Binary(std::move(truncated));
+      return std::optional<Literal>(Literal::Binary(std::move(truncated)));
     }
   }
-  return InvalidArgument("Cannot truncate upper bound for binary: all bytes are 0xFF");
+  return std::nullopt;
```
wgtmac, line 201: I think the original error message is worth keeping.

src/iceberg/util/truncate_util.cc
```diff
@@ -167,42 +168,46 @@ Literal TruncateLiteralImpl<TypeId::kBinary>(const Literal& literal, int32_t wid
...
+    const Literal& literal, int32_t width) {
   const auto& str = std::get<std::string>(literal.value());
-  ICEBERG_ASSIGN_OR_RAISE(std::string truncated,
-                          TruncateUtils::TruncateUTF8Max(str, width));
-  return Literal::String(std::move(truncated));
+  ICEBERG_ASSIGN_OR_RAISE(auto truncated, TruncateUtils::TruncateUTF8Max(str, width));
+  if (!truncated.has_value()) {
+    return std::nullopt;
+  }
+  return std::optional<Literal>(Literal::String(std::move(truncated.value())));
```
wgtmac, line 182: We can directly write `return truncated.transform([](std::string t) { return Literal::String(std::move(t)); });` to make it compact.

src/iceberg/util/truncate_util.cc
```diff
@@ -167,42 +168,46 @@ Literal TruncateLiteralImpl<TypeId::kBinary>(const Literal& literal, int32_t wid
...
 
 template <>
-Result<Literal> TruncateLiteralMaxImpl<TypeId::kBinary>(const Literal& literal,
-                                                        int32_t width) {
+Result<std::optional<Literal>> TruncateLiteralMaxImpl<TypeId::kBinary>(
+    const Literal& literal, int32_t width) {
   const auto& data = std::get<std::vector<uint8_t>>(literal.value());
   if (static_cast<int32_t>(data.size()) <= width) {
-    return literal;
+    return std::optional<Literal>(literal);
```
wgtmac, line 190: We don't need an extra wrapper of std::optional here, isn't it? Same for other changes in this file.

src/iceberg/util/truncate_util.cc
wgtmac, line 201: I think this error message is worth keeping.

src/iceberg/parquet/parquet_metrics.cc
```diff
@@ -0,0 +1,370 @@
...
+    return std::nullopt;
+  }
+
+  const auto& fm = it->second;
+  FieldMetrics result{.field_id = fm.field_id,
+                      .value_count = fm.value_count,
+                      .null_value_count = fm.null_value_count,
+                      .nan_value_count = fm.nan_value_count};
+
+  if (truncate_length > 0) {
```
wgtmac, line 190: Can we check primitive_type first to skip types that do not need truncation?

src/iceberg/test/parquet_metrics_test.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "iceberg/arrow/arrow_io_internal.h"
+#include "iceberg/arrow/arrow_status_internal.h"
+#include "iceberg/file_writer.h"
+#include "iceberg/parquet/parquet_register.h"
+#include "iceberg/test/metrics_test_base.h"
+#include "iceberg/util/checked_cast.h"
+
+namespace iceberg::test {
+
+class ParquetMetricsTest : public MetricsTestBase, public ::testing::Test {
```
wgtmac, line 36: The test suite is broad, but several cases are too forgiving and would miss real bugs.
- Decimal bounds are under-tested. MetricsForDecimals covers decimal-as-INT32, INT64, and fixed, but only checks that bounds exist. It should assert exact lower/upper values for each physical layout. This is exactly how the `EncodeMin()` vs Iceberg `Conversions::FromBytes` bug can slip through.
- “Multiple row group” tests may not actually test multiple row groups. MetricsForTopLevelWithMultipleRowGroup and nested equivalent write 201 rows, but Parquet’s `SupportsSmallRowGroups()` returns false, so row group count is not verified. Force multiple row groups or remove the claim.
- `field_metrics` override path is not tested. `ParquetMetrics::GetMetrics` supports precomputed `FieldMetrics`, but parquet_writer.cc always passes `{}`. Add direct tests where non-empty `field_metrics` overrides footer stats, carries NaN counts, and applies truncation.
- Bounds assertions ignore literal type. AssertBounds receives a `PrimitiveType` but does not validate the returned `Literal` type. So date vs int, timestamp vs long, etc. could pass accidentally. Assert the literal type as well as the value.
- Column sizes are barely verified. Several tests only check `column_sizes.empty()` or `!empty()`. They should assert exact field-id sets, especially for `none`, `counts`, nested-field overrides, and list/map exclusion behavior.
- Truncate integration is incomplete. Utility tests cover `std::nullopt` for impossible upper bounds, but there is no Parquet metrics test proving that an unrepresentable truncated upper bound omits only the upper bound while preserving counts/lower bound.

wgtmac (approved):
I've spotted various issues. Since this requires understanding of parquet internals, let me merge this as is and I will continue working on this. Thanks @WZhuo for initiating this!

PR #652, opened 2026-05-12
feat(manifest): add ManifestFilterManager and ManifestMergeManager

gty404:
Implement two manifest management classes for table write operations:

- ManifestFilterManager: filters manifest entries by row filter expression, file path, or partition value; supports FailMissingDeletePaths validation. Rewrites manifests that contain matching files, marking entries as DELETED; passes through manifests that cannot contain matching files unchanged.

- ManifestMergeManager: merges small manifests using greedy bin-packing, grouping by partition_spec_id (manifests with different specs are never merged). Oversized manifests pass through unchanged. ADDED entries from prior manifests become EXISTING when merged (matching Java semantics).


src/iceberg/manifest/manifest_filter_manager.cc
```diff
@@ -0,0 +1,248 @@
...
+  int32_t spec_id = file.partition_spec_id.value_or(manifest_spec_id);
+  if (drop_partitions_.contains(spec_id, file.partition)) {
+    return true;
+  }
+
+  // Expression-based check (inclusive metrics)
+  for (const auto& de : delete_exprs_) {
+    ICEBERG_ASSIGN_OR_RAISE(auto* eval, GetMetricsEvaluator(metadata, spec_id, de));
+    ICEBERG_ASSIGN_OR_RAISE(auto matches, eval->Evaluate(file));
+    if (matches) return true;
```
wgtmac: This row-filter delete path is broader than Java's behavior. Java's ManifestFilterManager first computes a partition residual and then uses both inclusive and strict metrics: files are deleted only when rowsMustMatch is true, and data files where only some rows may match raise `Cannot delete file where some, but not all, rows match filter`. Here, an inclusive `might match` result is enough to mark the whole file as DELETED, so a file whose bounds straddle the predicate would be incorrectly removed instead of rejected.

src/iceberg/manifest/manifest_filter_manager.cc
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  return false;
+}
+
+Result<std::vector<ManifestFile>> ManifestFilterManager::FilterManifests(
+    const TableMetadata& metadata, const std::shared_ptr<Snapshot>& base_snapshot,
+    const ManifestWriterFactory& writer_factory) {
+  // No base snapshot → nothing to filter
+  if (!base_snapshot) return std::vector<ManifestFile>{};
```
wgtmac: Java still validates required deletes when the manifest list is null or empty: `filterManifests` calls `validateRequiredDeletes()` before returning. With `FailMissingDeletePaths()` set, this early return silently succeeds for a missing path on an empty/no-base snapshot, while Java reports `Missing required files to delete`.

src/iceberg/manifest/manifest_filter_manager.cc
```diff
@@ -0,0 +1,248 @@
...
+    auto result = evaluator_ptr->Evaluate(manifest);
+    if (!result.has_value() || result.value()) return true;
+  }
+  return false;
+}
+
+bool ManifestFilterManager::CanContainDeletedFiles(const ManifestFile& manifest,
+                                                   const TableMetadata& metadata) {
+  // A manifest with no live files cannot contain files to delete.
+  bool has_live = (manifest.added_files_count.value_or(0) > 0) ||
```
wgtmac: This treats missing manifest counts as zero live files. In Iceberg/Java, null `addedFilesCount` or `existingFilesCount` means the count is unknown and `ManifestFile.hasAddedFiles/hasExistingFiles` returns true. Skipping such manifests can miss deletes for older or compatible metadata where counts are not populated.

src/iceberg/manifest/manifest_merge_manager.cc
```diff
@@ -0,0 +1,160 @@
...
+    const std::vector<ManifestFile>& new_manifests, int64_t snapshot_id,
+    const TableMetadata& metadata, std::shared_ptr<FileIO> file_io,
+    const ManifestWriterFactory& writer_factory) {
+  // Combine new then existing (new-first ordering is preserved in output)
+  std::vector<ManifestFile> all;
+  all.reserve(new_manifests.size() + existing_manifests.size());
+  all.insert(all.end(), new_manifests.begin(), new_manifests.end());
+  all.insert(all.end(), existing_manifests.begin(), existing_manifests.end());
+
+  if (!merge_enabled_ || std::cmp_less(all.size(), min_count_to_merge_)) {
```
wgtmac: This global `min_count_to_merge_` check does not match Java. Java only applies `minCountToMerge` to the bin containing the first/newest manifest; older bins that do not contain `first` can still be merged even when the total input count is below the threshold. Returning all manifests here can skip merges Java would perform.

src/iceberg/manifest/manifest_merge_manager.cc
```diff
@@ -0,0 +1,160 @@
...
+    const ManifestWriterFactory& writer_factory) {
+  // Collect bins using greedy first-fit packing.
+  std::vector<std::vector<ManifestFile>> bins;
+  std::vector<ManifestFile> current_bin;
+  int64_t bin_size = 0;
+
+  std::vector<ManifestFile> result;
+
+  for (const auto& manifest : group) {
+    if (manifest.manifest_length >= target_size_bytes_) {
```
wgtmac: Java uses `ListPacker.packEnd(group, ManifestFile::length)` with lookback 1 specifically to preserve manifest order and leave the under-filled bin at the beginning. This forward hand-rolled packing keeps the current small bin open across an oversized manifest, so a sequence like [small, oversized, small] can merge the two small manifests across the oversized one and reorder output relative to Java.

src/iceberg/manifest/manifest_merge_manager.cc
```diff
@@ -0,0 +1,160 @@
...
+
+  if (!merge_enabled_ || std::cmp_less(all.size(), min_count_to_merge_)) {
+    return all;
+  }
+
+  // The first (newest) manifest governs the per-bin minCountToMerge check.
+  const ManifestFile& first = all[0];
+
+  // Group manifests by partition_spec_id — never merge across specs
+  std::map<int32_t, std::vector<ManifestFile>> by_spec;
```
wgtmac: Java groups manifests by partition spec using a reverse-order TreeMap. This `std::map` iterates specs in ascending order, so the merged manifest list order can differ from Java. For v3 tables this ordering is observable because manifest-list writing assigns first-row IDs in output order.

src/iceberg/manifest/manifest_merge_manager.cc
```diff
@@ -0,0 +1,160 @@
...
+  all.reserve(new_manifests.size() + existing_manifests.size());
+  all.insert(all.end(), new_manifests.begin(), new_manifests.end());
+  all.insert(all.end(), existing_manifests.begin(), existing_manifests.end());
+
+  if (!merge_enabled_ || std::cmp_less(all.size(), min_count_to_merge_)) {
+    return all;
+  }
+
+  // The first (newest) manifest governs the per-bin minCountToMerge check.
+  const ManifestFile& first = all[0];
```
wgtmac: Please keep the Java empty-input behavior here before indexing into `all`. Java `ManifestMergeManager.mergeManifests` returns the input immediately when the iterator is empty, and `commit.manifest.min-count-to-merge=0` is a valid/aggressively-used setting in Java tests. With this implementation, an empty data/delete manifest input plus `min_count_to_merge_ <= 0` gets past the size check and reads `all[0]`, which is UB/crash instead of returning an empty manifest list.

src/iceberg/manifest/manifest_filter_manager.cc
```diff
@@ -0,0 +1,248 @@
...
+
+Result<bool> ManifestFilterManager::ShouldDelete(const ManifestEntry& entry,
+                                                 const TableMetadata& metadata,
+                                                 int32_t manifest_spec_id) {
+  if (!entry.data_file) return false;
+  const DataFile& file = *entry.data_file;
+
+  // Path-based check
+  if (delete_paths_.count(file.file_path)) {
+    pending_paths_.erase(file.file_path);
```
wgtmac: This consumes the required-delete tracking while filtering, which makes `FailMissingDeletePaths` non-idempotent across retries. Java keeps `deletePaths` immutable and `validateRequiredDeletes` recomputes the deleted file set from the filtered manifests on every call, so a retry against a changed base can still fail if the required path is no longer present. Here, once the first filtering pass erases the path from `pending_paths_`, a later `FilterManifests` call on the same manager can miss the validation failure because the path stays consumed.

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -0,0 +1,171 @@
...
+
+/// \brief Filters an existing snapshot's manifest list.
+///
+/// The manager accumulates delete conditions incrementally, then applies them all
+/// at once in a single FilterManifests() call.  Manifests that contain no deleted
+/// entries are returned unchanged (no I/O).  Manifests that do contain deleted
+/// entries are rewritten with those entries marked DELETED.
+///
+/// The manager is content-agnostic: pass ManifestContent::kData to process data
+/// manifests, or ManifestContent::kDeletes to process delete manifests.
```
wgtmac, line 51: If this manager is intended to cover delete manifests, it is still missing the Java `DeleteFileFilterManager` cleanup semantics. Java sets `dropDeleteFilesOlderThan(minDataSequenceNumber)` and `removeDanglingDeletesFor(filesToBeDeleted)`, then marks old delete files and dangling DVs as deleted while rewriting delete manifests. This implementation only applies path/partition/row-filter deletes, so a data-file delete or truncate can leave obsolete delete files/DVs live in the delete manifests, diverging from Java behavior and its DV cleanup tests. If delete-manifest cleanup is planned in a later layer, it would be good to avoid advertising `kDeletes` parity here or make the missing behavior explicit.

src/iceberg/manifest/manifest_merge_manager.cc
wgtmac: Could we keep data and delete manifests type-separated in this API? Java has distinct DataFileMergeManager and DeleteFileMergeManager, and MergingSnapshotProducer passes dataManifests/deleteManifests separately. Here the grouping key is only partition_spec_id, while FlushBin creates the writer from the first manifest content; if a caller accidentally gives a data and delete manifest with the same spec in one call, the bin can mix contents and fail mid-rewrite, a state Java typed managers make impossible. Either include content in the grouping/constructor or make the manager content-specific, and add a mixed-content / kDeletes merge test.

src/iceberg/manifest/manifest_merge_manager.h
```diff
@@ -0,0 +1,103 @@
...
+  /// \param existing_manifests Manifests already in the base snapshot
+  /// \param new_manifests Newly written manifests to incorporate
+  /// \param snapshot_id The ID of the snapshot being committed.  Used to preserve
+  ///   ADDED/DELETED status for entries written by this snapshot and to suppress
+  ///   stale DELETED tombstones from prior snapshots.
+  /// \param metadata Table metadata (provides specs and schema for readers)
+  /// \param file_io File IO used to open existing manifests for reading
+  /// \param writer_factory Factory to create new ManifestWriter instances
+  /// \return The merged manifest list, or an error
+  Result<std::vector<ManifestFile>> MergeManifests(
```
wgtmac, line 81: Could this manager expose the lifecycle state Java relies on for retries? Java ManifestMergeManager caches merged manifests, reports replacedManifestsCount(), and has cleanUncommitted(committed) so MergingSnapshotProducer can delete manifests produced by failed retry attempts and build created/kept/replaced summary accurately. This API only returns the final vector, so the later snapshot update layer has no way to distinguish committed vs uncommitted merged outputs or count replaced manifests consistently with Java tests such as transaction retry merge cleanup. The same lifecycle surface is also needed on the filter side for rewritten manifests.

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -0,0 +1,171 @@
...
+  /// \param expr The expression to match files against
+  /// \param case_sensitive Whether field name matching is case-sensitive
+  void DeleteByRowFilter(std::shared_ptr<Expression> expr, bool case_sensitive = true);
+
+  /// \brief Register an exact file path for deletion.
+  ///
+  /// Any manifest entry whose file_path matches this path will be marked DELETED.
+  ///
+  /// \param path The exact file path to delete
+  void DeleteFile(std::string_view path);
```
wgtmac, line 83: Could we keep the Java file-object delete path in the manager API instead of only accepting paths? Java ManifestFilterManager supports delete(F file) in addition to delete(CharSequence path); RewriteFiles.deleteFile(DataFile/DeleteFile) and StreamingDelete.deleteFile(DataFile) use that path, and the manager then retains file identity/manifest references for filesToBeDeleted(), buildSummary(), duplicate-delete accounting, and delete-manifest DV cleanup. With only DeleteFile(string_view), later layers have to degrade object deletes to paths and lose the state Java relies on. It would be good to add DataFile/DeleteFile object-delete coverage alongside the current path-only tests.

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -0,0 +1,171 @@
...
+#include "iceberg/type_fwd.h"
+#include "iceberg/util/partition_value_util.h"
+
+namespace iceberg {
+
+/// \brief Factory type for creating ManifestWriter instances during filtering/merging.
+///
+/// The factory receives the partition spec ID (to look up the spec) and the manifest
+/// content type, and returns a new ManifestWriter ready for writing.  The caller
+/// (i.e. MergingSnapshotUpdate in PR2) captures metadata, FileIO, and snapshot ID
```
zhjwpku: It seems `MergingSnapshotUpdate` isn't in the codebase. I don't mind you put it here if you plan to add it in a follow-up PR, but the `PR2` is very confusing.

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -0,0 +1,171 @@
...
+  const ManifestContent manifest_content_;
+  std::shared_ptr<FileIO> file_io_;
+
+  std::vector<DeleteExpr> delete_exprs_;
+  std::unordered_set<std::string> delete_paths_;
+  std::unordered_set<std::string> pending_paths_;
+  PartitionSet drop_partitions_;
+  bool fail_missing_delete_paths_{false};
+
+  // Cache: (spec_id, expr_index) → ManifestEvaluator
```
zhjwpku: I had previously suggested in another PR that we stick to ASCII characters in comments. However, after searching the codebase, it looks like this style is already in use, so I'm fine with it now.

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -0,0 +1,171 @@
...
+  /// \param spec_id The partition spec ID
+  /// \param partition The partition values to drop
+  void DropPartition(int32_t spec_id, PartitionValues partition);
+
+  /// \brief Set a flag that makes FilterManifests() fail if any registered
+  /// delete path was not found in any manifest entry.
+  void FailMissingDeletePaths();
+
+  /// \brief Returns true if any delete condition has been registered.
+  bool DeletesFiles() const;
```
zhjwpku: Should we use the following function names as Java impl?

```suggestion
  bool ContainsDeletes() const;
```

src/iceberg/manifest/manifest_merge_manager.h
```diff
@@ -0,0 +1,103 @@
...
+  ///   apply the minCountToMerge threshold on the bin that contains it.
+  Result<std::vector<ManifestFile>> MergeGroup(
+      const std::vector<ManifestFile>& group, const ManifestFile& first,
+      int64_t snapshot_id, const TableMetadata& metadata, std::shared_ptr<FileIO> file_io,
+      const ManifestWriterFactory& writer_factory);
+
+  /// \brief Write a merged manifest from all manifests in a bin.
+  ///
+  /// Entries are written snapshot-aware:
+  /// - ADDED from \p snapshot_id  → WriteAddedEntry (preserve status)
```
zhjwpku: Do we need the `\p`?

src/iceberg/manifest/manifest_merge_manager.h
```diff
@@ -0,0 +1,103 @@
...
+      const std::vector<ManifestFile>& existing_manifests,
+      const std::vector<ManifestFile>& new_manifests, int64_t snapshot_id,
+      const TableMetadata& metadata, std::shared_ptr<FileIO> file_io,
+      const ManifestWriterFactory& writer_factory);
+
+ private:
+  /// \brief Merge a group of manifests sharing the same spec_id.
+  ///
+  /// \param first The overall first (newest) manifest across all groups, used to
+  ///   apply the minCountToMerge threshold on the bin that contains it.
```
zhjwpku: nit, min_count_to_merge is a documented param, so I think we should use that. Other places using minCountToMerge are fine because those are not in document.

```suggestion
  ///   apply the min_count_to_merge threshold on the bin that contains it.
```

src/iceberg/manifest/manifest_filter_manager.h
gty404: I broke down the support for MergingSnapshotUpdate into multiple PRs. This is the first PR, and the second is the support for MergingSnapshotUpdate. Next are the commit interfaces, such as overwrite, append, and replace partitions. I will delete this comment when I submit PR2.

src/iceberg/manifest/manifest_filter_manager.h
gty404: This PR's ManifestFilterManager only handles path/partition/row-filter deletes. Java's dropDeleteFilesOlderThan(minDataSequenceNumber) and removeDanglingDeletesFor(filesToBeDeleted) / DV cleanup semantics are intentionally out of scope here. these delete-file-specific v2 semantics (minSequenceNumber cleanup, dangling DV detection) are deferred to (RowDelta), which is where BaseRowDelta.java's delete file management lives in Java. We can strengthen the note to name PR5 explicitly if that helps track it.

src/iceberg/manifest/manifest_merge_manager.cc
gty404: The grouping key was already changed to (partition_spec_id, content) in the review-fix commit, so data and delete manifests with the same spec can never land in the same bin. The writer factory call also uses first.content to ensure the output manifest has the correct content type.

src/iceberg/manifest/manifest_merge_manager.h
gty404: Fixed. Removed the \p markers from the bullet-list body (they only make sense in \param declarations).

src/iceberg/manifest/manifest_merge_manager.h
gty404: Fixed. Changed to min_count_to_merge to match the documented constructor parameter name.

src/iceberg/manifest/manifest_merge_manager.h
gty404: These lifecycle methods are deferred to PR2 (MergingSnapshotUpdate). The header \note at line 67–70 documents this explicitly: "Java's ManifestMergeManager additionally exposes replacedManifestsCount() and 
 cleanUncommitted(committed) for retry / rollback support. In this C++ implementation those responsibilities are handled by the caller (MergingSnapshotUpdate in PR2) via a tracked ManifestWriterFactory wrapper." The
 filter-side equivalent (cleanUncommitted for rewritten manifests) will live in the same PR2 caller layer.

src/iceberg/manifest/manifest_merge_manager.cc
gty404: Fixed. MergeGroup now mirrors ListPacker.packEnd(group, ManifestFile::length) with lookback=1 exactly

zhjwpku (approved):
LGTM, looking forward for your PR2.

wgtmac (approved):

wgtmac:
Thanks @gty404 for working on this and @zhjwpku for the review!

PR #653, opened 2026-05-17
feat: Add v3 timestamp nanosecond primitive types

zhjwpku:


zhjwpku:
I chose TypeId::kTimestampNs over TypeId::kTimestampNano (Java uses Nano) to align with the spec. @evindj Please help review the timestamp parsing part when you have time. I changed the fractional seconds handling a bit.

src/iceberg/util/bucket_util.cc
```diff
@@ -63,6 +63,16 @@ int32_t HashLiteral<TypeId::kTimestampTz>(const Literal& literal) {
...
 
+template <>
+int32_t HashLiteral<TypeId::kTimestampNs>(const Literal& literal) {
+  return BucketUtils::HashLong(std::get<int64_t>(literal.value()));
+}
+
+template <>
+int32_t HashLiteral<TypeId::kTimestampTzNs>(const Literal& literal) {
+  return BucketUtils::HashLong(std::get<int64_t>(literal.value()));
+}
```
wgtmac, line 77: According to the Iceberg V3 spec and the Java implementation (`BucketTimestampNano.java`), nanosecond timestamps must be converted to microseconds (divided by 1000) before hashing. This ensures that bucket partitioning is consistent between microsecond and nanosecond precision types for the same logical time.

```cpp
return BucketUtils::HashLong(std::get<int64_t>(literal.value()) / 1000);
```

src/iceberg/util/transform_util.cc
```diff
@@ -117,6 +199,21 @@ std::string TransformUtil::HumanTimestamp(int64_t timestamp_micros) {
   }
 }
 
+std::string TransformUtil::HumanTimestampNs(int64_t timestamp_nanos) {
+  auto tp = std::chrono::time_point<std::chrono::system_clock, std::chrono::seconds>{
+      std::chrono::seconds(timestamp_nanos / kNanosPerSecond)};
+  auto nanos = timestamp_nanos % kNanosPerSecond;
```
wgtmac: For negative timestamps (pre-1970), C++'s division (`/`) and modulo (`%`) operators truncate towards zero. This causes `ParseTimestampNs` and `HumanTimestampNs` to compute an incorrect base time point and a negative fractional part, breaking the string formatting and parsing.

For example, `1969-12-31T23:59:59.123456789` parses to `-876543211` nanos. Passing this back here yields `0` for seconds and `-876543211` for nanos, resulting in `1970-01-01T00:00:00.-876543211`.

Consider using `std::chrono::floor` to handle the negative values correctly (note: the original microsecond `HumanTimestamp` and `ParseTimestamp` also suffer from this issue).

src/iceberg/util/bucket_util.cc
zhjwpku: Fixed.

src/iceberg/util/transform_util.cc
zhjwpku: Good catch, fixed with some additional test cases.

wgtmac (commented):
Thanks for adding `timestamp_ns` and `timestamptz_ns`! Here are a few findings based on Java parity and the Iceberg Spec.

src/iceberg/expression/literal.cc
```diff
@@ -250,14 +263,27 @@ Result<Literal> LiteralCaster::CastFromString(
...
+      internal::checked_cast<const TimestampBase&>(*literal.type());
+  const bool source_is_nanos = source_timestamp.time_unit() == TimeUnit::kNanosecond;
 
   switch (target_type->type_id()) {
     case TypeId::kDate: {
       ICEBERG_ASSIGN_OR_RAISE(auto days, TemporalUtils::ExtractDay(literal));
       return Literal::Date(std::get<int32_t>(days.value()));
     }
+    case TypeId::kTimestamp:
+      return source_is_nanos ? Literal::Timestamp(timestamp_val / 1000)
```
wgtmac: C++ integer division truncates toward zero, causing incorrect results for negative timestamps (pre-1970) not evenly divisible by 1000. Java uses `Math.floorDiv`. We should use a floor division helper here.

src/iceberg/expression/literal.cc
```diff
@@ -250,14 +263,27 @@ Result<Literal> LiteralCaster::CastFromString(
...
+                             : Literal::Timestamp(timestamp_val);
     case TypeId::kTimestampTz:
-      return Literal::TimestampTz(timestamp_val);
+      return source_is_nanos ? Literal::TimestampTz(timestamp_val / 1000)
+                             : Literal::TimestampTz(timestamp_val);
+    case TypeId::kTimestampNs:
+      return source_is_nanos ? Literal::TimestampNs(timestamp_val)
+                             : Literal::TimestampNs(timestamp_val * 1000);
+    case TypeId::kTimestampTzNs:
```
wgtmac: Casting from `Timestamp(Ns)` to `TimestampTz(Ns)` is allowed here, but Java (`TimestampNanoLiteral.to(Type)`) explicitly returns `null` (blocking this promotion) because timezone information is missing. Should we return `NotSupported` to match Java?

src/iceberg/test/literal_test.cc
```diff
@@ -580,7 +580,15 @@ INSTANTIATE_TEST_SUITE_P(
...
                               .expected_type_id = TypeId::kTimestampTz,
-                              .expected_string = "1684137600000000"}),
+                              .expected_string = "1684137600000000"},
+        BasicLiteralTestParam{.test_name = "TimestampNs",
+                              .literal = Literal::TimestampNs(1684137600000000001LL),
+                              .expected_type_id = TypeId::kTimestampNs,
+                              .expected_string = "1684137600000000001"},
+        BasicLiteralTestParam{.test_name = "TimestampTzNs",
+                              .literal = Literal::TimestampTzNs(1684137600000000001LL),
+                              .expected_type_id = TypeId::kTimestampTzNs,
```
wgtmac, line 614: Consider adding cast tests for `TimestampNs` and `TimestampTzNs` (e.g., from String, and cross-casting between `TimestampNs` and `Timestamp`), especially for negative timestamps, to ensure rounding parity with Java.

src/iceberg/type_fwd.h
```diff
@@ -46,6 +46,8 @@ enum class TypeId {
   kTime,
   kTimestamp,
   kTimestampTz,
+  kTimestampNs,
+  kTimestampTzNs,
```
wgtmac, line 50: Should we sort them as 

```
  kTimestamp,
  kTimestampNs,
  kTimestampTz,
  kTimestampTzNs,
```

src/iceberg/type_fwd.h
```diff
@@ -77,6 +80,8 @@ class TimeType;
 class TimestampBase;
 class TimestampType;
 class TimestampTzType;
+class TimestampNsType;
```
wgtmac, line 83: ditto

src/iceberg/expression/literal.cc
```diff
@@ -395,8 +438,11 @@ bool Comparable(TypeId lhs, TypeId rhs) {
     case TypeId::kLong:
     case TypeId::kTimestamp:
     case TypeId::kTimestampTz:
+    case TypeId::kTimestampNs:
+    case TypeId::kTimestampTzNs:
       return rhs == TypeId::kLong || rhs == TypeId::kTimestamp ||
-             rhs == TypeId::kTimestampTz;
+             rhs == TypeId::kTimestampTz || rhs == TypeId::kTimestampNs ||
```
wgtmac: This looks incorrect to me. Should we be strict that only identical types are allowed to compare? It looks also dangerous to compare a timestamp value against a long value. Should we remove that support as well?

src/iceberg/util/bucket_util.cc
wgtmac: I think the original review comment had hallucination. Sorry about that.

The actual workflow is as below:

```
  private static class BucketTimestampNano extends Bucket<Long>
      implements SerializableFunction<Long, Integer> {

    private BucketTimestampNano(int numBuckets) {
      super(numBuckets);
    }

    @Override
    protected int hash(Long nanos) {
      return BucketUtil.hash(DateTimeUtil.nanosToMicros(nanos));
    }
  }
```

We can see that it also calls floorDiv inside:

```
  public static long nanosToMicros(long nanos) {
    return Math.floorDiv(nanos, NANOS_PER_MICRO);
  }

```

So my original (AI) suggestion was wrong. Please follow the same approach to use floorDiv here.

wgtmac: Perhaps it is worth adding a dedicated utility class/file for temporal types just like Java for reuse.

wgtmac (changes requested):

src/iceberg/util/bucket_util.cc
zhjwpku: We have a TemporalUtils/temporal_util.cc, I think we can enrich that.

src/iceberg/expression/literal.cc
zhjwpku: I checked the logic in java impl, ISTM zoned vs non-zoned is not distinguished at the literal cast level, in java zoned vs non-zoned share the same TypeID.

src/iceberg/expression/literal.cc
zhjwpku: Yeah, in Java, BaseLiteral<T>.equals requires the lhs and rhs to have the same getClass(). Since zoned and non-zoned timestamps appear to share the same class, I think we should allow kTimestamp/kTimestampTz and kTimestampNs/kTimestampTzNs to be comparable.

src/iceberg/type_fwd.h
zhjwpku: We don't seem to strictly follow alphabetical order, but rather the ordering defined in the spec?

src/iceberg/util/temporal_util.cc
```diff
@@ -21,19 +21,40 @@
...
 
+inline constexpr int64_t FloorDiv(int64_t dividend, int64_t divisor) {
+  const auto quotient = dividend / divisor;
+  if ((dividend ^ divisor) < 0 && quotient * divisor != dividend) {
+    return quotient - 1;
+  }
+  return quotient;
+}
+
+Result<int64_t> MultiplyExact(int64_t lhs, int64_t rhs) {
```
wgtmac, line 49: nit: we may introduce a dedicated math utility like https://github.com/apache/arrow/blob/96f80729c1581ee98f9d0a2cd9994ae50302491d/cpp/src/arrow/util/int_util_overflow.h#L184

src/iceberg/type_fwd.h
```diff
@@ -46,6 +46,8 @@ enum class TypeId {
   kTime,
   kTimestamp,
   kTimestampTz,
+  kTimestampNs,
```
wgtmac, line 49: This needs the same v3 gate as Java. Otherwise a v2 table can accept and write a v3-only schema.

src/iceberg/util/transform_util.cc
```diff
@@ -208,32 +312,32 @@ Result<int64_t> TransformUtil::ParseTimestamp(std::string_view str) {
...
-             (str[str.size() - 6] == '+' || str[str.size() - 6] == '-')) {
-    // Parse "+HH:mm" or "-HH:mm" offset suffix
-    ICEBERG_ASSIGN_OR_RAISE(offset_micros,
-                            ParseTimezoneOffset(str.substr(str.size() - 6)));
-    timestamp_part = str.substr(0, str.size() - 6);
-  } else {
-    return InvalidArgument("Invalid timestamptz string (missing timezone suffix): '{}'",
-                           str);
-  }
+  return static_cast<int64_t>(days) * kNanosPerDay + time_nanos;
```
wgtmac: This can overflow past the int64 nanos boundary. Please use checked arithmetic here and in the timezone offset path.

src/iceberg/util/temporal_util.cc
zhjwpku: Yeah, I also noticed that most of the utilities in `transform_util.h` seem to belong more naturally in a temporal_util. I'll create a separate PR to address these issues.

src/iceberg/type_fwd.h
zhjwpku: fixed.

src/iceberg/util/transform_util.cc
zhjwpku: fixed

src/iceberg/util/transform_util.cc
```diff
@@ -44,28 +49,124 @@ Result<int64_t> ParseTimezoneOffset(std::string_view offset) {
...
+  }
+
+  return std::make_pair(timestamp_part, offset_micros);
+}
+
+Result<int64_t> TimestampFromDayTime(int32_t days, int64_t time_units,
+                                     int64_t units_per_day, int64_t offset_micros,
+                                     int64_t units_per_micro) {
+  const auto offset_units =
+      static_cast<int128_t>(offset_micros) * static_cast<int128_t>(units_per_micro);
```
wgtmac, line 93: Should we use the new `MultiplyExact` instead of int128_t trick here? Depending on int128_t may occupy more cycles. If the next plan is to add a dedicate arithmetic utility for overflow check, I'm fine with that.

src/iceberg/expression/literal.cc
```diff
@@ -388,18 +451,19 @@ std::strong_ordering CompareFloat(T lhs, T rhs) {
...
-    default:
-      return lhs == rhs;
+  if (lhs == rhs) {
+    return true;
+  }
+  if ((lhs == TypeId::kInt || lhs == TypeId::kDate) &&
+      (rhs == TypeId::kInt || rhs == TypeId::kDate)) {
+    return true;
   }
+  if ((lhs == TypeId::kTimestamp || lhs == TypeId::kTimestampTz) &&
```
wgtmac, line 461: I checked that this aligns with the Java impl but I think that is because of Java's deficiency that timestamp type does not carry zone info. Not a strong opinion but I think it is much safer if we do not allow comparison between zoned and non-zoned.

wgtmac (approved):
I have left some minor comments and overall it is good to merge. Thanks @zhjwpku!

wgtmac:
Let me merge this PR since it has touched a lot of files and may conflict with other PRs.

src/iceberg/expression/literal.cc
zhjwpku: I create https://github.com/apache/iceberg-cpp/pull/676 to make it stricter

PR #656, opened 2026-05-18
fix(parquet): check compression codec availability

wgtmac:
## Summary
- Check Parquet compression codec availability with Arrow before opening the file writer.
- Return a clear `InvalidArgument` when a requested Parquet codec is not built into Arrow.
- Add a regression test that exercises an unavailable codec in the current Arrow build.

## Test Plan
- `cmake --build build --target parquet_test data_test && ctest --test-dir build -R 'parquet_test|data_test' --output-on-failure`
- `git diff --check`

### DIFF
--- a/src/iceberg/parquet/parquet_writer.cc
+++ b/src/iceberg/parquet/parquet_writer.cc
@@ -20,9 +20,11 @@
 #include "iceberg/parquet/parquet_writer.h"
 
 #include <memory>
+#include <string_view>
 
 #include <arrow/c/bridge.h>
 #include <arrow/record_batch.h>
+#include <arrow/util/compression.h>
 #include <arrow/util/key_value_metadata.h>
 #include <parquet/arrow/schema.h>
 #include <parquet/arrow/writer.h>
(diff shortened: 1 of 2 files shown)


zhjwpku (approved):
+1

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -45,21 +47,31 @@ Result<std::shared_ptr<::arrow::io::OutputStream>> OpenOutputStream(
...
   } else if (compression_name == "lz4") {
-    return ::arrow::Compression::LZ4;
+    compression = ::arrow::Compression::LZ4;
   } else if (compression_name == "zstd") {
-    return ::arrow::Compression::ZSTD;
+    compression = ::arrow::Compression::ZSTD;
   } else {
     return InvalidArgument("Unsupported Parquet compression codec: {}", compression_name);
   }
+  return compression;
```
WZhuo: It's unreachable

WZhuo (approved):

src/iceberg/parquet/parquet_writer.cc
zhjwpku: It's reachable. The `return`s were changed to assignments except the final `else` branch. But I have no idea if that's a better style :-)

src/iceberg/test/parquet_test.cc
```diff
@@ -124,6 +128,27 @@ void DoRoundtrip(std::shared_ptr<::arrow::Array> data, std::shared_ptr<Schema> s
   ASSERT_TRUE(out != nullptr) << "Reader.Next() returned no data";
 }
 
+struct ParquetCodec {
+  std::string name;
+  ::arrow::Compression::type compression;
+};
+
+std::optional<ParquetCodec> UnavailableParquetCodec() {
```
raulcd: LGTM should this function return the list of unavailable codecs instead of the first one? Otherwise I feel like the name could be slightly misleading, maybe `FirstUnavailableParquetCodec`.

raulcd (commented):
Just a minor nit! Thanks @wgtmac

raulcd (approved):

src/iceberg/test/parquet_test.cc
wgtmac: Renamed the helper to `FirstUnavailableParquetCodec` to match the first-match behavior.

src/iceberg/parquet/parquet_writer.cc
wgtmac: Updated `ParseCompression` to use early returns again while keeping the availability check after parsing.

raulcd (approved):

blackmwk (approved):

wgtmac:
Thanks all for the review!

PR #657, opened 2026-05-18
feat(data): add MOR file scan task reader

wgtmac:
Add FileScanTaskReader as the delete-aware Arrow stream entrypoint for file scan tasks, covering both no-delete reader passthrough and merge-on-read filtering with position and equality deletes.

Introduce reusable Arrow C Data utilities for stream wrapping and batch projection, including ProjectionContext caching and optional Arrow compute registration via arrow::RegisterAll. Move Arrow IO registration into arrow_register and remove FileScanTask::ToArrow.

Add coverage for projection behavior across nanoarrow and Arrow compute paths, plus end-to-end FileScanTaskReader tests for projected reads, position deletes, equality deletes, dropped equality fields, and fully deleted batches.


src/iceberg/arrow_c_data_util_internal.h
```diff
@@ -0,0 +1,239 @@
...
+
+  ProjectBatchFunction project_batch_function() const;
+
+  ProjectBatchState& project_batch_state();
+
+ private:
+  ProjectionContext() = default;
+
+  const Schema* input_schema_ = nullptr;
+  const Schema* output_schema_ = nullptr;
```
WZhuo, line 106: nit: Storing raw pointers here with the requirement that callers ensure the schemas outlive the context is fragile. A future refactor could easily break this invariant. Consider either storing `shared_ptr<const Schema>` or at minimum adding a prominent comment at each call site (e.g. in `FileScanTaskReader::Impl::Open`) explaining why the lifetime is guaranteed.

src/iceberg/arrow_c_data_util_internal.h
```diff
@@ -0,0 +1,239 @@
...
+#include <cerrno>
+#include <concepts>
+#include <cstdint>
+#include <cstring>
+#include <memory>
+#include <optional>
+#include <span>
+#include <string>
+#include <tuple>
+#include <utility>
```
WZhuo, line 30: nit: `<tuple>` does not appear to be used in this header. Can be removed.

src/iceberg/arrow_c_data_util_internal.h
```diff
@@ -0,0 +1,239 @@
...
+
+  ProjectBatchState& project_batch_state();
+
+ private:
+  ProjectionContext() = default;
+
+  const Schema* input_schema_ = nullptr;
+  const Schema* output_schema_ = nullptr;
+  std::vector<int> selected_field_indices_;
+  ArrowSchema input_arrow_schema_{};
```
WZhuo, line 108: nit: `selected_field_indices_` is `vector<int>` but `row_indices` in `ProjectBatch` is `span<const int32_t>`. Using `int32_t` consistently for both would avoid implicit conversions in the Arrow compute path.

src/iceberg/arrow_c_data_util_internal.h
```diff
@@ -0,0 +1,239 @@
...
+#include "iceberg/arrow_c_data.h"
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Cached state for ProjectBatch over one input/output schema pair.
+class ICEBERG_EXPORT ProjectionContext {
+ public:
```
WZhuo, line 44: nit: This is an internal header (suffix `_internal.h`) but uses `ICEBERG_EXPORT`. This is needed because the symbols are used across TUs within the library, but it is a bit surprising. A brief comment explaining why would help future readers.

src/iceberg/arrow_c_data_util.cc
```diff
@@ -0,0 +1,395 @@
...
+    const {
+  return project_batch_function_;
+}
+
+ProjectionContext::ProjectBatchState& ProjectionContext::project_batch_state() {
+  return project_batch_state_;
+}
+
+void ProjectionContext::RegisterProjectBatchFunction(
+    ProjectionContext::ProjectBatchFunction project_batch_function) {
```
WZhuo, line 312: The global function pointer with mutex is a process-wide singleton — if two libraries in the same process register different implementations, the last one wins silently. Worth documenting this "last writer wins" behavior, either here or in the header doc for `RegisterProjectBatchFunction`.

src/iceberg/arrow_c_data_util.cc
```diff
@@ -0,0 +1,395 @@
...
+}
+
+}  // namespace
+
+Result<ArrowArray> ProjectBatch(ArrowArray* input_batch,
+                                std::span<const int32_t> row_indices,
+                                ProjectionContext& projection) {
+  ICEBERG_PRECHECK(input_batch != nullptr, "input_batch must not be null");
+  internal::ArrowArrayGuard input_batch_guard(input_batch);
+
```
WZhuo, line 384: Suggestion: add a comment explaining why the `ArrowArrayGuard` here does not double-release when the Arrow compute path calls `ImportRecordBatch(input_batch, ...)`. The reason is that `ImportRecordBatch` zeroes the `release` pointer after import, so the guard becomes a no-op. Without this note, a reader might worry about double-free.

src/iceberg/arrow_c_data_util.cc
```diff
@@ -0,0 +1,395 @@
...
+  ArrowArrayView input_view;
+  ArrowError error;
+  ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+      ArrowArrayViewInitFromSchema(&input_view, &input_arrow_schema, &error), error);
+  internal::ArrowArrayViewGuard input_view_guard(&input_view);
+  ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+      ArrowArrayViewSetArray(&input_view, &input_batch, &error), error);
+  ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+      ArrowArrayViewValidate(&input_view, NANOARROW_VALIDATION_LEVEL_FULL, &error),
+      error);
```
WZhuo, line 346: `NANOARROW_VALIDATION_LEVEL_FULL` on every batch is expensive for production workloads. Consider making this configurable (e.g. via a debug flag or build option), or using `NANOARROW_VALIDATION_LEVEL_DEFAULT` by default and only enabling full validation in debug/test builds.

src/iceberg/arrow/arrow_c_data_util.cc
```diff
@@ -0,0 +1,130 @@
...
+                                            ProjectionContext& projection) {
+  ICEBERG_PRECHECK(input_batch != nullptr, "input_batch must not be null");
+  ICEBERG_ASSIGN_OR_RAISE(auto state, GetArrowProjectBatchState(projection));
+
+  ICEBERG_ARROW_ASSIGN_OR_RETURN(
+      auto input_record_batch,
+      ::arrow::ImportRecordBatch(input_batch, state->input_schema));
+
+  const int32_t empty_index = 0;
+  const int32_t* row_indices_data =
```
WZhuo, line 92: nit: `empty_index` is a bit misleading — it is just a non-null pointer to satisfy `Buffer::Wrap` when `row_indices` is empty (the value `0` is never actually used). A brief comment would clarify intent, e.g.:
```cpp
// Provide a valid pointer for Buffer::Wrap; the array length is 0 so this is never read.
```

src/iceberg/data/file_scan_task_reader.cc
```diff
@@ -0,0 +1,233 @@
...
+      return {};
+    }
+    return reader_->Close();
+  }
+
+  Result<std::optional<ArrowArray>> Next() {
+    if (!cached_schema_.has_value()) {
+      ICEBERG_ASSIGN_OR_RAISE(cached_schema_, reader_->Schema());
+    }
+    ArrowSchema& input_arrow_schema = cached_schema_.value();
```
WZhuo, line 90: The schema returned by `reader_->Schema()` is cached on first call and reused for all subsequent batches. This is correct for Parquet (schema is fixed), but if a future format reader could return varying schemas across batches, this would break. A brief comment noting the assumption would be helpful.

WZhuo (commented):
Nice work! This is a well-structured PR with solid test coverage. The design cleanly separates the nanoarrow fallback from the optional Arrow compute path, and the `ArrowArrayStreamProvider` concept is a good generalization. A few comments below — mostly minor suggestions and documentation requests.

src/iceberg/arrow_c_data_util_internal.h
wgtmac: Kept borrowed pointers and added lifetime comments in the header and FileScanTaskReader call site.

src/iceberg/arrow_c_data_util_internal.h
wgtmac: Removed.

src/iceberg/arrow_c_data_util_internal.h
wgtmac: Changed selected field indices to int32_t consistently.

src/iceberg/arrow_c_data_util_internal.h
wgtmac: Added a short export comment.

src/iceberg/arrow_c_data_util.cc
wgtmac: Documented process-wide last-writer-wins behavior in the header.

src/iceberg/arrow_c_data_util.cc
wgtmac: Added a comment explaining the Arrow import ownership transfer.

src/iceberg/arrow_c_data_util.cc
wgtmac: Switched the nanoarrow path to DEFAULT validation.

src/iceberg/arrow/arrow_c_data_util.cc
wgtmac: Added the Buffer::Wrap empty-input comment.

src/iceberg/data/file_scan_task_reader.cc
wgtmac: Added a comment for the stable stream schema assumption.

WZhuo (approved):
LGTM

zhjwpku (approved):

blackmwk (approved):

wgtmac:
Thanks all for the review!

PR #659, opened 2026-05-18
feat(io): add bulk delete API to FileIO.

### LINKED ISSUE #658: Support bulk and parallel file deletion for ExpireSnapshots
## Background

When running `ExpireSnapshots`, iceberg-cpp may need to clean up files that are no longer referenced by expired snapshots. These files can include:

- data files
- delete files
- manifest files
- manifest list files
- statistics files

Currently, file deletion in iceberg-cpp is primarily based on single-file deletion through:

```
FileIO::DeleteFile(...)
```

When a large number of files need to be deleted, deleting them one by one can be inefficient, especially for object stores or remote filesystems where each delete request may involve non-trivial network latency.

Java Iceberg already has a similar abstraction:

```
SupportsBulkOperations#deleteFiles
```

This allows cleanup logic to use bulk deletion when supported by the underlying FileIO, and fall back to regular per-file deletion otherwise.

iceberg-cpp should consider adding a similar mechanism.

## Current Problem
iceberg-cpp does not currently have a unified bulk deletion entry point.

`FileIO` currently exposes single-file deletion:
```
virtual Status DeleteFile(const std::string& file_location);
```
As a result:

1. Deleting many files can be slow.
2. `ExpireSnapshots` cannot take advantage of storage-native bulk deletion.
3. FileIO implementations do not have a common extension point for optimized deletion.
4. There is no clear API layer for adding parallel deletion fallback in the future.

## Proposed Approach
This can be implemented incrementally.

### Step 1: Add a bulk delete API to FileIO

Add a new bulk deletion entry point, for example:

```
virtual Status DeleteFiles(std::span<const std::string> file_locations);
```

The initial implementation can provide a backward-compatible default fallback:
```
virtual Status DeleteFiles(std::span<const std::string> file_locations) {
  for (const auto& file_location : file_locations) {
    auto status = DeleteFile(file_location);
    if (!status.has_value()) {
      return status;
    }
  }
  return {};
}
```

The goal of this step is to:

- add a unified API
- preserve backward compatibility
- avoid requiring every FileIO implementation to immediately support native bulk deletion
- prepare for future optimizations

This step should only add the API and sequential fallback. It should not introduce parallel deletion yet, and it does not need to modify `ExpireSnapshots`.

### Step 3: Add optimized deletion implementations

slfan1989:
## Summary

Add a new `FileIO::DeleteFiles(...)` API as a bulk deletion entry point.

The default implementation deletes files sequentially by calling the existing
`DeleteFile(...)` method and returns the first deletion error encountered.

This PR only adds the API and backward-compatible fallback behavior. It does not
yet update `ExpireSnapshots` to use `DeleteFiles(...)`, and it does not introduce
parallel deletion.

Fixed: #658 

## Motivation

`ExpireSnapshots` and other cleanup flows may need to delete many files. A
bulk deletion API gives FileIO implementations a common extension point for
future optimized deletion strategies, such as storage-native batch deletion or
parallel fallback deletion.

### DIFF
--- a/src/iceberg/file_io.h
+++ b/src/iceberg/file_io.h
@@ -26,6 +26,7 @@
 #include <span>
 #include <string>
 #include <string_view>
+#include <vector>
 
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
@@ -163,15 +164,7 @@ class ICEBERG_EXPORT FileIO {
   ///
   /// \param file_locations The locations of the files to delete.
   /// \return void if all deletes succeeded, an error code if any delete failed.
-  virtual Status DeleteFiles(std::span<const std::string> file_locations) {
-    for (const auto& file_location : file_locations) {
-      auto status = DeleteFile(file_location);
-      if (!status.has_value()) {
-        return status;
-      }
-    }
-    return {};
-  }
+  virtual Status DeleteFiles(const std::vector<std::string>& file_locations);
 };
 
 }  // namespace iceberg
--- a/src/iceberg/file_io.h
+++ b/src/iceberg/file_io.h
@@ -163,7 +163,7 @@ class ICEBERG_EXPORT FileIO {
   /// and returns the first error encountered.
   ///
   /// \param file_locations The locations of the files to delete.
-  /// \return void if all deletes succeeded, an error code if any delete failed.
+  /// \return void if all deletes succeed, or an error code if any delete fails.
   virtual Status DeleteFiles(const std::vector<std::string>& file_locations);
 };
 
--- a/src/iceberg/file_io.h
+++ b/src/iceberg/file_io.h
@@ -151,7 +151,7 @@ class ICEBERG_EXPORT FileIO {
   ///
   /// \param file_location The location of the file to delete.
   /// \return void if the delete succeeded, an error code if the delete failed.
-  virtual Status DeleteFile(const std::string&) {
+  virtual Status DeleteFile(const std::string& file_location) {
     return NotImplemented("DeleteFile not implemented");
   }
 
--- a/src/iceberg/file_io.h
+++ b/src/iceberg/file_io.h
@@ -151,8 +151,26 @@ class ICEBERG_EXPORT FileIO {
   ///
   /// \param file_location The location of the file to delete.
   /// \return void if the delete succeeded, an error code if the delete failed.
-  virtual Status DeleteFile(const std::string& file_location) {
+  virtual Status DeleteFile(const std::string&) {
     return NotImplemented("DeleteFile not implemented");
+  }
+
+  /// \brief Delete files at the given locations.
+  ///
+  /// Implementations that can delete multiple files efficiently should override this
+  /// method. The default implementation deletes files sequentially using DeleteFile
+  /// and returns the first error encountered.
+  ///
+  /// \param file_locations The locations of the files to delete.
+  /// \return void if all deletes succeeded, an error code if any delete failed.
+  virtual Status DeleteFiles(std::span<const std::string> file_locations) {
+    for (const auto& file_location : file_locations) {
+      auto status = DeleteFile(file_location);
+      if (!status.has_value()) {
+        return status;
+      }
+    }
+    return {};
   }
 };
 
--- a/src/iceberg/test/file_io_test.cc
+++ b/src/iceberg/test/file_io_test.cc
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/file_io.h"
+
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <gtest/gtest.h>
+
+#include "iceberg/test/matchers.h"
+
+namespace iceberg {
+namespace {
+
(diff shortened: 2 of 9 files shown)


slfan1989:
@wgtmac Could you please help review this PR? This is my first code contribution to the iceberg-cpp module, and I would really appreciate your guidance and help. Thank you very much!

src/iceberg/file_io.h
```diff
@@ -151,9 +151,27 @@ class ICEBERG_EXPORT FileIO {
   ///
   /// \param file_location The location of the file to delete.
   /// \return void if the delete succeeded, an error code if the delete failed.
-  virtual Status DeleteFile(const std::string& file_location) {
```
zhjwpku, line 154: Why change this line?

src/iceberg/file_io.h
```diff
@@ -151,9 +151,27 @@ class ICEBERG_EXPORT FileIO {
...
+
+  /// \brief Delete files at the given locations.
+  ///
+  /// Implementations that can delete multiple files efficiently should override this
+  /// method. The default implementation deletes files sequentially using DeleteFile
+  /// and returns the first error encountered.
+  ///
+  /// \param file_locations The locations of the files to delete.
+  /// \return void if all deletes succeeded, an error code if any delete failed.
+  virtual Status DeleteFiles(std::span<const std::string> file_locations) {
```
zhjwpku: It would be better if you could make some changes that utilize this new API. I saw you mentioned `ExpireSnapshots`.

src/iceberg/test/file_io_test.cc
```diff
@@ -0,0 +1,74 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <gtest/gtest.h>
+
+#include "iceberg/test/matchers.h"
+
+namespace iceberg {
+namespace {
```
zhjwpku, line 31: We don't need this or its counterpart. If you want to wrap `RecordingFileIO` in an anonymous namespace, you should shrink the scope instead.

```suggestion
```

src/iceberg/file_io.h
slfan1989: I changed this only to avoid the unused parameter warning in the default implementation, since `file_location` is not used when returning `NotImplemented`.

I thought it was a small cleanup while touching this area, but I agree it is not required for this PR. I can revert it if you prefer.

src/iceberg/file_io.h
slfan1989: Thanks, that makes sense. I originally planned to split this into two PRs: first add the `FileIO::DeleteFiles` API, then update `ExpireSnapshots` to use it.

I agree that this PR would be more useful if it also adopted the new API. I’ll take a look at updating `ExpireSnapshots` to use `DeleteFiles` for grouped file cleanup.

src/iceberg/test/file_io_test.cc
slfan1989: Makes sense. I’ll shrink the anonymous namespace scope to cover only the helper `RecordingFileIO` and keep the tests outside of the `iceberg` namespace.

src/iceberg/file_io.h
```diff
@@ -154,6 +154,24 @@ class ICEBERG_EXPORT FileIO {
...
+    for (const auto& file_location : file_locations) {
+      auto status = DeleteFile(file_location);
```
wgtmac: nit: put the implementation to file_io.cc and use `ICEBERG_RETURN_UNEXPECTED` to save some lines.

src/iceberg/file_io.h
```diff
@@ -154,6 +154,24 @@ class ICEBERG_EXPORT FileIO {
...
   }
+
+  /// \brief Delete files at the given locations.
+  ///
+  /// Implementations that can delete multiple files efficiently should override this
+  /// method. The default implementation deletes files sequentially using DeleteFile
+  /// and returns the first error encountered.
+  ///
+  /// \param file_locations The locations of the files to delete.
+  /// \return void if all deletes succeeded, an error code if any delete failed.
```
wgtmac:
```suggestion
  /// \return void if all deletes succeed, or an error code if any delete fails.
```

src/iceberg/file_io.h
```diff
@@ -154,6 +154,24 @@ class ICEBERG_EXPORT FileIO {
...
+  virtual Status DeleteFiles(std::span<const std::string> file_locations) {
```
wgtmac: To be complete, let's override this function in `arrow_io_internal.h` by adapting to https://github.com/apache/arrow/blob/eab3d11b484df5bb6ddd75630f9bfa17c0a3f491/cpp/src/arrow/filesystem/filesystem.h#L274

A relevant question is whether to use `const std::vector<std::string>&` as the parameter so no conversion is needed to call the Arrow equivalent.

src/iceberg/test/file_io_test.cc
wgtmac: Why not putting all cases in the `iceberg` namespace so we don't need to use `iceberg::` prefix everywhere?

slfan1989: Fixed. I moved the test cases into the iceberg namespace and kept RecordingFileIO in an anonymous namespace, so the iceberg:: prefixes are no longer needed.

src/iceberg/file_io.h
slfan1989: Will update it. I'll move the default `DeleteFiles` implementation to file_io.cc and use `ICEBERG_RETURN_UNEXPECTED` for error propagation.

src/iceberg/file_io.h
slfan1989: I changed FileIO::DeleteFiles to take const std::vector<std::string>& to match Arrow FileSystem::DeleteFiles, and added an ArrowFileSystemFileIO override that resolves Iceberg file locations and delegates to arrow_fs_->DeleteFiles(paths). I also added a LocalFileIO test for the bulk delete path.

src/iceberg/test/file_io_test.cc
```diff
@@ -0,0 +1,75 @@
...
+
+class RecordingFileIO : public FileIO {
+ public:
+  explicit RecordingFileIO(std::string failure_path = "")
```
wgtmac, line 35: The name `RecordingFileIO` is not that obvious about its purpose but this is fine in the test.

wgtmac (approved):

wgtmac:
Thanks @slfan1989 for working on this and @zhjwpku for the review!

slfan1989:
> Thanks @slfan1989 for working on this and @zhjwpku for the review!

@wgtmac @zhjwpku Thank you for the reviews and guidance! I really appreciate your time and help throughout the PR.

PR #661, opened 2026-05-19
fix: preserve snapshot ID width in UseRef

### LINKED ISSUE #660: TableScanBuilder::UseRef truncates snapshot IDs above INT32_MAX
## Problem

`TableScanBuilder::UseRef()` copies `SnapshotRef::snapshot_id` into an `int32_t` before looking up the snapshot and writing the scan context.

Snapshot IDs are stored as `int64_t` in `SnapshotRef`, `TableMetadata::SnapshotById`, and `TableScanContext`. A ref that points at a snapshot ID greater than `INT32_MAX` can be truncated, which can make branch/ref scans fail to find the snapshot or select the wrong ID.

## Expected

`UseRef()` should preserve the full `int64_t` snapshot ID when resolving a ref.

fallintoplace:
## Summary

- keep `UseRef()` snapshot IDs as `int64_t` while resolving refs
- add a regression test for a ref pointing at `INT32_MAX + 42`

## Root Cause

`TableScanBuilder::UseRef()` copied `SnapshotRef::snapshot_id` into an `int32_t`, even though snapshot IDs are modeled as `int64_t` in refs, metadata lookup, and scan context.

## Testing

- `uvx cmake --build build --target scan_test -j 8`
- `./build/src/iceberg/test/scan_test --gtest_filter=*UseRefPreservesInt64SnapshotIds*`
- `./build/src/iceberg/test/scan_test`

Fixes #660

### DIFF
--- a/src/iceberg/table_scan.cc
+++ b/src/iceberg/table_scan.cc
@@ -430,7 +430,7 @@ TableScanBuilder<ScanType>& TableScanBuilder<ScanType>::UseRef(const std::string
   auto iter = metadata_->refs.find(ref);
   ICEBERG_BUILDER_CHECK(iter != metadata_->refs.end(), "Cannot find ref {}", ref);
   ICEBERG_BUILDER_CHECK(iter->second != nullptr, "Ref {} is null", ref);
-  int32_t snapshot_id = iter->second->snapshot_id;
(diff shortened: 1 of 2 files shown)


wgtmac (approved):

PR #662, opened 2026-05-20
feat: support Iceberg v3 unknown type

### LINKED ISSUE #665: Support v3 `unknown` data type
`unknown` is a new data type added in [Iceberg table spec v3](https://iceberg.apache.org/spec/#version-3-extended-types-and-capabilities)

manuzhang:
Closes #665 

---

## Summary
- Add Iceberg v3 unknown primitive type and JSON serialization/deserialization support.
- Support unknown as null-only data across Arrow, Avro, Parquet, schema projection, and nested fields.
- Enforce required-field invariants for unknown/null-only projections and Arrow null imports.

## Validation
- ctest --test-dir build --output-on-failure

---

Co-authored-by: @codex

### DIFF
--- a/src/iceberg/avro/avro_schema_util.cc
+++ b/src/iceberg/avro/avro_schema_util.cc
@@ -766,7 +766,7 @@ Result<FieldProjection> ProjectList(const ListType& list_type,
   FieldProjection element_projection;
   ICEBERG_ASSIGN_OR_RAISE(element_projection,
                           ProjectField(expected_element_field, avro_node->leafAt(0),
-                                       size_t{0}, prune_source));
+                                       /*source_index*/size_t{0}, prune_source));
 
   FieldProjection result;
   result.children.emplace_back(std::move(element_projection));
--- a/src/iceberg/avro/avro_schema_util.cc
+++ b/src/iceberg/avro/avro_schema_util.cc
@@ -237,6 +237,11 @@ Status ToAvroNodeVisitor::Visit(const BinaryType& type, ::avro::NodePtr* node) {
   return {};
 }
 
+Status ToAvroNodeVisitor::Visit(const UnknownType&, ::avro::NodePtr* node) {
+  *node = std::make_shared<::avro::NodePrimitive>(::avro::AVRO_NULL);
+  return {};
+}
+
 Status ToAvroNodeVisitor::Visit(const StructType& type, ::avro::NodePtr* node) {
   *node = std::make_shared<::avro::NodeRecord>();
 
@@ -338,7 +343,7 @@ Status ToAvroNodeVisitor::Visit(const SchemaField& field, ::avro::NodePtr* node)
   field_ids_.push(field.field_id());
   ICEBERG_RETURN_UNEXPECTED(VisitTypeInline(*field.type(), /*visitor=*/this, node));
 
-  if (field.optional()) {
+  if (field.optional() && (*node)->type() != ::avro::AVRO_NULL) {
     ::avro::MultiLeaves union_types;
     union_types.add(std::make_shared<::avro::NodePrimitive>(::avro::AVRO_NULL));
     union_types.add(std::move(*node));
@@ -383,8 +388,8 @@ Status HasIdVisitor::Visit(const ::avro::NodePtr& node) {
     case ::avro::AVRO_STRING:
     case ::avro::AVRO_BYTES:
     case ::avro::AVRO_FIXED:
-      return {};
     case ::avro::AVRO_NULL:
+      return {};
     case ::avro::AVRO_ENUM:
     default:
       return InvalidSchema("Unsupported Avro type: {}", static_cast<int>(node->type()));
@@ -512,6 +517,10 @@ Result<int32_t> GetFieldId(const ::avro::NodePtr& node, size_t field_idx) {
 
 Status ValidateAvroSchemaEvolution(const Type& expected_type,
                                    const ::avro::NodePtr& avro_node) {
+  if (avro_node->type() == ::avro::AVRO_NULL) {
+    return {};
+  }
+
   switch (expected_type.type_id()) {
     case TypeId::kBoolean:
       if (avro_node->type() == ::avro::AVRO_BOOL) {
@@ -615,6 +624,8 @@ Status ValidateAvroSchemaEvolution(const Type& expected_type,
         return {};
       }
       break;
+    case TypeId::kUnknown:
+      return {};
     default:
       break;
   }
@@ -649,6 +660,35 @@ Status UnwrapUnion(const ::avro::NodePtr& node, ::avro::NodePtr* result) {
 Result<FieldProjection> ProjectNested(const Type& expected_type,
                                       const ::avro::NodePtr& avro_node,
                                       bool prune_source);
+
+Result<FieldProjection> ProjectField(const SchemaField& expected_field,
+                                     const ::avro::NodePtr& avro_node,
+                                     size_t source_index, bool prune_source) {
+  const Type& expected_type = *expected_field.type();
+  ::avro::NodePtr field_node;
+  ICEBERG_RETURN_UNEXPECTED(UnwrapUnion(avro_node, &field_node));
+
+  FieldProjection projection;
+  if (expected_type.type_id() == TypeId::kUnknown ||
+      field_node->type() == ::avro::AVRO_NULL) {
+    if (!expected_field.optional()) {
+      return InvalidSchema("Cannot project required field with ID: {} as null",
+                           expected_field.field_id());
+    }
+    projection.kind = FieldProjection::Kind::kNull;
+    return projection;
+  }
+
+  if (expected_type.is_nested()) {
+    ICEBERG_ASSIGN_OR_RAISE(projection,
+                            ProjectNested(expected_type, field_node, prune_source));
+  } else {
+    ICEBERG_RETURN_UNEXPECTED(ValidateAvroSchemaEvolution(expected_type, field_node));
+  }
+  projection.from = source_index;
+  projection.kind = FieldProjection::Kind::kProjected;
+  return projection;
+}
 
 Result<FieldProjection> ProjectStruct(const StructType& struct_type,
                                       const ::avro::NodePtr& avro_node,
@@ -685,18 +725,9 @@ Result<FieldProjection> ProjectStruct(const StructType& struct_type,
     FieldProjection child_projection;
 
     if (auto iter = node_info_map.find(field_id); iter != node_info_map.cend()) {
-      ::avro::NodePtr field_node;
-      ICEBERG_RETURN_UNEXPECTED(UnwrapUnion(iter->second.field_node, &field_node));
-      if (expected_field.type()->is_nested()) {
-        ICEBERG_ASSIGN_OR_RAISE(
-            child_projection,
-            ProjectNested(*expected_field.type(), field_node, prune_source));
-      } else {
-        ICEBERG_RETURN_UNEXPECTED(
-            ValidateAvroSchemaEvolution(*expected_field.type(), field_node));
-      }
-      child_projection.from = iter->second.local_index;
-      child_projection.kind = FieldProjection::Kind::kProjected;
+      ICEBERG_ASSIGN_OR_RAISE(child_projection,
+                              ProjectField(expected_field, iter->second.field_node,
+                                           iter->second.local_index, prune_source));
     } else if (MetadataColumns::IsMetadataColumn(field_id)) {
       child_projection.kind = FieldProjection::Kind::kMetadata;
     } else if (expected_field.optional()) {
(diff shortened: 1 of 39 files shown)


src/iceberg/schema_util.cc
```diff
@@ -79,6 +82,55 @@ Status ValidateSchemaEvolution(const Type& expected_type, const Type& source_typ
...
+    projection.kind = FieldProjection::Kind::kNull;
+    return projection;
+  }
+
+  if (source_field.type()->type_id() == TypeId::kUnknown && !expected_field.optional()) {
+    return InvalidSchema("Cannot project required field with id {} as null",
+                         expected_field.field_id());
+  }
+
+  if (expected_field.type()->is_nested()) {
```
wgtmac, line 112: `unknown` is still a primitive type, so it should not evolve to struct/list/map. This path should fail instead of projecting the whole nested value as null.

manuzhang: Can you point me to the spec or impl where it's not allowed?

src/iceberg/test/update_schema_test.cc
```diff
@@ -1054,6 +1054,36 @@ TEST_F(UpdateSchemaTest, UpdateColumnFloatToDouble) {
   EXPECT_EQ(*field_opt->get().type(), *float64());
 }
 
+TEST_F(UpdateSchemaTest, UpdateColumnUnknownToPrimitiveFails) {
```
wgtmac: Spec v3 allows `unknown` to promote to any primitive type. This should succeed; `unknown -> struct/list/map` should be rejected instead.

src/iceberg/parquet/parquet_writer.cc
```diff
@@ -71,6 +78,64 @@ Result<std::optional<int32_t>> ParseCodecLevel(const WriterProperties& propertie
...
+        if (auto pruned_field = PruneUnknownField(field)) {
+          fields.emplace_back(std::move(pruned_field.value()));
+        }
+      }
+      return std::make_shared<StructType>(std::move(fields));
+    }
+    case TypeId::kList: {
+      const auto& list_type = static_cast<const ListType&>(*type);
+      auto pruned_element = PruneUnknownField(list_type.element());
+      if (!pruned_element.has_value()) {
```
wgtmac: Pruning `list<unknown>` or `map<..., unknown>` drops the whole container, including list lengths or map keys. Java rejects these on write; only top-level/struct unknown fields should be skipped.

wgtmac:
Please rebase on the latest main as well.

src/iceberg/schema_util.cc
wgtmac: https://iceberg.apache.org/spec/#schema-evolution

It says `Valid primitive type promotions are` and `Evolving primitive types to structs is not allowed`.

manuzhang: Ok, I misunderstood your comments as unknown type is not allowed inside a nested type.

wgtmac:
Could you please rebase to resolve conflicts? PR for v3 nano precision timestamp types has been merged.

src/iceberg/test/schema_json_test.cc
```diff
@@ -134,6 +135,87 @@ TEST(SchemaJsonTest, RoundTrip) {
...
+  ASSERT_EQ(field.field_id(), 1);
+  ASSERT_EQ(field.name(), "mystery");
+  ASSERT_EQ(field.type()->type_id(), TypeId::kUnknown);
+  ASSERT_TRUE(field.optional());
+  ASSERT_EQ(ToJson(*schema).dump(), json);
+}
+
+TEST(SchemaJsonTest, NestedUnknownFieldsRoundTrip) {
+  constexpr std::string_view json =
+      R"({"fields":[{"id":1,"name":"profile","required":false,"type":{"fields":[{"id":2,"name":"mystery","required":false,"type":"unknown"}],"type":"struct"}},{"id":3,"name":"mysteries","required":false,"type":{"element":"unknown","element-id":4,"element-required":false,"type":"list"}},{"id":5,"name":"properties","required":false,"type":{"key":"string","key-id":6,"type":"map","value":"unknown","value-id":7,"value-required":false}}],"schema-id":1,"type":"struct"})";
```
zhjwpku: nit: It would be more readable if we made this multi-line.

src/iceberg/test/schema_json_test.cc
```diff
@@ -134,6 +135,87 @@ TEST(SchemaJsonTest, RoundTrip) {
...
+  ASSERT_THAT(schema_result, IsError(ErrorKind::kJsonParseError));
+  ASSERT_THAT(schema_result,
+              HasErrorMessage("Unknown type field 'element' must be optional"));
+}
+
+TEST(SchemaJsonTest, RejectUnknownMapKey) {
+  constexpr std::string_view json =
+      R"({"fields":[{"id":1,"name":"mysteries","required":false,"type":{"key":"unknown","key-id":2,"type":"map","value":"string","value-id":3,"value-required":false}}],"schema-id":1,"type":"struct"})";
+
+  auto schema_result = SchemaFromJson(nlohmann::json::parse(json));
```
zhjwpku: From the spec: Map keys are required and map values can be either optional or required. So I'm wondering should we report something like `Map 'key' can not be unknown type` rather than the current error msg?

src/iceberg/json_serde.cc
```diff
@@ -458,6 +470,10 @@ Result<std::unique_ptr<Type>> MapTypeFromJson(const nlohmann::json& json) {
   ICEBERG_ASSIGN_OR_RAISE(auto value_id, GetJsonValue<int32_t>(json, kValueId));
   ICEBERG_ASSIGN_OR_RAISE(auto value_required, GetJsonValue<bool>(json, kValueRequired));
 
+  ICEBERG_RETURN_UNEXPECTED(
```
zhjwpku: Maybe we can restrict `key_type` can't be `unknown` here, then we don't need to call ValidateUnknownFieldOptional for key_type.

wgtmac:
@manuzhang Do you have time to revive this?

manuzhang:
@wgtmac Yes, I'm on a trip this week and will update a bit later.

src/iceberg/avro/avro_writer.cc
```diff
@@ -81,6 +89,126 @@ Result<std::optional<int32_t>> ParseCodecLevel(const WriterProperties& propertie
...
+  if (field.type()->type_id() == TypeId::kUnknown) {
+    ICEBERG_PRECHECK(context != FieldContext::kMapKey,
+                     "Cannot write map key '{}' of unknown type because it has no "
+                     "physical Avro representation",
+                     field.name());
+    ICEBERG_PRECHECK(field.optional(), "Unknown type field '{}' must be optional",
+                     field.name());
+    if (context == FieldContext::kListElement || context == FieldContext::kMapValue) {
+      return field;
+    }
```
zhjwpku: This is different from the PruneUnknownField in parquet_writer.cc, is it intentional?

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -733,20 +764,9 @@ Result<FieldProjection> ProjectList(const ListType& list_type,
...
-    ICEBERG_RETURN_UNEXPECTED(
-        ValidateAvroSchemaEvolution(*expected_element_field.type(), element_node));
-  }
-
-  // Set the element projection metadata but preserve its children
-  element_projection.kind = FieldProjection::Kind::kProjected;
-  element_projection.from = size_t{0};
+  ICEBERG_ASSIGN_OR_RAISE(element_projection,
+                          ProjectField(expected_element_field, avro_node->leafAt(0),
+                                       size_t{0}, prune_source));
```
zhjwpku: nit:
```suggestion
                                       /*source_index*/size_t{0}, prune_source));
```

src/iceberg/avro/avro_writer.cc
zhjwpku: Sorry, I may have been misleading in my earlier comment. I think your original logic should be correct.

For Parquet, rejecting `list<unknown>` and `map<..., unknown>` makes sense because there is no physical value column to write. Pruning the unknown element or value would lose list cardinality or map keys.

Avro is different. This PR already maps `UnknownType` to `AVRO_NULL`, so `list<unknown>` can be represented as an Avro array with null items, and `map<string, unknown>` as an Avro map with null values. That preserves list lengths and map keys.

This also matches the Java impl: `TypeToSchema` maps `UNKNOWN` to `NULL_SCHEMA`, and the Spark Avro test suite includes coverage for both `testUnknownListType` and `testUnknownMapType`.

cc @wgtmac to double-check my understanding.

zhjwpku (approved):
LGTM, thanks for working on this.

manuzhang:
@wgtmac Do you have further comments?

src/iceberg/schema.cc
```diff
@@ -35,6 +35,45 @@
...
+                       "Map 'key' cannot be unknown type");
+      ICEBERG_PRECHECK(value.optional() || value.type()->type_id() != TypeId::kUnknown,
+                       "Unknown type field '{}' must be optional", value.name());
+      ICEBERG_RETURN_UNEXPECTED(ValidateUnknownFieldsOptional(*key.type()));
+      return ValidateUnknownFieldsOptional(*value.type());
+    }
+    default:
+      return {};
+  }
+}
```
wgtmac:
```suggestion
Status ValidateFieldNullability(const Type& type) {
  auto validate_field = [&](const SchemaField& field) -> Status {
    ICEBERG_PRECHECK(field.optional() || field.type()->type_id() != TypeId::kUnknown,
                     "Unknown type field '{}' must be optional", field.name());
    return ValidateFieldNullability(*field.type());
  };

  switch (type.type_id()) {
    case TypeId::kStruct: {
      const auto& struct_type = static_cast<const StructType&>(type);
      for (const auto& field : struct_type.fields()) {
        ICEBERG_RETURN_UNEXPECTED(validate_field(field));
      }
      return {};
    }
    case TypeId::kList: {
      const auto& list_type = static_cast<const ListType&>(type);
      const auto& element = list_type.element();
      return validate_field(element);
    }
    case TypeId::kMap: {
      const auto& map_type = static_cast<const MapType&>(type);
      const auto& key = map_type.key();
      const auto& value = map_type.value();
      ICEBERG_PRECHECK(key.type()->type_id() != TypeId::kUnknown,
                       "Map 'key' cannot be unknown type");
      ICEBERG_RETURN_UNEXPECTED(ValidateFieldNullability(*key.type()));
      return validate_field(value);
    }
    default:
      return {};
  }
}
```

I'd rather rewriting it as the above. In the next step, we can reuse it for checking v3 default values.

src/iceberg/test/update_schema_test.cc
```diff
@@ -1054,6 +1057,113 @@ TEST_F(UpdateSchemaTest, UpdateColumnFloatToDouble) {
   EXPECT_EQ(*field_opt->get().type(), *float64());
 }
 
+TEST_F(UpdateSchemaTest, UpdateColumnUnknownToPrimitive) {
+  ICEBERG_UNWRAP_OR_FAIL(auto update, table_->NewUpdateSchema());
+  update->AddColumn("mystery", unknown(), "A null-only placeholder");
```
wgtmac, line 1060: Doesn't it require format version v3 to succeed?

src/iceberg/avro/avro_schema_util.cc
```diff
@@ -650,6 +661,35 @@ Result<FieldProjection> ProjectNested(const Type& expected_type,
                                       const ::avro::NodePtr& avro_node,
                                       bool prune_source);
 
+Result<FieldProjection> ProjectField(const SchemaField& expected_field,
```
wgtmac, line 664: Nice refactoring!

src/iceberg/avro/avro_writer.cc
```diff
@@ -81,6 +89,126 @@ Result<std::optional<int32_t>> ParseCodecLevel(const WriterProperties& propertie
...
 
+enum class FieldContext {
+  kTopLevel,
+  kStruct,
+  kListElement,
+  kMapKey,
+  kMapValue,
+};
+
+Result<std::optional<SchemaField>> PruneUnknownField(const SchemaField& field,
```
wgtmac: I don't think it is worth this significant change but only top-level unknown fields are pruned. We should just simplify this by always writing AVRO_NULL for unknown types. Same for parquet writer.

src/iceberg/update/update_schema.h
```diff
@@ -180,23 +180,23 @@ class ICEBERG_EXPORT UpdateSchema : public PendingUpdate {
...
-  /// \param new_type Replacement type for the column (must be primitive).
+  /// \param new_type Replacement type for the column.
   /// \return Reference to this for method chaining.
   /// \note InvalidArgument will be reported if name doesn't identify a column in the
   /// schema or if
   ///       this change introduces a type incompatibility or if it conflicts with
   ///       other additions, renames, or updates.
-  UpdateSchema& UpdateColumn(std::string_view name,
-                             std::shared_ptr<PrimitiveType> new_type);
+  UpdateSchema& UpdateColumn(std::string_view name, std::shared_ptr<Type> new_type);
```
wgtmac: I don't think this is the right direction. As per the spec, unknown type can only be promoted to primitive type. The Java implementation follows this rule so it is consistent.

src/iceberg/json_serde.cc
```diff
@@ -441,12 +443,22 @@ Result<std::unique_ptr<Type>> StructTypeFromJson(const nlohmann::json& json) {
...
+  return {};
+}
+
 Result<std::unique_ptr<Type>> ListTypeFromJson(const nlohmann::json& json) {
   ICEBERG_ASSIGN_OR_RAISE(auto element_type, TypeFromJson(json[kElement]));
   ICEBERG_ASSIGN_OR_RAISE(auto element_id, GetJsonValue<int32_t>(json, kElementId));
   ICEBERG_ASSIGN_OR_RAISE(auto element_required,
                           GetJsonValue<bool>(json, kElementRequired));
 
+  ICEBERG_RETURN_UNEXPECTED(ValidateUnknownFieldOptional(*element_type, !element_required,
```
wgtmac: This check is duplicate since Schema::Validate will cover the same thing.

wgtmac (approved):
I've pushed some changes to simplify this. Thanks @manuzhang for adding this and @zhjwpku for the review!

PR #664, opened 2026-05-20
fix: evaluate ManifestGroup file filters

### LINKED ISSUE #663: ManifestGroup::FilterFiles silently ignores file filters
### Problem

`ManifestGroup::FilterFiles()` accepts a file-level expression and stores it in `file_filter_`, but `ReadEntries()` did not build or run an evaluator against each `DataFile`. As a result, non-true file filters were accepted but silently ignored.

This is primarily a public API correctness issue. Normal table scans still apply data and partition filtering through the existing scan path, but direct use of `ManifestGroup::FilterFiles()` can return entries that should have been filtered out.

### Expected behavior

`ManifestGroup::FilterFiles()` should evaluate supported predicates against `DataFile` metadata before returning entries. To stay aligned with Java `ManifestGroup`, file filters should bind against file metadata with an empty `partition` struct: callers should use `FilterData(...)` for logical data predicates, while concrete `partition.*` file filters should fail explicitly instead of being silently ignored.

### Reproduction idea

Create a manifest with two entries that differ by `record_count`, call `FilterFiles(record_count >= 10)`, and observe that entries below the threshold are still returned.

fallintoplace:
## Summary

Fixes #663.

This implements `ManifestGroup::FilterFiles()` instead of accepting the filter as a silent no-op. The change adds a `DataFile` `StructLike` wrapper, binds a file-level evaluator against file metadata, and applies it to each manifest entry before returning it.

The evaluator schema is aligned with Java `ManifestGroup`: concrete partition fields are not exposed under `partition`, so callers should use `FilterData(...)` for logical data predicates. Supported file metadata includes `spec_id`, and unsupported concrete partition filters fail during binding rather than being ignored.

The reader projection now includes file-filter referenced metadata columns when callers use `Select(...)`. Empty container-backed optional metadata is exposed as null so projected or missing metadata does not look present to the evaluator.

## Tests

- `uvx pre-commit run --files src/iceberg/manifest/manifest_group.cc src/iceberg/manifest/manifest_entry.h src/iceberg/row/manifest_wrapper.cc src/iceberg/row/manifest_wrapper.h src/iceberg/test/manifest_group_test.cc`
- `cmake --build build --target manifest_test -j 8`
- `./build/src/iceberg/test/manifest_test --gtest_filter='ManifestGroupVersions/ManifestGroupTest.FilterFiles*'`
- `./build/src/iceberg/test/manifest_test`

### DIFF
--- a/src/iceberg/manifest/manifest_group.cc
+++ b/src/iceberg/manifest/manifest_group.cc
@@ -370,20 +370,11 @@ ManifestGroup::ReadEntries() {
   const bool has_file_filter =
       file_filter_ && file_filter_->op() != Expression::Operation::kTrue;
   std::unique_ptr<Evaluator> data_file_evaluator;
-  auto get_data_file_evaluator = [&]() -> Result<Evaluator*> {
-    if (!has_file_filter) {
-      return nullptr;
-    }
-    if (data_file_evaluator != nullptr) {
-      return data_file_evaluator.get();
-    }
-
+  if (has_file_filter) {
     ICEBERG_ASSIGN_OR_RAISE(
         data_file_evaluator,
         Evaluator::Make(*DataFileFilterSchema(), file_filter_, case_sensitive_));
-
-    return data_file_evaluator.get();
-  };
+  }
 
   std::unordered_map<int32_t, std::vector<ManifestEntry>> result;
 
@@ -416,7 +407,6 @@ ManifestGroup::ReadEntries() {
     ICEBERG_ASSIGN_OR_RAISE(auto reader, MakeReader(manifest));
     ICEBERG_ASSIGN_OR_RAISE(auto entries,
                             ignore_deleted_ ? reader->LiveEntries() : reader->Entries());
-    ICEBERG_ASSIGN_OR_RAISE(auto data_file_evaluator, get_data_file_evaluator());
 
     for (auto& entry : entries) {
       if (ignore_existing_ && entry.status == ManifestStatus::kExisting) {
--- a/src/iceberg/manifest/manifest_group.cc
+++ b/src/iceberg/manifest/manifest_group.cc
@@ -20,9 +20,11 @@
 #include "iceberg/manifest/manifest_group.h"
 
 #include <algorithm>
+#include <memory>
 #include <string>
 #include <unordered_set>
 #include <utility>
+#include <vector>
 
 #include "iceberg/expression/binder.h"
 #include "iceberg/expression/evaluator.h"
@@ -36,11 +38,43 @@
 #include "iceberg/row/manifest_wrapper.h"
 #include "iceberg/schema.h"
 #include "iceberg/table_scan.h"
+#include "iceberg/type.h"
 #include "iceberg/util/checked_cast.h"
 #include "iceberg/util/content_file_util.h"
 #include "iceberg/util/macros.h"
 
 namespace iceberg {
+
+namespace {
+
+std::shared_ptr<Schema> DataFileFilterSchema() {
+  auto empty_partition_type = std::make_shared<StructType>(std::vector<SchemaField>{});
+  return std::make_shared<Schema>(std::vector<SchemaField>{
+      DataFile::kContent,
+      DataFile::kFilePath,
+      DataFile::kFileFormat,
+      DataFile::kSpecId,
+      SchemaField::MakeRequired(DataFile::kPartitionFieldId, DataFile::kPartitionField,
+                                std::move(empty_partition_type), DataFile::kPartitionDoc),
+      DataFile::kRecordCount,
+      DataFile::kFileSize,
+      DataFile::kColumnSizes,
+      DataFile::kValueCounts,
+      DataFile::kNullValueCounts,
+      DataFile::kNanValueCounts,
+      DataFile::kLowerBounds,
+      DataFile::kUpperBounds,
+      DataFile::kKeyMetadata,
+      DataFile::kSplitOffsets,
+      DataFile::kEqualityIds,
+      DataFile::kSortOrderId,
+      DataFile::kFirstRowId,
+      DataFile::kReferencedDataFile,
+      DataFile::kContentOffset,
+      DataFile::kContentSize});
+}
+
+}  // namespace
 
 Result<std::unique_ptr<ManifestGroup>> ManifestGroup::Make(
     std::shared_ptr<FileIO> io, std::shared_ptr<Schema> schema,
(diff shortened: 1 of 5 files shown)


src/iceberg/test/manifest_group_test.cc
```diff
@@ -404,6 +405,133 @@ TEST_P(ManifestGroupTest, CustomManifestEntriesFilter) {
...
+                MakeDataFile("/path/to/bucket1.parquet", partition_bucket_1,
+                             partitioned_spec_->spec_id()))};
+  auto data_manifest =
+      WriteDataManifest(version, kSnapshotId, std::move(data_entries), partitioned_spec_);
+
+  std::vector<ManifestFile> manifests = {data_manifest};
+  ICEBERG_UNWRAP_OR_FAIL(
+      auto group,
+      ManifestGroup::Make(file_io_, schema_, GetSpecsById(), std::move(manifests)));
+  group->FilterFiles(Expressions::Equal("partition.data_bucket_16_2", Literal::Int(1)));
```
wgtmac, line 455: I'm afraid that partition field name like `data_bucket_16_2` is not spec-compliant but implementation-specific.

Per the spec from https://iceberg.apache.org/spec/#partitioning, users are discouraged to directly write partition filters: `the partitioning can change and the correct partition filters are always derived from column predicates`.

Java `ManifestGroup` builds the file evaluator with `DataFile.getType(EMPTY_STRUCT)`, so it does not allow binding concrete partition fields under `partition`. For logical data predicates, callers should use `FilterData(...)`, letting projection derive the appropriate partition/manifest filters for each spec. I'd recommend to be aligned with the Java approach at this moment.

src/iceberg/row/manifest_wrapper.cc
```diff
@@ -134,4 +221,63 @@ std::unique_ptr<StructLike> FromManifestFile(const ManifestFile& file) {
   return std::make_unique<ManifestFileStructLike>(file);
 }
 
+Result<Scalar> DataFileStructLike::GetField(size_t pos) const {
+  if (pos >= num_fields()) {
+    return InvalidArgument("Invalid data file field index: {}", pos);
+  }
+
+  const auto& data_file = data_file_.get();
+  switch (static_cast<DataFileFieldPosition>(pos)) {
```
wgtmac, line 255: Should we expose `partition_spec_id` field as an optional field? FYI Java puts it at pos 3 (zero-based).

src/iceberg/row/manifest_wrapper.cc
```diff
@@ -134,4 +221,63 @@ std::unique_ptr<StructLike> FromManifestFile(const ManifestFile& file) {
...
+    case DataFileFieldPosition::kNullValueCounts:
+      return std::make_shared<IntMapLike<int64_t>>(data_file.null_value_counts);
+    case DataFileFieldPosition::kNanValueCounts:
+      return std::make_shared<IntMapLike<int64_t>>(data_file.nan_value_counts);
+    case DataFileFieldPosition::kLowerBounds:
+      return std::make_shared<IntMapLike<std::vector<uint8_t>>>(data_file.lower_bounds);
+    case DataFileFieldPosition::kUpperBounds:
+      return std::make_shared<IntMapLike<std::vector<uint8_t>>>(data_file.upper_bounds);
+    case DataFileFieldPosition::kKeyMetadata:
+      return ToView(data_file.key_metadata);
```
wgtmac: Do we want to return std::monostate if any std container is empty? The main point is that we do not use `std::optional` to represent a missing value for these fields. Besides, fields of `DataFileStructLike` might be missing when ManifestGroup has specified projection.

src/iceberg/row/manifest_wrapper.cc
```diff
@@ -39,6 +76,56 @@ Result<Scalar> FromOptional(const std::optional<T>& value) {
...
+    if (pos >= size()) {
+      return InvalidArgument("Invalid array index: {}", pos);
+    }
+    return ToScalar(values_.get()[pos]);
+  }
+
+  size_t size() const override { return values_.get().size(); }
+
+ private:
+  std::reference_wrapper<const std::vector<T>> values_;
```
wgtmac: Is it better to use `std::span<const T>` here?

src/iceberg/manifest/manifest_group.cc
```diff
@@ -265,10 +270,45 @@ Result<std::unique_ptr<ManifestReader>> ManifestGroup::MakeReader(
   ICEBERG_ASSIGN_OR_RAISE(auto reader,
                           ManifestReader::Make(manifest, io_, schema_, specs_by_id_));
 
+  auto columns = columns_;
+  if (file_filter_ && file_filter_->op() != Expression::Operation::kTrue &&
+      !columns.empty() &&
+      std::ranges::find(columns, Schema::kAllColumns) == columns.end()) {
+    auto spec_iter = specs_by_id_.find(manifest.partition_spec_id);
```
wgtmac: As my other comment, `partition_spec_id` varies among manifest files so let's remove its support for now.

fallintoplace:
Updated after the review feedback:

- bind file filters against `DataFile` metadata with an empty `partition` struct, matching Java `ManifestGroup`
- make concrete `partition.*` file filters fail during binding instead of exposing implementation-specific partition field names
- expose `spec_id` through `DataFileStructLike` for file-level filters
- return null for empty container-backed optional metadata
- use `std::span` in the vector-backed array wrapper
- refresh regression tests for record count, selected metadata columns, `spec_id`, and rejected partition metadata filters

Local validation passed with pre-commit, the focused `FilterFiles*` manifest tests, and the full `manifest_test` suite.

src/iceberg/test/manifest_group_test.cc
```diff
@@ -404,6 +405,123 @@ TEST_P(ManifestGroupTest, CustomManifestEntriesFilter) {
...
+
+  auto result = group->Entries();
+  EXPECT_THAT(result, IsError(ErrorKind::kInvalidExpression));
+  EXPECT_THAT(result, HasErrorMessage("Cannot find field 'partition.data_bucket_16_2'"));
```
wgtmac, line 459: Why we don't skip unknown predicates? Usually we just ignore unknown or malformed predicates in practice (e.g. assume they are always evaluated to match), otherwise any malformed predicate can prohibit any table scan to proceed.

src/iceberg/test/manifest_group_test.cc
```diff
@@ -404,6 +405,123 @@ TEST_P(ManifestGroupTest, CustomManifestEntriesFilter) {
...
+                MakeDataFile("/path/to/large.parquet", part_value,
+                             partitioned_spec_->spec_id(), /*record_count=*/15))};
+  auto data_manifest =
+      WriteDataManifest(version, kSnapshotId, std::move(data_entries), partitioned_spec_);
+
+  std::vector<ManifestFile> manifests = {data_manifest};
+  ICEBERG_UNWRAP_OR_FAIL(
+      auto group,
+      ManifestGroup::Make(file_io_, schema_, GetSpecsById(), std::move(manifests)));
+  group->FilterFiles(Expressions::GreaterThanOrEqual("record_count", Literal::Long(10)));
```
wgtmac, line 431: Does case sensitivity matter here? For example what if users set case-inventive and provide `RECORD_COUNT` as the column name?

wgtmac (commented):
Thanks for improving this! I have some questions with regards to the test.

src/iceberg/test/manifest_group_test.cc
fallintoplace: I checked the Java behavior before changing this. Java `ManifestGroup` builds the file evaluator with `DataFile.getType(EMPTY_STRUCT)`, and `Evaluator` binds the expression eagerly, so unknown fields fail binding rather than being treated as match-all.

I kept the C++ behavior aligned with that. In this path, silently ignoring malformed or unknown `FilterFiles(...)` predicates would broaden the scan and hide caller bugs, which felt riskier than failing fast.

For logical predicates on table columns, callers should use `FilterData(...)`, which can then be projected into the appropriate partition/manifest filters per spec.

wgtmac (approved):
Thank you for adding this!

PR #669, opened 2026-05-22
feat(update): add parallel deletion for ExpireSnapshots cleanup.

### LINKED ISSUE #668: Add parallel file deletion support for ExpireSnapshots cleanup
## Background

`ExpireSnapshots` may delete many files during cleanup, including data files, manifest files, manifest list files, and statistics files.

Currently, `ExpireSnapshots` already uses `FileIO ::DeleteFiles` for grouped deletion. However, the cleanup path itself still processes deletion groups sequentially. For large tables with many expired files, this may make snapshot expiration cleanup slower than necessary.

## Proposal

Add parallel file deletion support inside the `ExpireSnapshots` cleanup strategy.

The initial implementation can be intentionally scoped and conservative:

- Keep the public `ExpireSnapshots` API unchanged.
- Use an internal default parallelism value.
- Apply parallelism only inside the cleanup deletion path.
- Preserve existing best-effort cleanup semantics.
- Continue supporting custom `DeleteWith(...)` callbacks.

slfan1989:
## Summary

This PR adds parallel file deletion support to `ExpireSnapshots` cleanup.

The change is scoped to the internal cleanup path only:
- Keeps the public `ExpireSnapshots` API unchanged.
- Uses an internal default delete parallelism.
- Splits cleanup files into chunks and deletes them concurrently.
- Preserves the existing best-effort cleanup behavior.
- Keeps `DeleteWith(...)` supported for custom deletion logic.

Fixed: #668 

## Motivation

`ExpireSnapshots` may need to delete many files during cleanup, including data files, manifest files, manifest list files, and statistics files. Deleting these files sequentially can slow down snapshot expiration for large tables.

Parallelizing the cleanup deletion path improves throughput without changing the default behavior of `FileIO::DeleteFiles`.

### DIFF
--- a/src/iceberg/update/expire_snapshots.cc
+++ b/src/iceberg/update/expire_snapshots.cc
@@ -24,6 +24,7 @@
 #include <iterator>
 #include <memory>
 #include <string>
+#include <thread>
 #include <tuple>
 #include <unordered_set>
 #include <utility>
@@ -47,6 +48,8 @@
 namespace iceberg {
 
 namespace {
+
+constexpr size_t kDefaultDeleteParallelism = 4;
 
 Result<std::unique_ptr<ManifestReader>> MakeManifestReader(
     const ManifestFile& manifest, const std::shared_ptr<FileIO>& file_io,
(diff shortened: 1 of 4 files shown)


src/iceberg/update/expire_snapshots.cc
```diff
@@ -100,13 +103,39 @@ class FileCleanupStrategy {
...
+    }
+
+    std::vector<std::string> path_list(paths.begin(), paths.end());
+    auto parallelism = std::min(kDefaultDeleteParallelism, path_list.size());
+    if (parallelism <= 1) {
+      DeleteFileRange(path_list, 0, path_list.size());
+      return;
+    }
+
+    std::vector<std::thread> workers;
```
wgtmac, line 117: I think a better approach is to introduce a general purpose thread pool, which is assigned to @HuaHuaY as per 0.3.0 planning.

slfan1989: Thanks for the review and guidance. I agree that adding `std::thread` directly inside `ExpireSnapshots` is too local, and using the planned general-purpose thread pool would be a better direction.

I'll pause this PR for now and revisit the ExpireSnapshots parallel cleanup after the thread pool infrastructure lands.

PR #670, opened 2026-05-22
fix(ci): fix macOS ci to use provided fmt

slfan1989:
## Summary

Fix the S3 build on macOS 26 / AppleClang by disabling fmt's consteval format-string checks for vendored avro-cpp.

## Problem

The S3 CI job on AArch64 macOS 26 fails while compiling vendored avro-cpp:

```text
fmt/format-inl.h: error: call to consteval function ... is not a constant expression
```

https://github.com/apache/iceberg-cpp/actions/runs/26262392739/job/77298497335?pr=669

### DIFF
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -248,6 +248,11 @@ function(resolve_avro_dependency)
     set(AVRO_VENDORED TRUE)
     set_target_properties(avrocpp_s PROPERTIES OUTPUT_NAME "iceberg_vendored_avrocpp")
     set_target_properties(avrocpp_s PROPERTIES POSITION_INDEPENDENT_CODE ON)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+      # Work around AppleClang 21 rejecting fmt's consteval format-string checks
+      # while compiling vendored avro-cpp in C++23 mode.
+      target_compile_definitions(avrocpp_s PUBLIC FMT_USE_CONSTEVAL=0)
+    endif()
     install(TARGETS avrocpp_s
             EXPORT iceberg_targets
             RUNTIME DESTINATION "${ICEBERG_INSTALL_BINDIR}"
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -251,7 +251,7 @@ function(resolve_avro_dependency)
     if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
       # Work around AppleClang 21 rejecting fmt's consteval format-string checks
       # while compiling vendored avro-cpp in C++23 mode.
-      target_compile_definitions(avrocpp_s PRIVATE FMT_CONSTEVAL=)
+      target_compile_definitions(avrocpp_s PUBLIC FMT_CONSTEVAL=)
     endif()
     install(TARGETS avrocpp_s
             EXPORT iceberg_targets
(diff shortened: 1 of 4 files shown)


cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -248,6 +248,11 @@ function(resolve_avro_dependency)
     set(AVRO_VENDORED TRUE)
     set_target_properties(avrocpp_s PROPERTIES OUTPUT_NAME "iceberg_vendored_avrocpp")
     set_target_properties(avrocpp_s PROPERTIES POSITION_INDEPENDENT_CODE ON)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+      # Work around AppleClang 21 rejecting fmt's consteval format-string checks
+      # while compiling vendored avro-cpp in C++23 mode.
+      target_compile_definitions(avrocpp_s PUBLIC FMT_CONSTEVAL=)
```
zhjwpku: Can you try PRIVATE, so that `FMT_CONSTEVAL=` won't propagate. Not sure if that work :)

slfan1989: That makes sense. The current failure happens while compiling avrocpp_s itself, so PRIVATE may be sufficient and would avoid leaking the workaround to downstream targets. I’ll try switching it to PRIVATE and let CI verify whether Iceberg sources still compile without inheriting FMT_CONSTEVAL=.

zhjwpku: It seems PRIVATE won't work :(

slfan1989: I changed it back to PUBLIC.

HuaHuaY:
I think a simpler fix would be to temporarily introduce a higher version of `fmt` in CI environment or CMake. Latest `fmt` has fixed this issue https://github.com/fmtlib/fmt/issues/4396, but `avro` will fetch an outdated `fmt` if it doesn't find a `fmt` package. There is a PR https://github.com/apache/avro/pull/3788 in `avro` which tries to replace `fmt` with `std::format`. We can remove the `fmt` after `avro` accepts the PR.
```cmake
// lang/c++/CMakeLists.txt
find_package(fmt QUIET)
if (NOT fmt_FOUND)
    include(FetchContent)
    FetchContent_Declare(
            fmt
            GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
            GIT_TAG         10.2.1
            GIT_PROGRESS    TRUE
            USES_TERMINAL_DOWNLOAD TRUE
    )
    FetchContent_MakeAvailable(fmt)
endif ()
```

wgtmac:
> I think a simpler fix would be to temporarily introduce a higher version of `fmt` in CI environment or CMake. Latest `fmt` has fixed this issue [fmtlib/fmt#4396](https://github.com/fmtlib/fmt/issues/4396), but `avro` will fetch an outdated `fmt` if it doesn't find a `fmt` package. There is a PR in `avro` which tries to replace `fmt` with `std::format`. We can remove the `fmt` after `avro` accepts the PR.
> 
> ```cmake
> // lang/c++/CMakeLists.txt
> find_package(fmt QUIET)
> if (NOT fmt_FOUND)
>     include(FetchContent)
>     FetchContent_Declare(
>             fmt
>             GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
>             GIT_TAG         10.2.1
>             GIT_PROGRESS    TRUE
>             USES_TERMINAL_DOWNLOAD TRUE
>     )
>     FetchContent_MakeAvailable(fmt)
> endif ()
> ```

I think this is a better approach. We can see that homebrew is available on the macos-26 image: https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md. Homebrew has fmt 12.1.0 as the stable version: https://formulae.brew.sh/formula/fmt. We can add a step to call `brew install fmt` in the macOS related CI.

wgtmac (approved):

wgtmac:
Thank you for fixing it, @slfan1989!

slfan1989:
> Thank you for fixing it, @slfan1989!

@wgtmac @zhjwpku @HuaHuaY Thanks for the review and suggestion!

PR #674, opened 2026-05-22
fix: upgrade fmt to 11.1.4 to fix avro-cpp build on macOS 26

lishuxu:
fix: upgrade fmt to 11.1.4 to fix avro-cpp build on macOS 26


wgtmac:
This duplicates https://github.com/apache/iceberg-cpp/pull/670

lishuxu:
Yes, I didn't notice #670 until after I pushed — the two commits were very close in time. Closing in favor of #670.

PR #675, opened 2026-05-23
refactor: move temporal utilities out of transform util

zhjwpku:
Also add internal math helpers for floor division and checked multiplication. Keep the Human* formatting helpers in TransformUtil to stay consistent with the Java TransformUtil implementation.


src/iceberg/util/math_util_internal.h
```diff
@@ -0,0 +1,47 @@
...
+
+#pragma once
+
+#include <cstdint>
+#include <limits>
+
+#include "iceberg/result.h"
+#include "iceberg/util/int128.h"
+
+namespace iceberg::internal {
```
wgtmac: nit: it does not necessarily to be in the `internal` namespace.

wgtmac (approved):

src/iceberg/util/math_util_internal.h
zhjwpku: Let me fix this.

PR #682, opened 2026-05-26
feat: add MergingSnapshotUpdate

gty404:
Abstract base for merge-based snapshot operations (MergeAppend, OverwriteFiles, RowDelta, etc.), implementing the filter → write → merge pipeline consistent with Java's MergingSnapshotProducer.

Also fixes SnapshotSummaryBuilder manifest count fields and a use-after-free bug in SnapshotUpdate::DeleteFile.

### DIFF
--- a/src/iceberg/manifest/manifest_filter_manager.cc
+++ b/src/iceberg/manifest/manifest_filter_manager.cc
@@ -94,7 +94,6 @@ void ManifestFilterManager::DeleteFile(std::string_view path) {
 Status ManifestFilterManager::DeleteFile(std::shared_ptr<DataFile> file) {
   ICEBERG_PRECHECK(file != nullptr, "Cannot delete file: null");
   delete_paths_.insert(file->file_path);
-  delete_files_.insert(std::move(file));
   return {};
 }
 
@@ -132,9 +131,7 @@ Result<bool> ManifestFilterManager::CanContainDroppedFiles(const ManifestFile&) 
   // manifests once object-delete partitions are tracked separately.
   // Currently, DeleteFile(std::shared_ptr<DataFile>) degrades to a path-based delete,
   // which forces scanning all manifests.
-  // Also open delete manifests when a minimum sequence number is set for cleanup.
-  return !delete_paths_.empty() || !removed_data_file_paths_.empty() ||
-         (manifest_content_ == ManifestContent::kDeletes && min_sequence_number_ > 0);
+  return !delete_paths_.empty() || !removed_data_file_paths_.empty();
 }
 
 Result<bool> ManifestFilterManager::CanContainDroppedPartitions(
@@ -385,6 +382,16 @@ Status ManifestFilterManager::ValidateRequiredDeletes(
 Result<std::vector<ManifestFile>> ManifestFilterManager::FilterManifests(
     const TableMetadata& metadata, const std::shared_ptr<Snapshot>& base_snapshot,
     const ManifestWriterFactory& writer_factory) {
+  ICEBERG_ASSIGN_OR_RAISE(auto schema, metadata.Schema());
+  return FilterManifests(schema, metadata, base_snapshot, writer_factory);
+}
+
+Result<std::vector<ManifestFile>> ManifestFilterManager::FilterManifests(
+    const std::shared_ptr<Schema>& schema, const TableMetadata& metadata,
+    const std::shared_ptr<Snapshot>& base_snapshot,
+    const ManifestWriterFactory& writer_factory) {
+  delete_files_.clear();
+  replaced_manifests_count_ = 0;
   if (!base_snapshot) {
     ICEBERG_RETURN_UNEXPECTED(ValidateRequiredDeletes({}));
     return std::vector<ManifestFile>{};
@@ -402,7 +409,6 @@ Result<std::vector<ManifestFile>> ManifestFilterManager::FilterManifests(
     manifests.push_back(&manifest);
   }
 
-  ICEBERG_ASSIGN_OR_RAISE(auto schema, metadata.Schema());
   TableMetadataCache metadata_cache(&metadata);
   ICEBERG_ASSIGN_OR_RAISE(auto specs_by_id, metadata_cache.GetPartitionSpecsById());
 
@@ -426,7 +432,9 @@ Result<std::vector<ManifestFile>> ManifestFilterManager::FilterManifests(
   }
 
   std::unordered_set<std::string> found_paths;
+  delete_files_.clear();
   if (manifests.empty()) {
+    replaced_manifests_count_ = 0;
     ICEBERG_RETURN_UNEXPECTED(ValidateRequiredDeletes(found_paths));
     return std::vector<ManifestFile>{};
   }
--- a/src/iceberg/manifest/manifest_filter_manager.cc
+++ b/src/iceberg/manifest/manifest_filter_manager.cc
@@ -20,6 +20,7 @@
 #include "iceberg/manifest/manifest_filter_manager.h"
 
 #include <string>
+#include <tuple>
 #include <unordered_set>
 #include <vector>
 
@@ -29,6 +30,7 @@
 #include "iceberg/expression/manifest_evaluator.h"
 #include "iceberg/expression/residual_evaluator.h"
 #include "iceberg/expression/strict_metrics_evaluator.h"
+#include "iceberg/file_io.h"
 #include "iceberg/manifest/manifest_entry.h"
 #include "iceberg/manifest/manifest_list.h"
 #include "iceberg/manifest/manifest_reader.h"
@@ -63,37 +65,76 @@ Result<std::string> FormatPartitionPath(const PartitionSpecsById& specs_by_id,
   return spec->PartitionPath(file.partition);
 }
 
+void AddDeletedFileToManager(ManifestContent manifest_content, DataFileSet& data_files,
+                             DeleteFileSet& delete_files,
+                             std::vector<std::shared_ptr<DataFile>>& deleted_files,
+                             DataFileSet& deleted_data_file_set,
+                             DeleteFileSet& deleted_file_set,
+                             const std::shared_ptr<DataFile>& file) {
+  if (file == nullptr) {
+    return;
+  }
+
+  bool inserted;
+  if (manifest_content == ManifestContent::kData) {
+    data_files.insert(file);
+    inserted = deleted_data_file_set.insert(file).second;
+  } else {
+    delete_files.insert(file);
+    inserted = deleted_file_set.insert(file).second;
+  }
+  if (inserted) {
+    deleted_files.push_back(file);
+  }
+}
+
+bool AddDeletedFileToManifest(ManifestContent manifest_content,
+                              std::vector<std::shared_ptr<DataFile>>& deleted_files,
+                              DataFileSet& deleted_data_file_set,
+                              DeleteFileSet& deleted_file_set,
+                              const std::shared_ptr<DataFile>& file) {
+  if (file == nullptr) {
+    return false;
+  }
+  bool inserted = manifest_content == ManifestContent::kData
+                      ? deleted_data_file_set.insert(file).second
+                      : deleted_file_set.insert(file).second;
+  if (inserted) {
+    deleted_files.push_back(file);
+  }
+  return inserted;
+}
+
 }  // namespace
 
-size_t ManifestFilterManager::DeleteFileKeyHash::operator()(
-    const DeleteFileKey& key) const {
-  size_t hash = std::hash<std::string>{}(key.path);
-  auto combine = [&hash](const auto& value) {
-    size_t value_hash = value.has_value() ? std::hash<int64_t>{}(*value) : 0;
-    hash ^= value_hash + 0x9e3779b9 + (hash << 6) + (hash >> 2);
-  };
-  combine(key.content_offset);
-  combine(key.content_size_in_bytes);
-  return hash;
+Result<std::unique_ptr<ManifestFilterManager>> ManifestFilterManager::Make(
+    ManifestContent content, std::shared_ptr<FileIO> file_io,
+    std::function<Status(const std::string&)> delete_file) {
+  ICEBERG_PRECHECK(file_io != nullptr, "FileIO cannot be null");
+  return std::unique_ptr<ManifestFilterManager>(
+      new ManifestFilterManager(content, std::move(file_io), std::move(delete_file)));
 }
 
-ManifestFilterManager::DeleteFileKey ManifestFilterManager::MakeDeleteFileKey(
-    const DataFile& file) {
-  return DeleteFileKey{.path = file.file_path,
-                       .content_offset = file.content_offset,
-                       .content_size_in_bytes = file.content_size_in_bytes};
-}
-
-ManifestFilterManager::ManifestFilterManager(ManifestContent content,
-                                             std::shared_ptr<FileIO> file_io)
+ManifestFilterManager::ManifestFilterManager(
+    ManifestContent content, std::shared_ptr<FileIO> file_io,
+    std::function<Status(const std::string&)> delete_file)
     : manifest_content_(content),
       file_io_(std::move(file_io)),
-      delete_expr_(Expressions::AlwaysFalse()) {}
+      delete_file_(std::move(delete_file)),
+      delete_expr_(Expressions::AlwaysFalse()) {
+  ICEBERG_DCHECK(file_io_, "FileIO cannot be null");
+  if (delete_file_ == nullptr) {
+    delete_file_ = [this](const std::string& location) {
+      return file_io_->DeleteFile(location);
+    };
+  }
+}
 
 ManifestFilterManager::~ManifestFilterManager() = default;
 
 Status ManifestFilterManager::DeleteByRowFilter(std::shared_ptr<Expression> expr) {
   ICEBERG_PRECHECK(expr != nullptr, "Cannot delete files using filter: null");
+  ICEBERG_RETURN_UNEXPECTED(InvalidateFilteredCache());
   ICEBERG_ASSIGN_OR_RAISE(delete_expr_, Or::MakeFolded(delete_expr_, std::move(expr)));
   manifest_evaluator_cache_.clear();
   residual_evaluator_cache_.clear();
@@ -106,27 +147,58 @@ void ManifestFilterManager::CaseSensitive(bool case_sensitive) {
   residual_evaluator_cache_.clear();
 }
 
-void ManifestFilterManager::DeleteFile(std::string_view path) {
+Status ManifestFilterManager::DeleteFile(std::string_view path) {
+  ICEBERG_RETURN_UNEXPECTED(InvalidateFilteredCache());
   delete_paths_.insert(std::string(path));
+  return {};
 }
 
 Status ManifestFilterManager::DeleteFile(std::shared_ptr<DataFile> file) {
   ICEBERG_PRECHECK(file != nullptr, "Cannot delete file: null");
-  delete_file_keys_.insert(MakeDeleteFileKey(*file));
+  ICEBERG_RETURN_UNEXPECTED(InvalidateFilteredCache());
+  if (manifest_content_ == ManifestContent::kData) {
+    data_files_.insert(file);
+    data_files_to_delete_.insert(std::move(file));
+  } else {
+    delete_files_.insert(file);
+    delete_files_to_delete_.insert(std::move(file));
+  }
   return {};
 }
 
-const DataFileSet& ManifestFilterManager::FilesToBeDeleted() const {
-  return delete_files_;
-}
+const DataFileSet& ManifestFilterManager::FilesToBeDeleted() const { return data_files_; }
 
 const std::vector<std::shared_ptr<DataFile>>& ManifestFilterManager::DeletedFiles()
     const {
   return deleted_files_;
 }
 
-void ManifestFilterManager::DropPartition(int32_t spec_id, PartitionValues partition) {
+Result<SnapshotSummaryBuilder> ManifestFilterManager::BuildSummary(
+    const std::vector<ManifestFile>& manifests,
+    const PartitionSpecsById& specs_by_id) const {
+  SnapshotSummaryBuilder summary;
+  for (const auto& manifest : manifests) {
+    auto deleted_iter = filtered_manifest_to_deleted_files_.find(manifest);
+    if (deleted_iter == filtered_manifest_to_deleted_files_.end()) {
+      continue;
+    }
+
+    ICEBERG_ASSIGN_OR_RAISE(auto spec,
+                            PartitionSpecById(specs_by_id, manifest.partition_spec_id));
+    for (const auto& file : deleted_iter->second.files) {
+      if (file != nullptr) {
+        ICEBERG_RETURN_UNEXPECTED(summary.DeletedFile(*spec, *file));
+      }
+    }
+  }
+  summary.IncrementDuplicateDeletes(duplicate_deletes_count_);
+  return summary;
+}
+
+Status ManifestFilterManager::DropPartition(int32_t spec_id, PartitionValues partition) {
+  ICEBERG_RETURN_UNEXPECTED(InvalidateFilteredCache());
   drop_partitions_.add(spec_id, std::move(partition));
+  return {};
 }
 
 void ManifestFilterManager::FailMissingDeletePaths() {
@@ -137,24 +209,32 @@ void ManifestFilterManager::FailAnyDelete() { fail_any_delete_ = true; }
 
 bool ManifestFilterManager::ContainsDeletes() const {
   return HasRowFilterExpression(delete_expr_) || !delete_paths_.empty() ||
-         !delete_file_keys_.empty() || !drop_partitions_.empty();
+         !data_files_to_delete_.empty() || !delete_files_to_delete_.empty() ||
+         !drop_partitions_.empty();
 }
 
-void ManifestFilterManager::DropDeleteFilesOlderThan(int64_t sequence_number) {
+Status ManifestFilterManager::DropDeleteFilesOlderThan(int64_t sequence_number) {
+  ICEBERG_PRECHECK(sequence_number >= 0, "Invalid minimum data sequence number: {}",
+                   sequence_number);
   min_sequence_number_ = sequence_number;
+  return {};
 }
 
 void ManifestFilterManager::RemoveDanglingDeletesFor(const DataFileSet& deleted_files) {
+  std::unordered_set<std::string> removed_data_file_paths;
   for (const auto& file : deleted_files) {
-    removed_data_file_paths_.insert(file->file_path);
+    if (file != nullptr) {
+      removed_data_file_paths.insert(file->file_path);
+    }
   }
+  removed_data_file_paths_ = std::move(removed_data_file_paths);
 }
 
 Result<bool> ManifestFilterManager::CanContainDroppedFiles(const ManifestFile&) const {
-  // TODO(Guotao): Use the manifest descriptor to skip unrelated object-delete
-  // manifests once object-delete partitions are tracked separately.
-  return !delete_paths_.empty() || !delete_file_keys_.empty() ||
-         !removed_data_file_paths_.empty();
+  // TODO(Guotao): prune object deletes by partition once manifest partition
+  // summary checks are available.
+  return !delete_paths_.empty() || !data_files_to_delete_.empty() ||
+         !delete_files_to_delete_.empty() || !removed_data_file_paths_.empty();
 }
 
 Result<bool> ManifestFilterManager::CanContainDroppedPartitions(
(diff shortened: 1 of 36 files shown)


src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,870 @@
...
+  }
+  if (manifest.has_existing_files()) {
+    return InvalidArgument("Cannot append manifest with existing files: {}",
+                           manifest.manifest_path);
+  }
+  if (manifest.has_deleted_files()) {
+    return InvalidArgument("Cannot append manifest with deleted files: {}",
+                           manifest.manifest_path);
+  }
+  if (manifest.added_snapshot_id != kInvalidSnapshotId) {
```
wgtmac: Java copies append manifests that already have a snapshot id (or assigned first_row_id) instead of rejecting them. This should go through CopyManifest for parity.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,870 @@
...
+  summary_builder().Clear();
+  summary_builder().Merge(added_data_files_summary_);
+  summary_builder().Merge(added_delete_files_summary_);
+  summary_builder().Merge(appended_manifests_summary_);
+
+  auto tracked_factory = MakeTrackedWriterFactory();
+
+  // Step 1: Filter data manifests.
+  ICEBERG_ASSIGN_OR_RAISE(auto filtered_data,
+                          data_filter_manager_.FilterManifests(
```
wgtmac: Java uses SnapshotUtil.schemaFor(base, targetBranch()) here. Using metadata.Schema() underneath can evaluate filters with the wrong schema when committing to a branch behind current.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,870 @@
...
+                              metadata_to_update, snapshot, tracked_factory));
+
+  // Track deleted data files in the summary builder.
+  for (const auto& file : data_filter_manager_.FilesToBeDeleted()) {
```
wgtmac: This counts requested deletes, not actual filtered deletes. Java builds the summary from deleted entries in filtered manifests, so missing optional deletes should not affect totals.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,870 @@
...
+Status MergingSnapshotUpdate::ValidateAddedDataFiles(
+    const TableMetadata& metadata, int64_t starting_snapshot_id,
+    std::shared_ptr<Expression> filter, const std::shared_ptr<Snapshot>& parent,
+    std::shared_ptr<FileIO> io, bool case_sensitive) {
+  if (parent == nullptr) {
+    return {};
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto schema, metadata.Schema());
+  ICEBERG_ASSIGN_OR_RAISE(auto ancestors,
```
wgtmac: Java validationHistory fails if it cannot reach startingSnapshotId. These helpers should not silently validate a truncated ancestor range.

src/iceberg/manifest/manifest_filter_manager.cc
```diff
@@ -117,12 +117,24 @@ bool ManifestFilterManager::ContainsDeletes() const {
...
+
 Result<bool> ManifestFilterManager::CanContainDroppedFiles(const ManifestFile&) const {
   // TODO(Guotao): Use the manifest descriptor to skip unrelated object-delete
   // manifests once object-delete partitions are tracked separately.
   // Currently, DeleteFile(std::shared_ptr<DataFile>) degrades to a path-based delete,
   // which forces scanning all manifests.
-  return !delete_paths_.empty();
+  // Also open delete manifests when a minimum sequence number is set for cleanup.
+  return !delete_paths_.empty() || !removed_data_file_paths_.empty() ||
+         (manifest_content_ == ManifestContent::kDeletes && min_sequence_number_ > 0);
```
wgtmac: Java does not open delete manifests solely because minSequenceNumber is set; old deletes are dropped only when a delete manifest is already being filtered.

src/iceberg/manifest/manifest_merge_manager.cc
```diff
@@ -140,6 +142,8 @@ Result<std::vector<ManifestFile>> ManifestMergeManager::MergeGroup(
     } else {
       ICEBERG_ASSIGN_OR_RAISE(
           auto merged, FlushBin(bin, snapshot_id, metadata, file_io, writer_factory));
+      // Each manifest consumed into the merged output (beyond the 1 output) is replaced.
+      replaced_manifests_count_ += static_cast<int32_t>(bin.size()) - 1;
```
wgtmac: Java counts each previous-snapshot manifest consumed by the merge as replaced, not bin.size() - 1. This can undercount old-only bins and overcount current-snapshot inputs.

src/iceberg/update/merging_snapshot_update.cc
gty404: Good catch. I'll stop rejecting manifests that already carry a snapshot id or first_row_id and route them through CopyManifest so the behavior matches Java

src/iceberg/update/merging_snapshot_update.cc
gty404: Agreed. I'll switch this path to use the schema for the target branch instead of the current table schema so filter evaluation stays correct on non-current branches.

src/iceberg/update/merging_snapshot_update.cc
gty404: That makes sense. I'll change the summary accounting to use the files actually deleted in the filtered manifests rather than the requested delete set.

src/iceberg/update/merging_snapshot_update.cc
gty404: Agreed. I'll make these validation paths fail when the starting snapshot is not reachable instead of silently validating a truncated ancestry range.

src/iceberg/manifest/manifest_filter_manager.cc
gty404: You're right. I'll remove the eager scan triggered solely by minSequenceNumber and only drop old deletes when a delete manifest is already being rewritten for filtering.

src/iceberg/manifest/manifest_merge_manager.cc
gty404: Good point. I'll update the replaced-manifest accounting to count the previous-snapshot manifests consumed by the merge instead of using bin.size() - 1.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1026 @@
...
+  return {};
+}
+
+Status MergingSnapshotUpdate::ValidateNewDeleteFile(const DataFile& file) {
+  if (file.content == DataFile::Content::kData) {
+    return InvalidArgument("Expected a delete file but got a data file: {}",
+                           file.file_path);
+  }
+  const int8_t format_version = base().format_version;
+  const bool is_dv = file.referenced_data_file.has_value();
```
wgtmac: Use ContentFileUtil::IsDV(file) here. Java treats a DV as a PUFFIN delete file, not just a position delete with referenced_data_file set, so this currently accepts/rejects the wrong files across v2/v3.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1026 @@
...
+    }
+  }
+  return result;
+}
+
+Status MergingSnapshotUpdate::AddManifest(ManifestFile manifest) {
+  if (manifest.content != ManifestContent::kData) {
+    return InvalidArgument("Cannot append delete manifest: {}", manifest.manifest_path);
+  }
+  if (can_inherit_snapshot_id() && manifest.added_snapshot_id == kInvalidSnapshotId &&
```
wgtmac: Please keep Java parity here: when snapshot ID can be inherited and the manifest has no snapshot ID, Java rejects an assigned first_row_id instead of copying the manifest. Silently rewriting it can reassign row IDs.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1026 @@
...
+
+Status MergingSnapshotUpdate::ValidateNoNewDeletesForDataFiles(
+    const TableMetadata& metadata, int64_t starting_snapshot_id,
+    std::shared_ptr<Expression> data_filter, const DataFileSet& replaced_files,
+    const std::shared_ptr<Snapshot>& parent, std::shared_ptr<FileIO> io) {
+  if (parent == nullptr || replaced_files.empty() || metadata.format_version < 2) {
+    return {};
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto deletes, AddedDeleteFiles(metadata, starting_snapshot_id,
```
wgtmac: Pass data_filter into AddedDeleteFiles. Java filters inside DeleteFileIndex, so pruning happens while reading delete manifests; filtering ReferencedDeleteFiles later can report conflicts that Java would skip.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1026 @@
...
+    }
+  }
+  return {};
+}
+
+Status MergingSnapshotUpdate::ValidateNoNewDeleteFiles(
+    const TableMetadata& metadata, int64_t starting_snapshot_id,
+    std::shared_ptr<Expression> data_filter, const std::shared_ptr<Snapshot>& parent,
+    std::shared_ptr<FileIO> io) {
+  ICEBERG_ASSIGN_OR_RAISE(auto deletes, AddedDeleteFiles(metadata, starting_snapshot_id,
```
wgtmac: Same issue here: build the delete index with data_filter instead of filtering the already-built index. That keeps this aligned with Java addedDeleteFiles(..., dataFilter, ...).

src/iceberg/update/merging_snapshot_update.cc
gty404: Fixed.

src/iceberg/update/merging_snapshot_update.cc
gty404: Fixed.

src/iceberg/update/merging_snapshot_update.cc
gty404: Fixed.

src/iceberg/update/merging_snapshot_update.cc
gty404: Fixed.

src/iceberg/manifest/manifest_filter_manager.cc
```diff
@@ -94,7 +94,6 @@ void ManifestFilterManager::DeleteFile(std::string_view path) {
 Status ManifestFilterManager::DeleteFile(std::shared_ptr<DataFile> file) {
   ICEBERG_PRECHECK(file != nullptr, "Cannot delete file: null");
   delete_paths_.insert(file->file_path);
```
wgtmac: Keep object deletes distinct from path deletes. Java DeleteFileSet keys delete files by location + content offset + content size; path-only matching can remove every DV blob in the same Puffin file.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1004 @@
...
+    cached_new_delete_manifests_.reset();
+  }
+
+  if (cached_new_delete_manifests_.has_value()) {
+    return *cached_new_delete_manifests_;
+  }
+
+  // Group delete files by partition spec ID, mirroring WriteNewDataManifests().
+  std::unordered_map<int32_t, std::vector<PendingDeleteFile>> delete_files_by_spec;
+  for (const auto& pending_file : new_delete_files_) {
```
wgtmac: Normalize staged deletes before writing and summarizing. Java dedupes v2 deletes and merges DVs by referenced data file first; writing new_delete_files_ as-is can over-count duplicates and commit multiple DVs for one data file.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1004 @@
...
+                           delete_file->file_path);
+  }
+  return {};
+}
+
+Status MergingSnapshotUpdate::ValidateNoNewDeleteFiles(
+    const TableMetadata& metadata, int64_t starting_snapshot_id,
+    const PartitionSet& partition_set, const std::shared_ptr<Snapshot>& parent,
+    std::shared_ptr<FileIO> io) {
+  ICEBERG_ASSIGN_OR_RAISE(auto deletes, AddedDeleteFiles(metadata, starting_snapshot_id,
```
wgtmac: Pass partition_set into AddedDeleteFiles. Java filters delete manifests before indexing; post-filtering ReferencedDeleteFiles can still fail on unrelated deletes, for example duplicate DVs outside these partitions.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1004 @@
...
+    builder.FilterPartitions(std::move(partition_set));
+  }
+  return builder.Build();
+}
+
+Status MergingSnapshotUpdate::ValidateAddedDVs(
+    const TableMetadata& /*metadata*/, int64_t /*starting_snapshot_id*/,
+    std::shared_ptr<Expression> /*conflict_filter*/,
+    const std::shared_ptr<Snapshot>& /*parent*/, std::shared_ptr<FileIO> /*io*/) {
+  return NotImplemented(
```
wgtmac: Implement Java’s concurrent-DV check here. Returning NotImplemented leaves v3 row-delta style writers unable to validate the guard that rejects another DV for the same referenced data file.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1004 @@
...
+                          SnapshotUtil::SchemaFor(metadata_to_update, target_branch()));
+  auto tracked_factory = MakeTrackedWriterFactory(target_schema);
+
+  // Step 1: Filter data manifests.
+  ICEBERG_ASSIGN_OR_RAISE(auto filtered_data, data_filter_manager_.FilterManifests(
+                                                  target_schema, metadata_to_update,
+                                                  snapshot, tracked_factory));
+
+  // Track deleted data files in the summary builder.
+  for (const auto& file : data_filter_manager_.FilesToBeDeleted()) {
```
wgtmac: Also propagate duplicate-delete counts into the summary. Java reports deleted-duplicate-files when filtering removes the same file more than once; FilesToBeDeleted() dedupes by path, so that signal is lost.

gty404: Fixed

src/iceberg/update/merging_snapshot_update.cc
gty404: Fixed

src/iceberg/update/merging_snapshot_update.cc
gty404: Fixed

src/iceberg/update/merging_snapshot_update.cc
gty404: Fixed

src/iceberg/manifest/manifest_filter_manager.cc
gty404: Fixed

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1090 @@
...
+    if (!seen_delete_files.insert(key).second) {
+      continue;
+    }
+
+    if (ContentFileUtil::IsDV(*file)) {
+      ICEBERG_PRECHECK(file->referenced_data_file.has_value(),
+                       "DV must have a referenced data file: {}", file->file_path);
+      auto [it, inserted] =
+          dv_by_referenced_data_file.emplace(*file->referenced_data_file, key);
+      if (!inserted && it->second != key) {
```
wgtmac: Match Java DV handling here. MergingSnapshotProducer merges multiple staged DVs for the same referenced data file via DVUtil; returning NotImplemented rejects valid v3 commits instead of writing one merged DV.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1090 @@
...
+Status MergingSnapshotUpdate::ValidateAddedDVs(
+    const TableMetadata& metadata, int64_t starting_snapshot_id,
+    std::shared_ptr<Expression> conflict_filter, const std::shared_ptr<Snapshot>& parent,
+    std::shared_ptr<FileIO> io) {
+  if (parent == nullptr || metadata.format_version < 3) {
+    return {};
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto deletes, AddedDeleteFiles(metadata, starting_snapshot_id,
```
wgtmac: Avoid using the generic AddedDeleteFiles index for Java validateAddedDVs parity. Java scans overwrite/delete/replace history and only rejects DVs whose referenced data file is also in this operation’s staged DV set; this misses replace snapshots and rejects unrelated DVs.

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -116,9 +122,36 @@ class ICEBERG_EXPORT ManifestFilterManager {
...
+  /// and was rewritten with those entries marked DELETED.
+  int32_t ReplacedManifestsCount() const { return replaced_manifests_count_; }
+
   /// \brief Returns true if any delete condition has been registered.
   bool ContainsDeletes() const;
 
+  /// \brief Set the minimum data sequence number for delete files to retain.
+  ///
+  /// Only valid for ManifestContent::kDeletes managers.  Delete entries whose
+  /// data_sequence_number is positive and less than \p sequence_number will be
```
zhjwpku: nit: we don't use `\p` :) same for other places.

src/iceberg/test/merging_snapshot_update_test.cc
```diff
@@ -0,0 +1,1227 @@
...
+// -------------------------------------------------------------------------
+
+TEST_F(MergingSnapshotUpdateTest, CommitNewDataFile) {
+  ICEBERG_UNWRAP_OR_FAIL(auto op, NewMergeAppend());
+  EXPECT_THAT(op->AddFile(file_a_), IsOk());
+  EXPECT_THAT(op->Commit(), IsOk());
+
+  EXPECT_THAT(table_->Refresh(), IsOk());
+  ICEBERG_UNWRAP_OR_FAIL(auto snapshot, table_->current_snapshot());
+  EXPECT_EQ(snapshot->summary.at("added-data-files"), "1");
```
zhjwpku: nit: use the metrics const string listed in snapshot.h? not a strong opinion but it seems you are using them later in this test file.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1090 @@
...
+  if (!file) {
+    return InvalidArgument("Cannot add a null delete file");
+  }
+  ICEBERG_RETURN_UNEXPECTED(ValidateNewDeleteFile(*file));
+  if (!file->partition_spec_id.has_value()) {
+    return InvalidArgument("Delete file must have a partition spec ID");
+  }
+  ICEBERG_ASSIGN_OR_RAISE(auto spec,
+                          base().PartitionSpecById(file->partition_spec_id.value()));
+  (void)spec;
```
zhjwpku: Use `ICEBERG_RETURN_UNEXPECTED` if we don't use the returned value?

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -0,0 +1,1090 @@
...
+  added_delete_files_summary_.Clear();
+  for (const auto& pending_file : new_delete_files_) {
+    ICEBERG_ASSIGN_OR_RAISE(auto spec, metadata_to_update.PartitionSpecById(
+                                           *pending_file.file->partition_spec_id));
+    ICEBERG_RETURN_UNEXPECTED(
+        added_delete_files_summary_.AddedFile(*spec, *pending_file.file));
+  }
+
+  // Rebuild summary from stable sub-builders so that commit retries don't double-count.
+  summary_builder().Clear();
```
zhjwpku: Will this clear custom property set by user?

src/iceberg/update/merging_snapshot_update.cc
gty404: The current implementation fails closed because C++ does not yet have the Java DVUtil/Puffin write path needed to merge multiple staged DVs. I’ll narrow this PR to avoid claiming Java parity for staged DV merging.

src/iceberg/update/merging_snapshot_update.cc
gty404: Agreed. Using the generic AddedDeleteFiles index is too broad here. I’ll change this to match Java’s validation semantics: scan the relevant overwrite/delete/replace history and only reject DVs whose referenced data file is also in this operation’s staged DV set.

src/iceberg/manifest/manifest_filter_manager.h
gty404: Thanks, will update the doc comment to avoid \\p and check the nearby comments for the same style issue.

src/iceberg/update/merging_snapshot_update.cc
gty404: Good catch. This does clear properties set through Set(). I’ll persist custom summary properties separately and reapply them after rebuilding the summary builder in Apply().

src/iceberg/update/merging_snapshot_update.cc
gty404: Agreed. This is only validating that the spec exists, so I’ll replace the unused assignment with ICEBERG_RETURN_UNEXPECTED

src/iceberg/test/merging_snapshot_update_test.cc
gty404: Makes sense. I’ll switch these assertions to SnapshotSummaryFields constants to keep the test consistent with the rest of the file

src/iceberg/util/snapshot_util.cc
```diff
@@ -375,7 +375,7 @@ Result<std::shared_ptr<Schema>> SnapshotUtil::SchemaFor(const Table& table,
 
   const auto& metadata = table.metadata();
   auto it = metadata->refs.find(ref);
-  if (it == metadata->refs.cend() || it->second->type() == SnapshotRefType::kBranch) {
+  if (it == metadata->refs.cend()) {
```
wgtmac: Keep Java parity here: branches should return the current table schema, and only tags should use the referenced snapshot schema. Please keep the branch check in both SchemaFor(ref) overloads.

src/iceberg/update/merging_snapshot_update.h
```diff
@@ -0,0 +1,357 @@
...
+  /// data files.
+  Result<std::shared_ptr<PartitionSpec>> DataSpec() const;
+
+  /// \brief Returns all data files staged for addition.
+  std::vector<std::shared_ptr<DataFile>> AddedDataFiles() const;
+
+  /// \brief Return an error if any snapshot in [starting_snapshot_id+1, parent]
+  /// added a data file matching the given filter expression.
+  static Status ValidateAddedDataFiles(const TableMetadata& metadata,
+                                       int64_t starting_snapshot_id,
```
wgtmac: Java allows startingSnapshotId to be null and validates back to the first snapshot. Please make this optional through these validation helpers so default RowDelta/Overwrite/Rewrite validation matches Java.

src/iceberg/test/merging_snapshot_update_test.cc
```diff
@@ -0,0 +1,1278 @@
...
+  EXPECT_THAT(op->Commit(), IsOk());
+
+  EXPECT_THAT(table_->Refresh(), IsOk());
+  ICEBERG_UNWRAP_OR_FAIL(auto snapshot, table_->current_snapshot());
+  EXPECT_EQ(snapshot->summary.at(SnapshotSummaryFields::kAddedDataFiles), "1");
+  EXPECT_EQ(snapshot->summary.at(SnapshotSummaryFields::kTotalDataFiles), "1");
+}
+
+// -------------------------------------------------------------------------
+// ValidateNewDeleteFile format version tests
```
wgtmac, line 723: Please add Java-parity tests for the v3 gates: reject non-DV position deletes on v3, allow DVs on v3, and reject a v2-staged position delete after upgrading to v3 before commit.

src/iceberg/update/snapshot_update.cc
```diff
@@ -41,7 +41,7 @@ namespace iceberg {
 
 namespace {
 
-// The Java impl skips updating total if parsing fails. Here we choose to be strict.
+// Skips updating total if parsing fails would be lenient; here we choose to be strict.
```
wgtmac: This sentence is hard to parse. Consider: `Java skips updating totals if parsing fails; C++ treats parse failures as errors.`

src/iceberg/util/snapshot_util.cc
```diff
@@ -375,7 +375,7 @@ Result<std::shared_ptr<Schema>> SnapshotUtil::SchemaFor(const Table& table,
 
   const auto& metadata = table.metadata();
   auto it = metadata->refs.find(ref);
-  if (it == metadata->refs.cend() || it->second->type() == SnapshotRefType::kBranch) {
```
wgtmac, line 378: Why here and below are changed? I checked that they still exist in the Java implementation.

src/iceberg/snapshot.h
```diff
@@ -359,6 +368,10 @@ class ICEBERG_EXPORT SnapshotSummaryBuilder {
   int32_t max_changed_partitions_for_summaries_{0};
   int64_t deleted_duplicate_files_{0};
   bool trust_partition_metrics_{true};
+  bool manifests_counts_set_{false};
```
wgtmac: Java does not populate these fields in the summary builder and instead directly call `Set(const std::string& property, const std::string& value)` to set these values. Should we follow a similar pattern as these fields are not commonly used?

src/iceberg/util/snapshot_util.cc
gty404: Agreed. I’ll restore Java parity here by returning the current table schema for branch refs and only resolving the referenced snapshot schema for tags, in both SchemaFor(ref) overloads. I’ll also update the related docs/tests if needed

src/iceberg/update/merging_snapshot_update.h
gty404: I’ll make starting_snapshot_id optional across these validation helpers and update the history traversal so a missing starting snapshot validates back to the beginning of the snapshot lineage, matching Java’s default validation behavior

src/iceberg/test/merging_snapshot_update_test.cc
gty404: The production validation path already rechecks staged delete files during Apply, but the v3 gate coverage is incomplete. I’ll add Java-parity tests for rejecting non-DV position deletes on v3, allowing DVs on v3, and rejecting a v2-staged position delete after upgrading to v3 before commit

src/iceberg/update/snapshot_update.cc
gty404: I’ll replace the comment with the suggested clearer wording: Java skips updating totals if parsing fails; C++ treats parse failures as errors.

src/iceberg/snapshot.cc
```diff
@@ -491,6 +503,10 @@ void SnapshotSummaryBuilder::Merge(const SnapshotSummaryBuilder& other) {
   }
 
   deleted_duplicate_files_ += other.deleted_duplicate_files_;
+  // Manifest counts (manifests_counts_set_ / manifests_created_ / manifests_kept_ /
```
wgtmac: If we remove these fields, then these comments should be removed too.

src/iceberg/update/snapshot_update.h
```diff
@@ -122,6 +122,11 @@ class ICEBERG_EXPORT SnapshotUpdate : public PendingUpdate {
   Status Finalize(Result<const TableMetadata*> commit_result) override;
 
  protected:
+  struct DeleteManifestEntry {
```
wgtmac: Do we want to add a dedicated `DataFileWrapper` structure to equip it with optional fields like `data_sequence_number` and `file_sequence_number`? It can be relocated to `manifest_entry.h` as `DeleteManifestEntry` looks confusing.

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -214,11 +273,22 @@ class ICEBERG_EXPORT ManifestFilterManager {
...
+  std::unordered_set<DeleteFileKey, DeleteFileKeyHash> delete_file_keys_;
   DataFileSet delete_files_;
+  std::vector<std::shared_ptr<DataFile>> deleted_files_;
+  std::unordered_set<DeleteFileKey, DeleteFileKeyHash> deleted_file_keys_;
   PartitionSet drop_partitions_;
   bool fail_missing_delete_paths_{false};
   bool fail_any_delete_{false};
   bool case_sensitive_{true};
+  int32_t duplicate_deletes_count_{0};
+
```
wgtmac:
```suggestion
```

src/iceberg/manifest/manifest_filter_manager.h
```diff
@@ -214,11 +273,22 @@ class ICEBERG_EXPORT ManifestFilterManager {
 
   std::shared_ptr<Expression> delete_expr_;
   std::unordered_set<std::string> delete_paths_;
+  std::unordered_set<DeleteFileKey, DeleteFileKeyHash> delete_file_keys_;
```
wgtmac: Why not using `DataFileSet`? Do `content_offset` and `content_size_in_bytes` really matter in this case? I don't see Java ManifestFilterManager has this special handling.

wgtmac (approved):
Thanks @gty404 for working on this and @zhjwpku for the review! This is quite a large change and it took me a long time to understand the whole workflow.

I pushed a follow-up cleanup commit to align the PR more closely with Java:

- Aligned MergingSnapshotUpdate apply/cleanup/validation flow with Java.
- Reworked ManifestFilterManager and ManifestMergeManager caching/cleanup semantics.
- Added DeleteFileSet support and tests.
- Fixed v3 first_row_id / manifest reader parity issues.
- Added/updated targeted tests for filter, merge, delete file set, manifest reader, and snapshot update behavior.

PR #684, opened 2026-05-27
fix(meson): add missing install headers for meson

wgtmac:


blackmwk (approved):

wgtmac:
Thanks for the review, @blackmwk!

PR #686, opened 2026-05-27
fix: correct StrictMetricsEvaluator::CanContainNulls/CanContainNaNs when field missing from stats map

### LINKED ISSUE #685: bug: StrictMetricsEvaluator returns incorrect result when null/NaN counts are missing for a field.
**Summary**

`StrictMetricsEvaluator::CanContainNulls` and `CanContainNaNs` incorrectly return `false` when the `null_value_counts` / `nan_value_counts` map is non-empty but does not contain an entry for the queried field. This causes the evaluator to erroneously return `kRowsMustMatch`, potentially skipping row-level filtering and returning rows that do not satisfy the predicate.

**Root Cause**

In `src/iceberg/expression/strict_metrics_evaluator.cc`:

```cpp
bool CanContainNulls(int32_t id) {
  if (data_file_.null_value_counts.empty()) {
  return true;
  }
  auto it = data_file_.null_value_counts.find(id);
  return it != data_file_.null_value_counts.cend() && it->second > 0; 
  //       ^^^ when field is missing from map, this evaluates to false
}
```

The same pattern exists in CanContainNaNs.

**Reproduction**

```cpp
  auto data_file = std::make_shared<DataFile>();
  data_file->record_count = 50;
  data_file->value_counts = {{14, 50L}};
  data_file->null_value_counts = {{4, 0L}, {5, 0L}};  // field 14 missing
  data_file->nan_value_counts = {{8, 0L}};             // field 14 missing
  data_file->upper_bounds = {{14, Literal::Double(100.0).Serialize().value()}};
  data_file->lower_bounds = {{14, Literal::Double(1.0).Serialize().value()}};

  // Evaluating: no_nan_stats < 200.0
  // Expected: kRowsMightNotMatch (null count unknown)
  // Actual:   kRowsMustMatch (incorrectly skips filtering)
```

**Proposed Fix**

CanContainNulls: if the field is required per schema, return false; if the field is not found in a non-empty map, return true (conservative).
CanContainNaNs: if the field type is not float/double, return false; if the field is not found in a non-empty map, return true (conservative).

This aligns with Java's StrictMetricsEvaluator.canContainNulls() / canContainNaNs() which return true when the field is missing from the map.

sentomk:
## Summary

- Fix `StrictMetricsEvaluator::CanContainNulls` to return `true` (conservative) when a field is absent from a non-empty `null_value_counts` map, and `false` for required fields per schema
- Fix `StrictMetricsEvaluator::CanContainNaNs` to return `true` (conservative) when a field is absent from a non-empty `nan_value_counts` map, and `false` for non-floating-point types
- Add regression tests covering the missing-entry scenario

Close #685 

## Motivation

When `null_value_counts` or `nan_value_counts` is non-empty but does not contain an entry for the queried field, the evaluator incorrectly returned `false`. This caused comparison operators to erroneously return `kRowsMustMatch`, potentially skipping row-level filtering and returning rows that do not satisfy the predicate.

## Test plan

- All 40 existing + new StrictMetrics tests pass
- pre-commit (clang-format, trailing whitespace, etc.) passes

### DIFF
--- a/src/iceberg/expression/strict_metrics_evaluator.cc
+++ b/src/iceberg/expression/strict_metrics_evaluator.cc
@@ -142,7 +142,9 @@ class StrictMetricsVisitor : public BoundVisitor<bool> {
       return kRowsMightNotMatch;
     }
 
-    if (CanContainNulls(id) || CanContainNaNs(id)) {
+    ICEBERG_ASSIGN_OR_RAISE(auto has_nulls, CanContainNulls(id));
+    ICEBERG_ASSIGN_OR_RAISE(auto has_nans, CanContainNaNs(id));
+    if (has_nulls || has_nans) {
       return kRowsMightNotMatch;
     }
 
@@ -168,7 +170,9 @@ class StrictMetricsVisitor : public BoundVisitor<bool> {
       return kRowsMightNotMatch;
     }
 
-    if (CanContainNulls(id) || CanContainNaNs(id)) {
+    ICEBERG_ASSIGN_OR_RAISE(auto has_nulls, CanContainNulls(id));
+    ICEBERG_ASSIGN_OR_RAISE(auto has_nans, CanContainNaNs(id));
+    if (has_nulls || has_nans) {
       return kRowsMightNotMatch;
     }
 
@@ -194,7 +198,9 @@ class StrictMetricsVisitor : public BoundVisitor<bool> {
       return kRowsMightNotMatch;
     }
 
-    if (CanContainNulls(id) || CanContainNaNs(id)) {
+    ICEBERG_ASSIGN_OR_RAISE(auto has_nulls, CanContainNulls(id));
+    ICEBERG_ASSIGN_OR_RAISE(auto has_nans, CanContainNaNs(id));
+    if (has_nulls || has_nans) {
       return kRowsMightNotMatch;
     }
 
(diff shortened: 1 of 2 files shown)


src/iceberg/expression/strict_metrics_evaluator.cc
```diff
@@ -436,18 +436,39 @@ class StrictMetricsVisitor : public BoundVisitor<bool> {
   }
 
   bool CanContainNulls(int32_t id) {
+    auto field_result = schema_.GetFieldById(id);
+    if (field_result.has_value() && field_result->has_value() &&
```
wgtmac: We should preserve error returned by `schema_.GetFieldById`, so we need to change the return type to `Result<bool>`. BTW, why calling `GetFieldById` instead of `FindFieldById`? They are semantically different.

src/iceberg/expression/strict_metrics_evaluator.cc
```diff
@@ -436,18 +436,39 @@ class StrictMetricsVisitor : public BoundVisitor<bool> {
...
     auto it = data_file_.null_value_counts.find(id);
-    return it != data_file_.null_value_counts.cend() && it->second > 0;
+    if (it == data_file_.null_value_counts.cend()) {
+      return true;
+    }
+    return it->second > 0;
   }
 
   bool CanContainNaNs(int32_t id) {
-    // nan counts might be null for early version writers when nan counters are not
```
wgtmac, line 447: I think the original implementation matches the Java parity exactly, which assumes null_value_counts should populate all fields but nan_value_counts don't (when they are not empty). I agree current PR is a nice fix. I still suggest keeping this comment since it still holds.

src/iceberg/test/strict_metrics_evaluator_test.cc
```diff
@@ -846,4 +846,50 @@ TEST_F(StrictMetricsEvaluatorMigratedTest, EvaluateOnNestedColumnWithStats) {
   ExpectShouldRead(Expressions::NotNull("struct.nested_col_with_stats"), false);
 }
 
+TEST_F(StrictMetricsEvaluatorMigratedTest, MissingNullCountForField) {
```
wgtmac: This is not migrated from Java implementation so please do not use this confusing name.

wgtmac (changes requested):

wgtmac (approved):
LGTM. Thank you, @sentomk!

PR #687, opened 2026-05-28
feat: add executor pool support

HuaHuaY:


src/iceberg/util/task_group.cc
```diff
@@ -0,0 +1,100 @@
...
+  return std::unexpected(Error{.kind = kind, .message = std::move(message)});
+}
+
+Result<std::future<Status>> SubmitTask(Executor& executor, FnOnce<Status()> task) {
+  std::promise<Status> promise;
+  auto future = promise.get_future();
+
+  ExecutorTask executor_task(
+      [promise = std::move(promise), task = std::move(task)]() mutable {
+        promise.set_value(std::move(task)());
```
wgtmac, line 55: If a task throws instead of returning a `Status`, the promise is never set and the exception escapes on the worker thread, while `future.get()` sees a broken promise. The repo uses `Result`/`Status` rather than exceptions, so this is unlikely in our own code, but `Executor` is a public extension point and users may plug in pools where it matters. Wrapping the call in try/catch and turning the exception into a `Status` would be safer.

src/iceberg/util/task_group.h
```diff
@@ -0,0 +1,109 @@
...
+      }
+    }());
+    return std::forward<decltype(self)>(self);
+  }
+
+  Status Run() && {
+    if (!executor_.has_value()) {
+      return internal::RunTasksSingleThreaded(std::move(tasks_));
+    }
+    return internal::RunTasksParallel(executor_->get(), std::move(tasks_));
```
wgtmac, line 99: `Run()` blocks on `future.get()` from the calling thread. That's fine for the current sequential call sites, but `PlanWith` is now public API. If someone drives a `TaskGroup` from a worker thread of the same bounded executor, it can deadlock (pool saturated while waiting on a task queued behind it). Worth documenting that the driving thread must not be one of the executor's own workers.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -376,57 +399,73 @@ ManifestGroup::ReadEntries() {
...
       }
+      return {};
+    });
+  }
+  ICEBERG_RETURN_UNEXPECTED(std::move(read_tasks).Run());
 
-      result[spec_id].push_back(std::move(entry));
+  std::unordered_map<int32_t, std::vector<ManifestEntry>> result;
+  for (auto& manifest_result : manifest_results) {
+    result.merge(manifest_result);
```
wgtmac: This depends on `merge()` moving out the non-conflicting nodes and leaving only the conflicting keys, which are then appended. It's correct, but takes a moment to parse. A plain loop doing `result[spec_id].insert(...)` for every entry reads more directly, at negligible cost.

wgtmac (commented):
Nice work. The abstraction is clean and the test coverage is solid. Since the default executor is `nullopt`, existing paths stay single-threaded, so the risk is well contained.

A couple of things before merge:

- The PR has no description. Worth adding the motivation and a short design note, ideally with the parallel/serial numbers that justify it.
- The `RetryRunner` change from a runtime fluent API (`OnlyRetryOn`/`StopRetryOn`) to the compile-time `RetryRunner<Policy>` is a breaking API change that's independent of executor support. Consider splitting it into its own PR so it gets reviewed on its own merits.

Left a few inline notes.

src/iceberg/manifest/manifest_filter_manager.cc
```diff
@@ -401,19 +451,35 @@ Result<std::vector<ManifestFile>> ManifestFilterManager::FilterManifests(
...
+  }
+  {
+    std::lock_guard lock(residual_evaluator_cache_mutex_);
+    residual_evaluator_cache_.clear();
+  }
+
+  std::vector<FilterManifestResult> filter_results(manifests.size());
+  auto filter_tasks = TaskGroup().SetExecutor(executor_);
+  for (auto&& [manifest, result] : std::views::zip(manifests, filter_results)) {
+    filter_tasks.Submit([&]() -> Status {
```
wgtmac: This is safe today: the bindings refer to elements of `manifests` and `filter_results`, both of which outlive `Run()`. But capturing a loop's structured bindings by `[&]` into a deferred closure is an easy footgun. Capturing explicitly (e.g. `[&manifest, &result, ...]`) makes the intent clearer and harder to break later.

src/iceberg/update/snapshot_update.cc
```diff
@@ -250,13 +251,17 @@ Result<SnapshotUpdate::ApplyResult> SnapshotUpdate::Apply() {
...
 
   ICEBERG_ASSIGN_OR_RAISE(auto manifests, Apply(base(), parent_snapshot));
+  auto metadata_tasks = TaskGroup().SetExecutor(plan_executor_);
   for (auto& manifest : manifests) {
     if (manifest.added_snapshot_id != kInvalidSnapshotId) {
       continue;
     }
-    // TODO(xxx): read in parallel and cache enriched manifests for retries
-    ICEBERG_ASSIGN_OR_RAISE(manifest, AddMetadata(manifest, ctx_->table->io(), base()));
+    metadata_tasks.Submit([&manifest, this]() -> Status {
```
wgtmac: This one is nicely done: each task captures a distinct element of `manifests` (not an iterator or index), so there's no aliasing, and the `manifest_count_`/`attempt_` switch to atomics is backed by the `ConcurrentManifestPaths` test.

src/iceberg/util/retry_util.h
```diff
@@ -69,76 +53,104 @@ struct ICEBERG_EXPORT RetryConfig {
...
+    }
   }
+};
 
-  /// \brief Specify a single error type that should stop retries immediately.
-  ///
-  /// \note OnlyRetryOn takes priority over StopRetryOn. If OnlyRetryOn is set,
-  /// StopRetryOn is ignored.
-  RetryRunner& StopRetryOn(ErrorKind error_kind) { return StopRetryOn({error_kind}); }
+using NoRetry = RetryPolicy<RetryPolicyMode::kNoRetry>;
```
wgtmac, line 111: Note this drops a couple of runtime behaviors the old API had: `OnlyRetryOn` taking precedence over `StopRetryOn`, and the "policy must be configured" runtime error. They're now compile-time `static_assert`s plus `NoRetry`. The only caller (`MakeCommitRetryRunner`) is updated so it's functionally fine, but it's a user-visible tightening worth calling out in the PR description.

wgtmac (commented):
A few more minor notes, non-blocking.

src/iceberg/util/task_group.cc
HuaHuaY: Errors resulting from violating calling conventions are the user's responsibility, not the library's.

src/iceberg/manifest/manifest_group.cc
HuaHuaY: `merge()` does not involve memory allocation and is more efficient.

src/iceberg/util/task_group.h
HuaHuaY: The issue of a thread pool's tasks spawning more tasks and then blocking themselves is a problem that I believe any thread pool with a thread limit will encounter, not just a problem with this particular PR.

src/iceberg/manifest/manifest_filter_manager.cc
HuaHuaY: There are too many variables captured by this lambda. It is inappropriate to write the names clearly.

wgtmac (commented):
Another pass, this time on the interface design and extensibility.

The overall shape is good: one virtual `Executor` that engines implement, threaded through the builders via `PlanWith`. The Arrow adapter test is a nice proof that an external pool drops in with basically one line, so "bring your own threadpool" is well covered.

On future async directions (C++23 coroutines, P2300 `std::execution`): the model here is synchronous and blocking, `TaskGroup::Run()` fans out and blocks on `std::future::get()`, and the planning APIs return `Result<...>` directly. So this is a parallel-for primitive, not a step toward a sender/receiver pipeline. That's a reasonable scope for now, I'd just flag it explicitly so nobody expects this interface to extend into async later, it'll be a separate one. Details inline.

src/iceberg/util/executor.h
```diff
@@ -0,0 +1,43 @@
...
+namespace iceberg {
+
+using ExecutorTask = FnOnce<void()>;
+
+class ICEBERG_EXPORT Executor {
+ public:
+  virtual ~Executor() = default;
+
+  /// \brief Schedule a task for execution.
+  virtual Status Submit(ExecutorTask task) = 0;
```
wgtmac, line 48: This is a fire-and-forget `execute`-style primitive (closer to the abandoned P0443 `executor.execute` than to P2300's scheduler/sender). Completion is tracked outside, via the `std::promise`/`future` plumbing in `RunTasksParallel`. Fine for a blocking parallel-for, but it doesn't lay groundwork for coroutines or `std::execution`: those need the executor to hand back something awaitable/composable, and the planning APIs (`PlanFiles() -> Result<...>`) are synchronous anyway. Going async later would be a separate interface, not an extension of this one. Worth stating in the header that `Executor` is a parallel-dispatch primitive, not an async scheduler.

Separately: `ExecutorTask` being move-only is the right call and matches Arrow, but pools whose submit takes a copyable `std::function` will need `std::move_only_function` or a small shim to adapt.

src/iceberg/util/executor.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+using OptionalExecutor = std::optional<std::reference_wrapper<Executor>>;
```
wgtmac, line 51: `std::optional<std::reference_wrapper<Executor>>` is a little awkward to use (`executor_->get()` at the call sites). A plain `Executor*` expresses "nullable, non-owning borrow" just as well and reads cleaner everywhere it's passed. Minor.

src/iceberg/util/task_group.cc
```diff
@@ -0,0 +1,100 @@
...
+  }
+  return AggregateTaskErrors(std::move(errors));
+}
+
+Status RunTasksParallel(Executor& executor, std::vector<FnOnce<Status()>> tasks) {
+  std::vector<std::future<Status>> futures;
+  futures.reserve(tasks.size());
+
+  std::vector<Error> errors;
+  for (auto& task : tasks) {
```
wgtmac, line 81: All tasks are submitted up front, each with its own promise/future. For a handful of manifests that's fine, but as a general primitive there's no concurrency bound and no fail-fast: N tasks always queue N and allocate N futures, and if one fails early the rest still run to completion. Probably out of scope for this PR, but worth a note if engines may push large fan-outs through here.

src/iceberg/update/snapshot_update.h
```diff
@@ -77,6 +79,16 @@ class ICEBERG_EXPORT SnapshotUpdate : public PendingUpdate {
     return self;
   }
 
+  /// \brief Configure an executor for manifest planning work.
+  ///
+  /// The executor is borrowed and must outlive this update. Planning callbacks may be
+  /// called concurrently; callers must synchronize shared mutable state captured by
+  /// those callbacks.
+  auto& ScanManifestsWith(this auto& self, Executor& executor) {
```
wgtmac: Naming: every other entry point uses `PlanWith` (TableScanBuilder, ExpireSnapshots, ManifestGroup, ManifestFilterManager, ManifestMergeManager), but this one is `ScanManifestsWith`. Worth settling on one name for a consistent public surface.

src/iceberg/manifest/manifest_merge_manager.h
```diff
@@ -56,6 +57,9 @@ class ICEBERG_EXPORT ManifestMergeManager {
   ManifestMergeManager(const ManifestMergeManager&) = delete;
   ManifestMergeManager& operator=(const ManifestMergeManager&) = delete;
 
+  /// \brief Configure an optional executor for manifest merging.
+  ManifestMergeManager& PlanWith(OptionalExecutor executor);
```
wgtmac: Once an executor is set here, the user-supplied `ManifestWriterFactory` gets invoked concurrently from worker threads (parallel `FlushBin`). The tests already guard against this with an atomic path counter + barrier, which confirms the intent, but the requirement isn't documented. A downstream engine implementing a factory won't know it has to be thread-safe. Worth a `\note` on this `PlanWith` (and the other `PlanWith` entry points / the `ManifestWriterFactory` typedef) stating the factory must be safe to call concurrently when an executor is configured.

wgtmac (commented):
One more pass. The main thing I found is a documentation gap around a new concurrency contract that now leaks onto user-supplied callbacks.

Once an executor is set, the user's `ManifestWriterFactory` and the shared `FileIO` get called from multiple worker threads. The tests already account for this (the factories use an atomic counter plus a barrier), so the requirement is understood, it's just not written down anywhere a downstream engine would see it. Worth documenting on the public surface.

src/iceberg/delete_file_index.cc
```diff
@@ -528,107 +531,159 @@ DeleteFileIndex::Builder& DeleteFileIndex::Builder::IgnoreResiduals() {
...
+                                                filter, spec, *schema_, case_sensitive_));
+    auto [inserted_iter, _] = eval_cache.emplace(spec_id, std::move(evaluator));
+    return inserted_iter->second.get();
+  };
+
+  std::vector<std::vector<ManifestEntry>> manifest_results(delete_manifests_.size());
+  auto read_tasks = TaskGroup().SetExecutor(executor_);
+  for (auto&& [manifest, manifest_result] :
+       std::views::zip(delete_manifests_, manifest_results)) {
+    read_tasks.Submit([&]() -> Status {
```
wgtmac: Same point on the read side: every task shares `io_` via `ManifestReader::Make`, so parallel planning assumes `FileIO` is safe for concurrent reads. `FileIO` doesn't document any thread-safety guarantee today. Either state the requirement on `FileIO` or note it at the `PlanWith` call sites, so implementers know what's expected when they pass an executor.

src/iceberg/util/executor.h
HuaHuaY: Holding a pointer requires a check to determine whether it needs to be destructed, which is detrimental to future code maintainability.

src/iceberg/update/snapshot_update.h
HuaHuaY: This is the behavior of `apache/iceberg` repo.

src/iceberg/manifest/manifest_merge_manager.h
HuaHuaY: The documentation is on the Executor interface. There are too many `PlanWith` and putting comments there will result in too much redundancy.

src/iceberg/delete_file_index.cc
HuaHuaY: ditto

src/iceberg/util/functional.h
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Borrowed the file from Apache Arrow:
+// https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/functional.h
```
wgtmac, line 21: We need to update LICENSE file

wgtmac (approved):

PR #693, opened 2026-05-31
feat(catalog): add SQL catalog

zhjwpku:
Add a relational database backed SqlCatalog with a CatalogStore abstraction, built-in sqlpp23 stores for SQLite, PostgreSQL, and MySQL, and Java-compatible catalog tables.


.github/workflows/test.yml
```diff
@@ -107,7 +107,7 @@ jobs:
       - name: Install dependencies
         shell: pwsh
         run: |
-          vcpkg install zlib:x64-windows nlohmann-json:x64-windows nanoarrow:x64-windows roaring:x64-windows cpr:x64-windows
+          vcpkg install zlib:x64-windows nlohmann-json:x64-windows nanoarrow:x64-windows roaring:x64-windows cpr:x64-windows sqlite3:x64-windows
```
wgtmac: Is it better to add a dedicated ci workflow for the sql catalog? We can trigger it only when files related to sql catalog have been changed to reduce resource usage.

CMakeLists.txt
```diff
@@ -46,6 +46,12 @@ option(ICEBERG_BUILD_BUNDLE "Build the battery included library" ON)
 option(ICEBERG_BUILD_REST "Build rest catalog client" ON)
 option(ICEBERG_BUILD_REST_INTEGRATION_TESTS "Build rest catalog integration tests" OFF)
 option(ICEBERG_BUILD_HIVE "Build hive (HMS) catalog client" OFF)
+option(ICEBERG_BUILD_SQL_CATALOG "Build SQL catalog client" ON)
```
wgtmac: Should we make it off by default? This follows the pattern used by rest catalog library.

src/iceberg/catalog/sql/README.md
```diff
@@ -0,0 +1,158 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+-->
+
+# SQL Catalog
```
wgtmac, line 20: Perhaps it is good to add to the iceberg-cpp site as well.

src/iceberg/catalog/sql/sql_catalog.h
```diff
@@ -0,0 +1,190 @@
...
+
+  Status RenameTable(const TableIdentifier& from, const TableIdentifier& to) override;
+
+  Result<std::shared_ptr<Table>> LoadTable(const TableIdentifier& identifier) override;
+
+  Result<std::shared_ptr<Table>> RegisterTable(
+      const TableIdentifier& identifier,
+      const std::string& metadata_file_location) override;
+
+#ifdef BUILD_SQLITE3_CONNECTOR
```
wgtmac: Would it be good to always make these functions visible to users? We can check `BUILD_SQLITE3_CONNECTOR` and other macros in the sql_catalog.cc file and return NotSupported if not built.

.github/workflows/test.yml
zhjwpku: Agreed. I've added a dedicated CI workflow, but I haven't configured `paths` filtering for SQL catalog-related changes yet. Since it's a bit tricky to define a clean filter, I'd rather address that in a separate PR.

CMakeLists.txt
zhjwpku: REST catalog is `ON` by default, but I'd agree that we should make SQL catalog OFF, changed.

src/iceberg/catalog/sql/README.md
zhjwpku: Add a new page, thanks.

src/iceberg/catalog/sql/sql_catalog.h
zhjwpku: Agreed, done that way.

.github/workflows/cpp-linter.yml
```diff
@@ -41,14 +41,20 @@ jobs:
...
+          sudo apt-get install -y libcurl4-openssl-dev libsqlite3-dev libpq-dev default-libmysqlclient-dev
       - name: Run build
         env:
           CC: gcc-14
           CXX: g++-14
         run: |
           mkdir build && cd build
-          cmake .. -G Ninja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+          cmake .. -G Ninja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+            -DICEBERG_BUILD_SQL_CATALOG=ON \
```
wgtmac, line 54: It would be nice to enhance this in the future by enabling specific modules only when their paths are affected.

wgtmac (approved):
Thank you for adding this, @zhjwpku! I think this is a great enrichment to iceberg-cpp and its downstream users. I'll count on you to maintain this module in the future.

PR #694, opened 2026-06-01
feat(hive): vendor Hive 3.1 metastore + fb303 Thrift IDL

MisterRaindrop:
Vendor the Apache Hive 3.1 standalone-metastore IDL and the fb303 helper IDL it includes into third_party/hive_metastore/. These files are the input for the C++ HMS client bindings, generated by a follow-up commit that invokes `thrift --gen cpp` at build time.

Provenance:
* hive_metastore.thrift       - apache/hive @ branch-3.1, standalone-metastore
* share/fb303/if/fb303.thrift - apache/thrift @ master, contrib/fb303

Both upstream files retain their Apache 2.0 license headers; only trailing whitespace and final newlines were normalized by the repository's pre-commit hooks. third_party/hive_metastore/NOTICE records the upstream sources, and the project root NOTICE references it. .github/.licenserc.yaml gains third_party/** to paths-ignore so the license-eye check skips the vendored tree.

### DIFF
--- a/.github/.licenserc.yaml
+++ b/.github/.licenserc.yaml
@@ -31,5 +31,6 @@ header:
     - 'requirements.txt'
     - 'src/iceberg/util/murmurhash3_internal.*'
     - 'src/iceberg/test/resources/**'
+    - 'thirdparty/**'
 
   comment: on-failure
--- a/thirdparty/hive_metastore/README.md
+++ b/thirdparty/hive_metastore/README.md
@@ -1,0 +1,49 @@
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+-->
+
+# Vendored Hive Metastore Thrift IDL
+
+This directory contains vendored Thrift interface definitions used to generate
+the C++ Hive Metastore (HMS) client consumed by the `iceberg_hive` library. The
+files are copied verbatim from upstream Apache projects and remain under the
+Apache License, Version 2.0 (see `LICENSE` in the repository root). The only
+modification is the repository's pre-commit normalization (trailing whitespace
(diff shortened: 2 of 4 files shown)


.github/.licenserc.yaml
```diff
@@ -31,5 +31,6 @@ header:
     - 'requirements.txt'
     - 'src/iceberg/util/murmurhash3_internal.*'
     - 'src/iceberg/test/resources/**'
+    - 'third_party/**'
```
wgtmac: Should we rename `third_party` to `thirdparty` which is more widely used?

third_party/hive_metastore/NOTICE
```diff
@@ -0,0 +1,26 @@
...
+`iceberg_hive` library. The interface files are sourced verbatim from
+upstream Apache projects (only trailing whitespace and final newlines
+were normalized by the repository's pre-commit hooks) and remain under
+the Apache License, Version 2.0 (see LICENSE in the repository root).
+
+Vendored sources
+================
+
+* hive_metastore.thrift
+    Apache Hive 3.1 standalone-metastore.
```
wgtmac: Why this specific version? How do we want to upgrade or maintain multiple versions in the future?

NOTICE
```diff
@@ -3,3 +3,8 @@ Copyright 2024-2026 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
+
+Third-party Thrift IDLs vendored under third_party/hive_metastore/ are
```
wgtmac: IIUC, ASF projects are exempted here. cc expert @jbonofre

wgtmac (commented):
Thanks for importing Hive related files. I've left some minor comments. BTW, the title is not core. Perhaps rename to `feat(hive): vendor Hive 3.1 metastore + fb303 Thrift IDL`

third_party/hive_metastore/NOTICE
MisterRaindrop: Good question. I selected Hive 3.1.3 because it is a mature and widely deployed HMS
version. I also considered the precedent from iceberg-rust, which maintains a single
client generated from a Hive 2.3 IDL and integration-tests it against Hive 3.1.3.

For future maintenance, I propose keeping a single vendored IDL pinned to an
immutable Hive release tag or commit, rather than maintaining separate generated
clients for each Hive version. The implementation should use RPCs shared across the
supported versions whenever possible.

If future Hive releases introduce incompatible RPC changes, we can add narrowly
scoped runtime adapters or fallback logic, allowing one build of iceberg-cpp to
support multiple Hive versions. We should validate and document the supported
versions through a CI compatibility matrix before claiming compatibility.

third_party/hive_metastore/NOTICE
```diff
@@ -0,0 +1,26 @@
...
+    Source: https://github.com/apache/hive/blob/branch-3.1/standalone-metastore/src/main/thrift/hive_metastore.thrift
```
gty404: Should we also pin the exact upstream commit SHAs (or tags) for the vendored files in third_party/hive_metastore/NOTICE or a README so provenance is reproducible and future updates are deterministic.

wgtmac (approved):

wgtmac:
Thank you, @MisterRaindrop and @gty404!

PR #697, opened 2026-06-05
feat: parallelize reading manifests

HuaHuaY:


.pre-commit-config.yaml
```diff
@@ -33,7 +33,7 @@ repos:
         exclude: ^src/iceberg/catalog/hive/gen-cpp/
 
   - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v20.1.8
+    rev: v22.1.5
```
wgtmac, line 36: It would be better to use a separate PR for this unrelated change.

src/iceberg/util/executor_util_internal.h
```diff
@@ -0,0 +1,221 @@
...
+template <auto... Options, typename... Args>
+  requires(sizeof...(Args) >= 2 && sizeof...(Args) % 2 == 0 &&
+           []<std::size_t... I>(std::index_sequence<I...>) consteval {
+             return (internal::ParallelCollectible<
+                         typename internal::ParallelCollectTraits<I, Args...>::input_type,
+                         typename internal::ParallelCollectTraits<I, Args...>::task_type,
+                         Options...> &&
+                     ...);
+           }(std::make_index_sequence<sizeof...(Args) / 2>{}))
+auto ParallelCollect(OptionalExecutor executor, Args&&... args) {
```
wgtmac, line 163: `Collect` is more like reduce semantics. We might need a better name.

wgtmac: BTW, it would be good to add some comments to this file since they are hard to understand but public to users.

wgtmac (approved):

.pre-commit-config.yaml
HuaHuaY: The pre-commit CI and cpplint CI are conflicting here and the pre-commit CI must be upgraded.

wgtmac (approved):

PR #699, opened 2026-06-06
feat: add merge append

zhjwpku:
While adding merge append tests, it was discovered that FastAppend did not populate manifest count summary fields, thus add SnapshotUpdate::BuildManifestCountSummary and use it from FastAppend and MergingSnapshotUpdate, which is consistent with java implementation.

### DIFF
--- a/src/iceberg/table.h
+++ b/src/iceberg/table.h
@@ -246,6 +246,8 @@ class ICEBERG_EXPORT StaticTable : public Table {
 
   Result<std::shared_ptr<FastAppend>> NewFastAppend() override;
 
+  Result<std::shared_ptr<MergeAppend>> NewMergeAppend() override;
+
   Result<std::shared_ptr<SnapshotManager>> NewSnapshotManager() override;
 
  private:
--- a/src/iceberg/table.h
+++ b/src/iceberg/table.h
@@ -176,6 +176,9 @@ class ICEBERG_EXPORT Table : public std::enable_shared_from_this<Table> {
   /// \brief Create a new FastAppend to append data files and commit the changes.
   virtual Result<std::shared_ptr<FastAppend>> NewFastAppend();
 
+  /// \brief Create a new MergeAppend to append data files and merge manifests.
+  virtual Result<std::shared_ptr<MergeAppend>> NewMergeAppend();
+
   /// \brief Create a new SnapshotManager to manage snapshots and snapshot references.
   virtual Result<std::shared_ptr<SnapshotManager>> NewSnapshotManager();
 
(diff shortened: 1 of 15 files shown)


src/iceberg/table.h
```diff
@@ -176,6 +176,9 @@ class ICEBERG_EXPORT Table : public std::enable_shared_from_this<Table> {
   /// \brief Create a new FastAppend to append data files and commit the changes.
   virtual Result<std::shared_ptr<FastAppend>> NewFastAppend();
 
+  /// \brief Create a new MergeAppend to append data files and merge manifests.
+  virtual Result<std::shared_ptr<MergeAppend>> NewMergeAppend();
```
wgtmac, line 180: StaticTable also needs to override this and return NotSupported. Otherwise a read-only StaticTable inherits Table::NewMergeAppend(), creates a mutating snapshot update, and later hits the null catalog path during commit. The existing static-table test should cover this alongside NewFastAppend.

src/iceberg/update/merge_append.cc
```diff
@@ -0,0 +1,64 @@
...
+  ICEBERG_BUILDER_CHECK(!manifest.has_existing_files(),
+                        "Cannot append manifest with existing files");
+  ICEBERG_BUILDER_CHECK(!manifest.has_deleted_files(),
+                        "Cannot append manifest with deleted files");
+  ICEBERG_BUILDER_CHECK(manifest.added_snapshot_id == kInvalidSnapshotId,
+                        "Snapshot id must be assigned during commit");
+  ICEBERG_BUILDER_CHECK(manifest.sequence_number == kInvalidSequenceNumber,
+                        "Sequence number must be assigned during commit");
+
+  ICEBERG_BUILDER_RETURN_IF_ERROR(AddManifest(manifest));
```
wgtmac, line 58: This path can reuse a deleted copied manifest on retry. In v1, AddManifest immediately copies into rewritten_append_manifests_. If a commit retry refreshes to a new base, SnapshotUpdate::Apply first calls CleanUncommitted({}), which deletes those copied manifests but keeps the vector. The next Apply can then write a manifest list that points at the deleted path, or fail when merge reads it. FastAppend has the same pattern, but this PR exposes it through MergeAppend; this needs either clearing/re-copying or delaying cleanup until retry is done.

wgtmac:
Let's rebase to resolve conflicts and perhaps address above findings.

src/iceberg/table.h
zhjwpku: Fixed.

src/iceberg/update/merge_append.cc
zhjwpku: Yeah, this was addressed by #707 , let me mark this as resolved.

wgtmac (approved):

PR #700, opened 2026-06-06
fix: align manifest partition reads with reassigned IDs

wgtmac:
Use unknown types for missing partition source fields when building manifest partition schemas. Add Java-style schema ID reassignment maps and use them to map raw manifest partition field IDs back to their original IDs.

Also add transform result type support so unbound partition specs can derive partition field types consistently with Java.


zhjwpku (approved):
LGTM

blackmwk (approved):

wgtmac:
Thank you, @zhjwpku @blackmwk!

PR #701, opened 2026-06-07
feat: metrics integration for commit and scan

evindj:
This change implements part 2 of the core metrics. it covers RestCatalog spec, integration with commit and scan workflows. It leaves the logging metrics reporter out of scope, it will be added in a subsequent diff. 
some other differences is that some of the http calls for the RestCatalogMetrics reporter is still synchronous a subsequent change will cover that.


src/iceberg/catalog/memory/in_memory_catalog.h
```diff
@@ -19,12 +19,15 @@
...
 #pragma once
 
+#include <memory>
 #include <shared_mutex>
 
 #include "iceberg/catalog.h"
 
 namespace iceberg {
 
+class MetricsReporter;
```
wgtmac: Should we add this forward declaration to `iceberg/type_fwd.h`?

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -185,7 +188,7 @@ Result<std::shared_ptr<RestCatalog>> RestCatalog::Make(
 }
 
 RestCatalog::RestCatalog(RestCatalogProperties config, std::shared_ptr<FileIO> file_io,
-                         std::unique_ptr<HttpClient> client,
+                         std::shared_ptr<HttpClient> client,
```
wgtmac, line 456: Why this change is required?

src/iceberg/table_scan.cc
```diff
@@ -538,11 +565,39 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
...
+    std::vector<std::string> projected_field_names;
+    for (const auto& field : schema_ptr->fields()) {
+      projected_field_ids.push_back(field.field_id());
+      projected_field_names.emplace_back(field.name());
+    }
+
+    ScanReport report{
+        .table_name = context_.table_name,
+        .snapshot_id = snapshot->snapshot_id,
+        .filter = context_.filter,
```
wgtmac: This sends the raw scan filter to metrics reporters. Java sanitizes the filter before reporting, so literal predicate values are not exposed. With the current code, a scan like `email = 'x'` posts that value to the REST metrics endpoint or any custom reporter.

src/iceberg/update/snapshot_update.h
```diff
@@ -54,6 +55,15 @@ class ICEBERG_EXPORT SnapshotUpdate : public PendingUpdate {
   Kind kind() const override { return Kind::kUpdateSnapshot; }
   bool IsRetryable() const override { return true; }
 
+  /// \brief Set the metrics reporter for this snapshot update.
+  ///
+  /// \param reporter The metrics reporter to use.
+  /// \return Reference to this for method chaining.
+  auto& ReportWith(this auto& self, std::shared_ptr<MetricsReporter> reporter) {
+    static_cast<SnapshotUpdate&>(self).reporter_ = std::move(reporter);
```
wgtmac, line 69: This stores the reporter on `SnapshotUpdate`, but nothing reads `reporter_` when building the `CommitReport`; `Transaction::Commit` uses `ctx_->table->reporter()` instead. A caller that does `fast_append->ReportWith(custom_reporter)` will not receive a commit report unless the table already has that reporter. Java's `SnapshotProducer.reportWith` overrides the operation reporter.

src/iceberg/table_scan.cc
```diff
@@ -538,11 +565,39 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
...
+  ICEBERG_ASSIGN_OR_RAISE(auto tasks, manifest_group->PlanFiles());
+
+  timed.Stop();
+
+  if (context_.metrics_reporter) {
+    ICEBERG_ASSIGN_OR_RAISE(auto projected_schema, ResolveProjectedSchema());
+    const auto& schema_ptr = projected_schema.get();
+    std::vector<int32_t> projected_field_ids;
+    std::vector<std::string> projected_field_names;
+    for (const auto& field : schema_ptr->fields()) {
```
wgtmac: This only reports top-level projected fields. For nested projections, Java uses `TypeUtil.getProjectedIds(schema())` and includes nested child IDs/names. The C++ scan already uses `GetProjectedIdsVisitor` when resolving selected columns; the report should use the same recursive IDs plus `FindColumnNameById`.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -220,6 +227,24 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> ManifestGroup::PlanFiles() {
   for (auto& task : tasks) {
     file_tasks.push_back(internal::checked_pointer_cast<FileScanTask>(task));
   }
+
+  if (scan_metrics_) {
+    for (const auto& task : file_tasks) {
+      for (const auto& df : task->delete_files()) {
+        scan_metrics_->result_delete_files->Increment(1);
```
wgtmac: These counters are updated per `FileScanTask`, so the same delete file is counted once for every data file it applies to. Java increments indexed/type delete counters once while building `DeleteFileIndex`. A global or partition delete file that matches N data files will inflate `indexed_delete_files` and the delete type counters by N here.

src/iceberg/delete_file_index.cc
```diff
@@ -624,6 +638,8 @@ Result<std::vector<ManifestEntry>> DeleteFileIndex::Builder::LoadDeleteFiles() {
                                               file.equality_ids.end());
         ContentFileUtil::DropUnselectedStats(*entry.data_file, columns);
         files.emplace_back(std::move(entry));
+      } else {
+        if (scan_metrics_) scan_metrics_->skipped_delete_files->Increment(1);
```
wgtmac: This counts delete files older than `min_sequence_number_` as `skipped_delete_files`. Java drops those files in the same sequence-number filter without counting them as skipped. Snapshots with old delete files will report extra skipped delete files.

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -201,10 +204,33 @@ RestCatalog::RestCatalog(RestCatalogProperties config, std::shared_ptr<FileIO> f
       catalog_session_(std::move(catalog_session)),
       snapshot_mode_(snapshot_mode) {
   ICEBERG_DCHECK(catalog_session_ != nullptr, "catalog_session must not be null");
+  const auto& props = config_.configs();
+  auto it = props.find(std::string(kMetricsReporterImpl));
+  if (it != props.end() && !it->second.empty() &&
+      it->second != kMetricsReporterTypeNoop) {
+    if (auto r = MetricsReporters::Load(props); r.has_value()) {
```
wgtmac: `MetricsReporters::Load` errors are silently ignored. If `metrics-reporter-impl` is misspelled or the factory fails, the REST catalog still initializes with no custom reporter and no diagnostic. Java fails catalog initialization for an invalid metrics reporter; this should propagate the load error.

src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -351,7 +352,15 @@ InMemoryCatalog::InMemoryCatalog(
       properties_(std::move(properties)),
       file_io_(std::move(file_io)),
       warehouse_location_(std::move(warehouse_location)),
-      root_namespace_(std::make_unique<InMemoryNamespace>()) {}
+      root_namespace_(std::make_unique<InMemoryNamespace>()) {
+  auto it = properties_.find(std::string(kMetricsReporterImpl));
+  if (it != properties_.end() && !it->second.empty() &&
+      it->second != kMetricsReporterTypeNoop) {
+    if (auto r = MetricsReporters::Load(properties_); r.has_value()) {
```
wgtmac: Same issue here: `MetricsReporters::Load` errors are ignored, so an invalid `metrics-reporter-impl` silently disables reporting for the in-memory catalog. This should propagate the load failure.

src/iceberg/catalog/sql/sql_catalog.cc
```diff
@@ -128,9 +129,15 @@ Result<std::string> ResolveTableLocation(
...
 SqlCatalog::SqlCatalog(SqlCatalogConfig config, std::shared_ptr<FileIO> file_io,
                        std::shared_ptr<CatalogStore> store)
-    : config_(std::move(config)),
-      file_io_(std::move(file_io)),
-      store_(std::move(store)) {}
+    : config_(std::move(config)), file_io_(std::move(file_io)), store_(std::move(store)) {
+  auto it = config_.props.find(std::string(kMetricsReporterImpl));
+  if (it != config_.props.end() && !it->second.empty() &&
+      it->second != kMetricsReporterTypeNoop) {
+    if (auto r = MetricsReporters::Load(config_.props); r.has_value()) {
```
wgtmac: Same issue here: `MetricsReporters::Load` errors are ignored, so an invalid `metrics-reporter-impl` silently disables reporting for the SQL catalog. This should propagate the load failure.

src/iceberg/table.cc
```diff
@@ -151,12 +154,23 @@ const std::shared_ptr<TableMetadata>& Table::metadata() const { return metadata_
...
+}
+
 Result<std::unique_ptr<LocationProvider>> Table::location_provider() const {
   return LocationProvider::Make(metadata_->location, metadata_->properties);
 }
 
 Result<std::unique_ptr<DataTableScanBuilder>> Table::NewScan() const {
-  return DataTableScanBuilder::Make(metadata_, io_);
+  ICEBERG_ASSIGN_OR_RAISE(auto builder, DataTableScanBuilder::Make(metadata_, io_));
+  builder->TableName(identifier_.ToString());
```
wgtmac: This uses the table identifier, not the fully qualified table name. Java reports `BaseTable.name()`, and REST builds that as `catalog.namespace.table`. Scan reports from two catalogs with the same identifier will collide and differ from Java REST payloads.

src/iceberg/transaction.cc
```diff
@@ -378,6 +392,27 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
+  // commits).
+  const auto& reporter = ctx_->table->reporter();
+  if (reporter) {
+    auto snapshot_result = ctx_->table->metadata()->Snapshot();
+    if (snapshot_result.has_value() && snapshot_result.value() &&
+        snapshot_result.value()->snapshot_id != pre_commit_snapshot_id) {
+      const auto& snapshot = snapshot_result.value();
+      const auto op = snapshot->Operation();
+      CommitReport report{
+          .table_name = ctx_->table->name().ToString(),
```
wgtmac: Same table-name issue for commit reports: this uses `TableIdentifier::ToString()`, so the catalog name is missing. Java commit reports use the table's full name (`catalog.namespace.table`), which avoids collisions across catalogs.

src/iceberg/catalog/rest/rest_catalog.cc
evindj: because MakeTableReporter() hands the same client to a per-table RestMetricsReporter that must keep posting metrics independently of the catalog's lifetime. alternatively we can make the table reporter own an httpClient but I feel like the burden of initialization could be high

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -483,6 +495,14 @@ RestCatalog::RestCatalog(RestCatalogProperties config, std::shared_ptr<FileIO> f
       snapshot_mode_(snapshot_mode),
       default_context_(std::move(default_context)) {
   ICEBERG_DCHECK(catalog_session_ != nullptr, "catalog_session must not be null");
+  const auto& props = config_.configs();
+  auto it = props.find(std::string(kMetricsReporterImpl));
+  if (it != props.end() && !it->second.empty() &&
+      it->second != kMetricsReporterTypeNoop) {
+    if (auto r = MetricsReporters::Load(props); r.has_value()) {
```
wgtmac: The REST catalog loads the custom metrics reporter twice: once in the constructor here and again in `RestCatalog::Make` after construction. That means a valid reporter factory is constructed and initialized twice, and the first instance is immediately discarded. Reporter initialization can have side effects or allocate resources; Java loads and initializes the configured reporter once. The constructor load should be removed and the `Make` path should be the single source of truth.

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -524,6 +544,21 @@ Result<std::shared_ptr<FileIO>> RestCatalog::TableFileIO(
...
 
+std::shared_ptr<MetricsReporter> RestCatalog::MakeTableReporter(
+    const TableIdentifier& identifier) const {
+  auto enabled = config_.Get(RestCatalogProperties::kMetricsReportingEnabled);
+  if (StringUtils::ToLower(enabled) == "true" &&
+      supported_endpoints_.contains(Endpoint::ReportMetrics())) {
+    auto path = paths_->Metrics(identifier);
+    if (path.has_value()) {
+      auto rest_reporter =
+          std::make_shared<RestMetricsReporter>(client_, *path, catalog_session_);
```
wgtmac: This `RestMetricsReporter` still uses the catalog-level auth session even after `MakeTableFromLoadResult` creates a table-scoped session for the table. Java wires the REST metrics reporter with the table REST client, so metrics POSTs use the same table-scoped credentials/headers as table operations. With a REST catalog that returns table-specific credentials, scan/commit metrics can be posted with the wrong auth and get silently dropped.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -408,19 +438,27 @@ ManifestGroup::ReadEntries() {
                                 manifest_evaluator->Evaluate(manifest));
         if (!should_match) {
           // Skip this manifest because it doesn't match partition filter
+          if (scan_metrics_) {
+            scan_metrics_->skipped_data_manifests->Increment(1);
+          }
           return {};
         }
+        if (scan_metrics_) {
```
wgtmac: `scanned_data_manifests` is incremented before the later `ignore_deleted_` / `ignore_existing_` manifest filters run. Java counts scanned data manifests only after those manifest-level filters, so a manifest skipped by either filter will be reported as scanned here.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -408,19 +438,27 @@ ManifestGroup::ReadEntries() {
...
+          scan_metrics_->scanned_data_manifests->Increment(1);
+        }
 
         if (ignore_deleted_) {
           // only scan manifests that have entries other than deletes
           if (!manifest.has_added_files() && !manifest.has_existing_files()) {
+            if (scan_metrics_) scan_metrics_->skipped_data_files->Increment(1);
```
wgtmac: This is a manifest-level skip, not a skipped data file. Java increments `skippedDataManifests` for `ignoreDeleted` manifest pruning, so a manifest containing only deleted entries should not add to `skipped_data_files`.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -408,19 +438,27 @@ ManifestGroup::ReadEntries() {
...
             return {};
           }
         }
 
         if (ignore_existing_) {
           // only scan manifests that have entries other than existing
           if (!manifest.has_added_files() && !manifest.has_deleted_files()) {
+            if (scan_metrics_) scan_metrics_->skipped_data_files->Increment(1);
```
wgtmac: Same counter issue for `ignoreExisting`: this skips the whole manifest, so it should increment `skipped_data_manifests` instead of `skipped_data_files`.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -346,6 +372,10 @@ Result<std::unique_ptr<ManifestReader>> ManifestGroup::MakeReader(
       .CaseSensitive(case_sensitive_)
       .Select(std::move(columns));
 
+  if (scan_metrics_) {
```
wgtmac, line 370: `SkipCounter` only covers filtering done inside `ManifestReader`. The later local filters in `ReadEntries` (`ignore_existing_`, `data_file_evaluator`, and `manifest_entry_predicate_`) still `continue` without incrementing `skipped_data_files`, unlike Java's entry-level filters wrapped with `skippedDataFiles`.

src/iceberg/catalog/rest/meson.build
```diff
@@ -30,6 +30,7 @@ iceberg_rest_sources = files(
     'resource_paths.cc',
     'rest_catalog.cc',
     'rest_file_io.cc',
+    'rest_metrics_reporter.cc',
```
wgtmac, line 33: Since this adds `RestMetricsReporter`, the Meson install list should also include `rest_metrics_reporter.h`; otherwise CMake installs the new public header but Meson installs do not.

src/iceberg/test/CMakeLists.txt
```diff
@@ -301,6 +302,7 @@ if(ICEBERG_BUILD_REST)
                         endpoint_test.cc
                         rest_file_io_test.cc
                         rest_json_serde_test.cc
+                        rest_metrics_reporter_test.cc
```
wgtmac, line 304: CMake now builds `rest_metrics_reporter_test.cc`, but `src/iceberg/test/meson.build` is not updated. Meson test runs will miss this coverage.

src/iceberg/meson.build
```diff
@@ -62,6 +62,7 @@ iceberg_sources = files(
     'expression/projections.cc',
     'expression/residual_evaluator.cc',
     'expression/rewrite_not.cc',
+    'expression/sanitize_expression.cc',
```
wgtmac, line 91: `sanitize_expression.h` is a new public header, but `src/iceberg/expression/meson.build` does not install it. Meson installs will miss this API even though the source is built.

src/iceberg/catalog/rest/rest_metrics_reporter.cc
```diff
@@ -0,0 +1,69 @@
...
+    return {};
+  }
+
+  // Inject "report-type" required by the REST spec (not included in core ToJson).
+  auto& json = json_result.value();
+  json[kReportType] =
+      std::holds_alternative<ScanReport>(report) ? kScanReportType : kCommitReportType;
+
+  // POST to the metrics endpoint; suppress errors to match Java fire-and-forget behavior.
+  std::ignore = client_->Post(metrics_endpoint_, json.dump(), /*headers=*/{},
```
wgtmac: `Report()` must not throw, but `json.dump()` and the HTTP call can throw here. Since scan/commit callers ignore the returned status, this should catch exceptions locally and keep the fire-and-forget metrics path from unwinding user operations.

src/iceberg/expression/sanitize_expression.cc
```diff
@@ -0,0 +1,347 @@
...
+}
+
+std::string SanitizeTimestamp(int64_t micros, int64_t now) {
+  constexpr int64_t kMicrosPerHour = 60LL * 60LL * 1'000'000LL;
+  constexpr int64_t kFiveMinutesInMicros = 5LL * 60LL * 1'000'000LL;
+  constexpr int64_t kThreeDaysInHours = 3LL * 24LL;
+  constexpr int64_t kNinetyDaysInHours = 90LL * 24LL;
+
+  std::string is_past = now > micros ? "ago" : "from-now";
+  int64_t diff = std::abs(now - micros);
```
wgtmac: This subtraction can overflow before `std::abs` when the sanitized timestamp is near the `int64_t` limits. It is an edge case, but using a saturating or unsigned distance would avoid undefined behavior in the sanitizer.

src/iceberg/table_scan.cc
```diff
@@ -545,11 +574,50 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
...
+                                *schema_ptr, /*include_struct_ids=*/true));
+    std::vector<int32_t> projected_field_ids(projected_id_set.begin(),
+                                             projected_id_set.end());
+    std::ranges::sort(projected_field_ids);
+
+    std::vector<std::string> projected_field_names;
+    projected_field_names.reserve(projected_field_ids.size());
+    for (int32_t field_id : projected_field_ids) {
+      ICEBERG_ASSIGN_OR_RAISE(auto field_name, schema_ptr->FindColumnNameById(field_id));
+      projected_field_names.emplace_back(field_name.value_or(""));
```
wgtmac: `field_id` comes from the projected schema, so failing to resolve its name would indicate a schema/cache bug. Emitting an empty field name would hide that invariant violation in the scan report; this should return an error instead of `value_or("")`.

src/iceberg/transaction.cc
```diff
@@ -394,6 +412,30 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
   ctx_->table = std::move(commit_result.value());
 
+  // Fire CommitReport only when a new snapshot was produced (not for property-only
+  // commits). A SnapshotUpdate's own ReportWith() override (captured into
+  // snapshot_reporter_ by ApplyUpdateSnapshot()) takes precedence over the table's
+  // reporter.
+  std::shared_ptr<MetricsReporter> reporter =
+      snapshot_reporter_ ? snapshot_reporter_ : ctx_->table->reporter();
+  if (reporter) {
+    auto snapshot_result = ctx_->table->metadata()->Snapshot();
```
wgtmac: This derives the commit report from the table's current snapshot after the commit. That misses `StageOnly()` updates, where the new snapshot is not current, and it can report rollback/set-current operations as if they produced a new commit. Java reports from the snapshot-producing operation event, so this should use the `SnapshotUpdate` result instead of the post-commit current snapshot.

src/iceberg/transaction.cc
```diff
@@ -275,6 +277,10 @@ Status Transaction::ApplyUpdateSnapshot(SnapshotUpdate& update) {
 
   ICEBERG_ASSIGN_OR_RAISE(auto result, update.Apply());
 
+  if (const auto& override_reporter = update.reporter()) {
```
wgtmac: This stores the per-update reporter on the transaction and never clears it. If one snapshot update calls `ReportWith(custom)`, later snapshot updates in the same transaction inherit that reporter even when they did not opt in. The override should be scoped to the update that produced the snapshot.

src/iceberg/table.cc
```diff
@@ -219,7 +255,11 @@ Result<std::shared_ptr<UpdateLocation>> Table::NewUpdateLocation() {
 Result<std::shared_ptr<FastAppend>> Table::NewFastAppend() {
   ICEBERG_ASSIGN_OR_RAISE(
       auto ctx, TransactionContext::Make(shared_from_this(), TransactionKind::kUpdate));
-  return FastAppend::Make(name().name, std::move(ctx));
+  ICEBERG_ASSIGN_OR_RAISE(auto op, FastAppend::Make(name().name, std::move(ctx)));
+  if (reporter_) {
+    op->ReportWith(reporter_);
```
wgtmac: This only propagates the table reporter to fast append. Other snapshot-producing operations such as merge append, row delta, overwrite, delete files, and rewrite files will not emit commit reports unless the caller sets a reporter on each operation. Java table operations inherit the table reporter consistently.

src/iceberg/delete_file_index.cc
```diff
@@ -638,8 +644,10 @@ Result<std::vector<ManifestEntry>> DeleteFileIndex::Builder::LoadDeleteFiles() {
           ICEBERG_ASSIGN_OR_RAISE(auto should_match,
                                   manifest_evaluator->Evaluate(manifest));
           if (!should_match) {
+            if (scan_metrics_) scan_metrics_->skipped_delete_manifests->Increment(1);
```
wgtmac, line 648: This only counts delete manifests skipped by the partition evaluator. A delete manifest with no added or existing files returns earlier and is not counted as `skipped_delete_manifests`, even though Java counts that manifest-level skip.

src/iceberg/delete_file_index.cc
```diff
@@ -638,8 +644,10 @@ Result<std::vector<ManifestEntry>> DeleteFileIndex::Builder::LoadDeleteFiles() {
...
             return manifest_result;
           }
+          if (scan_metrics_) scan_metrics_->scanned_delete_manifests->Increment(1);
```
wgtmac: The delete manifest entry filters below are still not wired to a `SkipCounter`, so delete files filtered by partition stats or the partition set are not counted in `skipped_delete_files`. Java wraps those entry-level filters with the skipped-delete-files counter.

src/iceberg/table_scan.cc
```diff
@@ -557,11 +586,52 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
...
+    std::vector<std::string> projected_field_names;
+    projected_field_names.reserve(projected_field_ids.size());
+    for (int32_t field_id : projected_field_ids) {
+      ICEBERG_ASSIGN_OR_RAISE(auto field_name, schema_ptr->FindColumnNameById(field_id));
+      ICEBERG_CHECK(field_name.has_value(), "Projected field {} not found in schema",
+                    field_id);
+      projected_field_names.emplace_back(*field_name);
+    }
+
+    ICEBERG_ASSIGN_OR_RAISE(auto sanitized_filter,
```
wgtmac: This sanitizes the raw unbound filter without the table schema or case-sensitivity setting. Java binds the expression before sanitizing, which preserves resolved field names and transform terms. Case-insensitive scans or filters with renamed/case-variant fields can report a different sanitized filter here.

src/iceberg/util/content_file_util.cc
```diff
@@ -83,6 +83,10 @@ std::string ContentFileUtil::DVDesc(const DataFile& file) {
                      file.referenced_data_file.value_or(""));
 }
 
+int64_t ContentFileUtil::ContentSizeInBytes(const DataFile& file) {
```
wgtmac, line 86: This uses `content_size_in_bytes` for every content file when it is present. Java only uses content size for deletion vectors; normal data/delete files should report `file_size_in_bytes`. Otherwise any non-DV file that carries this optional field will inflate or change delete metrics.

src/iceberg/catalog/rest/rest_metrics_reporter.cc
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    // Inject "report-type" required by the REST spec (not included in core ToJson).
+    auto& json = json_result.value();
+    json[kReportType] =
+        std::holds_alternative<ScanReport>(report) ? kScanReportType : kCommitReportType;
+
+    // POST to the metrics endpoint; suppress errors to match Java fire-and-forget
+    // behavior.
+    std::ignore = client_->Post(metrics_endpoint_, json.dump(), /*headers=*/{},
```
wgtmac: Nit/TODO: this POST is still synchronous. Java sends REST metrics asynchronously and fire-and-forget, so a slow metrics endpoint should not block scan planning or commits.

src/iceberg/test/CMakeLists.txt
```diff
@@ -223,6 +223,7 @@ if(ICEBERG_BUILD_BUNDLE)
                    file_scan_task_test.cc
                    incremental_append_scan_test.cc
                    incremental_changelog_scan_test.cc
+                   scan_planning_metrics_test.cc
```
wgtmac, line 219: This adds `scan_planning_metrics_test.cc` to CMake only. `src/iceberg/test/meson.build` should include the same test, otherwise Meson test runs miss this coverage.

src/iceberg/expression/sanitize_expression.cc
```diff
@@ -0,0 +1,358 @@
...
+std::string SanitizeTimestamp(int64_t micros, int64_t now) {
+  constexpr int64_t kMicrosPerHour = 60LL * 60LL * 1'000'000LL;
+  constexpr int64_t kFiveMinutesInMicros = 5LL * 60LL * 1'000'000LL;
+  constexpr int64_t kThreeDaysInHours = 3LL * 24LL;
+  constexpr int64_t kNinetyDaysInHours = 90LL * 24LL;
+
+  std::string is_past = now > micros ? "ago" : "from-now";
+  // See SanitizeDate() for why this avoids `std::abs(now - micros)`: the subtraction
+  // can overflow int64_t (UB) when the values are near the type's limits.
+  int64_t diff = now > micros ? static_cast<int64_t>(static_cast<uint64_t>(now) -
```
wgtmac: The unsigned subtraction avoids overflow, but casting the result back to `int64_t` can wrap when the distance is greater than `INT64_MAX`. Extreme timestamp/date values can then look negative and be classified as `about-now` or as a small distance. This should saturate or keep the distance unsigned before comparing.

src/iceberg/test/rest_metrics_reporter_test.cc
```diff
@@ -0,0 +1,74 @@
...
+    session_ = auth::AuthSession::MakeDefault({});
+  }
+
+  std::shared_ptr<HttpClient> client_;
+  std::shared_ptr<auth::AuthSession> session_;
+};
+
+// Report() must return OK even when the HTTP call fails (connection refused).
+// This validates the fire-and-forget error-suppression contract matching Java behavior.
+TEST_F(RestMetricsReporterTest, ReportSuppressesHttpErrorsForScanReport) {
```
wgtmac: This only verifies that errors are suppressed. It does not assert that the reporter posts to the configured metrics endpoint, includes the required `report-type`, or serializes scan vs commit bodies correctly. A mock or local test server would catch REST payload regressions here.

src/iceberg/table.cc
```diff
@@ -219,7 +255,8 @@ Result<std::shared_ptr<UpdateLocation>> Table::NewUpdateLocation() {
 Result<std::shared_ptr<FastAppend>> Table::NewFastAppend() {
   ICEBERG_ASSIGN_OR_RAISE(
       auto ctx, TransactionContext::Make(shared_from_this(), TransactionKind::kUpdate));
-  return FastAppend::Make(name().name, std::move(ctx));
+  ICEBERG_ASSIGN_OR_RAISE(auto op, FastAppend::Make(name().name, std::move(ctx)));
```
evindj: possible merge conflicts stuff I will have a look.

wgtmac:
I haven't finished reading this PR yet. Codex published my pending comments by accident. Will let you know when I'm done.

wgtmac:
Thanks for the work on this PR! I did a detailed code review and pushed the changes directly to the branch to avoid a long review-and-revision loop.

Summary of the changes:

- Aligned table full-name handling with Java’s catalog-owned design and preserved REST’s catalog-specific naming behavior.
- Removed unnecessary public APIs, including reporter-combination and legacy table-scan builder APIs.
- Simplified commit reporting so each snapshot update owns its metrics and reports once, only after a successful standalone or transaction commit.
- Reused retry-runner attempt accounting and removed redundant commit-attempt state.
- Aligned scan metrics accounting and naming across table scans, manifest planning, delete-file indexing, and deletion vectors.
- Improved Java parity for expression sanitization, including temporal values, hashes, transforms, unary predicates, and overflow handling.
- Reworked catalog reporter initialization and executor handling. REST reporting now supports both asynchronous execution with a user-provided executor and synchronous fallback without one.
- Propagated catalog reporter construction errors through `Result` instead of silently ignoring them.
- Strengthened and simplified the related tests, fixtures, names, and comments.

wgtmac (approved):

PR #702, opened 2026-06-07
fix(rest): use "ref" field name for AssertRefSnapshotId requirement

plusplusjiajia:
The `AssertRefSnapshotId` table requirement was serialized and deserialized with the JSON field name `ref-name`. Per the Iceberg REST OpenAPI spec (`rest-catalog-open-api.yaml`), this requirement's field is named `ref`. Only the `SetSnapshotRef` / `RemoveSnapshotRef` *updates* use `ref-name`


yingcai-cy (approved):

wgtmac (approved):
Nice catch!

PR #703, opened 2026-06-07
feat(arrow): resolve S3-compatible schemes in ArrowFileSystemFileIO

plusplusjiajia:
`ArrowFileSystemFileIO::ResolvePath` passed the full URI to the wrapped FileSystem's `PathFromUri`, which only accepts that FileSystem's native scheme(`s3://` for `S3FileSystem`). S3-compatible object stores are commonly addressed with other schemes — `s3a`/`s3n`, or vendor schemes such as `gs://` (GCS) and`oss://` (Alibaba OSS) — served by an `arrow::fs::S3FileSystem` configured with an `endpoint_override`. Every read and write of such a location failed with:
```
expected a URI with one of the schemes (s3) but received <scheme>://...
```

### DIFF
--- a/src/iceberg/arrow/arrow_io.cc
+++ b/src/iceberg/arrow/arrow_io.cc
@@ -473,16 +473,26 @@ class ArrowOutputFile : public OutputFile {
 }  // namespace
 
 Result<std::string> ArrowFileSystemFileIO::ResolvePath(const std::string& file_location) {
-  if (auto pos = file_location.find("://"); pos != std::string::npos) {
-    auto path = arrow_fs_->PathFromUri(file_location);
-    if (path.ok()) {
-      return path.ValueOrDie();
-    }
-    // PathFromUri rejects S3-compatible schemes (s3a/s3n, gs://, oss://);
-    // fall back to the scheme-less bucket/key.
-    return file_location.substr(pos + 3);
+  const auto pos = file_location.find("://");
+  if (pos == std::string::npos) {
+    return file_location;
   }
-  return file_location;
+
+  auto path = arrow_fs_->PathFromUri(file_location);
+  if (path.ok()) {
+    return std::move(path).ValueOrDie();
+  }
+
+  // Only fall back for Arrow's scheme-mismatch error; propagate anything else.
+  const auto& status = path.status();
+  if (status.message().find("expected a URI with one of the schemes") ==
+      std::string::npos) {
+    return std::unexpected<Error>{
+        {.kind = ToErrorKind(status), .message = status.ToString()}};
+  }
+  // Scheme-less bucket/key, dropping any ?query / #fragment.
+  std::string bucket_key = file_location.substr(pos + 3);
+  return bucket_key.substr(0, bucket_key.find_first_of("?#"));
 }
 
 Result<std::shared_ptr<::arrow::io::RandomAccessFile>> OpenArrowInputStream(
--- a/src/iceberg/arrow/arrow_io.cc
+++ b/src/iceberg/arrow/arrow_io.cc
@@ -30,6 +30,7 @@
 #include <arrow/io/interfaces.h>
 #include <arrow/result.h>
 #include <arrow/status.h>
+#include <arrow/util/uri.h>
 
 #include "iceberg/arrow/arrow_io_internal.h"
 #include "iceberg/arrow/arrow_io_util.h"
@@ -490,9 +491,10 @@ Result<std::string> ArrowFileSystemFileIO::ResolvePath(const std::string& file_l
     return std::unexpected<Error>{
         {.kind = ToErrorKind(status), .message = status.ToString()}};
   }
-  // Scheme-less bucket/key, dropping any ?query / #fragment.
+  // Scheme-less bucket/key: drop ?query / #fragment and percent-decode to match s3://.
   std::string bucket_key = file_location.substr(pos + 3);
-  return bucket_key.substr(0, bucket_key.find_first_of("?#"));
+  bucket_key = bucket_key.substr(0, bucket_key.find_first_of("?#"));
+  return ::arrow::util::UriUnescape(bucket_key);
 }
 
 Result<std::shared_ptr<::arrow::io::RandomAccessFile>> OpenArrowInputStream(
--- a/src/iceberg/arrow/arrow_io.cc
+++ b/src/iceberg/arrow/arrow_io.cc
@@ -484,14 +484,13 @@ Result<std::string> ArrowFileSystemFileIO::ResolvePath(const std::string& file_l
     return std::move(path).ValueOrDie();
   }
 
-  // Only fall back for Arrow's scheme-mismatch error; propagate anything else.
-  const auto& status = path.status();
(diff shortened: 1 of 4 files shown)


src/iceberg/arrow/arrow_io.cc
```diff
@@ -473,9 +473,14 @@ class ArrowOutputFile : public OutputFile {
...
-    ICEBERG_ARROW_ASSIGN_OR_RETURN(auto path, arrow_fs_->PathFromUri(file_location));
-    return path;
+  if (auto pos = file_location.find("://"); pos != std::string::npos) {
+    auto path = arrow_fs_->PathFromUri(file_location);
+    if (path.ok()) {
+      return path.ValueOrDie();
+    }
+    // PathFromUri rejects S3-compatible schemes (s3a/s3n, gs://, oss://);
+    // fall back to the scheme-less bucket/key.
+    return file_location.substr(pos + 3);
```
MisterRaindrop: I think this fallback is too broad. As implemented, any URI scheme rejected by PathFromUri will be stri
  pped and interpreted as a path in the wrapped filesystem. For example, with an S3FileSystem, a mismatch ed URI such as file://bucket/prefix would be interpreted as bucket/prefix instead of failing fast.

Can we make the fallback explicit and narrow?

  1. The scheme should be in an allowlist for the wrapped filesystem. For S3FileSystem, I think we can
  support well-known aliases such as s3a and s3n. Vendor schemes such as oss or gs should probably be opt- in or discussed separately, because they may also represent native OSS/GCS locations.
  2. The PathFromUri failure should match the expected scheme-mismatch / unsupported-scheme case. Other errors, such as malformed URI, authority handling errors, or local-path errors, should still be returned to the caller.

  In other words, this should be guarded by both a scheme allowlist and the expected error condition,
  rather than falling back for every PathFromUri failure.

plusplusjiajia: @MisterRaindrop Agreed — the scheme allowlist belongs at the selection layer (like Java's ResolvingFileIO.SCHEME_TO_FILE_IO), so DetectBuiltinFileIO now accepts s3a/s3n. For the error condition: ResolvePath now only falls back on Arrow's scheme-mismatch error — any other PathFromUri failure (malformed URI, unsupported authority, …) is propagated — and it strips ?query/#fragment. oss/gs left for separate discussion, as you suggested.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -473,9 +473,14 @@ class ArrowOutputFile : public OutputFile {
...
   }
```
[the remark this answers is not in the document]
plusplusjiajia: Addressed: the fallback now triggers only on Arrow's scheme-mismatch error; other PathFromUri failures (malformed URI, unsupported authority, …) propagate. ?query/#fragment are stripped. Added tests for the propagate path and for preserving percent-encoded keys.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -473,11 +473,26 @@ class ArrowOutputFile : public OutputFile {
...
+  // Only fall back for Arrow's scheme-mismatch error; propagate anything else.
+  const auto& status = path.status();
+  if (status.message().find("expected a URI with one of the schemes") ==
+      std::string::npos) {
+    return std::unexpected<Error>{
+        {.kind = ToErrorKind(status), .message = status.ToString()}};
+  }
+  // Scheme-less bucket/key, dropping any ?query / #fragment.
+  std::string bucket_key = file_location.substr(pos + 3);
+  return bucket_key.substr(0, bucket_key.find_first_of("?#"));
```
MisterRaindrop: One subtle concern: s3:// and s3a:// currently use different parsing semantics. s3:// goes through Arrow's PathFromUri, while s3a/s3n use the fallback substring parser. This means percent-encoded keys may resolve differently.
example:
s3://bucket/a%20b   -> Arrow PathFromUri -> bucket/a b
  s3a://bucket/a%20b  -> fallback          -> bucket/a%20b

plusplusjiajia: @MisterRaindrop Good catch — fixed. The fallback now percent-decodes the key via arrow::util::UriUnescape, so s3a/s3n resolve identically to native s3:// (a%20b → a b).

src/iceberg/arrow/arrow_io.cc
```diff
@@ -473,11 +474,27 @@ class ArrowOutputFile : public OutputFile {
...
-  return file_location;
+
+  auto path = arrow_fs_->PathFromUri(file_location);
+  if (path.ok()) {
+    return std::move(path).ValueOrDie();
+  }
+
+  // Only fall back for Arrow's scheme-mismatch error; propagate anything else.
+  const auto& status = path.status();
+  if (status.message().find("expected a URI with one of the schemes") ==
```
MisterRaindrop: This approach is problematic. It relies on Arrow's error messages, which will break if Arrow changes the message text. It does not simultaneously check status.code()

plusplusjiajia: @MisterRaindrop Good point — dropped the error-text match. ResolvePath now validates the URI with arrow::util::Uri::Parse (a malformed URI fails there and propagates) and then percent-decodes the scheme-less bucket/key, so it no longer depends on Arrow's message text or status code.

MisterRaindrop:
Thanks for @plusplusjiajia contribution. There are a few minor issues with this PR.
I agree that the allowlist in DetectBuiltinFileIO covers the initial catalog-level FileIO selection from warehouse/io-impl.

My remaining concern is that later metadata/manifest/delete locations may reach the already-created FileIO directly, so
they are not necessarily guarded by DetectBuiltinFileIO. That said, I also agree that ResolvePath is not the right place to implement global scheme routing or rejection.

For the scope of this PR, this looks acceptable to me. I think we should follow up with a separate issue to discuss a higher-level resolving FileIO, similar to Java's ResolvingFileIO, for per-location scheme dispatch/enforcement.

plusplusjiajia:
> Thanks for @plusplusjiajia contribution. There are a few minor issues with this PR. I agree that the allowlist in DetectBuiltinFileIO covers the initial catalog-level FileIO selection from warehouse/io-impl.
> 
> My remaining concern is that later metadata/manifest/delete locations may reach the already-created FileIO directly, so they are not necessarily guarded by DetectBuiltinFileIO. That said, I also agree that ResolvePath is not the right place to implement global scheme routing or rejection.
> 
> For the scope of this PR, this looks acceptable to me. I think we should follow up with a separate issue to discuss a higher-level resolving FileIO, similar to Java's ResolvingFileIO, for per-location scheme dispatch/enforcement.

Thanks @MisterRaindrop! Agreed on both. I'll open a follow-up issue to track it. Thanks for the careful review!

MisterRaindrop (approved):
LGTM

wgtmac (approved):
Thanks @plusplusjiajia for working on this and @MisterRaindrop for the review! I think this PR adds support for s3 scheme extension but still do not support other S3-compatible schemes like `oss://`, `cos://`, etc. But that may take more work including scheme allow-list and vendor-specific configuration.

PR #705, opened 2026-06-08
chore: release candidate version string

zhjwpku:


manuzhang:
Is it possible to catch this in `verify_rc.sh`?

zhjwpku:
> Is it possible to catch this in `verify_rc.sh`?

Yeah, we can catch this by checking the contents of `ICEBERG_VERSION_STRING`. I'm planning to fix it after this release cycle. We currently don't have `ICEBERG_VERSION_STRING` defined in the Meson build, which is another issue that needs to be addressed.

wgtmac (approved):
Let's move forward to unblock release process. Thanks @zhjwpku and @manuzhang!

PR #708, opened 2026-06-09
feat: MergingSnapshotUpdate merges multiple DVs

HuaHuaY:


src/iceberg/puffin/puffin_writer.cc
```diff
@@ -152,10 +152,10 @@ Status PuffinWriter::Finish() {
 
   ICEBERG_ASSIGN_OR_RAISE(auto end_pos, stream_->Position());
   footer_size_ = end_pos - footer_start;
+  file_size_ = end_pos;
   footer_written_ = true;
   ICEBERG_RETURN_UNEXPECTED(stream_->Flush());
   ICEBERG_RETURN_UNEXPECTED(stream_->Close());
-  ICEBERG_ASSIGN_OR_RAISE(file_size_, stream_->Position());
```
wgtmac: I remember that this is an intentional design because Java puffin writer will append some bytes after close in some cases that we haven't implemented so far.

src/iceberg/meson.build
```diff
@@ -49,6 +49,7 @@ iceberg_sources = files(
     'catalog/session_catalog.cc',
     'catalog/session_context.cc',
     'delete_file_index.cc',
+    'deletes/roaring_position_bitmap.cc',
```
wgtmac: Perhaps we need to move it out of the `deletes` subdirectory.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -757,27 +767,137 @@ MergingSnapshotUpdate::MergeDVs() const {
...
+             blob.length == dv.file->content_size_in_bytes.value();
+    });
+    ICEBERG_PRECHECK(blob_it != metadata.blobs.end(),
+                     "Cannot find DV blob at offset {} with length {} in {}",
+                     dv.file->content_offset.value(),
+                     dv.file->content_size_in_bytes.value(), dv.file->file_path);
+    ICEBERG_ASSIGN_OR_RAISE(auto blob, reader->ReadBlob(*blob_it));
+    ICEBERG_ASSIGN_OR_RAISE(
+        auto dv_bitmap,
+        RoaringPositionBitmap::Deserialize(std::string_view(
```
wgtmac: Java/spec-compliant deletion-vector-v1 blobs are not raw Roaring payloads: they include the 4-byte length, magic bytes, bitmap, and CRC, and Java validates CRC/cardinality during deserialize. `PuffinReader::ReadBlob` returns the stored blob bytes unchanged, so passing them directly to `RoaringPositionBitmap::Deserialize()` will reject or misread DVs written by Java. The writer below has the symmetric issue because `bitmap.Serialize()` writes only the raw bitmap. Please add DV-level framing/parsing here.

src/iceberg/update/merging_snapshot_update.cc
```diff
@@ -757,27 +767,137 @@ MergingSnapshotUpdate::MergeDVs() const {
...
+            reinterpret_cast<const char*>(blob.second.data()), blob.second.size())));
+    bitmap.Or(dv_bitmap);
+  }
+
+  bitmap.Optimize();
+  ICEBERG_ASSIGN_OR_RAISE(auto serialized, bitmap.Serialize());
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto blob_metadata,
+      writer.Write(puffin::Blob{
```
wgtmac: This Puffin blob metadata does not match Java or the Puffin DV spec. Java `BaseDVFileWriter` writes `fields=[ROW_POSITION]`, `snapshot-id=-1`, `sequence-number=-1`, and `referenced-data-file`/`cardinality` properties. Writing empty fields, concrete snapshot/sequence values, and no properties produces invalid DV footer metadata for other implementations. Please mirror the Java DV writer metadata.

src/iceberg/puffin/puffin_writer.cc
HuaHuaY: Revert this and it's fixed in PR https://github.com/apache/iceberg-cpp/pull/773 that `Position` can't be called after `Close` in some kinds of stream.

wgtmac:
@zhaoxuan1994 is working on adding puffin reader/writer for v3 DVs. Let's wait for it and then rebase on it later.

wgtmac:
I've made several changes to this PR:
- Kept puffin and dv code in `iceberg-data` library to make the core library IO-less still.
- Refactored DV logic based on a recent commit which added DV writer and reader support.
- Removed `misc-include-cleaner` from clang-tidy since I found it is useless but adds a lot of noise by blocking every ci until we add a lot of lengthy include statements.
- Removed ci workflow for sigv4 with meson. Recently vcpkg update introduced a breaking change of aws sdk. I think this workflow is fragile but there is no active Meson expert. Considering that Meson is still partially supported, I don't want to waste time and burn tokens to fix it. So I directly removed this workflow and may eventually remove all Meson code if unmaintained for a long time (perhaps a year?).

wgtmac (approved):

.clang-tidy
```diff
@@ -21,7 +21,6 @@ Checks: |
   clang-analyzer-*,
   google-*,
   modernize-*,
-  misc-include-cleaner,
```
wgtmac, line 24: https://github.com/apache/iceberg-cpp/pull/820 has been reverted via this change because it does not add too much value but a lot of noise. cc @yadavay-amzn @zhjwpku

PR #709, opened 2026-06-09
feat: add DeleteFiles update API

manuzhang:
## Summary
- Add a DeleteFiles snapshot update API and wire it through table and transaction update flows.
- Add DeleteFiles implementation/build integration.
- Add coverage for path matching, case-insensitive row filters, empty delete commits, and strict-projection partial-match rejection.

## Validation
- `cmake --build build --target table_update_test`
- `./build/src/iceberg/test/table_update_test '--gtest_filter=DeleteFilesTest.*'`

### DIFF
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -99,6 +99,7 @@ set(ICEBERG_SOURCES
     transform.cc
     transform_function.cc
     type.cc
+    update/delete_files.cc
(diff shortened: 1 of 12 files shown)


wgtmac (approved):

wgtmac:
Thanks for adding this! @manuzhang

manuzhang:
Thanks for review! @wgtmac

PR #710, opened 2026-06-09
build: install generated version header

manuzhang:
Install the generated iceberg/version.h with the rest of the public headers so installed packages expose the version macros.


wgtmac (approved):
Nice catch!

Issue #502, opened 2026-01-09
Add support for Arrow LargeListArray in Parquet data projection


callmepandey:
## Summary

The `ProjectRecordBatch` function in `parquet_data_util.cc` only supports `::arrow::ListArray` (32-bit offsets) but not `::arrow::LargeListArray` (64-bit offsets). This limitation is marked with a FIXME comment at line 151.

## Problem

Arrow's `LargeListArray` uses 64-bit offsets instead of 32-bit, allowing it to handle lists with more than 2^31-1 total child elements. Currently, attempting to project a `LargeListArray` would fail with an error like:
```
Expected list type, got: large_list<...>
```

## Proposed Solution

1. **Add templated `ProjectListArrayImpl<>` function** - Generic implementation that works with both `ListArray` and `LargeListArray`

2. **Add `ProjectLargeListArray` wrapper** - Calls the template with `LargeListArray` and `LargeListType`

3. **Update `ProjectNestedArray`** - Handle both `::arrow::Type::LIST` and `::arrow::Type::LARGE_LIST` in the `TypeId::kList` case

4. **Add test case** - Verify `LargeListArray` projection works correctly

## Files to Change

- `src/iceberg/parquet/parquet_data_util.cc`
- `src/iceberg/test/parquet_data_test.cc`

## References

- FIXME comment: `src/iceberg/parquet/parquet_data_util.cc:151`
- Arrow LargeListArray docs: https://arrow.apache.org/docs/cpp/api/array.html

PR #714, opened 2026-06-10
feat(parquet): support reading list columns as Arrow large_list

### LINKED ISSUE #513: Enable LargeListArray support in Parquet reader schema validation
## Summary

Follow-up to #502. The data conversion layer now supports `LargeListArray` (64-bit offsets) via `ProjectRecordBatch`, but the Parquet reader's schema validation still rejects `LARGE_LIST` types. Additionally, the reader needs to expose Arrow's `list_type` property to allow users to request `LargeListArray` output.

## Problem

1. `ValidateParquetSchemaEvolution` in `parquet_schema_util.cc:177-180` only accepts `::arrow::Type::LIST`:

```cpp
case TypeId::kList:
  if (arrow_type->id() == ::arrow::Type::LIST) {
    return {};
  }
  break;
```

2. Arrow's Parquet reader defaults to `Type::LIST` output. Without exposing `ArrowReaderProperties::set_list_type()`, users cannot request `LargeListArray` output.

### 1. Update schema validation to accept both list types

```cpp
case TypeId::kList:
  if (arrow_type->id() == ::arrow::Type::LIST ||
      arrow_type->id() == ::arrow::Type::LARGE_LIST) {
    return {};
  }
  break;
```

### 2. Add `kListType` to `ReaderProperties`

Expose a property to configure the Arrow list type preference.

### 3. Pass through to Arrow reader

In `ParquetReader::Impl::Open()`, call `arrow_reader_properties.set_list_type()` with the configured value.

## Why This Is Safe

1. Iceberg's `ListType` doesn't distinguish between LIST and LARGE_LIST
2. The projection layer (`ProjectRecordBatch`) already handles both via templated `ProjectListArrayImpl<>`
3. Both represent the same logical "list" concept, just with different offset sizes

## Files to Change

- `src/iceberg/parquet/parquet_schema_util.cc` - Update `ValidateParquetSchemaEvolution`
- `src/iceberg/parquet/parquet_reader.cc` - Pass `list_type` to `ArrowReaderProperties`
- `src/iceberg/reader.h` - Add `kListType` to `ReaderProperties`
- `src/iceberg/test/parquet_test.cc` - Add integration tests

## Related

- Closes the remaining work from #502

rahulsmahadev:
## Summary

Implements the remaining `LargeListArray` gaps from #502, following the design suggested by @wgtmac in #513:

1. `ValidateParquetSchemaEvolution` now accepts `LARGE_LIST` wherever `LIST` is accepted, so schema projection works when the Arrow reader presents 64-bit offset list types.
2. A new reader property `read.arrow.use-large-list` (default: `false`) configures the Parquet reader via `ArrowReaderProperties::set_list_type(::arrow::Type::LARGE_LIST)` to decode list columns as `large_list`.

Since `ToArrowSchema` builds the reader's output schema with 32-bit lists (and `ProjectRecordBatch` dispatches on the output schema type), enabling the property also rewrites list fields in the output Arrow schema to `large_list` so the projection layer takes the `ProjectLargeListArray` path added in #502. The rewrite is local to the Parquet reader to avoid changing the `ToArrowSchema` signature used across writers and manifest readers.

Closes #513

## Changes

- `src/iceberg/parquet/parquet_schema_util.cc`: accept `LARGE_LIST` for `TypeId::kList` in schema evolution validation.
- `src/iceberg/file_reader.h`: add `ReaderProperties::kArrowUseLargeList` (`read.arrow.use-large-list`, default `false`), following the `kBatchSize` pattern.
- `src/iceberg/parquet/parquet_reader.cc`: set `ArrowReaderProperties::set_list_type` when the property is enabled, and align the output Arrow schema (lists nested in structs/maps included) with the large_list arrays produced by the reader.

## Test plan

- `ParquetSchemaProjectionTest.ValidateSchemaEvolutionAllowsLargeList`: `large_list` Arrow type validates against an Iceberg `ListType`.
- `ParquetSchemaProjectionTest.ProjectLargeListType`: projection over a `SchemaManifest` built with `set_list_type(LARGE_LIST)` (the same path `BuildProjection` uses in the reader).
- `ParquetReaderTest.ReadListType`: default behavior unchanged — list columns read as 32-bit offset `list`.
- `ParquetReaderTest.ReadListAsLargeList`: with `read.arrow.use-large-list=true`, the output schema exposes `large_list` and values round-trip correctly (verified via array slices since JSON parsing creates regular `ListArray`).

Note: my local environment lacks a C++23 toolchain (cmake 3.16/gcc 10), so I could not build locally; relying on CI to verify. All Arrow APIs used (`set_list_type`, `large_list(field)`, `MapType(key_field, item_field, keys_sorted)`, `Field::WithType`) were checked against the pinned Arrow 24.0.0 headers.

### DIFF
--- a/src/iceberg/parquet/parquet_reader.cc
+++ b/src/iceberg/parquet/parquet_reader.cc
@@ -19,6 +19,7 @@
 
 #include "iceberg/parquet/parquet_reader.h"
 
+#include <algorithm>
 #include <numeric>
 
 #include <arrow/c/bridge.h>
@@ -85,6 +86,7 @@ class EmptyRecordBatchReader : public ::arrow::RecordBatchReader {
   }
 };
 
+// forward declaration to unblock cycle dependence.
 std::shared_ptr<::arrow::Field> UseLargeListField(
     const std::shared_ptr<::arrow::Field>& field);
 
@@ -118,6 +120,42 @@ std::shared_ptr<::arrow::DataType> UseLargeListType(
 std::shared_ptr<::arrow::Field> UseLargeListField(
     const std::shared_ptr<::arrow::Field>& field) {
   return field->WithType(UseLargeListType(field->type()));
+}
+
+// Rewrite all fields in a field vector to use large_list instead of list.
+::arrow::FieldVector UseLargeListFields(const ::arrow::FieldVector& fields) {
+  ::arrow::FieldVector rewritten;
+  rewritten.reserve(fields.size());
+  for (const auto& field : fields) {
+    rewritten.push_back(UseLargeListField(field));
+  }
+  return rewritten;
+}
+
+// Returns true if the type contains a large_list, at any level of nesting.
+bool ContainsLargeList(const ::arrow::DataType& type) {
+  if (type.id() == ::arrow::Type::LARGE_LIST) {
+    return true;
+  }
+  return std::ranges::any_of(
+      type.fields(), [](const auto& field) { return ContainsLargeList(*field->type()); });
+}
+
+// Returns true if the reader produces large_list arrays.
+//
+// Arrow honors the requested large_list type only when it derives the Arrow schema from
+// the Parquet schema. A file that carries serialized ARROW:schema metadata keeps its
+// original list type instead, so whether large lists are produced can only be told from
+// the schema of the reader.
+bool ProducesLargeList(const ::arrow::RecordBatchReader& reader) {
+  const auto& schema = reader.schema();
+  if (schema == nullptr) {
+    // an empty reader produces no arrays to be described
+    return false;
+  }
+  return std::ranges::any_of(schema->fields(), [](const auto& field) {
+    return ContainsLargeList(*field->type());
+  });
 }
 
 }  // namespace
@@ -252,23 +290,6 @@ class ParquetReader::Impl {
   Status InitReadContext() {
     context_ = std::make_unique<ReadContext>();
 
-    // Build the output Arrow schema
-    ArrowSchema arrow_schema;
-    ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*read_schema_, &arrow_schema));
-    ICEBERG_ARROW_ASSIGN_OR_RETURN(context_->output_arrow_schema_,
-                                   ::arrow::ImportSchema(&arrow_schema));
-    if (use_large_list_) {
-      // Align the output schema with the large_list arrays produced by the
-      // Parquet reader when kArrowUseLargeList is enabled.
-      ::arrow::FieldVector fields;
-      fields.reserve(context_->output_arrow_schema_->fields().size());
-      for (const auto& field : context_->output_arrow_schema_->fields()) {
-        fields.push_back(UseLargeListField(field));
-      }
-      context_->output_arrow_schema_ =
-          ::arrow::schema(std::move(fields), context_->output_arrow_schema_->metadata());
-    }
-
     // Row group pruning based on the split
     // TODO(gangwu): add row group filtering based on zone map, bloom filter, etc.
     std::vector<int> row_group_indices;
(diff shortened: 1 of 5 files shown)


rahulsmahadev:
@wgtmac this implements the design you suggested in #513 — would appreciate a review when you get a chance.

One note beyond the two items in the issue: the reader's output Arrow schema (built by `ToArrowSchema`, which uses 32-bit lists) also needed to be aligned so `ProjectRecordBatch` dispatches to the large_list path from #502. I kept that rewrite local to the Parquet reader to avoid changing the `ToArrowSchema` signature, but happy to take a different approach if you'd prefer.

wgtmac:
Thanks for improving this, @rahulsmahadev! I think the long-term shape is to make this output-schema driven, ideally allowing callers to choose `list` vs `large_list` per projected column. Arrow’s Parquet reader only exposes a global `set_list_type` today, so keeping a global flag for this PR seems reasonable.

One API concern: since this option lives in generic `ReaderProperties`, it reads as format-independent. Either other readers such as Avro should honor it too, or the option should be clearly scoped/documented as Parquet-only.

One thing that is worth noting: when `read.arrow.use-large-list` is enabled, the reader always rewrites the output schema to `large_list`, but Arrow ignores `set_list_type(LARGE_LIST)` when the Parquet file contains serialized `ARROW:schema` metadata. In that case the actual batch can still contain `ListArray`, while `ProjectRecordBatch` dispatches as `LargeListArray`, causing a bad cast/crash. Please base the output schema on the actual reader schema or add a guard, and cover this with an `ARROW:schema` regression test.

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -84,6 +85,41 @@ class EmptyRecordBatchReader : public ::arrow::RecordBatchReader {
   }
 };
 
+std::shared_ptr<::arrow::Field> UseLargeListField(
```
kamcheungting-db, line 90: add comment: 
```suggestion
// forward declaration to unblock cycle dependence.
std::shared_ptr<::arrow::Field> UseLargeListField(
```

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -214,6 +254,17 @@ class ParquetReader::Impl {
...
     ICEBERG_ARROW_ASSIGN_OR_RETURN(context_->output_arrow_schema_,
                                    ::arrow::ImportSchema(&arrow_schema));
+    if (use_large_list_) {
+      // Align the output schema with the large_list arrays produced by the
+      // Parquet reader when kArrowUseLargeList is enabled.
+      ::arrow::FieldVector fields;
+      fields.reserve(context_->output_arrow_schema_->fields().size());
+      for (const auto& field : context_->output_arrow_schema_->fields()) {
+        fields.push_back(UseLargeListField(field));
+      }
```
kamcheungting-db: this is duplicated logic.
let's add a method.
```
::arrow::FieldVector UseLargeListFields(const ::arrow::FieldVector& fields) {
  ::arrow::FieldVector rewritten;
  rewritten.reserve(fields.size());
  for (const auto& field : fields) {
    rewritten.push_back(UseLargeListField(field));
  }
  return rewritten;
}
```

rahulsmahadev:
@wgtmac can you help review

rahulsmahadev:
@wgtmac have addressed comments, let me know if this PR is good

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -250,6 +322,24 @@ class ParquetReader::Impl {
...
 
+    // Build the output Arrow schema from the projected Iceberg schema. This schema is the
+    // target of ProjectRecordBatch, so it must describe the projected schema rather than
+    // the schema of the file.
+    ArrowSchema arrow_schema;
+    ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*read_schema_, &arrow_schema));
+    ICEBERG_ARROW_ASSIGN_OR_RETURN(context_->output_arrow_schema_,
+                                   ::arrow::ImportSchema(&arrow_schema));
+
+    if (use_large_list_ && ProducesLargeList(*context_->record_batch_reader_)) {
```
wgtmac: Please align the output schema with the reader schema even when this option is false. Arrow restores a stored `large_list` regardless of `set_list_type`, so this path currently leaves the output as `list` and casts a `LargeListArray` as `ListArray`.

src/iceberg/parquet/parquet_reader.cc
```diff
@@ -250,6 +322,24 @@ class ParquetReader::Impl {
...
+      // Align the output schema with the large_list arrays produced by the Parquet
+      // reader. Note that Arrow ignores the requested list type when the file carries
+      // serialized ARROW:schema metadata, in which case the reader keeps producing plain
+      // list arrays and the output schema must keep describing them as such.
+      context_->output_arrow_schema_ =
+          ::arrow::schema(UseLargeListFields(context_->output_arrow_schema_->fields()),
```
wgtmac: This rewrite is global: one `large_list` makes every projected list a `large_list`. Stored Arrow schemas can contain both types, so please map each projected field recursively from the corresponding reader field instead of using `ProducesLargeList()` as a single flag. For missing or default fields, use the configured output preference.

src/iceberg/test/parquet_test.cc
```diff
@@ -267,6 +293,38 @@ class ParquetReaderTest : public TempFileTestBase {
...
+                                               R"([[1, [1, 2]], [2, [3]]])")
+                .ValueOrDie())
+            .ValueOrDie();
+    auto table = ::arrow::Table::FromRecordBatches(arrow_schema, {batch}).ValueOrDie();
+
+    auto io = internal::checked_cast<arrow::ArrowFileSystemFileIO&>(*file_io_);
+    auto outfile = io.fs()->OpenOutputStream(temp_parquet_file_).ValueOrDie();
+
+    // write a single row group so that one batch holds every row
+    ASSERT_TRUE(::parquet::arrow::WriteTable(*table, ::arrow::default_memory_pool(),
```
wgtmac: `WriteTable` does not store `ARROW:schema` by default. Please pass `ArrowWriterProperties::Builder().store_schema()->build()`, assert that the footer key exists, and cover a stored `large_list` with the default reader option plus a mixed `list`/`large_list` schema.

wgtmac (changes requested):
Thanks for updating this PR! I still have some comments. LMKWYT.

wgtmac (approved):
Thanks for addressing feedback and sorry for the delay!

PR #715, opened 2026-06-10
fix: YearTransform passes wrong TransformType kTruncate instead of kYear

WZhuo:
## Summary

**Bug:** YearTransform constructor passed TransformType::kTruncate to the TransformFunction base class instead of TransformType::kYear. This caused transform_type() to return kTruncate on a bound YearTransform, affecting code paths that dispatch based on transform_type() (e.g., Transform::Project, Transform::ProjectStrict, Transform::SatisfiesOrderOf).

**Fix:** Changed kTruncate to kYear in the YearTransform constructor.

**Tests:** Added TransformFunctionTest cases for Year, Month, Day, Hour, and Void transforms to verify transform_type() after Bind(). The existing suite only tested Identity, Bucket, and Truncate.

### DIFF
--- a/src/iceberg/test/transform_test.cc
+++ b/src/iceberg/test/transform_test.cc
@@ -71,6 +71,56 @@ TEST(TransformFunctionTest, CreateTruncateTransform) {
   EXPECT_EQ(transformPtr.value()->transform_type(), TransformType::kTruncate);
 }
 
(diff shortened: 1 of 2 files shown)


zhjwpku (approved):
Good catch.

wgtmac (approved):

PR #716, opened 2026-06-10
fix: align rest table update serde with java/spec

wgtmac:


src/iceberg/catalog/rest/json_serde.cc
```diff
@@ -903,13 +903,21 @@ Result<CommitTableRequest> CommitTableRequestFromJson(const nlohmann::json& json
 
   ICEBERG_ASSIGN_OR_RAISE(auto requirements_json,
```
zhjwpku, line 904: I took a look at Java's `UpdateTableRequest.fromJson`. It doesn't appear to error when `REQUIREMENTS` or `UPDATES` is missing from the JSON object. In contrast, we're using `GetJsonValue` here, which would result in a `JsonParseError` under the same conditions. Should we align with the Java(using GetJsonValueOptional)?

zhjwpku (approved):

blackmwk (approved):

PR #718, opened 2026-06-10
Revert "chore: remove spdlog dependency (#459)"

kamcheungting-db:
## What
Re-adds spdlog (v1.15.3, configured with SPDLOG_USE_STD_FORMAT) as a build dependency, reverting #459.

## Why
#459 removed spdlog because nothing used it at the time, with the note that it could be added back "when it is actually needed." It is now needed: this is the foundation for the upcoming logging system for iceberg-cpp. Restoring the dependency wiring in its own change keeps that follow-up focused purely on the logger itself.


wgtmac (approved):
Thank you, @kamcheungting-db! Just a notice that we are releasing 0.3.0 so I cannot merge it immediately. Will merge after release is complete.

PR #719, opened 2026-06-10
feat(rest): parse storage-credentials and bind a per-table FileIO from vended credentials

plusplusjiajia:
What

  Parse storage-credentials from the REST LoadTableResponse and use the vended credentials to bind a per-table FileIO, so a loaded Table can read/write its (S3-compatible) storage without any manual wiring.

  Why

  REST catalogs hand out short-lived, scoped storage access via vended credentials (STS). Until now LoadTableResult.storage_credentials was dropped on the floor and every table fell back to the catalog's default FileIO, so a vended-credential table could not actually access storage.

  How

  - Serde: add storage-credentials (prefix + config) to the LoadTableResponse (de)serializer, aligned with Java LoadCredentialsResponseParser / CredentialParser.
  - Credential selection: SelectS3StorageCredential picks the longest matching s3-family prefix, matching Java VendedCredentialsProvider.
  - Per-table FileIO: implement the REST catalog's TableFileIO hook added in #750 — when the load response carries an s3-family vended credential, merge catalog + table config + the credential and build an arrow-fs-s3 FileIO (resolving oss:// and other S3-compatible schemes), replacing the fail-closed ValidateNoFileIOConfig
  path; otherwise keep the catalog default. This fills the TODO(gangwu) left in #750.
  - S3 option alignment: prefer the standard client.region (fall back to legacy s3.region), strip the scheme from a full-URI s3.endpoint, and honor s3.path-style-access for virtual-hosted addressing


plusplusjiajia:
Rebased onto main now that #750 merged. The per-table FileIO binding now lives in the TableFileIO hook — it builds an arrow-fs-s3 FileIO from the vended s3 credential, else keeps the fail-closed path, filling the TODO(gangwu) from #750. @wgtmac @gangwu PTAL.

src/iceberg/catalog/rest/rest_file_io.cc
```diff
@@ -92,4 +95,30 @@ Result<std::unique_ptr<FileIO>> MakeCatalogFileIO(const RestCatalogProperties& c
   return FileIORegistry::Load(io_impl, config.configs());
 }
 
+const StorageCredential* SelectS3StorageCredential(
+    const std::vector<StorageCredential>& credentials) {
+  const StorageCredential* best = nullptr;
+  for (const auto& cred : credentials) {
+    if (cred.prefix.starts_with("s3") &&
```
wgtmac: Prefix selection needs the actual storage path. The REST spec and Java S3FileIO choose the longest prefix that matches each path; picking the longest S3 prefix globally makes unmatched paths and other S3 prefixes use the wrong credential.

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -878,7 +892,8 @@ Result<std::shared_ptr<Table>> RestCatalog::MakeTableFromCommitResponse(
   // TODO(gangwu): If the REST commit response grows table config or
   // storage credentials, derive a replacement table session/FileIO from that
   // response. The current table commit response does not define config.
-  ICEBERG_ASSIGN_OR_RAISE(auto table_io, TableFileIO(context, table_config));
+  ICEBERG_ASSIGN_OR_RAISE(auto table_io,
+                          TableFileIO(context, table_config, /*storage_credentials=*/{}));
```
wgtmac: This rebuilds the returned table with an empty credential list after commit. CommitTableResponse does not carry storage-credentials, so a table loaded with vended credentials loses its FileIO credentials after the first commit.

src/iceberg/catalog/rest/rest_file_io.cc
```diff
@@ -92,4 +95,30 @@ Result<std::unique_ptr<FileIO>> MakeCatalogFileIO(const RestCatalogProperties& c
...
+
+Result<std::unique_ptr<FileIO>> MakeS3FileIOFromCredential(
+    const std::unordered_map<std::string, std::string>& catalog_config,
+    const std::unordered_map<std::string, std::string>& table_config,
+    const StorageCredential& s3_cred) {
+  auto properties = catalog_config;
+  for (const auto& [key, value] : table_config) {
+    properties[key] = value;
+  }
+  for (const auto& [key, value] : s3_cred.config) {
```
wgtmac: This makes the STS values static. Java S3FileIO keeps the credential set and refreshes through credentials.uri when s3.session-token-expires-at-ms is present; long scans or reused tables will fail after the vended token expires.

src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -516,7 +516,17 @@ Result<std::shared_ptr<auth::AuthSession>> RestCatalog::TableAuthSession(
 
 Result<std::shared_ptr<FileIO>> RestCatalog::TableFileIO(
     const SessionContext& /*context*/,
-    const std::unordered_map<std::string, std::string>& table_config) const {
+    const std::unordered_map<std::string, std::string>& table_config,
+    const std::vector<StorageCredential>& storage_credentials) const {
+  // Longest-prefix "s3" vended credential. Java's VendedCredentialsProvider
+  // resolves per path against the table location; we bind one at load time
+  // (fine for the common one-credential-per-table case).
+  if (const StorageCredential* s3_cred = SelectS3StorageCredential(storage_credentials);
```
wgtmac: storage-credentials can also carry GCS/ADLS credentials. This path silently ignores non-S3 credentials and falls back to the catalog FileIO; if S3-only is intentional for now, please add an explicit TODO or fail-fast note.

src/iceberg/catalog/rest/rest_file_io.cc
plusplusjiajia: @wgtmac Good catch. Fixed: SelectS3StorageCredential now picks the longest credential prefix that is actually a prefix of the table's storage location (passed down from metadata->location), instead of the longest s3 prefix globally — matching the REST spec / Java S3FileIO. It also treats s3/s3a/s3n as equivalent so an s3:// vended prefix still matches an s3a:// location.

src/iceberg/catalog/rest/rest_catalog.cc
plusplusjiajia: @wgtmac Fixed. MakeTableFromCommitResponse now reuses the FileIO bound at load (threaded through TableScopedCatalog) instead of rebuilding it with an empty credential list.

src/iceberg/catalog/rest/rest_catalog.cc
plusplusjiajia: @wgtmac Addressed. When the server vends only non-S3 credentials (GCS/ADLS), TableFileIO now fails fast with NotSupported listing the prefixes, rather than silently using the catalog FileIO. If S3 credentials are vended but none covers this table's location, it falls through to the catalog FileIO (which may use ambient creds). S3-only support is intentional for now — see the TODO in TableFileIO.

src/iceberg/catalog/rest/rest_file_io.cc
plusplusjiajia: @wgtmac You're right that the vended STS values are static and will expire on long scans / reused tables. Implementing refresh properly needs a credential-aware FileIO with a background refresher that re-fetches via credentials.uri (mirroring Java's VendedCredentialsProvider + S3FileIO#scheduleCredentialRefresh) — a larger  change than this PR. I've left an explicit TODO and would prefer to handle it as a follow-up. Shall I open an issue to track it? Happy to scope it there.

src/iceberg/catalog/rest/types.h
```diff
@@ -180,12 +180,23 @@ struct ICEBERG_REST_EXPORT CreateTableRequest {
...
+};
+
 /// \brief Result body for table create/load/register APIs.
 struct ICEBERG_REST_EXPORT LoadTableResult {
   std::string metadata_location;
   std::shared_ptr<TableMetadata> metadata;  // required
   std::unordered_map<std::string, std::string> config;
-  // TODO(Li Feiyang): Add std::shared_ptr<StorageCredential> storage_credential;
+  /// \brief Vended storage credentials, one per URI prefix; empty if none.
+  std::vector<StorageCredential> storage_credentials;
```
wgtmac, line 190: Please validate these credentials at the model boundary too. `LoadTableResultFromJson` rejects empty `prefix` or `config`, and Java `Credential.validate()` does the same, but `LoadTableResult::Validate()` still only checks metadata. C++ can now serialize a credential it cannot parse back.

plusplusjiajia: @wgtmac  Done. Added StorageCredential::Validate() (non-empty prefix and config, mirroring Java Credential.validate()), and LoadTableResult::Validate() now validates each vended credential — so the model boundary enforces the same invariant as the JSON parser.

wgtmac (approved):
Thanks for working on this PR. One small suggestion for future changes: please spend a bit more time on self-review before asking for review. It would help to check whether the change fits the codebase design, not only whether it works locally. A few examples from this PR: the storage credential logic was a bit too storage-specific, table FileIO config handling needed follow-up fixes and prefix matching needed more careful validation. Reviewer time is limited, so catching more of these during self-review would really help make future PRs easier to review and merge.

plusplusjiajia:
@wgtmac Thanks a lot for the careful review and for getting this merged — and the point about self-review is well taken, I'll be more thorough going forward. The ArrowS3FileIO refactor reads really cleanly, too.
While going back over the merged credential path, I ran into something I wanted to check with you. It looks like MakeTableFileIO chooses the FileIO impl from DetectBuiltinFileIO(warehouse), so when the warehouse isn't an s3:// path it resolves to arrow-fs-local, and binding the vended credentials then returns Configured FileIO does not support vended storage credentials. That seems to affect the catalogs that vend credentials, since their warehouse is usually a non-URI identifier (a bucket ARN for S3 Tables, a catalog/account id for Glue, a catalog name for Unity / Polaris) rather than an s3:// path — so I'm not quite sure how it's meant to work when the warehouse isn't s3://.
I may be missing some context here — could you share how you'd suggest handling this? I'd be glad to put together a follow-up once we settle on an approach.

wgtmac:
I agree with your concern that it is wrong to deduce the file system scheme based on `warehouse`. I remember that this is just a temporary workaround and should be removed after we introduce a C++ `ResolvingFileIO` or the like. Perhaps now it is the right time to fix this. Maybe we can get started with expanding the S3-compatible object storage support with AWS SDK and report unsupported error for other storage system. Then we can add `ResolvingFileIO` based on this?

plusplusjiajia:
@wgtmac Sounds good to me — solidifying the S3-compatible support first and then adding a ResolvingFileIO on top is the right direction. Looking forward to seeing it usable soon. Thanks for the quick response!

wgtmac:
@plusplusjiajia Feel free to open a PR if you want to work on this. Thanks!

PR #721, opened 2026-06-10
feat: add row delta update

manuzhang:
Implements the RowDelta update builder, table and transaction factory methods, and focused tests for row-level add/delete flows.

### DIFF
--- a/src/iceberg/table.cc
+++ b/src/iceberg/table.cc
@@ -35,6 +35,7 @@
 #include "iceberg/update/expire_snapshots.h"
 #include "iceberg/update/fast_append.h"
 #include "iceberg/update/merge_append.h"
+#include "iceberg/update/row_delta.h"
 #include "iceberg/update/set_snapshot.h"
 #include "iceberg/update/snapshot_manager.h"
 #include "iceberg/update/update_location.h"
@@ -231,6 +232,12 @@ Result<std::shared_ptr<DeleteFiles>> Table::NewDeleteFiles() {
   return DeleteFiles::Make(name().name, std::move(ctx));
 }
 
+Result<std::shared_ptr<RowDelta>> Table::NewRowDelta() {
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto ctx, TransactionContext::Make(shared_from_this(), TransactionKind::kUpdate));
+  return RowDelta::Make(name().name, std::move(ctx));
+}
+
 Result<std::shared_ptr<UpdateStatistics>> Table::NewUpdateStatistics() {
   ICEBERG_ASSIGN_OR_RAISE(
       auto ctx, TransactionContext::Make(shared_from_this(), TransactionKind::kUpdate));
@@ -332,6 +339,14 @@ Result<std::shared_ptr<FastAppend>> StaticTable::NewFastAppend() {
 
 Result<std::shared_ptr<MergeAppend>> StaticTable::NewMergeAppend() {
   return NotSupported("Cannot create a merge append for a static table");
+}
+
+Result<std::shared_ptr<DeleteFiles>> StaticTable::NewDeleteFiles() {
+  return NotSupported("Cannot create delete files for a static table");
+}
+
+Result<std::shared_ptr<RowDelta>> StaticTable::NewRowDelta() {
+  return NotSupported("Cannot create a row delta for a static table");
 }
 
 Result<std::shared_ptr<SnapshotManager>> StaticTable::NewSnapshotManager() {
--- a/src/iceberg/test/row_delta_test.cc
+++ b/src/iceberg/test/row_delta_test.cc
@@ -274,7 +274,7 @@ TEST_F(RowDeltaTest, ValidateNoConflictingDeleteFilesFailsForConcurrentDelete) {
   EXPECT_THAT(result, HasErrorMessage(delete_file->file_path));
 }
 
-TEST_F(RowDeltaTest, ValidateDataFilesExistFailsForConcurrentDelete) {
+TEST_F(RowDeltaTest, ValidateDataFilesExistSkipsConcurrentDeleteByDefault) {
   CommitFileA();
   ICEBERG_UNWRAP_OR_FAIL(auto starting_snapshot, table_->current_snapshot());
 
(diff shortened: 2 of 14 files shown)


src/iceberg/update/row_delta.cc
```diff
@@ -0,0 +1,196 @@
...
+
+RowDelta& RowDelta::AddDeletes(const std::shared_ptr<DataFile>& deletes) {
+  ICEBERG_BUILDER_RETURN_IF_ERROR(AddDeleteFile(deletes));
+  return *this;
+}
+
+RowDelta& RowDelta::RemoveRows(const std::shared_ptr<DataFile>& file) {
+  ICEBERG_BUILDER_CHECK(file != nullptr, "Cannot remove a null data file");
+  removed_data_files_.insert(file);
+  ICEBERG_BUILDER_RETURN_IF_ERROR(DeleteDataFile(file));
```
WZhuo:
```suggestion
  ICEBERG_BUILDER_RETURN_IF_ERROR(DeleteDataFile(file));
  removed_data_files_.insert(file);
```

src/iceberg/update/merging_snapshot_update.h
```diff
@@ -288,6 +288,14 @@ class ICEBERG_EXPORT MergingSnapshotUpdate : public SnapshotUpdate {
       const std::shared_ptr<Snapshot>& parent, std::shared_ptr<FileIO> io,
       bool case_sensitive = true);
 
+  /// \brief Return an error if a staged deletion vector conflicts with a deletion
+  /// vector added since starting_snapshot_id.
+  Status ValidateAddedDVs(const TableMetadata& metadata,
```
WZhuo, line 293: `protected` is sufficient and keeps this internal validation helper from leaking into the public surface of every merge-based update.

src/iceberg/test/row_delta_test.cc
```diff
@@ -0,0 +1,287 @@
...
+#include "iceberg/snapshot.h"
+#include "iceberg/table.h"
+#include "iceberg/table_metadata.h"
+#include "iceberg/test/matchers.h"
+#include "iceberg/test/update_test_base.h"
+#include "iceberg/update/fast_append.h"
+
+namespace iceberg {
+
+class RowDeltaTest : public MinimalUpdateTestBase {
```
WZhuo, line 45: It could be better add some tests that commit a conflicting data/delete file after the starting snapshot and assert ValidationFailed from each validator.

src/iceberg/update/row_delta.cc
```diff
@@ -0,0 +1,196 @@
...
+    for (size_t i = 0; i < conflicting_files.size(); ++i) {
+      if (i > 0) {
+        files += ", ";
+      }
+      files += conflicting_files[i];
+    }
+    files += "]";
+    return ValidationFailed(
+        "Cannot delete data files {} that are referenced by new delete files", files);
+  }
```
WZhuo, line 188: `FormatRange` defined in iceberg/util/formatter_internal.h could be used.

src/iceberg/update/merging_snapshot_update.h
manuzhang: It's already protected.

src/iceberg/update/row_delta.cc
```diff
@@ -0,0 +1,189 @@
...
+
+  ICEBERG_RETURN_UNEXPECTED(ValidateNoConflictingFileAndPositionDeletes());
+
+  return MergingSnapshotUpdate::ValidateAddedDVs(
+      current_metadata, starting_snapshot_id_, conflict_detection_filter_, snapshot, io);
+}
+
+Status RowDelta::ValidateNoConflictingFileAndPositionDeletes() const {
+  std::vector<std::string> conflicting_files;
+  for (const auto& file : removed_data_files_) {
```
WZhuo, line 178: Suggestion: `conflicting_files` could use `std::vector<std::string_view>` instead of `std::vector<std::string>` to avoid unnecessary string copies.

This is safe because:
- `file->file_path` is a `std::string` member of `DataFile`
- `removed_data_files_` holds `shared_ptr<DataFile>`, so the backing strings live through the function
- `conflicting_files` is a local only consumed by `FormatRange` before returning — the `string_view`s never escape the scope

src/iceberg/update/row_delta.cc
```diff
@@ -0,0 +1,191 @@
...
+}
+
+RowDelta& RowDelta::ValidateNoConflictingDeleteFiles() {
+  validate_new_delete_files_ = true;
+  return *this;
+}
+
+std::string RowDelta::operation() {
+  if (AddsDataFiles() && !AddsDeleteFiles() && !DeletesDataFiles() &&
+      !DeletesDeleteFiles()) {
```
wgtmac: Java's BaseRowDelta still returns APPEND when rows are added and only delete files are removed. This extra check makes AddRows + RemoveDeletes produce OVERWRITE, which breaks parity.

src/iceberg/table.cc
```diff
@@ -231,6 +232,12 @@ Result<std::shared_ptr<DeleteFiles>> Table::NewDeleteFiles() {
   return DeleteFiles::Make(name().name, std::move(ctx));
 }
 
+Result<std::shared_ptr<RowDelta>> Table::NewRowDelta() {
```
wgtmac, line 235: StaticTable should override this and return NotSupported, like the other update factories. Otherwise a read-only table can create a RowDelta builder.

src/iceberg/update/row_delta.cc
```diff
@@ -0,0 +1,191 @@
...
+                                                  starting_snapshot_id_.value()));
+    ICEBERG_CHECK(is_ancestor, "Snapshot {} is not an ancestor of {}",
+                  starting_snapshot_id_.value(), snapshot->snapshot_id);
+  }
+
+  auto io = ctx_->table->io();
+  if (!referenced_data_files_.empty()) {
+    ICEBERG_RETURN_UNEXPECTED(MergingSnapshotUpdate::ValidateDataFilesExist(
+        current_metadata, starting_snapshot_id_, referenced_data_files_,
+        /*skip_deletes=*/false, conflict_detection_filter_, snapshot, io,
```
wgtmac: This should stay tied to validate_deletes_. Java only checks DELETE snapshots here after validateDeletedFiles() is enabled; otherwise validateDataFilesExist skips concurrent deletes.

wgtmac (approved):

PR #722, opened 2026-06-10
feat(logging): add LogLevel severity enum (1/6)

kamcheungting-db:
## What

First of PR that adds a logging system to iceberg-cpp.

This PR adds LogLevel — the severity scale every later piece builds on — and nothing else.

- enum class LogLevel { kTrace, kDebug, kInfo, kWarn, kError, kCritical, kFatal, kOff }, ordered most-to-least verbose so "should this log?" is a plain level >= threshold check. kOff is the max sentinel for disabling logging.
- constexpr ToString(LogLevel) and a case-insensitive LogLevelFromString returning Result<LogLevel> — same idiom as CounterUnit in metrics/counter.h.

Header-only: this PR adds src/iceberg/logging/log_level.h, wires the new src/iceberg/logging/ directory into CMake, and adds log_level_test.cc (ToString/FromString round-trip, case-insensitivity, unknown-input error, ordering).

## Why

iceberg-cpp has no logging today — only error-as-value Result. This stack introduces a pluggable logger; levels are the foundational vocabulary, landed on their own so the
rest reviews cleanly.

## Testing

compiled and run with clang++ -std=c++23 -stdlib=libc++ (clang 18).

The logging module builds clean and a behavior driver passes (levels round-trip, macro formatting + source location, gate filtering, disabled-arg elision, never-throws fallback, FATAL emit-then-abort, compile-time stripping).

Full ctest -R logging_test + the spdlog-ON build still want a C++23 toolchain with network (gtest/spdlog via FetchContent) in CI.


wgtmac:
Do you want to add it to meson.build as well?

wgtmac (approved):

kamcheungting-db:
> Do you want to add it to meson.build as well?

Done

wgtmac (approved):

PR #723, opened 2026-06-10
feat(logging): add Logger interface and default logger (2/6)

kamcheungting-db:
Part 2 of the logging stack (builds on #722). Adds the logging API and a swappable default logger — the foundation the backends and macros plug into.

**What's here**
- `Logger`: the pluggable sink interface (`ShouldLog` / `Log` / `SetLevel` / `Flush` / `Initialize`). `ShouldLog()` is the single source of truth for runtime filtering.
- `LogMessage` owns its formatted text so a sink can safely keep a record; reserves an attributes field for future structured logging.
- Process-global default logger: `GetDefaultLogger` / `SetDefaultLogger` / `SetDefaultLevel`, with a lock-free thread-local fast path so logging stays cheap.
- `Initialize` applies the `level` property, so config-driven levels actually work. `CurrentLogger()` is safe to call even from a `thread_local` destructor during thread shutdown.
- `logger.h` stays backend-agnostic (never includes the build config header), so consumers see one stable API regardless of backend.

**Examples** — using the API directly (the `LOG_*` macros that wrap it arrive in #725):

```cpp
// A custom sink, installed as the process default.
class MySink : public Logger {
 public:
  bool ShouldLog(LogLevel level) const override { return level >= level_; }
  void Log(LogMessage&& m) noexcept override { write_line(m.message); }
  void SetLevel(LogLevel level) override { level_ = level; }
  LogLevel level() const override { return level_; }
 private:
  std::atomic<LogLevel> level_{LogLevel::kInfo};
};

SetDefaultLogger(std::make_shared<MySink>());   // install process-wide
SetDefaultLevel(LogLevel::kDebug);              // adjust the threshold

auto logger = GetDefaultLogger();               // borrow the current default
if (logger->ShouldLog(LogLevel::kInfo)) {
  logger->Log(LogMessage{.level = LogLevel::kInfo, .message = "scan ready"});
}

// Or configure from catalog-style properties (applies the "level" key):
auto sink = std::make_shared<MySink>();
auto status = sink->Initialize({{std::string(kLevelProperty), "warn"}});  // -> kWarn
```

The same example is documented inline in `logger.h`.

**Tests** — `logger_test`: default-logger API, level-from-property, invalid level rejected, concurrent swap/read, and logging during thread teardown. Built and run with clang/libc++ (spdlog ON and OFF).

### DIFF
--- a/src/iceberg/logging/logger.cc
+++ b/src/iceberg/logging/logger.cc
@@ -123,7 +123,7 @@ void SetDefaultLevel(LogLevel level) {
   slot.logger->SetLevel(level);
 }
 
-namespace detail {
+namespace internal {
 
 const std::shared_ptr<Logger>& CurrentLogger() noexcept {
   // The per-thread cache is reached through a trivially-destructible thread_local
@@ -162,6 +162,6 @@ void EmitFormatError(Logger& logger, LogLevel level,
                         .attributes = {}});
 }
 
-}  // namespace detail
+}  // namespace internal
 
 }  // namespace iceberg
--- a/src/iceberg/logging/logger.cc
+++ b/src/iceberg/logging/logger.cc
@@ -1,0 +1,146 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/logging/logger.h"
+
+#include <atomic>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <utility>
+
+namespace iceberg {
+
+namespace {
+
+/// \brief Logger that drops every record.
+class NoopLogger final : public Logger {
+ public:
+  bool ShouldLog(LogLevel /*level*/) const override { return false; }
+  void Log(LogMessage&& /*message*/) noexcept override {}
+  void SetLevel(LogLevel /*level*/) override {}
+  LogLevel level() const override { return LogLevel::kOff; }
+  bool IsNoop() const override { return true; }
+};
+
+/// \brief Construct the process default logger for this build configuration.
+///
+/// This block ships only the interface and the no-op logger; the concrete
+/// std::cerr and spdlog sinks (and the build-config selection between them)
+/// arrive in later blocks, which update this factory.
+std::shared_ptr<Logger> MakeDefaultLogger() { return std::make_shared<NoopLogger>(); }
+
+/// \brief The process-global default-logger slot.
+struct DefaultSlot {
+  std::mutex mtx;
+  std::shared_ptr<Logger> logger;
+  // Seeded to 1 so a fresh thread (tls_gen == 0) always refreshes on first use.
+  std::atomic<uint64_t> gen{1};
+
+  DefaultSlot() : logger(MakeDefaultLogger()) {}
+};
+
+/// \brief Immortal (leaked, hence reachable -> LSan-clean) accessor for the slot.
+DefaultSlot& Slot() {
+  static auto* slot = new DefaultSlot();
+  return *slot;
+}
+
+}  // namespace
+
+std::shared_ptr<Logger> Logger::Noop() {
+  // Intentionally leaked: reachable via the function-local static (LSan-clean)
+  // and never destroyed, so logging during static teardown stays safe.
+  static auto* instance = new std::shared_ptr<Logger>(std::make_shared<NoopLogger>());
+  return *instance;
+}
+
+std::shared_ptr<Logger> GetDefaultLogger() {
+  DefaultSlot& slot = Slot();
+  std::lock_guard<std::mutex> lock(slot.mtx);
+  return slot.logger;
+}
+
+void SetDefaultLogger(std::shared_ptr<Logger> logger) {
+  if (!logger) {
+    logger = Logger::Noop();
+  }
+  DefaultSlot& slot = Slot();
+  std::lock_guard<std::mutex> lock(slot.mtx);
+  slot.logger = std::move(logger);
+  // Publish the swap; the mutex provides the happens-before, gen is a detector.
+  slot.gen.fetch_add(1, std::memory_order_relaxed);
+}
+
+void SetDefaultLevel(LogLevel level) {
+  DefaultSlot& slot = Slot();
+  std::lock_guard<std::mutex> lock(slot.mtx);
+  slot.logger->SetLevel(level);
+}
+
+namespace detail {
+
+const std::shared_ptr<Logger>& CurrentLogger() noexcept {
+  static thread_local std::shared_ptr<Logger> tls;
+  static thread_local uint64_t tls_gen = 0;
+  // Sentinel whose destructor marks the cache dead at thread exit. It is
+  // declared after tls/tls_gen, so it is destroyed FIRST (reverse order); once
+  // dead, a log from any later-destroyed thread_local destructor must not touch
+  // the (about-to-be / already) destroyed tls slot.
+  static thread_local struct AliveFlag {
+    bool value = true;
+    ~AliveFlag() { value = false; }
+  } alive;
+  if (!alive.value) {
+    // Thread teardown: the TLS cache is unsafe. Fall back to an immortal logger
+    // (leaked, never destroyed) so logging during teardown stays safe.
+    static const std::shared_ptr<Logger> kFallback = Logger::Noop();
+    return kFallback;
+  }
+  DefaultSlot& slot = Slot();
+  uint64_t current = slot.gen.load(std::memory_order_relaxed);
+  if (current != tls_gen) {
+    std::lock_guard<std::mutex> lock(slot.mtx);
+    tls = slot.logger;
+    tls_gen = current;
+  }
+  return tls;
+}
+
+void Emit(Logger& logger, LogLevel level, const std::source_location& location,
+          std::string&& message) {
+  logger.Log(LogMessage{.level = level,
+                        .message = std::move(message),
+                        .location = location,
+                        .attributes = {}});
+}
+
+void EmitFormatError(Logger& logger, LogLevel level,
+                     const std::source_location& location) noexcept {
+  // Fixed short literal (<= 15 bytes, fits SSO on libstdc++/libc++/MSVC -> no heap
+  // allocation), no std::format, no retry. Cannot throw or recurse.
+  logger.Log(LogMessage{.level = level,
+                        .message = std::string("<fmt error>"),
+                        .location = location,
+                        .attributes = {}});
+}
+
+}  // namespace detail
+
+}  // namespace iceberg
--- a/src/iceberg/logging/logger.cc
+++ b/src/iceberg/logging/logger.cc
@@ -24,7 +24,6 @@
 #include <memory>
 #include <mutex>
 #include <utility>
-#include <vector>
 
 namespace iceberg {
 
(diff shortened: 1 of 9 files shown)


src/iceberg/test/logger_test.cc
```diff
@@ -0,0 +1,149 @@
...
+  std::atomic<bool> saw_null{false};
+  std::vector<std::thread> readers;
+  for (int i = 0; i < 6; ++i) {
+    readers.emplace_back([&stop, &saw_null] {
+      // ASSERT_* doesn't propagate from non-main threads; record via a flag.
+      while (!stop.load(std::memory_order_relaxed)) {
+        const auto& l = detail::CurrentLogger();
+        if (!l) saw_null.store(true, std::memory_order_relaxed);
+        (void)l->ShouldLog(LogLevel::kError);
+        (void)GetDefaultLogger();
```
wgtmac:
```suggestion
        std::ignore = l->ShouldLog(LogLevel::kError);
        std::ignore = GetDefaultLogger();
```

src/iceberg/test/logger_test.cc
```diff
@@ -0,0 +1,149 @@
...
+  ASSERT_TRUE(status.has_value());
+  EXPECT_EQ(logger.level(), LogLevel::kError);
+}
+
+TEST(LoggerTest, InitializeRejectsInvalidLevel) {
+  CapturingLogger logger;
+  auto status =
+      logger.Initialize({{std::string(kLevelProperty), std::string("not-a-level")}});
+  ASSERT_FALSE(status.has_value());
+  EXPECT_EQ(status.error().kind, ErrorKind::kInvalidArgument);
```
wgtmac: Let's use `IsError` matcher.

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,213 @@
...
+#include "iceberg/result.h"
+
+namespace iceberg {
+
+/// \brief A structured key/value attribute attached to a log record.
+///
+/// Both key and value are owned so a sink may retain the record safely.
+/// Unused in v1; reserved so structured logging can be added without an ABI
+/// break to LogMessage.
+struct LogAttribute {
```
wgtmac: Do we need to add `ICEBERG_EXPORT` to this and below?

src/iceberg/logging/logger.cc
```diff
@@ -0,0 +1,146 @@
...
+
+void SetDefaultLevel(LogLevel level) {
+  DefaultSlot& slot = Slot();
+  std::lock_guard<std::mutex> lock(slot.mtx);
+  slot.logger->SetLevel(level);
+}
+
+namespace detail {
+
+const std::shared_ptr<Logger>& CurrentLogger() noexcept {
```
wgtmac, line 143: I think this is still not safe for thread_local teardown.

One case: a thread creates another thread_local object first. Then it calls CurrentLogger() in normal code. At thread exit, alive is destroyed before that earlier object. If that earlier object logs in its destructor, we read alive.value after alive is already destroyed. This is UB.

The test only checks first use from a TLS destructor. Maybe we need a design without destructible TLS state, or leak the per-thread cache.

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,213 @@
...
+    if (auto it = properties.find(std::string(kLevelProperty)); it != properties.end()) {
+      auto parsed = LogLevelFromString(it->second);
+      if (!parsed) return std::unexpected(parsed.error());
+      SetLevel(*parsed);
+    }
+    return {};
+  }
+
+  /// \brief Cheap check whether a record at \p level would be emitted.
+  virtual bool ShouldLog(LogLevel level) const = 0;
```
wgtmac: The doc says logger implementations must not throw. ShouldLog() is also in the hot path.

But the virtual method is not noexcept. A custom engine logger may throw here by mistake. Then logging can throw from the filter path. Adding noexcept later is also an API/ABI change.

Can we mark ShouldLog() noexcept now? Maybe SetLevel() and level() too, if we want the same rule for all logger methods.

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,213 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/logging/logger.h
+/// \brief Pluggable logging interface, the process-global default logger, and
```
wgtmac: Small doc issue. This header says it has logging macros, but this PR does not add them.

The PR text says macros are in a later block. Users who read the installed header may be confused.

Maybe say this header has the logger API and macro helper hooks. Or move this sentence to the later PR.

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,213 @@
...
+
+  /// \brief Return a shared, immortal no-op logger singleton.
+  static std::shared_ptr<Logger> Noop();
+};
+
+/// \brief Return the process-global default logger (never null).
+///
+/// Off the hot path -- acquires the slot lock and returns an owning copy. The
+/// logging macros use the cheaper internal hot-path accessor instead.
+ICEBERG_EXPORT std::shared_ptr<Logger> GetDefaultLogger();
```
wgtmac, line 191: I think only having a process-global default logger may be hard for engine integration.

In engines, users may want different loggers for different catalog, session, query, or task. If all code only reads one global logger, it is hard to route logs to the engine logging context.

Can we consider a way to pass or bind a logger from a higher level context? The global logger can still be the fallback.

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,213 @@
...
+/// only when the default logger has changed (no lock / no refcount churn in
+/// steady state). The reference is valid for the duration of the calling
+/// statement.
+ICEBERG_EXPORT const std::shared_ptr<Logger>& CurrentLogger() noexcept;
+
+/// \brief Build a LogMessage from the already-formatted text and dispatch it.
+///
+/// Declared ICEBERG_EXPORT because the logging macros expand into this call in
+/// consumer translation units.
+ICEBERG_EXPORT void Emit(Logger& logger, LogLevel level,
```
wgtmac, line 290: I think we should keep a lazy format-style API here.

Now Emit takes an already formatted std::string. This means the string must be built before calling the sink. It is OK after ShouldLog(), but it still makes the API string-based. Some engine loggers may want to keep format string + args, or format only inside their own backend.

Can we provide an API closer to std::format style, like log(level, fmt, args...), and only build the string after the level check? This also makes normal use easier for callers.

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,213 @@
...
+/// \brief A single log record handed to a Logger.
+///
+/// The formatted message is owned (moved in by the logging macros), so a sink
+/// may safely retain the record beyond the Log() call. The member set must not
+/// depend on the build's logging backend (the spdlog backend never appears here).
+struct LogMessage {
+  LogLevel level = LogLevel::kOff;
+  std::string message;
+  std::source_location location = std::source_location::current();
+  std::vector<LogAttribute> attributes;
```
wgtmac, line 68: I think the attributes field is useful, but now there is no easy API to fill it.

For engine loggers, structured fields are often important. For example query id, task id, table name, snapshot id, file path, etc. If the first public API only supports message text in practice, it may be harder to add good structured logging later.

Can we add a small helper or builder path for attributes, or at least define how future macros will pass them?

src/iceberg/test/logger_test.cc
kamcheungting-db: Done

src/iceberg/test/logger_test.cc
kamcheungting-db: Done

src/iceberg/logging/logger.h
kamcheungting-db: Done

src/iceberg/logging/logger.h
kamcheungting-db: done

src/iceberg/logging/logger.h
kamcheungting-db: done.

src/iceberg/logging/logger.h
kamcheungting-db: Done

src/iceberg/logging/logger.h
kamcheungting-db: Done

kamcheungting-db: we now support:
`iceberg::Log(LogLevel::kInfo, "loaded {} files", n);`
and 
`iceberg::Log(logger, LogLevel::kWarn, "retry {}", attempt);`

src/iceberg/logging/logger.h
kamcheungting-db: Done, we now support
```
auto record = LogMessage::Builder(LogLevel::kInfo)
                  .Message("scan finished")
                  .Attribute("table", table_name)
                  .Attribute("snapshot_id", std::to_string(id))
                  .Build();
logger->Log(std::move(record));

```

src/iceberg/logging/logger.cc
kamcheungting-db: Done.
The new thread local is implemented and guaranteed teardown logic safe.
The tear down logic is protected by a thread local `Guard`.
```
ThreadCache* AccessThreadCache(bool create) noexcept {
  static thread_local bool dead = false;
  static thread_local ThreadCache* cache = nullptr;
  static thread_local struct Guard {
    ~Guard() {
      dead = true;  // mark BEFORE freeing, so a re-entrant log hits the fallback
      delete cache;
      cache = nullptr;
    }
  } guard;
  std::ignore = guard;  // mark the thread_local as intentionally used (its dtor is
                        // registered by reaching the declaration above)

  if (dead) return nullptr;
  if (cache == nullptr && create) cache = new ThreadCache();
  return cache;
}
```

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,369 @@
...
+}
+
+}  // namespace internal
+
+/// \brief Log to the process-default logger, std::format style. Formats only if
+/// the level is enabled; never throws.
+///
+/// Example: `iceberg::Log(LogLevel::kInfo, "loaded {} files", n);`
+template <typename... Args>
+void Log(LogLevel level, internal::FmtWithLoc<std::type_identity_t<Args>...> fmt,
```
wgtmac, line 354: The comment says this API never throws, but the public Log overload itself is not marked noexcept. The helper is noexcept, so the contract is only missing at the public boundary.

Can we mark both public Log overloads noexcept to make the no-throw contract explicit?

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,369 @@
...
+///                     .Attribute("table", table_name)
+///                     .Attribute("snapshot_id", std::to_string(id))
+///                     .Build();
+///   logger->Log(std::move(record));
+///
+/// The location defaults to where the Builder is constructed; override it with
+/// Location() (e.g. to forward a caller's std::source_location).
+class ICEBERG_EXPORT LogMessage::Builder {
+ public:
+  explicit Builder(LogLevel level) : level_(level) {}
```
wgtmac: This default location is not the user call site. Since location_ is initialized in the header, Builder(...).Build() records logger.h instead of the caller file/line.

Can we take std::source_location in the Builder constructor, with current() as the default argument, and add a test that checks the recorded line is the call site?

src/iceberg/logging/logger.h
kamcheungting-db: Done.
Now, both public Log() overloads are now noexcept.

src/iceberg/logging/logger.cc
kamcheungting-db: Done

src/iceberg/logging/logger.h
kamcheungting-db: Done

src/iceberg/logging/logger.h
```diff
@@ -0,0 +1,371 @@
...
+///                     .Attribute("table", table_name)
+///                     .Attribute("snapshot_id", std::to_string(id))
+///                     .Build();
+///   logger->Log(std::move(record));
+///
+/// The location defaults to where the Builder is constructed; override it with
+/// Location() (e.g. to forward a caller's std::source_location).
+class ICEBERG_EXPORT LogMessage::Builder {
+ public:
+  explicit Builder(LogLevel level)
```
wgtmac: I think this still gives the wrong default location. Now the test makes the default point to logger.h / Builder, but users will expect the record to point to their call site.

Can we make the constructor take std::source_location with current() as a default argument, and test that the default location is the Builder call site?

Example:

explicit Builder(LogLevel level,
                 std::source_location location = std::source_location::current())
    : level_(level), location_(location) {}

kamcheungting-db: You are right.

 ```
 explicit Builder(LogLevel level,
                   std::source_location location = std::source_location::current())
      : level_(level), location_(location) {}
```

```
TEST(LoggerTest, BuilderDefaultLocationIsCallerSite) {
  auto here = std::source_location::current();
  auto record = LogMessage::Builder(LogLevel::kInfo).Message("m").Build();
  EXPECT_STREQ(record.location.file_name(), here.file_name());
  EXPECT_EQ(record.location.line(), here.line() + 1);
}
```

kamcheungting-db: Done

wgtmac (approved):
LGTM, thanks @kamcheungting-db!

PR #724, opened 2026-06-10
feat(logging): add CerrLogger std::cerr backend (3/6)

kamcheungting-db:
Part 3 of the logging stack — now based on `main` (#723 merged). Adds the first concrete sink: a dependency-free `std::cerr` logger, also the default backend when spdlog is off.

**What's here**
- `CerrLogger` writes one line per record: `YYYY-MM-DDThh:mm:ss.mmmZ LEVEL [tid] [file:line] message`.
- UTC millisecond timestamps, OS-native cached thread id (extracted to `util/thread_util.h`), lock-free level check, and a whole-line mutex so concurrent records never interleave.
- Pure standard library, always available. `Log`/`Flush` never throw. Inherits the base `Initialize("level")`; `pattern` is ignored (fixed layout; only the spdlog backend honors `pattern`).

**Tests** (`cerr_logger_test`): level filtering, the fixed line layout incl. the `.mmm` field and `[file:line]`, inherited `Initialize`, `Flush`, never-throw when the sink itself throws, and concurrent-write non-interleaving. Built/run under clang-libc++; ASan + TSan clean.


src/iceberg/logging/cerr_logger.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace iceberg {
+
+/// \brief Logger that writes one line per record to std::cerr.
+///
+/// Line layout: `YYYY-MM-DDThh:mm:ss.mmmZ LEVEL [tid] file:line] message`.
+/// The minimum level is held in a lock-free atomic; a mutex serializes the
+/// whole-line write so concurrent records never interleave. Pure standard
+/// library -- always compiled, regardless of ICEBERG_SPDLOG.
+class ICEBERG_EXPORT CerrLogger : public Logger {
```
wgtmac, line 40: CerrLogger is a formatting sink now, but it does not handle `kPatternProperty`. The public logger API says `pattern` is honored by CerrLogger. Please either implement `Initialize()` for the pattern here, or remove this promise until the later PR. A small test for `{"pattern", ...}` would help.

src/iceberg/logging/cerr_logger.cc
```diff
@@ -0,0 +1,105 @@
...
+std::string_view Basename(std::string_view path) noexcept {
+  auto pos = path.find_last_of("/\\");
+  return pos == std::string_view::npos ? path : path.substr(pos + 1);
+}
+
+/// \brief Format a record into a single newline-terminated line.
+std::string FormatLine(const LogMessage& message) {
+  auto now =
+      std::chrono::floor<std::chrono::milliseconds>(std::chrono::system_clock::now());
+  return std::format("{:%Y-%m-%dT%H:%M:%S}Z {} [{}] {}:{}] {}\n", now,
```
wgtmac: The format has `file:line]`, with a closing `]` but no opening `[`. This looks like a typo. Please use `[file:line]` or `file:line`, and add a test that checks this part of the layout.

src/iceberg/logging/cerr_logger.cc
```diff
@@ -0,0 +1,106 @@
...
```
kamcheungting-db:
```suggestion
  return std::format("{:%Y-%m-%dT%H:%M:%S}Z {} [{}] [{}:{}] {}\n", now,
```

src/iceberg/logging/cerr_logger.h
kamcheungting-db: Rethink about this part. It is not necessary to implement "pattern" on CerrLogger.
I have updated the documentation inside logger.h.

src/iceberg/logging/cerr_logger.cc
kamcheungting-db: Done

src/iceberg/logging/cerr_logger.cc
```diff
@@ -0,0 +1,77 @@
...
```
wgtmac: The PR description says the layout is `[file:line]`, but this still emits `file:line]`. Please update the format string, the header doc, and the regex test to use the same layout.

src/iceberg/logging/logger.h
```diff
@@ -127,8 +127,8 @@ class ICEBERG_EXPORT LogMessage::Builder {
 /// \brief Well-known Logger::Initialize() property keys.
 ///
 /// `level` is honored by the base Logger::Initialize (parsed via
-/// LogLevelFromString). `pattern` is honored by the formatting sinks
-/// (CerrLogger, SpdLogger).
+/// LogLevelFromString) on every backend. `pattern` is honored only by the
+/// spdlog backend; CerrLogger uses a fixed layout and ignores it.
```
wgtmac, line 131: This line is clear, but the `Initialize()` comment below still says formatting sinks apply `pattern`. CerrLogger is a formatting sink and now ignores `pattern`. Please narrow that later comment to the spdlog backend too.

src/iceberg/CMakeLists.txt
```diff
@@ -131,6 +132,7 @@ set(ICEBERG_SOURCES
     util/struct_like_set.cc
     util/task_group.cc
     util/temporal_util.cc
+    util/thread_util_internal.cc
```
wgtmac: We can just use `thread_util.cc` as its name without internal suffix.

src/iceberg/logging/cerr_logger.cc
wgtmac: Seems not resolved. See my comment below.

src/iceberg/logging/cerr_logger.cc
```diff
@@ -0,0 +1,77 @@
...
+
+#include "iceberg/util/thread_util_internal.h"
+
+namespace iceberg {
+
+namespace {
+
+/// \brief Trailing path component of a source file path.
+std::string_view Basename(std::string_view path) noexcept {
+  auto pos = path.find_last_of("/\\");
```
wgtmac, line 37: Is this platform compatible? Should we use std::filesystem?

src/iceberg/logging/cerr_logger.cc
kamcheungting-db: Done. uploaded a wrong version.

src/iceberg/logging/cerr_logger.cc
kamcheungting-db: fixed

src/iceberg/logging/logger.h
kamcheungting-db: Updated

src/iceberg/CMakeLists.txt
kamcheungting-db: ok

src/iceberg/logging/cerr_logger.cc
kamcheungting-db: Done

wgtmac (approved):

PR #725, opened 2026-06-11
feat(logging): add logging macros (4/6)

kamcheungting-db:
Part 4 of the logging stack (builds on #724). Adds the application-facing macros — the part most callers actually use.

**What's here**
- `ICEBERG_LOG_{TRACE,DEBUG,INFO,WARN,ERROR,CRITICAL,FATAL}`, the generic `ICEBERG_LOG(level, …)`, `ICEBERG_LOG_TO(logger, …)`, and `ICEBERG_LOG_RUNTIME_FMT` for runtime format strings.
- `ICEBERG_LOG_ACTIVE_LEVEL` is a compile-time floor: statements below it are removed entirely. `ICEBERG_LOG_FATAL` always emits, flushes, then aborts.
- Filtering is decided only by `ShouldLog()`; formatting happens only on the taken path and never throws (a bad format falls back safely).
- Generic/runtime-format macros use the same lock-free fast path as the fixed-severity ones (the slot lock is taken only on the fatal/abort path).
- Bare `LOG_*` aliases are opt-in via `ICEBERG_LOG_SHORT_MACROS`.

**Build note** — sets `/Zc:preprocessor` on MSVC, required for the `__VA_OPT__` used by the macros.

**Examples** — every macro takes a `std::format` string + args. The rendered line depends on the active backend; with the default `CerrLogger` it looks like:

```
ICEBERG_LOG_TRACE("entering scan for {}", table);
  2026-06-16T10:59:41.186Z trace [12345] table_scan.cc:88] entering scan for db.t
ICEBERG_LOG_DEBUG("cache miss key={}", key);
  2026-06-16T10:59:41.186Z debug [12345] cache.cc:42] cache miss key=manifest-7
ICEBERG_LOG_INFO("loaded {} manifests in {} ms", n, ms);
  2026-06-16T10:59:41.186Z info [12345] table_scan.cc:91] loaded 5 manifests in 12 ms
ICEBERG_LOG_WARN("retry {} after {}", attempt, err);
  2026-06-16T10:59:41.186Z warn [12345] io.cc:51] retry 2 after timeout
ICEBERG_LOG_ERROR("commit failed: {}", status);
  2026-06-16T10:59:41.186Z error [12345] txn.cc:77] commit failed: conflict
ICEBERG_LOG_CRITICAL("metadata unreadable at {}", path);
  2026-06-16T10:59:41.186Z critical [12345] meta.cc:30] metadata unreadable at s3://b/m.json
ICEBERG_LOG_FATAL("unrecoverable: {}", reason);   // emits, flushes, then std::abort()
  2026-06-16T10:59:41.186Z fatal [12345] boot.cc:19] unrecoverable: bad config
```

Less common forms: `ICEBERG_LOG(level, …)` (runtime severity), `ICEBERG_LOG_TO(logger, level, …)` (explicit logger), `ICEBERG_LOG_RUNTIME_FMT(level, fmt_string, args…)` (non-literal format). The same examples are documented inline in `logger.h`.

**Tests** — `macros_test` / `macros_active_level_test`: formatting, level gating, "argument not evaluated when disabled", compile-time stripping, never-throws, and FATAL-aborts death tests. clang/libc++.

### DIFF
--- a/src/iceberg/logging/logger.h
+++ b/src/iceberg/logging/logger.h
@@ -28,6 +28,7 @@
 
 #include <concepts>
 #include <cstdlib>
+#include <exception>
 #include <format>
 #include <memory>
 #include <source_location>
@@ -339,7 +340,7 @@ void FormatAndEmit(Logger& logger, LogLevel level, const std::source_location& l
   if (!logger.ShouldLog(level)) return;
   try {
     Emit(logger, level, loc, std::format(fmt, std::forward<Args>(args)...));
-  } catch (...) {
+  } catch (const std::exception&) {
     EmitFormatError(logger, level, loc);
   }
 }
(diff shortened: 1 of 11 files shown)


src/iceberg/logging/logger.h
```diff
@@ -371,3 +371,179 @@ void Log(Logger& logger, LogLevel level,
...
+  ICEBERG_INTERNAL_LOG(::iceberg::LogLevel::kError, __VA_ARGS__)
+#define ICEBERG_LOG_CRITICAL(...) \
+  ICEBERG_INTERNAL_LOG(::iceberg::LogLevel::kCritical, __VA_ARGS__)
+
+// FATAL: emit if enabled (never compile-stripped), then ALWAYS flush + abort.
+// Acquires the default logger ONCE and uses the same instance for emit and flush
+// so a concurrent SetDefaultLogger cannot flush a different logger than it emitted to.
+#define ICEBERG_LOG_FATAL(FMT_, ...)                                                   \
+  do {                                                                                 \
+    auto _ib_logger = ::iceberg::GetDefaultLogger();                                   \
```
wgtmac: This should use the current scoped logger rather than the process default. `ScopedLogger` says `LOG_*` routes through the scoped binding, but this path would send a fatal record to the global default and then abort.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +371,179 @@ void Log(Logger& logger, LogLevel level,
...
+// formatting only on the taken path, never throwing.
+#define ICEBERG_INTERNAL_LOG(level_, FMT_, ...)                                      \
+  do {                                                                               \
+    if constexpr ((level_) >= ICEBERG_LOG_ACTIVE_LEVEL) {                            \
+      const auto& _ib_logger = ::iceberg::internal::CurrentLogger();                 \
+      if (_ib_logger && _ib_logger->ShouldLog(level_)) {                             \
+        try {                                                                        \
+          ::iceberg::internal::Emit(*_ib_logger, (level_),                           \
+                                    ::std::source_location::current(),               \
+                                    ::std::format(FMT_ __VA_OPT__(, ) __VA_ARGS__)); \
```
wgtmac: **Performance/Heap Allocation:** `std::format` returns a `std::string`. If the formatted message exceeds the Small String Optimization (SSO) limit, this will trigger a heap allocation before the logger even processes it. For a high-performance system like Iceberg, this could become a bottleneck on hot paths. Consider using a stack/thread-local buffer (like `fmt::memory_buffer` in `fmtlib` or `spdlog`) to achieve zero-allocation formatting.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +371,179 @@ void Log(Logger& logger, LogLevel level,
...
+/// are never emitted). Defaults to keeping everything. ICEBERG_LOG_FATAL is never
+/// gated by this floor -- its abort is always compiled in.
+#ifndef ICEBERG_LOG_ACTIVE_LEVEL
+#  define ICEBERG_LOG_ACTIVE_LEVEL ::iceberg::LogLevel::kTrace
+#endif
+
+// Internal: fixed-severity emit with compile-time floor then the authoritative
+// Logger::ShouldLog (the single source of truth for runtime filtering), with
+// formatting only on the taken path, never throwing.
+#define ICEBERG_INTERNAL_LOG(level_, FMT_, ...)                                      \
```
wgtmac: **Binary Bloat:** Expanding this entire block (including `if`, `try-catch`, and `ShouldLog`) at every log call site can significantly bloat the binary size and reduce instruction cache hit rates. A common optimization is to use type erasure (e.g., `std::make_format_args`) and push the actual `try-catch` execution into a non-inline function in the `.cc` file.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +371,179 @@ void Log(Logger& logger, LogLevel level,
...
+  do {                                                                               \
+    if constexpr ((level_) >= ICEBERG_LOG_ACTIVE_LEVEL) {                            \
+      const auto& _ib_logger = ::iceberg::internal::CurrentLogger();                 \
+      if (_ib_logger && _ib_logger->ShouldLog(level_)) {                             \
+        try {                                                                        \
+          ::iceberg::internal::Emit(*_ib_logger, (level_),                           \
+                                    ::std::source_location::current(),               \
+                                    ::std::format(FMT_ __VA_OPT__(, ) __VA_ARGS__)); \
+        } catch (...) {                                                              \
```
wgtmac: **Exception Handling Scope:** `catch (...)` is very broad and will swallow all exceptions, including `std::bad_alloc` (OOM) or thread cancellation exceptions (e.g., POSIX `abi::__forced_unwind`). Since `std::format` typically throws `std::format_error`, it is safer to catch `const std::format_error&` or `const std::exception&` so fatal system exceptions can propagate.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +371,179 @@ void Log(Logger& logger, LogLevel level,
...
+        ::iceberg::internal::Emit(*_ib_logger, ::iceberg::LogLevel::kFatal,            \
+                                  ::std::source_location::current(),                   \
+                                  ::std::format(FMT_ __VA_OPT__(, ) __VA_ARGS__));     \
+      } catch (...) {                                                                  \
+        ::iceberg::internal::EmitFormatError(*_ib_logger, ::iceberg::LogLevel::kFatal, \
+                                             ::std::source_location::current());       \
+      }                                                                                \
+    }                                                                                  \
+    if (_ib_logger) _ib_logger->Flush();                                               \
+    ::std::abort();                                                                    \
```
wgtmac: **Extensibility:** Hardcoding `std::abort()` removes the ability for the host application (like a Java JNI or Python process embedding Iceberg) to handle fatal errors gracefully (e.g., flushing resources or printing a stack trace). Consider allowing the user to register a custom `FatalHandler` callback before falling back to `abort()`.

wgtmac (commented):
Great work on the logging macros! I have a few suggestions regarding performance, binary size, and exception handling for your consideration.

src/iceberg/logging/logger.h
kamcheungting-db: Good catch on the SSO/heap-allocation concern. Rather than pull the stack/thread-local buffer work into this PR, I split it out into #795, which makes the format→dispatch path allocation-free (sinks receive a non-owning `LogRecord` view instead of an owned message). Keeping this PR simple and focused so the two concerns can be reviewed independently.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +375,182 @@ void Log(Logger& logger, LogLevel level,
...
+#define ICEBERG_INTERNAL_LOG(level_, FMT_, ...)                                      \
+  do {                                                                               \
+    if constexpr ((level_) >= ICEBERG_LOG_ACTIVE_LEVEL) {                            \
+      const auto& _ib_logger = ::iceberg::internal::CurrentLogger();                 \
+      if (_ib_logger && _ib_logger->ShouldLog(level_)) {                             \
+        try {                                                                        \
+          ::iceberg::internal::Emit(*_ib_logger, (level_),                           \
+                                    ::std::source_location::current(),               \
+                                    ::std::format(FMT_ __VA_OPT__(, ) __VA_ARGS__)); \
+        } catch (const std::exception&) {                                            \
```
wgtmac: This no longer fully matches the “logging never throws” guarantee. `std::format` can call user-defined formatters, and those can throw non-`std::exception` types; with this catch they escape (or hit `noexcept` in the function API). Either keep the catch-all or narrow the guarantee.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +375,182 @@ void Log(Logger& logger, LogLevel level,
...
+//   ICEBERG_LOG_TO(logger, level, "to an explicit logger {}", y);
+//   ICEBERG_LOG_RUNTIME_FMT(level, fmt_string, args...);       // non-literal format
+//
+// With ICEBERG_LOG_SHORT_MACROS defined, bare aliases (LOG_INFO, ...) are also
+// available. A format string is mandatory; zero extra args is fine
+// (ICEBERG_LOG_INFO("done")).
+// ---------------------------------------------------------------------------
+
+/// \brief Compile-time severity floor: statements below this level are removed
+/// entirely from the build (their format call sites and source_location literals
```
wgtmac: `if constexpr` does not remove this at preprocessing time. Even when the branch is discarded, the log statement still has to be syntactically/semantically valid, so bad format strings or non-formattable args may still fail to compile. Maybe soften “removed entirely” here.

src/iceberg/test/CMakeLists.txt
```diff
@@ -64,7 +64,9 @@ function(add_iceberg_test test_name)
   endif()
 
   if(MSVC_TOOLCHAIN)
-    target_compile_options(${test_name} PRIVATE /bigobj)
+    # /Zc:preprocessor: conforming preprocessor for the __VA_OPT__ in the logging
+    # macros (MSVC's traditional preprocessor rejects it).
+    target_compile_options(${test_name} PRIVATE /bigobj /Zc:preprocessor)
```
wgtmac, line 67: Meson adds `/Zc:preprocessor` project-wide, but CMake only adds it to tests. Since `logger.h` is public and `logger.cc` includes it, the library targets (and ideally consumers) also need the conforming preprocessor option on MSVC.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +375,182 @@ void Log(Logger& logger, LogLevel level,
...
+#define ICEBERG_LOG_CRITICAL(...) \
+  ICEBERG_INTERNAL_LOG(::iceberg::LogLevel::kCritical, __VA_ARGS__)
+
+// FATAL: emit if enabled (never compile-stripped), then ALWAYS flush + abort.
+// Acquires the effective (scoped-or-default) logger ONCE -- honoring any active
+// ScopedLogger binding -- and uses that same instance for emit and flush, so a
+// concurrent SetDefaultLogger cannot flush a different logger than it emitted to.
+#define ICEBERG_LOG_FATAL(FMT_, ...)                                                   \
+  do {                                                                                 \
+    auto _ib_logger = ::iceberg::GetCurrentLogger();                                   \
```
wgtmac: Can we add a small death/regression test for this path under `ScopedLogger`? That would lock in the switch from `GetDefaultLogger()` to `GetCurrentLogger()` for fixed `ICEBERG_LOG_FATAL`.

src/iceberg/logging/logger.h
```diff
@@ -371,3 +375,182 @@ void Log(Logger& logger, LogLevel level,
...
+// ---------------------------------------------------------------------------
+// Logging macros.
+//
+// Every macro takes a std::format string followed by its arguments. The
+// rendered line depends on the active backend (see cerr_logger.h for the
+// std::cerr layout, or the spdlog pattern); the examples below show the call
+// site and, for the default CerrLogger, the line it produces.
+//
+//   ICEBERG_LOG_TRACE("entering scan for {}", table);
+//     2026-06-16T10:59:41.186Z trace [12345] table_scan.cc:88] entering scan for db.t
```
wgtmac: The example output is missing the opening `[` before the source location. `CerrLogger` formats it as `[file:line]`.

wgtmac:
I put together a small draft PR with the macro shape I had in mind: #824.

My concern here is mostly about the structure, not individual lines. Right now `logger.h` exposes macros to every consumer, short `LOG_*` names are on the default path, and the fixed-level/runtime/explicit/fatal paths duplicate formatting, filtering, and fatal handling logic.

The prototype keeps `logger.h` as the C++ API only, moves `ICEBERG_*` macros to `log_macros.h`, puts `LOG_*` aliases behind an opt-in header, and uses preprocessor active-level gating. Not suggesting it should replace this PR; it is just a concrete reference for the direction I think would be easier to maintain.

kamcheungting-db:
Thanks — #824 was a helpful concrete reference. I've folded the structural direction into this PR:

**Adopted from #824**
- **Header split:** `logger.h` is now the C++ API only. `ICEBERG_LOG_*` macros moved to a new `log_macros.h`; the bare `LOG_*` aliases moved behind an opt-in `short_log_macros.h` (replacing the `ICEBERG_LOG_SHORT_MACROS` define). Consumers of just the C++ API no longer pull in the macros or the conforming-preprocessor requirement.
- **Deduplication:** the five macro bodies now delegate to shared `internal::` helpers (`EmitIfEnabled` + `LogToCurrent` / `LogToCurrentRuntime` / `LogToExplicitRuntime` / `LogFatal`) taking a lazy `[&]()->std::string` message thunk invoked only past `ShouldLog`, so disabled logs still don't evaluate their arguments. `VFormat` moved alongside the macros (only `ICEBERG_LOG_RUNTIME_FMT` uses it).

**One deliberate difference:** I kept the `if constexpr` compile-time floor rather than the preprocessor `#if … ((void)0)` gating. With `if constexpr`, a disabled fixed-level log is still parsed and type-checked, so a bad format string / non-formattable arg is caught immediately instead of hiding until the floor is lowered — consistent with your earlier note that the statement must remain well-formed. Happy to switch if you'd prefer the spdlog-style "not even compiled" guarantee.

Also addressed the rest of the review in this round:
- `ICEBERG_LOG_FATAL` routes through `GetCurrentLogger()` (scoped-aware), with a `FatalRoutesThroughScopedLogger` regression death test.
- Reverted to `catch (...)` to keep the noexcept "logging never throws" guarantee for user formatters that throw non-`std::exception` types.
- `/Zc:preprocessor` is now applied PUBLIC to the library targets on MSVC (not just tests).
- Softened the compile-time-floor doc wording; fixed the example output to `[file:line]`.
- **Extensibility:** added a registerable `SetFatalHandler` invoked on the fatal path (after emit+flush, before `abort()`), so a JNI/Python host or crash reporter can flush resources / print a stack trace; the handler receives the source location and formatted message. Death tests cover it.

The owned-`std::string` message is unchanged here; the zero-allocation `LogRecord`/reused-buffer path is in #795, and I'll reconcile the thunk with it when that lands.

wgtmac (approved):
Thanks for reworking this. The structure looks much better now, and this looks good to me.\n\nI still slightly prefer preprocessor gating for the active-level macros, but I don't think that needs to block this PR.

kamcheungting-db:
> Thanks for reworking this. The structure looks much better now, and this looks good to me.\n\nI still slightly prefer preprocessor gating for the active-level macros, but I don't think that needs to block this PR.

Acked. I would move to preprocessor gating in next PR.

wgtmac:
Let's merge this to unblock followup PRs. Thank you, @kamcheungting-db!

PR #726, opened 2026-06-11
feat(logging): add spdlog backend behind ICEBERG_SPDLOG (5/6)

kamcheungting-db:
Part 5 of the logging stack (builds on #725). Adds the spdlog backend and the build option that selects it — the default in production builds.

**What's here**
- `SpdLogger` wraps `spdlog::logger` (synchronous), maps levels (fatal/critical → spdlog critical; the abort stays in the macro layer), and forwards source location. Default sink is a colored stderr sink.
- New `ICEBERG_SPDLOG` CMake option and matching Meson `spdlog` feature option (enabled by default). When disabled, the build has no spdlog dependency and `CerrLogger` is the default.
- `pattern` property is honored here via `spdlog set_pattern` (the cerr backend keeps its fixed layout); `level` works on both backends.
- `SpdLogger` lives in `logging/spdlog_logger_internal.h` and is not installed — apps get it via the default logger.
- Runtime `kFatal` macro paths now use the same fatal-handler and abort sequence as `ICEBERG_LOG_FATAL`.

**Tests** — logging tests cover default-backend selection, level filtering and mapping including fatal→critical, source-location forwarding, the `pattern` property, and runtime fatal-handler routing.


wgtmac:
It's time to rebase it :)

src/iceberg/logging/internal/spdlog_logger.cc
```diff
@@ -0,0 +1,107 @@
...
+}
+
+void SpdLogger::Log(LogMessage&& message) noexcept {
+  try {
+    spdlog::source_loc loc{message.location.file_name(),
+                           static_cast<int>(message.location.line()),
+                           message.location.function_name()};
+    // Pass the pre-formatted text as an argument ("{}") so any braces in the
+    // message are not re-interpreted as a format string.
+    logger_->log(loc, ToSpdLevel(message.level), "{}", message.message);
```
wgtmac: This re-formats an already formatted message through fmt for every emitted record, adding another full copy and possibly an allocation for long messages. Please call the raw-message overload with `spdlog::string_view_t{message.message.data(), message.message.size()}` instead.

src/iceberg/logging/internal/spdlog_logger.cc
```diff
@@ -0,0 +1,107 @@
...
+  // Apply "level" via the base implementation.
+  return Logger::Initialize(properties);
+}
+
+SpdLogger::SpdLogger(std::shared_ptr<spdlog::logger> logger, LogLevel level)
+    : logger_(std::move(logger)), level_(level) {
+  // logger_ is a hard precondition: SpdLogger is not consumer-constructible (it is
+  // obtained via the default logger or the "spdlog" registry factory, both of which
+  // pass a real logger), so Initialize/Log/Flush may dereference it unconditionally.
+  ICEBERG_DCHECK(logger_ != nullptr, "SpdLogger requires a non-null spdlog::logger");
```
wgtmac: `ICEBERG_DCHECK` disappears under `NDEBUG`, and the constructor dereferences `logger_` immediately afterwards. Since `shared_ptr` is nullable and the header only documents the synchronous requirement, an empty input becomes a release crash. Please reject null in a factory or encode the non-null requirement in the API.

src/iceberg/logging/meson.build
```diff
@@ -15,7 +15,23 @@
 # specific language governing permissions and limitations
 # under the License.
 
+# Generate the .cc-only logging backend config header. The meson build always
+# links spdlog, so ICEBERG_HAS_SPDLOG is always defined here. Generated into
+# build/src/iceberg/logging/config.h (resolved via include_directories('..'),
+# which exposes both the source and build trees); not installed.
+logging_config_data = configuration_data()
+logging_config_data.set('ICEBERG_HAS_SPDLOG', 1)
```
wgtmac: CMake supports `ICEBERG_SPDLOG=OFF`, but Meson always enables this backend and unconditionally links spdlog. Please add a matching Meson feature option so the no-spdlog/Cerr configuration is available in both supported build systems.

src/iceberg/test/spdlog_logger_test.cc
```diff
@@ -0,0 +1,106 @@
...
+}  // namespace
+
+TEST(SpdLoggerTest, DefaultLevelIsInfo) {
+  internal::SpdLogger logger;
+  EXPECT_EQ(logger.level(), LogLevel::kInfo);
+  EXPECT_FALSE(logger.ShouldLog(LogLevel::kDebug));
+  EXPECT_TRUE(logger.ShouldLog(LogLevel::kError));
+}
+
+TEST(SpdLoggerTest, ForwardsMessageToSink) {
```
wgtmac, line 103: This only verifies the payload, so the advertised source-location forwarding is untested. Please use a `%s:%# %! %v` pattern and assert the file, line, and function fields.

wgtmac (approved):

wgtmac:
I did another cleanup pass to align this PR with repository conventions: kept SpdLogger behind a non-installed internal header, made it own spdlog::logger by value, simplified the CMake/Meson feature gating, and switched the tests to an injected ostream sink for Windows portability. This keeps spdlog out of the public API while preserving clean no-spdlog builds.

Thanks for working on this, @kamcheungting-db!

PR #727, opened 2026-06-11
feat: Collect Parquet NaN metrics during writes

WZhuo:
Collects NaN value counts for float and double columns during Parquet writes, since the Parquet footer statistics do not track NaN counts.

## Changes

- **Write-side NaN metric collection** (`FieldMetricsCollector`): A visitor that walks each record batch before writing, accumulating value counts, null counts, NaN counts, and NaN-excluding lower/upper bounds for float/double fields.
- **MetricsConfig-aware skipping**: Fields whose `MetricsMode` is `kNone` are skipped entirely, avoiding wasted work.
- **Integration with existing footer metrics**: Write-side `FieldMetrics` take precedence over footer statistics in `ParquetMetrics::GetMetrics`, so NaN counts are populated while counts/bounds still fall back to footer stats when write-side data isn't available.
- **Tests**: `ParquetMetricsTest` now overrides `ReportsNanCounts()` to `true`, and existing NaN test cases verify NaN counts alongside existing value/null count assertions.

## Behavior alignment with Java

- Fields nested inside lists/maps do not get NaN metrics (both Java and C++ agree — Java collects then discards; C++ skips collection entirely).
- NaN values are excluded from lower/upper bounds in both implementations.
- Float/double fields with all-NaN values correctly set `nan_value_count` without setting bounds.

### DIFF
--- a/src/iceberg/parquet/parquet_writer.cc
+++ b/src/iceberg/parquet/parquet_writer.cc
@@ -19,10 +19,16 @@
 
 #include "iceberg/parquet/parquet_writer.h"
 
+#include <cmath>
+#include <cstdint>
 #include <memory>
+#include <optional>
 #include <string_view>
+#include <type_traits>
+#include <unordered_map>
 #include <vector>
 
+#include <arrow/array.h>
 #include <arrow/c/bridge.h>
 #include <arrow/record_batch.h>
 #include <arrow/util/compression.h>
@@ -36,7 +42,9 @@
 #include "iceberg/arrow/arrow_status_internal.h"
 #include "iceberg/parquet/parquet_metrics_internal.h"
 #include "iceberg/schema_internal.h"
+#include "iceberg/type.h"
 #include "iceberg/util/macros.h"
+#include "iceberg/util/visit_type.h"
 
 namespace iceberg::parquet {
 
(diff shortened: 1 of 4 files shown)


src/iceberg/parquet/parquet_writer.cc
```diff
@@ -74,6 +82,144 @@ Status CheckCompressionAvailable(std::string_view compression_name,
...
+    if (!field_metrics.upper_bound.has_value() ||
+        literal > field_metrics.upper_bound.value()) {
+      field_metrics.upper_bound = std::move(literal);
+    }
+  }
+
+  return {};
+}
+
+std::optional<std::vector<uint8_t>> BuildValidRows(const ::arrow::Array& array,
```
wgtmac, line 140: Non-blocking: this could be made cheaper by carrying Arrow validity bitmaps instead of materializing a `std::vector<uint8_t>` and calling `array.IsNull(i)` for every row at each struct level. The main detail is that this needs to preserve the current parent-validity behavior: child arrays of a null `StructArray` do not necessarily have those parent nulls reflected in their own null bitmap, so the effective validity should be `parent_validity AND array.null_bitmap()`.

One possible approach is to keep a bitmap/buffer plus offset for the current effective validity, and when descending into a field, use Arrow bitmap utilities to AND it with the child array validity bitmap (or just reuse the parent bitmap when the child has `null_count() == 0`). Then the float/double collector can iterate only the effective bitmap bits instead of indexing a byte vector. This should avoid per-level byte-vector allocation and reduce the per-row validity checks for wide/deep struct writes.

wgtmac (approved):

PR #737, opened 2026-06-14
feat(logging): add Loggers registry (6/6)

kamcheungting-db:
Part 6 of the logging stack (builds on #726). Lets applications pick a logging backend from configuration and install it as the process default — the same pattern as `MetricsReporters`.

**How it works**

Choose a backend with the `logger-impl` property; built-ins are `noop`, `cerr`, and `spdlog` (when compiled in).

```cpp
// Build a logger from properties and install it as the process default.
auto logger = iceberg::Loggers::Load({{"logger-impl", "spdlog"}, {"level", "info"}});
if (logger) {
  iceberg::SetDefaultLogger(
      std::shared_ptr<iceberg::Logger>(std::move(*logger)));
}

// ...then anywhere in the code:
ICEBERG_LOG_INFO("scan planned: {} files", n);
```

**API**
- `Loggers::Load(properties)` — build a logger, choosing the type from `logger-impl`.
- `Loggers::Register(type, factory)` — register a custom backend under a new name.

With no `logger-impl` set, the default is `spdlog` when compiled in, otherwise `cerr` — so logging works out of the box.

**Tests** — `loggers_test` drives the full path an app uses: configure a backend + level via properties, install it, log through the macros, and check the real output. Covers configured level handling and a macro reaching a real spdlog sink, with spdlog ON and OFF.


manuzhang:
@kamcheungting-db can you please rebase your PR?

wgtmac (approved):
Thank you for adding the registry!

PR #738, opened 2026-06-14
fix: fix URL encoding for non-ASCII bytes

fallintoplace:
## Summary

- Treat URL encoder input bytes as unsigned before splitting them into hex nibbles.
- Add UTF-8 regression coverage for `café` encoding to `caf%C3%A9`.

## Why

On platforms where `char` is signed, bytes >= 0x80 can become negative before the hex lookup. That can lead to undefined behavior or incorrect percent encoding for non-ASCII strings.

## Testing

- `cmake -S . -B build -G Ninja -DICEBERG_BUILD_BUNDLE=OFF -DICEBERG_BUILD_REST=OFF -DICEBERG_BUILD_HIVE=OFF -DICEBERG_BUILD_SQL_CATALOG=OFF`
- `cmake --build build --target util_test`
- `ctest --test-dir build -R util_test --output-on-failure`


zhjwpku (approved):
Good catch, for non-ASCII input bytes, kHexChars[c >> 4] can lead to out-of-bounds access and undefined behavior. +1 for this fix.

wgtmac (approved):

wgtmac:
Thanks @fallintoplace for fixing this and @zhjwpku for the review!

PR #739, opened 2026-06-14
fix: validate truncate widths in utility entrypoints

fallintoplace:
## Summary

- Reject non-positive widths in the `TruncateLiteral` and `TruncateLiteralMax` utility entrypoints.
- Reject zero width in `TruncateUTF8Max`, which returns a `Result` and can report invalid input directly.
- Add regression coverage for zero and negative widths across numeric, decimal, string, and binary literal truncation.

## Why

The truncate transform validates its width before creating a transform function, but the utility functions can also be called directly. Without a guard there, invalid widths can reach modulo operations or iterator arithmetic in the type-specific implementations.

## Testing

- `cmake --build build --target util_test`
- `ctest --test-dir build -R util_test --output-on-failure`

### DIFF
--- a/src/iceberg/test/truncate_util_test.cc
+++ b/src/iceberg/test/truncate_util_test.cc
@@ -51,6 +51,26 @@ TEST(TruncateUtilTest, TruncateLiteral) {
             Literal::Binary(std::vector<uint8_t>(expected.begin(), expected.end())));
 }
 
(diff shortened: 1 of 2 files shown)


wgtmac (approved):

wgtmac:
Thank you, @fallintoplace!

PR #741, opened 2026-06-14
feat(update): add OverwriteFiles for overwrite snapshot commits

lishuxu:
Summary:
Add a production OverwriteFiles builder that brings iceberg-cpp to semantic parity with Java's BaseOverwriteFiles. It supports explicit file replacement (DeleteFile + AddFile) and range-based replacement (OverwriteByRowFilter + AddFile) with the same family of pre-commit concurrency validations. The builder is a thin subclass of MergingSnapshotUpdate and reuses the existing commit kernel (Apply/summary/retry/cleanup) unchanged.

Changes:
- New OverwriteFiles class (src/iceberg/update/overwrite_files.{h,cc}) and Table::NewOverwrite() / Transaction::NewOverwrite() entry points.
- Builder surface: AddFile, DeleteFile, bulk DeleteFiles, OverwriteByRowFilter, ValidateFromSnapshot, ConflictDetectionFilter, ValidateNoConflictingData, ValidateNoConflictingDeletes, ValidateAddedFilesMatchOverwriteFilter, WithCaseSensitivity.
- Validate(): conflict-filter resolution, concurrent add/delete conflict checks, and strict added-file range validation (projection + StrictMetricsEvaluator).
- Tests (overwrite_files_test.cc, 45 cases) and CMake/meson wiring.

Behavior alignment with Java:
- operation() returns append/delete/overwrite from builder content.
- Conflict-filter resolution mirrors BaseOverwriteFiles (explicit -> row filter -> AlwaysTrue); replaced-file delete checks honor ConflictDetectionFilter.
- Strict added-file validation uses a single DataSpec(), rejecting multi-spec and empty added-file sets.
- Deviations: public WithCaseSensitivity (vs caseSensitive) to avoid a protected-name clash; ValidateFromSnapshot rejects negative ids early.

### DIFF
--- a/src/iceberg/table.cc
+++ b/src/iceberg/table.cc
@@ -35,8 +35,8 @@
 #include "iceberg/update/expire_snapshots.h"
 #include "iceberg/update/fast_append.h"
 #include "iceberg/update/merge_append.h"
-#include "iceberg/update/row_delta.h"
 #include "iceberg/update/overwrite_files.h"
+#include "iceberg/update/row_delta.h"
 #include "iceberg/update/set_snapshot.h"
 #include "iceberg/update/snapshot_manager.h"
 #include "iceberg/update/update_location.h"
@@ -354,6 +354,10 @@ Result<std::shared_ptr<DeleteFiles>> StaticTable::NewDeleteFiles() {
 
 Result<std::shared_ptr<RowDelta>> StaticTable::NewRowDelta() {
   return NotSupported("Cannot create a row delta for a static table");
+}
+
+Result<std::shared_ptr<OverwriteFiles>> StaticTable::NewOverwrite() {
+  return NotSupported("Cannot create an overwrite for a static table");
 }
 
 Result<std::shared_ptr<SnapshotManager>> StaticTable::NewSnapshotManager() {
--- a/src/iceberg/table.cc
+++ b/src/iceberg/table.cc
@@ -36,6 +36,7 @@
 #include "iceberg/update/fast_append.h"
 #include "iceberg/update/merge_append.h"
 #include "iceberg/update/row_delta.h"
+#include "iceberg/update/overwrite_files.h"
 #include "iceberg/update/set_snapshot.h"
 #include "iceberg/update/snapshot_manager.h"
 #include "iceberg/update/update_location.h"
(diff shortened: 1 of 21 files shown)


src/iceberg/table.cc
```diff
@@ -217,6 +218,12 @@ Result<std::shared_ptr<FastAppend>> Table::NewFastAppend() {
   return FastAppend::Make(name().name, std::move(ctx));
 }
 
+Result<std::shared_ptr<OverwriteFiles>> Table::NewOverwrite() {
```
wgtmac, line 242: This adds `Table::NewOverwrite()` for all `Table` subclasses, but `StaticTable` does not override it. Java `BaseReadOnlyTable.newOverwrite()` rejects read-only/static tables, and the existing C++ `StaticTable` overrides other mutating builders for the same reason. Please add a `StaticTable::NewOverwrite()` override returning `NotSupported` and cover it in `StaticTableTest.NewMutatingOperationsAreNotSupported` so static tables cannot expose a writable overwrite API.

src/iceberg/update/overwrite_files.h
```diff
@@ -0,0 +1,135 @@
...
+  /// overwrite range.
+  ///
+  /// \return Reference to this for method chaining
+  OverwriteFiles& ValidateAddedFilesMatchOverwriteFilter();
+
+  /// \brief Set case sensitivity for binding, projection, and metrics evaluation.
+  ///
+  /// \param case_sensitive Whether matching should be case-sensitive
+  /// \return Reference to this for method chaining
+  OverwriteFiles& WithCaseSensitivity(bool case_sensitive);
```
wgtmac: The Java API is `caseSensitive(boolean)`, and the surrounding C++ update builders expose this as `CaseSensitive(bool)` (`DeleteFiles`, `UpdateSchema`, `TableScanBuilder`, etc.). Naming this `WithCaseSensitivity` makes this new builder inconsistent with both Java parity and local C++ API conventions, so callers porting overwrite code will not find/compile the expected method. Please expose this as `CaseSensitive(bool)`.

src/iceberg/update/overwrite_files.h
```diff
@@ -0,0 +1,135 @@
...
+
+  std::string operation() override;
+
+  Status Validate(const TableMetadata& current_metadata,
```
wgtmac, line 183: `Validate` is an internal snapshot-update hook in the existing C++ hierarchy (`DeleteFiles::Validate` is protected) and Java keeps `BaseOverwriteFiles.validate(...)` protected as well. Exposing it publicly here adds a public API surface that users can call directly and that will be hard to remove later. Please keep this override non-public, and use a test helper/subclass if tests need to invoke it directly.

wgtmac (approved):

wgtmac:
Thanks @lishuxu

PR #742, opened 2026-06-15
feat: implement RenameTable in InMemoryCatalog

WZhuo:
Replace the NotImplemented stub with a full implementation that:
- Returns early as a no-op when renaming a table to itself
- Returns NoSuchTable when the source table does not exist
- Returns AlreadyExists when the destination table already exists
- Moves the metadata location from source to destination identifier

Add unit test covering rename to self, nonexistent source, destination conflict, and full rename + load verification.


zhjwpku (approved):
LGTM

wgtmac (approved):

PR #744, opened 2026-06-15
feat: implement purge support in DropTable for InMemoryCatalog and SqlCatalog

WZhuo:
When purge=true, delete all metadata files (current + metadata log entries) before unregistering the table. Old log files are deleted first so the current metadata file remains as an anchor for retries.

### DIFF
--- a/src/iceberg/catalog/memory/in_memory_catalog.cc
+++ b/src/iceberg/catalog/memory/in_memory_catalog.cc
@@ -22,6 +22,7 @@
 #include <algorithm>
 #include <iterator>
 
+#include "iceberg/file_io.h"
 #include "iceberg/table.h"
 #include "iceberg/table_identifier.h"
 #include "iceberg/table_metadata.h"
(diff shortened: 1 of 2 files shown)


src/iceberg/catalog/memory/in_memory_catalog.cc
```diff
@@ -511,7 +512,19 @@ Result<bool> InMemoryCatalog::TableExists(const TableIdentifier& identifier) con
...
-  // TODO(Guotao): Delete all metadata files if purge is true.
+  if (purge && file_io_) {
+    ICEBERG_ASSIGN_OR_RAISE(auto metadata_location,
+                            root_namespace_->GetTableMetadataLocation(identifier));
+    ICEBERG_ASSIGN_OR_RAISE(auto metadata,
+                            TableMetadataUtil::Read(*file_io_, metadata_location));
+    // Delete previous metadata files from the log first, so that if deletion
+    // fails and is retried, the current metadata file still exists as an
+    // anchor to locate any remaining old files.
+    for (const auto& entry : metadata->metadata_log) {
```
zhjwpku, line 525: Not a blocking comment, but should we use the newly introduced DeleteFiles API so that batch deletion can be leveraged when supported by the underlying FileIO implementation?

zhjwpku (approved):

zhjwpku (approved):

wgtmac (approved):

PR #745, opened 2026-06-15
fix: reject tag refs for main snapshot ref

zhjwpku:
Validate that the reserved main snapshot ref is always a branch, and preserve snapshot log timestamps for newly added snapshots without overwriting table metadata update time.

This behavior is aligned with Java's TableMetadata.Builder.setRef


huan233usc (approved):

wgtmac (approved):

wgtmac:
Thanks @zhjwpku for the fix and @huan233usc for the review!

PR #731, opened 2026-06-11
[wip]feat: support column default values (initial-default / write-default)

huan233usc:
Closes #730 (item 2 of #637).

This end-to-end proof of concept has been split into four smaller pull requests:

1. #746 — Schema model, JSON serialization, and v3 validation
2. #792 — Parquet read-path support
3. #800 — Avro read-path support
4. #793 — Schema-evolution support in `UpdateSchema`

Please review the split pull requests above. This PR is retained only as the original end-to-end proof of concept.

Issue #730, opened 2026-06-11
Support column default values (initial-default / write-default)


huan233usc:
Sub-issue of #637 (item 2: Column Default Values).
Iceberg v3 allows schema evolution with column default values (spec — Default values (https://iceberg.apache.org/spec/#default-values)):
  - initial-default: returned for rows that predate the column's addition,  i.e. applied on the read path when a data file does not contain the field.
  - write-default: applied on the write path when a writer does not supply a value or the column.

 Planned scope:
 - [x] Schema model: carry initial-default / write-default on SchemaField
 - [x] JSON serde: parse/write the two fields using single-value serialization
 - [x] Read path: project missing columns as the initial-default constant instead of null when reading data files written before the column existed
 - [x] Write path: apply write-default where values are omitted; validation (e.g. adding a required column requires an initial-default, v3+ gating)
I plan to work on this and will send PR(s) referencing this issue.

Issue #637, opened 2026-05-05
Track Iceberg v3 Feature Support (Roadmap for 0.4.0)


zhjwpku:
We already have **partial support for Iceberg v3** in the codebase (e.g., Row Lineage and partition statistics). However, a number of key v3 features are still missing.

This issue serves as a **parent tracking issue** for the remaining v3 work we plan to (partially) include in the **0.3.0 release**.

If you are interested in working on any of the items below, please **open a dedicated sub-issue** and link it back to this one.

### 1. Schema-Level Support for New v3 Types

Add support for all newly introduced v3 types at the schema level, including:

- [x] `timestamp_ns`, `timestamptz_ns` @zhjwpku 
- [x] `unknown` @manuzhang 
- [ ] `variant` @zhjwpku rely on arrow, defered to 0.5.0
- [x] `geometry`, `geography` @wgtmac 

### 2. Column Default Values (Schema Evolution)

Enable schema evolution with default values:

- [x] Support `initial-default` and `write-default` @huan233usc
- [ ] Apply defaults correctly during read and write paths @huan233usc  #942 

### 4. Binary Deletion Vectors

Add support for deletion vectors (DV):

- [x] Puffin-based DV parsing @zhaoxuan1994 

### 5. Table-Level Encryption Metadata (Not planned for 0.4.0) @shangxinli 

Support encryption-related metadata defined in v3:

- [ ] Parse and preserve encryption metadata

### 6. Puffin-Based Statistics (Not planned for 0.4.0)

Support table and partition statistics stored in Puffin files:

- [ ] Metadata parsing
- [ ] Optional integration into scan planning

## Table update and maintenance: #364

- [x] OverwriteFiles @lishuxu 
- [x] RowDelta @manuzhang 
- [x] MergeAppend @zhjwpku 
- [x] DeleteFiles @manuzhang 
- [x] RewriteFiles @WZhuo 
- [x] ReplacePartitions @shangxinli 
- [ ] RewriteManifests @zhjwpku #774 
- [ ] CherryPickOperation @shangxinli #926 

## Miscellaneous

- [x] Metadata table inspection @All-less @WZhuo 
- [x] LRU cache @HuaHuaY 
- [x] Logging interface and spdlog integration @kamcheungting-db

If anything is missing or you have suggestions, feel free to comment or open additional sub-issues.

manuzhang:
[...]

PR #746, opened 2026-06-15
feat(schema): represent, serialize and validate v3 column default values (1/4)

huan233usc:
Part 1 of a multi-part split of #730 (column default values, item 2 of #637). The full
end-to-end implementation is in #731, kept open as the proof-of-concept; this series
lands it in reviewable pieces.

This PR is the **schema foundation** — representing, serializing and validating v3
column default values. It is purely additive and changes no read or write behavior on
its own.

## What's in this PR

- **`SchemaField`** carries `initial-default` / `write-default`, stored as
  `std::shared_ptr<const Literal>` (immutable payload shared across copies, like the
  adjacent `type_`; the C++ analog of Java's `final Literal<?>`). They are set via the
  constructor. Getters return `std::optional<std::reference_wrapper<const Literal>>` for
  reading (the `Schema::FindFieldByName` idiom); `initial_default_ptr()` /
  `write_default_ptr()` expose the shared pointer so a rebuilt field (e.g. ID
  reassignment) shares the value instead of copying it.
- **JSON serde**: parse/write `initial-default` / `write-default` using the existing
  single-value serialization (all primitive types).
- **`Schema::Validate`**: version-gates the `initial-default` to format v3
  (`kMinFormatVersionDefaultValues`) — it reinterprets how existing data files are read,
  so it requires the v3 reader contract. The `write-default` only affects values written
  going forward and is **not** version-gated (matching Java's `Schema.checkCompatibility`,
  which gates only the initial default). Both defaults are otherwise validated to be
  non-null primitive literals matching the field type.
- **Generic projection**: a column missing from a data file with an `initial-default`
  maps to `FieldProjection::Kind::kDefault` carrying the literal (the per-format readers
  consume this in the follow-up PRs).

## Follow-ups (stacked on this PR)

- read path — Parquet (`literal_util` + parquet projection/materialization)
- read path — Avro
- schema evolution (`UpdateSchema` add/update column defaults)

## Testing

Added tests


src/iceberg/schema_field.h
```diff
@@ -100,6 +142,11 @@ class ICEBERG_EXPORT SchemaField : public iceberg::util::Formattable {
   std::shared_ptr<Type> type_;
   bool optional_;
   std::string doc_;
+  // Default values are owned by this field and never mutated after being set; copies
+  // of the field share the same payload (reference-counted) instead of deep-copying,
+  // like `type_` above. Sharing is unobservable because the payload is immutable.
+  std::shared_ptr<const Literal> initial_default_;
+  std::shared_ptr<const Literal> write_default_;
```
WZhuo, line 119: `ReassignField` constructs a new `SchemaField` via the 5-argument constructor which initializes `initial_default_` and `write_default_` to `nullptr`. When schema IDs are reassigned (e.g., copying a schema with fresh IDs via the `Schema(get_id)` path), all default values on fields are silently lost. We should copy all field properties including `initialDefault` and `writeDefault`.

src/iceberg/json_serde.cc
```diff
@@ -552,9 +562,23 @@ Result<std::unique_ptr<SchemaField>> FieldFromJson(const nlohmann::json& json) {
...
+  if (initial_default_json.has_value()) {
+    ICEBERG_ASSIGN_OR_RAISE(Literal literal,
+                            LiteralFromJson(*initial_default_json, field.type().get()));
+    field = field.WithInitialDefault(std::move(literal));
+  }
+  if (write_default_json.has_value()) {
+    ICEBERG_ASSIGN_OR_RAISE(Literal literal,
+                            LiteralFromJson(*write_default_json, field.type().get()));
+    field = field.WithWriteDefault(std::move(literal));
+  }
```
WZhuo, line 694: The deserialization first constructs a bare `SchemaField`, then conditionally calls `WithInitialDefault`/`WithWriteDefault`, each of which copies the entire field (including the `shared_ptr<Type>`). This is an unnecessary intermediate copy.

src/iceberg/schema_field.cc
```diff
@@ -55,13 +58,86 @@ bool SchemaField::optional() const { return optional_; }
...
+    return std::nullopt;
+  }
+  return std::cref(*write_default_);
+}
+
+SchemaField SchemaField::WithInitialDefault(Literal initial_default) const {
+  SchemaField copy = *this;
+  copy.initial_default_ = std::make_shared<const Literal>(std::move(initial_default));
+  return copy;
+}
```
WZhuo: I don't think it's need to copy the whole `SchemaField`, can we just set the `initial_default_` field and return *this.
Also the following `With` methods.

src/iceberg/schema_field.h
huan233usc: Good catch, confirmed. Defaults are now constructor args, and `ReassignField` passes the source field's `initial_default_ptr()`/`write_default_ptr()` through, so they're shared with the reassigned field, not lost. Added `ReassignIdsPreservesDefaultValues`.

src/iceberg/json_serde.cc
huan233usc: Agreed — `FieldFromJson` now parses the defaults first and builds the field in one construction. Intermediate copy gone.

src/iceberg/schema_field.cc
huan233usc: Agreed — moved defaults into the constructor and removed both `With...` methods, so construction no longer copies the field.

WZhuo (approved):
LGTM

src/iceberg/schema_field.cc
```diff
@@ -55,13 +62,64 @@ bool SchemaField::optional() const { return optional_; }
...
+Status ValidateDefault(const SchemaField& field, const Literal& value,
+                       std::string_view kind) {
+  if (value.IsNull() || value.IsAboveMax() || value.IsBelowMin()) {
+    return InvalidSchema("Invalid {} value for {}: must be a non-null value", kind,
+                         field.name());
+  }
+  if (field.type() == nullptr || !field.type()->is_primitive()) {
+    return InvalidSchema("Invalid {} value for {}: {} (must be null)", kind, field.name(),
+                         value);
+  }
```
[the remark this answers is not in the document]
huan233usc: Fixed — the non-primitive case now reports "default values are only supported for primitive types" instead of the misleading "(must be null)".

src/iceberg/schema.cc
```diff
@@ -447,7 +453,21 @@ Status Schema::Validate(int32_t format_version) const {
...
+        format_version < TableMetadata::kMinFormatVersionDefaultValues) {
+      return InvalidSchema(
+          "Invalid initial default for {}: non-null default ({}) is not supported "
+          "until v{}",
+          field.name(), field.initial_default()->get(),
+          TableMetadata::kMinFormatVersionDefaultValues);
+    }
+    if (field.initial_default().has_value() || field.write_default().has_value()) {
+      ICEBERG_RETURN_UNEXPECTED(field.Validate());
+    }
```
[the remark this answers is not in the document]
huan233usc: The narrower gating is intentional: only `initial-default` is version-gated (it reinterprets how existing data is read), while `write-default` is allowed below v3 — matching Java's `Schema.checkCompatibility`, which gates only the initial default. Updated the PR description to state this explicitly.

src/iceberg/test/resources/TableMetadataV3Valid.json
```diff
@@ -0,0 +1,123 @@
+{
```
zhjwpku: This metadata file seems not being used anywhere.

src/iceberg/json_serde.cc
```diff
@@ -307,6 +300,15 @@ nlohmann::json ToJson(const SchemaField& field) {
   if (!field.doc().empty()) {
     json[kDoc] = field.doc();
   }
+  // Defaults are validated to be primitive literals matching the field type, so
```
zhjwpku: Should we add a round-trip serde test for default values?

huan233usc: Added default-value serde round-trip tests in `json_serde_test.cc` (9690529), covering all primitive types.

src/iceberg/test/resources/TableMetadataV3Valid.json
huan233usc: Good catch — it was unused. Removed in 511d9c8.

zhjwpku (approved):

huan233usc:
Thank you @zhjwpku and @WZhuo for review!
Hi @wgtmac, do you mind take another look as you requested copilot to review on your behalf. Thanks

src/iceberg/json_serde.cc
```diff
@@ -561,9 +571,27 @@ Result<std::unique_ptr<SchemaField>> FieldFromJson(const nlohmann::json& json) {
...
   ICEBERG_ASSIGN_OR_RAISE(auto doc, GetJsonValueOrDefault<std::string>(json, kDoc));
+  ICEBERG_ASSIGN_OR_RAISE(std::optional<nlohmann::json> initial_default_json,
+                          GetJsonValueOptional<nlohmann::json>(json, kInitialDefault));
+  ICEBERG_ASSIGN_OR_RAISE(std::optional<nlohmann::json> write_default_json,
+                          GetJsonValueOptional<nlohmann::json>(json, kWriteDefault));
+
+  std::shared_ptr<const Literal> initial_default;
+  if (initial_default_json.has_value()) {
+    ICEBERG_ASSIGN_OR_RAISE(Literal literal,
+                            LiteralFromJson(*initial_default_json, type.get()));
```
wgtmac, line 684: Default-value parsing should enforce the JSON single-value rule for `timestamptz` and `timestamptz_ns`. The shared timestamp parser accepts offsets such as `+05:00`, but the spec and Java `SingleValueParser` only accept UTC (`+00:00`) for these default values. As written, C++ can accept schema metadata that Java rejects and then silently normalize it on write.

src/iceberg/schema_field.cc
```diff
@@ -55,13 +62,65 @@ bool SchemaField::optional() const { return optional_; }
...
+
+Status ValidateDefault(const SchemaField& field, const Literal& value,
+                       std::string_view kind) {
+  if (value.IsNull() || value.IsAboveMax() || value.IsBelowMin()) {
+    return InvalidSchema("Invalid {} value for {}: must be a non-null value", kind,
+                         field.name());
+  }
+  if (field.type() == nullptr || !field.type()->is_primitive()) {
+    return InvalidSchema(
+        "Invalid {} value for {}: default values are only supported for primitive types",
```
wgtmac, line 118: This rejects all non-primitive defaults. The spec allows JSON single-value defaults for structs/lists/maps, and a struct field may have a non-null empty struct default (`{}`) while sub-field defaults live in field metadata. If C++ is intentionally matching the current Java model for this first patch, please call out that this remains a spec gap; otherwise this will reject legal v3 schema metadata.

src/iceberg/schema_field.cc
```diff
@@ -55,13 +62,65 @@ bool SchemaField::optional() const { return optional_; }
...
+        kind, field.name());
+  }
+  if (*value.type() != *field.type()) {
```
wgtmac, line 123: Java casts the provided default literal to the field type in `Types.NestedField` (`Literal.of(34)` can become a long/date default, string literals can become date/time/timestamp/uuid, etc.). This check requires exact type equality, so schemas built through the C++ API can be rejected even when Java accepts and normalizes the same default.

src/iceberg/test/json_serde_test.cc
```diff
@@ -86,6 +95,66 @@ void TestJsonConversion(const T& obj, const nlohmann::json& expected_json) {
...
+  cases.emplace_back(int64(), Literal::Long(1234567890123LL));
+  cases.emplace_back(float32(), Literal::Float(1.5f));
+  cases.emplace_back(float64(), Literal::Double(2.5));
+  cases.emplace_back(date(), Literal::Date(19738));
+  cases.emplace_back(timestamp(), Literal::Timestamp(1719446400000000LL));
+  cases.emplace_back(string(), Literal::String("hello"));
+  cases.emplace_back(decimal(9, 2), Literal::Decimal(12345, 9, 2));
+  cases.emplace_back(fixed(3), Literal::Fixed({0x01, 0x02, 0x03}));
+  cases.emplace_back(binary(), Literal::Binary({0xDE, 0xAD, 0xBE, 0xEF}));
+  cases.emplace_back(uuid(), Literal::UUID(uuid_value));
```
wgtmac, line 161: This round-trip list should include `time`, `timestamptz`, `timestamp_ns`, and `timestamptz_ns`, plus a negative case for non-UTC timestamptz defaults. Those are the encodings most likely to drift from Java/spec behavior; the current cases do not cover the offset rule above.

wgtmac (commented):
Static review notes; I did not build or run tests.

src/iceberg/json_serde.cc
huan233usc: Fixed in 180a9f9 — `FieldFromJson` now rejects non-UTC offsets for `timestamptz`/`timestamptz_ns` default values (new `TemporalUtils::IsUtcOffset`, which reuses the existing timezone-suffix parser). This follows Java iiuc

src/iceberg/schema_field.cc
huan233usc: Intentional for this patch — it matches Java's current primitive-only model. Added a comment in `ValidateDefault` marking struct/list/map single-value defaults as a known spec gap / follow-up (180a9f9).

src/iceberg/schema_field.cc
huan233usc: Fixed in 180a9f9 — `ValidateDefault` now casts the default to the field type via `Literal::CastTo` (e.g. int -> long) instead of requiring exact type equality, matching Java's `Types.NestedField`. Only uncastable or out-of-range defaults are rejected.

src/iceberg/test/json_serde_test.cc
huan233usc: Done in 180a9f9 — added `time`/`timestamptz`/`timestamp_ns`/`timestamptz_ns` to the round-trip cases, plus a negative test (`SchemaFieldRejectsNonUtcTimestamptzDefault`) for non-UTC `timestamptz` defaults.

wgtmac (commented):
Two remaining static review notes; I did not build or run tests.

src/iceberg/schema_field.cc
```diff
@@ -55,13 +62,76 @@ bool SchemaField::optional() const { return optional_; }
...
+        "Invalid {} value for {}: default values are only supported for primitive types",
+        kind, field.name());
+  }
+  // Match Java (Types.NestedField), which casts the default literal to the field type
+  // instead of requiring an exact type match (e.g. an int default on a long field, or
+  // a string default on a date/timestamp/uuid field). Reject only defaults that cannot
+  // be cast to the field type or fall outside its range (CastTo signals out-of-range as
+  // an above-max/below-min sentinel).
+  auto field_type = std::static_pointer_cast<PrimitiveType>(field.type());
+  auto cast = value.CastTo(field_type);
```
wgtmac: This validates the cast but keeps storing the original literal. Java stores the result of `castDefault`, so a programmatic default like `Literal::Long(...)` on a timestamp field would validate here, then later serialize or project as a long instead of a timestamp default. Either the field needs to retain the casted literal, or this should keep requiring an exact type match.

src/iceberg/schema_field.cc
```diff
@@ -55,13 +62,76 @@ bool SchemaField::optional() const { return optional_; }
...
+
+const std::shared_ptr<const Literal>& SchemaField::write_default_ptr() const {
+  return write_default_;
+}
+
+namespace {
+
+Status ValidateDefault(const SchemaField& field, const Literal& value,
+                       std::string_view kind) {
+  if (value.IsNull() || value.IsAboveMax() || value.IsBelowMin()) {
```
wgtmac: This rejects explicit null defaults. The spec allows optional field defaults to be null and says they may be explicitly set. Java mostly models a null default as absence, but C++ now has a present-null `Literal` path that will fail validation and, for `initial-default`, also trips the format-version gate by presence rather than by non-null value.

src/iceberg/schema_field.h
```diff
@@ -71,6 +79,32 @@ class ICEBERG_EXPORT SchemaField : public iceberg::util::Formattable {
   /// \brief Get the field documentation.
   std::string_view doc() const;
 
+  /// \brief Get the default value for this field used when reading rows written
+  /// before the field existed (v3 `initial-default`). Empty if absent.
+  ///
+  /// The returned reference is a non-owning view into a value owned by this field;
+  /// it remains valid for the lifetime of this SchemaField.
+  [[nodiscard]] std::optional<std::reference_wrapper<const Literal>> initial_default()
```
wgtmac: I think we should remove `[[nodiscard]]` because users should always know what they are doing when these are called. I think we need also clean up existing functions as well.

src/iceberg/schema_field.h
```diff
@@ -71,6 +79,32 @@ class ICEBERG_EXPORT SchemaField : public iceberg::util::Formattable {
...
+  ///
+  /// The returned reference is a non-owning view into a value owned by this field;
+  /// it remains valid for the lifetime of this SchemaField.
+  [[nodiscard]] std::optional<std::reference_wrapper<const Literal>> write_default()
+      const;
+
+  /// \brief Get the shared owning pointer to the `initial-default` value, or null if
+  /// absent. Prefer initial_default() for reading; this exists so a rebuilt field can
+  /// share the (immutable) value rather than copy it.
+  [[nodiscard]] const std::shared_ptr<const Literal>& initial_default_ptr() const;
```
wgtmac: It looks odd to provide both `initial_default` and `initial_default_ptr`. Should we keep one, like `const std::shared_ptr<const Literal>& initial_default() const`?

src/iceberg/schema_field.h
```diff
@@ -100,6 +134,11 @@ class ICEBERG_EXPORT SchemaField : public iceberg::util::Formattable {
   std::shared_ptr<Type> type_;
   bool optional_;
   std::string doc_;
+  // Default values are owned by this field and never mutated after being set; copies
```
wgtmac: Not a strong opinion but the comment looks too verbose.

src/iceberg/util/temporal_util.cc
```diff
@@ -444,6 +444,11 @@ Result<int64_t> TemporalUtils::ParseTimestampNsWithZone(std::string_view str) {
                               /*units_per_micro=*/internal::kNanosPerMicro);
 }
 
+Result<bool> TemporalUtils::IsUtcOffset(std::string_view str) {
+  ICEBERG_ASSIGN_OR_RAISE(auto timestamp_with_offset, ParseTimestampWithZoneSuffix(str));
```
wgtmac, line 448: This allows `Z` and `-00:00` as well. This is not allowed by the spec, right? However, Java impl seems doing the same thing. Perhaps we need to document this behavior explicitly?

src/iceberg/schema_field.cc
```diff
@@ -21,19 +21,26 @@
...
 #include <format>
 #include <string_view>
+#include <utility>
 
+#include "iceberg/expression/literal.h"
 #include "iceberg/type.h"
 #include "iceberg/util/formatter.h"  // IWYU pragma: keep
+#include "iceberg/util/macros.h"
 
 namespace iceberg {
```
wgtmac, line 32: This always create a new literal even when their types are the same which is unlikely?

wgtmac, line 32: Should we return `Result<std::shared_ptr<const Literal>>` so we don't swallow any unexpected error?

src/iceberg/schema_field.cc
```diff
@@ -55,13 +62,76 @@ bool SchemaField::optional() const { return optional_; }
...
+    return InvalidSchema(
+        "Invalid {} value for {}: default values are only supported for primitive types",
+        kind, field.name());
+  }
+  // Match Java (Types.NestedField), which casts the default literal to the field type
+  // instead of requiring an exact type match (e.g. an int default on a long field, or
+  // a string default on a date/timestamp/uuid field). Reject only defaults that cannot
+  // be cast to the field type or fall outside its range (CastTo signals out-of-range as
+  // an above-max/below-min sentinel).
+  auto field_type = std::static_pointer_cast<PrimitiveType>(field.type());
```
wgtmac: nit: use `internal::checked_pointer_cast`

src/iceberg/schema_field.cc
```diff
@@ -55,13 +62,76 @@ bool SchemaField::optional() const { return optional_; }
...
+  auto cast = value.CastTo(field_type);
+  if (!cast.has_value() || cast->IsAboveMax() || cast->IsBelowMin()) {
```
wgtmac: Perhaps it is better to use `ICEBERG_ASSIGN_OR_RAISE` to respect the original error instead of checking `cast.has_value()`?

src/iceberg/json_serde.cc
```diff
@@ -307,6 +310,15 @@ nlohmann::json ToJson(const SchemaField& field) {
   if (!field.doc().empty()) {
     json[kDoc] = field.doc();
   }
+  // Defaults are validated to be primitive literals matching the field type, so
+  // single-value serialization cannot fail here.
+  if (field.initial_default().has_value()) {
+    ICEBERG_ASSIGN_OR_THROW(json[kInitialDefault],
```
wgtmac: Though it cannot not fail, we shouldn't use `ICEBERG_ASSIGN_OR_THROW` here. Let's change the signature to return `Result<nlohmann::json>` just in case any error.

src/iceberg/json_serde.cc
```diff
@@ -554,16 +565,67 @@ Result<std::unique_ptr<Type>> TypeFromJson(const nlohmann::json& json) {
...
+  if (!value.is_string()) {
+    // Let LiteralFromJson report the type mismatch.
+    return {};
+  }
+  const auto str = value.get<std::string>();
+  ICEBERG_ASSIGN_OR_RAISE(bool is_utc, TemporalUtils::IsUtcOffset(str));
+  if (!is_utc) {
+    return JsonParseError(
+        "Invalid timestamptz default '{}' for {}: default values must use UTC "
+        "(offset 'Z' or '+00:00')",
```
wgtmac: As I have commented earlier, this can also accept `-00:00` so the error message is not consistent.

src/iceberg/json_serde.cc
```diff
@@ -554,16 +565,67 @@ Result<std::unique_ptr<Type>> TypeFromJson(const nlohmann::json& json) {
...
+// silently normalizes to UTC, which would let C++ accept default metadata that Java
+// rejects and then rewrite the offset on serialization. Enforce UTC for these
+// defaults at parse time, where the original offset is still visible.
+Status ValidateTimestamptzDefaultIsUtc(const Type& type, const nlohmann::json& value) {
+  const auto type_id = type.type_id();
+  if (type_id != TypeId::kTimestampTz && type_id != TypeId::kTimestampTzNs) {
+    return {};
+  }
+  if (!value.is_string()) {
+    // Let LiteralFromJson report the type mismatch.
```
wgtmac: Why not directly error out here?

src/iceberg/json_serde.cc
```diff
@@ -554,16 +565,67 @@ Result<std::unique_ptr<Type>> TypeFromJson(const nlohmann::json& json) {
...
+}  // namespace
+
 Result<std::unique_ptr<SchemaField>> FieldFromJson(const nlohmann::json& json) {
   ICEBERG_ASSIGN_OR_RAISE(
       auto type, GetJsonValue<nlohmann::json>(json, kType).and_then(TypeFromJson));
   ICEBERG_ASSIGN_OR_RAISE(auto field_id, GetJsonValue<int32_t>(json, kId));
   ICEBERG_ASSIGN_OR_RAISE(auto name, GetJsonValue<std::string>(json, kName));
   ICEBERG_ASSIGN_OR_RAISE(auto required, GetJsonValue<bool>(json, kRequired));
   ICEBERG_ASSIGN_OR_RAISE(auto doc, GetJsonValueOrDefault<std::string>(json, kDoc));
+  ICEBERG_ASSIGN_OR_RAISE(std::optional<nlohmann::json> initial_default_json,
```
wgtmac: Why not use `auto`?

wgtmac (commented):
I've just finished a human review over non-test files. Let me know what you think. Thanks!

src/iceberg/json_serde.cc
huan233usc: Thanks @wgtmac, this is in general a good call. But the changing the signature to return Result<nlohmann::json> just in case any error requires changing lots of places that is not related on default values. 

I updated the pr but also create a pure refactor in https://github.com/apache/iceberg-cpp/pull/785

wgtmac: Thanks! Merged that PR.

src/iceberg/schema_field.cc
```diff
@@ -72,9 +143,23 @@ std::string SchemaField::ToString() const {
...
 }
 
+namespace {
+
+bool DefaultEquals(const std::shared_ptr<const Literal>& lhs,
+                   const std::shared_ptr<const Literal>& rhs) {
+  if (lhs == nullptr || rhs == nullptr) {
+    return lhs == rhs;
+  }
+  return *lhs == *rhs;
```
wgtmac, line 163: Just a reminder that `Literal::operator<=>` returns unordered when any side `IsNull()` returns true so two null defaults do not equal. I think we should fix `Literal::operator<=>` to be null safe?

wgtmac: Or add a overload `Literal::NullSafeEquals()`

src/iceberg/util/temporal_util.cc
huan233usc: Thanks for raising this — I dug into both the spec and the reference implementation, and I think accepting `Z` / `+00:00` / `-00:00` is correct here.

**What the spec says:** default values use the single-value JSON serialization (Appendix D). For `timestamptz` / `timestamptz_ns` the value must be **UTC**, serialized as an ISO-8601 timestamp with a `+00:00` offset (e.g. `2017-11-16T22:31:08.123456+00:00`). The requirement is that the offset is UTC; `Z` and `-00:00` are just equivalent zero-offset spellings of the same instant.

**What the reference impl does:** `SingleValueParser` gates timestamptz defaults on `DateTimeUtil.isUTCTimestamptz`, which is:

```java
OffsetDateTime odt = OffsetDateTime.parse(s, DateTimeFormatter.ISO_DATE_TIME);
return odt.getOffset().equals(ZoneOffset.UTC);
```

So it checks the parsed offset equals UTC — `Z`, `+00:00` and `-00:00` all pass, and only a genuinely non-zero offset (e.g. `+05:00`) is rejected. That's exactly what `IsUtcOffset` does here. (Java's error string reads "offset must be +00:00", but the underlying check is offset == UTC.)

Refs:
- Spec, Appendix D — Single-value serialization: https://iceberg.apache.org/spec/#appendix-d-single-value-serialization
- `DateTimeUtil.isUTCTimestamptz`: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/util/DateTimeUtil.java

Happy to tighten it to `+00:00`-only if you'd prefer, though that would reject UTC values the Java reference accepts.

src/iceberg/schema_field.cc
huan233usc: Done, remove the normalization method

huan233usc: Done

wgtmac (approved):
Thanks for addressing all the feedback! Now this looks good to me.

One minor issue is that `ValidateDefault` now rejects any default whose literal type is not exactly the field type. That still diverges from Java: `Types.NestedField.castDefault` casts the provided literal to the field type and stores the converted literal. I don't think this is a blocker for now.

PR #747, opened 2026-06-15
fix: Handle unknown record_count better

### LINKED ISSUE #732: StrictMetricsEvaluator treats unknown record_count as rows must match
`StrictMetricsEvaluator::Evaluate` currently returns `ROWS_MUST_MATCH` when `data_file.record_count <= 0`:

https://github.com/apache/iceberg-cpp/blob/c0c6b01393070b0813f49b0d6220c98256379cef/src/iceberg/expression/strict_metrics_evaluator.cc#L534-L537

This is correct for `record_count == 0`, but not for `record_count == -1`. iceberg-cpp uses `-1` as an unknown row-count sentinel when writer metrics do not include `row_count`:

- data writer: https://github.com/apache/iceberg-cpp/blob/c0c6b01393070b0813f49b0d6220c98256379cef/src/iceberg/data/data_writer.cc#L80-L86
- equality delete writer: https://github.com/apache/iceberg-cpp/blob/c0c6b01393070b0813f49b0d6220c98256379cef/src/iceberg/data/equality_delete_writer.cc#L80-L86
- position delete writer: https://github.com/apache/iceberg-cpp/blob/c0c6b01393070b0813f49b0d6220c98256379cef/src/iceberg/data/position_delete_writer.cc#L140-L146

Other code already treats negative record count as unknown/missing:

- inclusive metrics: https://github.com/apache/iceberg-cpp/blob/c0c6b01393070b0813f49b0d6220c98256379cef/src/iceberg/expression/inclusive_metrics_evaluator.cc#L507-L516
- count aggregate: https://github.com/apache/iceberg-cpp/blob/c0c6b01393070b0813f49b0d6220c98256379cef/src/iceberg/expression/aggregate.cc#L379-L388

Impact: a file with unknown row count can be treated as if every row must match, even for predicates like `AlwaysFalse`.

Suggested fix: only special-case `record_count == 0`; let negative/unknown counts fall through to normal strict metrics evaluation.

A focused regression test with `record_count = -1` and `AlwaysFalse` fails before changing `<= 0` to `== 0`, and passes after.

goel-skd:
Fixes #732  Only special-case ```record_count == 0;``` 
Let negative/unknown counts fall through
to normal strict metrics evaluation.

### DIFF
--- a/src/iceberg/expression/strict_metrics_evaluator.cc
+++ b/src/iceberg/expression/strict_metrics_evaluator.cc
@@ -532,9 +532,12 @@ Result<std::unique_ptr<StrictMetricsEvaluator>> StrictMetricsEvaluator::Make(
 }
 
 Result<bool> StrictMetricsEvaluator::Evaluate(const DataFile& data_file) const {
-  if (data_file.record_count <= 0) {
+  if (data_file.record_count == 0) {
     return kRowsMustMatch;
   }
+  // A negative record count is used as a sentinel for an unknown row count (e.g. when
+  // writer metrics do not include a row count). Treat it as unknown and fall through to
+  // normal strict metrics evaluation rather than assuming all rows must match.
   StrictMetricsVisitor visitor(data_file, *schema_);
   return Visit<bool, StrictMetricsVisitor>(expr_, visitor);
 }
--- a/src/iceberg/expression/strict_metrics_evaluator.cc
+++ b/src/iceberg/expression/strict_metrics_evaluator.cc
@@ -535,9 +535,13 @@ Result<bool> StrictMetricsEvaluator::Evaluate(const DataFile& data_file) const {
   if (data_file.record_count == 0) {
     return kRowsMustMatch;
   }
-  // A negative record count is used as a sentinel for an unknown row count (e.g. when
-  // writer metrics do not include a row count). Treat it as unknown and fall through to
-  // normal strict metrics evaluation rather than assuming all rows must match.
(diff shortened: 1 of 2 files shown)


zhjwpku:
You may want to change the PR title to start with `fix:`. It's not strictly required, but we generally follow the `Conventional Commits` convention.

src/iceberg/expression/strict_metrics_evaluator.cc
```diff
@@ -532,9 +532,12 @@ Result<std::unique_ptr<StrictMetricsEvaluator>> StrictMetricsEvaluator::Make(
 }
 
 Result<bool> StrictMetricsEvaluator::Evaluate(const DataFile& data_file) const {
-  if (data_file.record_count <= 0) {
+  if (data_file.record_count == 0) {
     return kRowsMustMatch;
   }
+  // A negative record count is used as a sentinel for an unknown row count (e.g. when
```
wgtmac: Specifically, `-1` is the sentinel value. So is it better to error out on other negative values just in case?

goel-skd: Good point. Updated so only -1 is treated as the unknown sentinel. Anything below -1 is invalid metadata and now returns ```InvalidArgument``` rather than falling through. Also added a test for the invalid case.

wgtmac (approved):

wgtmac:
Thank you, @goel-skd and @zhjwpku!

goel-skd:
> Thank you, @goel-skd and @zhjwpku!

My first commit to this repo..yayy!! :)

Issue #613, opened 2026-04-01
case insensitive field matching behavior different from iceberg-python and iceberg-java


nvartolomei:
https://github.com/apache/iceberg-cpp/pull/180 added support for `GetFieldByName` with `case_sensitive` argument. Its behavior is different from Python and Java Iceberg implementations which could result in inconsistent behavior when used together.

Python/Java are unicode/multibyte aware. iceberg-cpp is not. An example where this breaks e.g. column names containing e.g. "Latin Capital Letter I with Dot Above" https://www.compart.com/en/unicode/U+0130.

Does not affect me directly at this moment.

wgtmac:
Thanks for reporting this issue. I agree that currently case sensitivity only supports ASCII characters. We might need to introduce https://github.com/unicode-org/icu to address this due to lack support of the C++ standard library.

nvartolomei:
Using iceberg-cpp with glue rest catalog could produce incorrect results. Python script which tests Glue behavior [test_column_names.py](https://github.com/user-attachments/files/26434447/test_column_names.py)

```
============================================================
TEST: mixed_case
  table: glue_column_test.col_test_mixed_case_04236e66
  requested columns: ['userId', 'FirstName', 'last_NAME', 'isActive', 'XMLPayload']
  returned columns:  ['userid', 'firstname', 'last_name', 'isactive', 'xmlpayload']
  RESULT: MISMATCH
                          'userId' -> 'userid' <-- CHANGED
                       'FirstName' -> 'firstname' <-- CHANGED
                       'last_NAME' -> 'last_name' <-- CHANGED
                        'isActive' -> 'isactive' <-- CHANGED
                      'XMLPayload' -> 'xmlpayload' <-- CHANGED

============================================================
TEST: non_ascii
  table: glue_column_test.col_test_non_ascii_04236e66
  requested columns: ['naïve_count', 'café_id', '日本語カラム', 'Ціна', 'straße', 'İstanbul_İd', 'GROẞE_id']
  returned columns:  ['naïve_count', 'café_id', '日本語カラム', 'ціна', 'straße', 'i̇stanbul_i̇d', 'große_id']
  RESULT: MISMATCH
                     'naïve_count' -> 'naïve_count'
                         'café_id' -> 'café_id'
                          '日本語カラム' -> '日本語カラム'
                            'Ціна' -> 'ціна' <-- CHANGED
                          'straße' -> 'straße'
[...]

PR #748, opened 2026-06-16
fix: cast to unsigned char in ASCII case conversion (1/2)

goel-skd:
fix: cast to unsigned char in ASCII case conversion 

`std::tolower`/`toupper` are UB when passed a negative `char`, which is what you get for any non-ASCII byte in a signed `char`. `StringUtils::ToLower`, `ToUpper`, and `EqualsIgnoreCase`
  all did this. Cast through `unsigned char`.

  Behavior for ASCII is unchanged; this doesn't add Unicode case folding, just fixes the UB.

  Related: #613

### DIFF
--- a/src/iceberg/util/string_util.h
+++ b/src/iceberg/util/string_util.h
@@ -41,8 +41,8 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
 
 class ICEBERG_EXPORT StringUtils {
  public:
-  // NOTE: These only convert ASCII letters (under the default C locale); non-ASCII
-  // (multibyte UTF-8) bytes are left unchanged.
+  // NOTE: These convert ASCII letters only; all other bytes, including non-ASCII
+  // (multibyte UTF-8) bytes, are passed through unchanged.
   // See https://github.com/apache/iceberg-cpp/issues/613.
   static std::string ToLower(std::string_view str) {
     return str | std::ranges::views::transform(ToLowerAscii) |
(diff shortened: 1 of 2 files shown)


src/iceberg/util/string_util.h
```diff
@@ -128,6 +132,18 @@ class ICEBERG_EXPORT StringUtils {
     }
     return value;
   }
+
+ private:
+  // std::tolower/std::toupper require their argument to be representable as
+  // unsigned char (or EOF); passing a raw char with a non-ASCII (negative) value is
+  // undefined behavior, so cast through unsigned char first.
```
zhjwpku: +1, Quoting cppreference:

Like all other functions from [\<cctype\>](https://en.cppreference.com/cpp/header/cctype), the behavior of std::tolower is undefined if the argument's value is neither representable as unsigned char nor equal to EOF. To use these functions safely with plain chars (or signed chars), the argument should first be converted to unsigned char:

```
char my_tolower(char ch)
{
    return static_cast<char>(std::tolower(static_cast<unsigned char>(ch)));
}
```

zhjwpku (approved):

goel-skd:
@zhjwpku @wgtmac - Thanks for the review. The locale dependency was in the original code too. This PR doesn't introduce or worsen it. Fixing UB is the only intention with this PR.

But I will include copilot's suggestions to tighten up things.

wgtmac (approved):

wgtmac:
Thank you, @goel-skd and @zhjwpku!

PR #749, opened 2026-06-16
feat(hive): vendor Hive Metastore IDL and generated bindings

MisterRaindrop:
## Summary

- Update vendored Hive Metastore IDL to Hive 4.0.1.
- Check in generated HMS/fb303 C++ bindings.
- Add a regeneration script plus lint/license exclusions for generated sources.


zhjwpku:
Do we need to check in the generated code? It might be cleaner to have the build system generate these files on the fly instead, since the IDL is the source of truth and the generated code is just a derived artifact.

wgtmac:
> Do we need to check in the generated code? It might be cleaner to have the build system generate these files on the fly instead, since the IDL is the source of truth and the generated code is just a derived artifact.

I don't think so. These files are not changed constantly so it wastes resource and complicates the ci workflow if we built them dynamically. Arrow C++ does the same thing for parquet thrift source code.

zhjwpku:
@MisterRaindrop Just curious, how long does it take to generate these files?

MisterRaindrop:
> @MisterRaindrop Just curious, how long does it take to generate these files?

The gen cpp speed is very fast, around one second.

zhjwpku (approved):
LGTM

wgtmac (approved):

wgtmac:
Thanks @MisterRaindrop for adding this and @zhjwpku for the review!

PR #750, opened 2026-06-16
feat(rest): add session-aware REST catalog

wgtmac:
Add SessionContext and SessionCatalog APIs, make RestCatalog a session-aware root, and bind REST catalog operations through default, contextual, and table-scoped catalog views.

Route contextual and table auth through AuthManager, preserve table response config for scoped catalogs, and fail closed on unsupported table FileIO config.

### DIFF
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -21,6 +21,7 @@ set(ICEBERG_SOURCES
     arrow_c_data_util.cc
     arrow_c_data_guard_internal.cc
     catalog/memory/in_memory_catalog.cc
+    catalog/session_context.cc
     delete_file_index.cc
     expression/aggregate.cc
     expression/binder.cc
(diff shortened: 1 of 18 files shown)


wgtmac:
You might want to take a look as you've contributed to rest catalog recently. @plusplusjiajia @lishuxu

plusplusjiajia:
Thanks for the ping @wgtmac. Read through it — the session-aware split (a root with context/table-scoped views) reads well, and it's a cleaner home for table-scoped FileIO than what I did in [#719](https://github.com/apache/iceberg-cpp/pull/719).

#719 already covers the TODO(gangwu) here: parsing storage-credentials into LoadTableResult (separate files, no conflict) + building the per-table S3 FileIO. I'd land the parsing independently, then rebase the binding onto this as the TableFileIO impl once it merges.

One design question: ValidateNoFileIOConfig fails closed on any non-auth key, but LoadTableResponse.config can carry non-auth, non-FileIO settings too — reject only recognized FileIO/storage keys and pass the rest through?

wgtmac:
Thanks @plusplusjiajia for the quick review!

> ValidateNoFileIOConfig fails closed on any non-auth key, but LoadTableResponse.config can carry non-auth, non-FileIO settings too — reject only recognized FileIO/storage keys and pass the rest through?

This is just a temporary workaround and should be removed together with table-level FileIO support.

blackmwk (approved):

PR #751, opened 2026-06-16
feat: implement RewriteFiles operation

WZhuo:
This PR implements the RewriteFiles operation for the Iceberg C++ library.

## Changes

- Added RewriteFiles update class in src/iceberg/update/rewrite_files.h and .cc
- Added RewriteFiles to the type forward declarations in src/iceberg/type_fwd.h
- Registered RewiteFiles as a merging snapshot update in src/iceberg/table.h and .cc
- Added RewriteFiles support in transaction (src/iceberg/transaction.h and .cc)
- Added comprehensive tests in src/iceberg/test/rewrite_files_test.cc
- Added merge test for RewriteFiles in src/iceberg/test/merging_snapshot_update_test.cc


src/iceberg/table.cc
```diff
@@ -217,6 +218,12 @@ Result<std::shared_ptr<FastAppend>> Table::NewFastAppend() {
   return FastAppend::Make(name().name, std::move(ctx));
 }
 
+Result<std::shared_ptr<RewriteFiles>> Table::NewRewriteFiles() {
```
wgtmac, line 249: StaticTable inherits this method too. That lets a read-only table create a rewrite update, then fail later when Commit reaches the null catalog. Java rejects newRewrite() in BaseReadOnlyTable; StaticTable should override this like the other mutating APIs and return NotSupported.

src/iceberg/update/rewrite_files.cc
```diff
@@ -0,0 +1,176 @@
...
+  ICEBERG_PRECHECK(!table_name.empty(), "Table name cannot be empty");
+  ICEBERG_PRECHECK(ctx != nullptr, "Cannot create RewriteFiles without a context");
+  return std::unique_ptr<RewriteFiles>(
+      new RewriteFiles(std::move(table_name), std::move(ctx)));
+}
+
+RewriteFiles& RewriteFiles::DeleteDataFile(const std::shared_ptr<DataFile>& data_file) {
+  ICEBERG_BUILDER_RETURN_IF_ERROR(MergingSnapshotUpdate::DeleteDataFile(data_file));
+  // Track replaced data files for conflict detection
+  replaced_data_files_.insert(data_file);
```
wgtmac: This keeps the caller-owned DataFile as the conflict identity. DataFile is mutable and DataFileSet indexes by file_path, so a caller can change or reuse the object after staging and make the rewrite validate the wrong path. Freeze the deleted-file identity here, for example by storing a copy or an immutable path/key, like the add path stages a copy.

src/iceberg/test/rewrite_files_test.cc
```diff
@@ -0,0 +1,420 @@
...
+    EXPECT_THAT(result, HasErrorMessage("Missing required files to delete"));
+  }
+}
+
+// ============================================================================
+// TODO(WZhuo): Tests blocked on missing infrastructure in iceberg-cpp.
+// ============================================================================
+//
+// TODO(RewriteDataAndDeleteFiles):
+//   Blocked by: RowDelta not yet ported.
```
zhjwpku: Just FYI, RowDelta has been merged.

src/iceberg/update/rewrite_files.h
```diff
@@ -0,0 +1,185 @@
...
+/// Note that the new state of the table after each rewrite must be logically
+/// equivalent to the original table state.
+class ICEBERG_EXPORT RewriteFiles : public MergingSnapshotUpdate {
+ public:
+  /// \brief Create a new RewriteFiles operation.
+  ///
+  /// \param table_name The name of the table
+  /// \param ctx The transaction context
+  /// \return A unique pointer to the new RewriteFiles operation
+  static Result<std::unique_ptr<RewriteFiles>> Make(
```
zhjwpku, line 58: I noticed that some update operations return a std::unique_ptr (e.g., RowDelta), while others return a std::shared_ptr (e.g., OverwriteFiles). I'm not sure what the standard is here. Should we make them consistent? cc @wgtmac

src/iceberg/test/rewrite_files_test.cc
```diff
@@ -0,0 +1,420 @@
...
+#include "iceberg/row/partition_values.h"
+#include "iceberg/snapshot.h"
+#include "iceberg/table.h"
+#include "iceberg/test/matchers.h"
+#include "iceberg/test/update_test_base.h"
+#include "iceberg/update/fast_append.h"
+
+namespace iceberg {
+
+class RewriteFilesTest : public MinimalUpdateTestBase {
```
zhjwpku, line 51: Should we consider the table format version in these tests? I handled it in MergeAppendTestBase, but the resulting code structure feels a bit awkward. I'm wondering if we should refactor UpdateTestBase to make it easier to write tests for different table format versions.

src/iceberg/test/rewrite_files_test.cc
```diff
@@ -0,0 +1,420 @@
...
+//   Verifies manifest entry statuses (ADDED/DELETED/EXISTING), snapshot IDs, file
+//   identity.
+//
+// TODO(RewriteDataAndAssignOldSequenceNumber):
+//   Blocked by: RowDelta not yet ported.
+//   Creates delete files via RowDelta, rewrites data files with old sequence number.
+//   Verifies sequence number propagation to new manifests.
+//
+// TODO(Failure):
+//   Blocked by: commit fail injection (failCommits) not yet ported.
```
zhjwpku: I think this is already supported. See `BindTableWithFailingCommits` in `merge_append_test.cc`. The implementation is a bit ugly though, feel free to improve it.

src/iceberg/update/rewrite_files.h
wgtmac: We should return unique_ptr here and in similar places.

src/iceberg/table.cc
WZhuo: Good catch. `StaticTable::NewRewriteFiles()` already overrides this and returns `NotSupported("Cannot create a rewrite files for a static table")` — matching the pattern used by `NewOverwrite()`, `NewMergeAppend()`, etc.

src/iceberg/test/rewrite_files_test.cc
WZhuo: Good suggestion. I parameterized all tests using a `RewriteFilesFormatVersionTest` fixture (with `::testing::WithParamInterface<int8_t>`) that runs each test across v1 and v2. Tests matching Java's `assumeThat(formatVersion >= 2)` guards use `AssumesFormatVersionAtLeast(2)` which calls `GTEST_SKIP()` on v1. The structure mirrors your `MergeAppendTestBase` approach without requiring a refactor of `UpdateTestBase` — the format version is set in `SetUp()` before each test.

src/iceberg/test/rewrite_files_test.cc
WZhuo: You're right, thanks for the pointer. I adopted the `BindTableWithFailingCommits` pattern from `merge_append_test.cc` and used it to implement `Failure`, `Recovery`, and the data+delete variants. The tests inject commit failures via `MockCatalog` and verify retry behavior — failures exhaust retries return `CommitFailed`, transient failures within the budget eventually succeed.

src/iceberg/test/rewrite_files_test.cc
WZhuo: Thanks for the heads up! I rebased on top of the merged RowDelta and used it to implement `RewriteDataAndDeleteFiles`, `ReplaceEqualityDeletesWithPositionDeletes`, `RemoveAllDeletes`, `NewDeleteFile`, and the data+delete failure/recovery tests. The TODOs that were blocked on RowDelta are all resolved now.

wgtmac (approved):

wgtmac:
Thanks @WZhuo for adding this and @zhjwpku for the review!

PR #753, opened 2026-06-17
feat(hive): add iceberg_hive library with bundled/system thrift dependency

MisterRaindrop:
## Summary
- Add `iceberg_hive` library wiring for the checked-in HMS/fb303 bindings.
- Add stub `HiveCatalog` and HMS connection properties.
- Link against Arrow's bundled Thrift runtime.


cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -742,3 +742,14 @@ endif()
 if(ICEBERG_BUILD_SQL_CATALOG)
   resolve_sql_catalog_dependencies()
 endif()
+
+# Arrow's bundled build creates the Thrift C++ runtime as a `thrift` target
```
wgtmac: I'd recommend follow `resolve_aws_sdk_dependency` and `ICEBERG_AWSSDK_BUNDLED` to either bundle or use system thrift.

wgtmac (approved):
I've verified both vendored and system thrift locally and pushed some minor fixes. It seems that we don't have any ci workflow to cover hive catalog yet. But I'm fine to add them later in a separate PR just like what SQL catalog or AWS S3/SigV4 do. Thanks @MisterRaindrop for adding this!

PR #754, opened 2026-06-18
feat: add uniqueValue optimization for notEq/notIn in InclusiveMetricsEvaluator

WZhuo:
When a column has a single unique value (no nulls, no NaNs, and lower bound == upper bound), notEq(X) and notIn({...X...}) can prune files where that unique value matches the predicate, returning ROWS_CANNOT_MATCH. Previously these always returned ROWS_MIGHT_MATCH, never pruning.

This mirrors the upstream Java InclusiveMetricsEvaluator.uniqueValue() logic. The StrictMetricsEvaluator is unchanged: its existing bounds-based notEq/notIn logic already covers the single-value case, matching upstream Java which has no uniqueValue helper there.

Ports the corresponding testNotEqWithSingleValue/testNotInWithSingleValue cases from the Java test suite.

### DIFF
--- a/src/iceberg/expression/inclusive_metrics_evaluator.cc
+++ b/src/iceberg/expression/inclusive_metrics_evaluator.cc
@@ -450,7 +450,7 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
       return std::nullopt;
     }
 
-    if (!(lower.value() == upper.value())) {
+    if (lower.value() != upper.value()) {
       return std::nullopt;
     }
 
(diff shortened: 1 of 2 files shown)


src/iceberg/expression/inclusive_metrics_evaluator.cc
```diff
@@ -416,6 +429,34 @@ class InclusiveMetricsVisitor : public BoundVisitor<bool> {
...
+        lower->IsNaN() || upper->IsNaN()) {
+      return std::nullopt;
+    }
+
+    auto nan_it = data_file_.nan_value_counts.find(id);
+    if (nan_it != data_file_.nan_value_counts.cend() && nan_it->second != 0) {
+      return std::nullopt;
+    }
+
+    if (!(lower.value() == upper.value())) {
```
zhjwpku: nit: can we simplify this?
```suggestion
    if (lower.value() != upper.value()) {
```

zhjwpku (approved):
I checked the java's impl, LGTM

wgtmac (approved):

PR #755, opened 2026-06-18
fix: avoid signed overflow in integer truncation

fallintoplace:
## Summary

- evaluate integer truncation arithmetic in a wider signed type to avoid signed overflow
- preserve modulo conversion when the mathematical result falls below the source type's signed range
- add regression coverage for large positive widths and `int32_t` / `int64_t` minimum values

## Testing

- `cmake -S . -B build-truncate -G Ninja -DICEBERG_BUILD_BUNDLE=OFF -DICEBERG_BUILD_REST=OFF -DICEBERG_BUILD_SHARED=OFF -DICEBERG_ENABLE_UBSAN=ON`
- `UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1 ctest --test-dir build-truncate -R util_test --output-on-failure`

### DIFF
--- a/src/iceberg/util/truncate_util.h
+++ b/src/iceberg/util/truncate_util.h
@@ -22,11 +22,13 @@
 #include <cstdint>
 #include <optional>
 #include <string>
+#include <type_traits>
 #include <utility>
 
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
 #include "iceberg/type_fwd.h"
+#include "iceberg/util/int128.h"
 
 namespace iceberg {
 
@@ -84,7 +86,15 @@ class ICEBERG_EXPORT TruncateUtils {
   template <typename T>
     requires std::is_same_v<T, int32_t> || std::is_same_v<T, int64_t>
   static inline T TruncateInteger(T v, int32_t W) {
-    return v - (((v % W) + W) % W);
+    using WideInt = std::conditional_t<std::is_same_v<T, int32_t>, int64_t, int128_t>;
+    using UnsignedT = std::make_unsigned_t<T>;
+
+    const auto value = static_cast<WideInt>(v);
+    const auto width = static_cast<WideInt>(W);
+    const auto remainder = ((value % width) + width) % width;
+    const auto truncated = value - remainder;
+    // Preserve modulo conversion when the mathematical result is below the signed range.
+    return static_cast<T>(static_cast<UnsignedT>(truncated));
   }
 
   /// \brief Truncate a Decimal to a specified width.
--- a/src/iceberg/test/truncate_util_test.cc
+++ b/src/iceberg/test/truncate_util_test.cc
@@ -19,6 +19,8 @@
 
 #include "iceberg/util/truncate_util.h"
 
+#include <limits>
+
 #include <gtest/gtest.h>
 
 #include "iceberg/expression/literal.h"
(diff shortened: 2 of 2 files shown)


src/iceberg/util/truncate_util.h
```diff
@@ -84,7 +86,15 @@ class ICEBERG_EXPORT TruncateUtils {
   template <typename T>
     requires std::is_same_v<T, int32_t> || std::is_same_v<T, int64_t>
   static inline T TruncateInteger(T v, int32_t W) {
-    return v - (((v % W) + W) % W);
```
zhjwpku, line 87: This was from the spec:

```
The remainder, v % W, must be positive. For languages where % can produce negative values, the correct truncate function is: v - (((v % W) + W) % W)
```

I don't think we need to use int128_t since W is alway a int32_t and greater than 0.

fallintoplace: Yep, the formula is from the spec. The issue is not the spec math, but evaluating it with signed C++ integer types. For int32_t, (v % W) + W can overflow when W is large. For int64_t, the final v - remainder can overflow at INT64_MIN. The wider type keeps the spec formula while avoiding signed UB.

wgtmac: This seems deviate from the Java implementation with some very large values. Perhaps it is worth raising a discussion in the dev@iceberg.apache.org?

zhjwpku: I tried this on Compiler Explorer. For `W > INT32_MAX / 2`, `(v % W) + W` can overflow, causing the remainder to become negative, which violates the spec that the remainder must be positive.

For values close to `std::numeric_limits<int64_t>::min()`, our current implementation produces the same result as the modified version, but it triggers

`runtime error: signed integer overflow`

when compiled with `-fsanitize=undefined`.

I think Java has a similar issue with the first `int32_t` overflow case due to wraparound arithmetic, no idea if it suffers from the `signed integer overflow` problem that C++ has(maybe not).

If we change this, I'd suggest keep `v - (((v % W) + W) % W)` as a comment and clarify why the current form. 

I agree this is worth discussing on the dev mailing list.

fallintoplace: I am not very experienced about this process, how do we proceed?

zhjwpku: You may want to check https://lists.apache.org/list.html?dev@iceberg.apache.org.

wgtmac: cc @manuzhang

manuzhang: +1 to open a discussion on the dev list. Actually, this boundary behavior has been reported in the Java repository before:

  - [apache/iceberg#9767](https://github.com/apache/iceberg/issues/9767) reproduces `INT_MIN` wrapping to a large positive partition value and causing incorrect partition pruning.
  - [apache/iceberg#9768](https://github.com/apache/iceberg/issues/9768) notes that the specification is unclear when the mathematical truncate result is outside the source type.
  - [apache/iceberg#13105](https://github.com/apache/iceberg/issues/13105) reproduces the same `LONG_MIN`, width `10` case handled here: the result becomes `LONG_MAX - 1`, and a query such as `a < 0` misses the row.
  - [apache/iceberg#12915](https://github.com/apache/iceberg/issues/12915) covers the analogous decimal boundary and also mentions the integer minimum-value problem.

PR #756, opened 2026-06-18
fix: skip only null partition value rows in ParseDataFile

WZhuo:
## Summary

`ParseDataFile()` in `manifest_reader.cc` iterated over rows for each partition child field and used `break` to skip null partition values. `break` exits the entire inner row loop, so once a null value was encountered, **all subsequent rows for that partition column were skipped** and never had their partition values parsed.

The fix changes `break` to `continue`, so only the current null row is skipped while the remaining rows are still processed.

## Changes

- `src/iceberg/manifest/manifest_reader.cc`: `break` → `continue` in the partition-value row loop.
- `src/iceberg/test/manifest_reader_test.cc`: add regression test `NullPartitionValueDoesNotSkipSubsequentRows` that writes a manifest with a null partition value before non-null ones and verifies the later rows are parsed correctly.

## Testing

- New test passes for manifest versions V1/V2/V3.
- Confirmed the test fails with the original `break` (`partition.num_fields()` is 0 instead of 1 for rows after the null one).
- Full `manifest_test` suite passes (171 tests).

### DIFF
--- a/src/iceberg/manifest/manifest_reader.cc
+++ b/src/iceberg/manifest/manifest_reader.cc
@@ -477,7 +477,7 @@ Status ParseDataFile(const std::shared_ptr<StructType>& data_file_schema,
           auto part_view = field_view->children[part_idx];
           for (int64_t row_idx = 0; row_idx < part_view->length; row_idx++) {
             if (ArrowArrayViewIsNull(part_view, row_idx)) {
-              break;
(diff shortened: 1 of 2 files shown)


wgtmac (approved):

zhjwpku (approved):
LGTM

PR #757, opened 2026-06-18
fix: validate transform parameters in metadata

fallintoplace:
## Summary
- add `Transform::Validate` to reuse bind-time transform checks during metadata validation
- reject invalid bucket/truncate parameters when parsing transform strings
- make partition specs and sort orders reject non-positive bucket/truncate parameters during schema-bound validation

## Validation
- `cmake -S . -B build-transform-params -G Ninja -DICEBERG_BUILD_BUNDLE=OFF -DICEBERG_BUILD_REST=OFF -DICEBERG_BUILD_SHARED=OFF`
- `ctest --test-dir build-transform-params -R schema_test --output-on-failure`


src/iceberg/transform.cc
```diff
@@ -232,6 +232,18 @@ bool Transform::CanTransform(const Type& source_type) const {
...
 
+Status Transform::Validate(const std::shared_ptr<Type>& source_type) const {
+  if (!source_type) {
+    return InvalidArgument("Source type cannot be null for transform {}", ToString());
+  }
+  if (!CanTransform(*source_type)) {
+    return InvalidArgument("Invalid source type {} for transform {}",
+                           source_type->ToString(), ToString());
+  }
+  ICEBERG_RETURN_UNEXPECTED(Bind(source_type));
```
huan233usc, line 244: Worth a one-line comment on why both `CanTransform` and `Bind` are called here: `CanTransform` keeps type mismatches as `InvalidArgument`/"Invalid source type", whereas `Bind`→`Make` would report them as `NotSupported`. Otherwise a future cleanup may drop `CanTransform` as redundant and silently change the error kind.

src/iceberg/transform.cc
```diff
@@ -554,9 +566,15 @@ Result<std::shared_ptr<Transform>> TransformFromString(std::string_view transfor
                             StringUtils::ParseNumber<int32_t>(match[2].str()));
 
     if (type_str == kBucketName) {
+      if (param <= 0) {
```
huan233usc: The bucket/truncate positivity rule (and its message) is now duplicated here and in `BucketTransform::Make`/`TruncateTransform::Make`. Fine for fail-fast on unbound transforms, but easy to drift — consider a shared helper if it grows.

huan233usc (approved):
LGTM. Correct and well-targeted: metadata validation now reaches the bind-time bucket/truncate parameter checks via `Validate`→`Bind`→`Make`, and the new tests genuinely exercise the parameter path (source types are valid for their transforms). Two minor, non-blocking nits inline.

src/iceberg/transform.cc
```diff
@@ -234,6 +233,19 @@ bool Transform::CanTransform(const Type& source_type) const {
   std::unreachable();
 }
 
+Status Transform::Validate(const std::shared_ptr<Type>& source_type) const {
+  if (!source_type) {
+    return InvalidArgument("Source type cannot be null for transform {}", ToString());
```
wgtmac, line 237: Why not just use `const Type&` as the input to save this check?

src/iceberg/transform_internal.h
```diff
@@ -21,16 +21,22 @@
...
-/// gettid, macOS pthread_threadid_np, Windows GetCurrentThreadId), not the
-/// opaque std::thread::id -- and so avoids the std::formatter<std::thread::id>
-/// (P2693) minimum-toolchain dependency. The first call on a thread performs
-/// the OS query; later calls return the cached value (no syscall per call).
-/// Declared here so the platform headers stay confined to the .cc.
-uint64_t OsThreadId() noexcept;
+namespace iceberg::internal {
 
-}  // namespace iceberg
+inline Status ValidateBucketTransformParameter(int32_t num_buckets) {
```
wgtmac: Can we avoid adding this new file? We can put these functions into `src/iceberg/util/transform_util.h`

src/iceberg/transform.cc
wgtmac: +1

wgtmac (approved):

wgtmac:
I went ahead to address all comments and now CIs are green. Let me merge this. Thanks @fallintoplace and @huan233usc!

PR #758, opened 2026-06-18
fix: invalidate table scan schema cache

fallintoplace:
## Summary
- invalidate the cached scan snapshot schema whenever snapshot/ref/time-travel state changes
- preserve requested column-stat names and re-resolve them during Build so stats IDs match the final snapshot schema
- add regression coverage for IncludeColumnStats before UseSnapshot and non-main UseRef with an evolved schema

## Tests
- cmake --build build --target scan_test
- ctest --test-dir build -R scan_test --output-on-failure

### DIFF
--- a/src/iceberg/table_scan.cc
+++ b/src/iceberg/table_scan.cc
@@ -239,6 +239,8 @@ TableScanBuilder<ScanType>& TableScanBuilder<ScanType>::CaseSensitive(
 template <typename ScanType>
 TableScanBuilder<ScanType>& TableScanBuilder<ScanType>::IncludeColumnStats() {
   context_.return_column_stats = true;
+  context_.columns_to_keep_stats.clear();
+  requested_column_stats_.reset();
   return *this;
 }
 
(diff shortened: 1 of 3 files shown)


src/iceberg/table_scan.cc
```diff
@@ -239,24 +239,17 @@ TableScanBuilder<ScanType>& TableScanBuilder<ScanType>::CaseSensitive(
...
-  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto schema_ref, ResolveSnapshotSchema());
-  const auto& schema = schema_ref.get();
-  for (const auto& column_name : requested_columns) {
-    ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto field, schema->FindFieldByName(column_name));
-    if (field.has_value()) {
-      context_.columns_to_keep_stats.insert(field.value().get().field_id());
-    }
-  }
+  requested_column_stats_ = requested_columns;
+  ICEBERG_BUILDER_RETURN_IF_ERROR(ResolveColumnStatsSelection());
```
wgtmac: Could we avoid eager schema resolution here? If `IncludeColumnStats(requested_columns)` only stores the requested names, and `Build()` resolves them against the final snapshot schema, then we don't need to remember to call `InvalidateSnapshotSchema()` in every snapshot/ref setter. That seems simpler and less error-prone.

wgtmac (changes requested):

wgtmac (approved):

wgtmac:
Thank you, @fallintoplace!

PR #759, opened 2026-06-18
fix(scan): reject missing column stats fields

fallintoplace:
## Summary
- validate `IncludeColumnStats()` column names against the resolved snapshot schema during scan build
- return a validation error when requested stats columns are missing instead of silently dropping them
- leave `Schema::Select()` and regular scan projection semantics unchanged

## Tests
- `pre-commit run --all-files`
- `cmake --build build --target scan_test`
- `ctest --test-dir build -R '^scan_test$' --output-on-failure`

### DIFF
--- a/src/iceberg/table_scan.cc
+++ b/src/iceberg/table_scan.cc
@@ -400,9 +400,8 @@ Status TableScanBuilder<ScanType>::ResolveColumnStatsSelection() {
   const auto& schema = schema_ref.get();
   for (const auto& column_name : *requested_column_stats_) {
     ICEBERG_ASSIGN_OR_RAISE(auto field, schema->FindFieldByName(column_name));
-    if (field.has_value()) {
-      context_.columns_to_keep_stats.insert(field.value().get().field_id());
-    }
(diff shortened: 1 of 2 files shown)


wgtmac:
Ah, I didn't notice this PR before my agent raised https://github.com/apache/iceberg-cpp/pull/786. By practice we should respect the 1st PR that works on the same issue. However I think this PR chose the wrong direction (changed the semantics of Schema::Select) and applied a more complex fix (relied on heavy call to Schema::Select for validation). Do you mind changing your PR to use the approach in https://github.com/apache/iceberg-cpp/pull/786? Then I can merge this one and close that one.

fallintoplace:
Thank you @wgtmac, I will push the commit in a few hours.

fallintoplace:
Thanks @wgtmac, pushed the narrower version following #786’s approach.

This now validates missing `IncludeColumnStats()` columns during scan build and leaves `Schema::Select()` semantics unchanged. I also updated the PR title/description. CI is green now.

wgtmac (approved):

wgtmac:
Thank you, @fallintoplace!

PR #760, opened 2026-06-19
feat: make ToLower Unicode-aware via utf8proc

goel-skd:
Replaces the ASCII-only `StringUtils::ToLower` with a Unicode-aware implementation backed by [utf8proc](https://github.com/JuliaStrings/utf8proc), aligning case-insensitive name handling with Iceberg Java's `toLowerCase(Locale.ROOT)` for simple (1:1) case mappings. The few code points where simple and full case mapping differ (e.g. `İ` U+0130) are tracked in #808.

- `ToLower` lower-cases UTF-8 using utf8proc simple (1:1) case mapping (e.g. `CAFÉ` → `café`, `GROẞE` → `große`). Pure-ASCII input takes a byte-wise fast path and never touches utf8proc.
- Invalid UTF-8 is handled as a correctness guarantee (raised in review): `ToLower` is total — a byte that does not begin a valid UTF-8 sequence is passed through unchanged and decoding resumes at the next byte, so the valid code points around it are still lower-cased. The comparisons below therefore compare invalid bytes verbatim.
- `EqualsIgnoreCase` and `StartsWithIgnoreCase` fold through `ToLower`, so they are case-insensitive for non-ASCII letters too. Both take an allocation-free ASCII fast path and fall back to `ToLower` only when a non-ASCII byte is present.
- `StartsWithIgnoreCase` no longer assumes case mapping preserves byte length — the old byte-slice guard mishandled length-changing maps (e.g. `İ` U+0130 → `i`). Now `"İx"` starts with `"i"`, and `"i"` starts with `"İ"`.
- `ToUpper` is intentionally left ASCII-only — it only normalizes ASCII enum/codec strings, and simple case mapping would be wrong for some letters (e.g. `ß` would stay unchanged instead of becoming `SS`).
- utf8proc is pinned to 2.10.0 and wired into both the CMake (vendored via FetchContent / system package) and Meson (`subprojects/utf8proc.wrap`) builds, with matching source hashes and correct static-vs-shared linkage on Windows.

## Testing

- `string_util_test.cc`: `ToLowerUnicode` (incl. invalid / truncated / stray-continuation bytes), `ToUpperAsciiOnly`, and Unicode + invalid-UTF-8 cases for `EqualsIgnoreCase` / `StartsWithIgnoreCase`.
- `IgnoreCaseAgreesWithToLowerOracle` exhaustively checks both ASCII fast paths against the `ToLower` oracle over all short strings drawn from an alphabet with length-changing maps (`İ`→`i`, `K`→`k`) and invalid UTF-8.

## Remaining work (tracked in #808)

- Full case-mapping parity with Java for the code points where simple ≠ full mapping (e.g. `İ` → `i̇`).
- A streaming (allocation-free) non-ASCII path for `EqualsIgnoreCase` / `StartsWithIgnoreCase`.

Part of #613 — does not fully fix it; #808 holds the design and remaining plan.
Follow-up to #748.

### DIFF
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -422,6 +422,62 @@ function(resolve_croaring_dependency)
 endfunction()
 
 # ----------------------------------------------------------------------
+# utf8proc
+
+function(resolve_utf8proc_dependency)
+  prepare_fetchcontent()
+
+  if(DEFINED ENV{ICEBERG_UTF8PROC_URL})
+    set(UTF8PROC_URL "$ENV{ICEBERG_UTF8PROC_URL}")
+  else()
+    set(UTF8PROC_URL
+        "https://github.com/JuliaStrings/utf8proc/archive/refs/tags/v2.10.0.tar.gz")
+  endif()
+
+  fetchcontent_declare(utf8proc
+                       ${FC_DECLARE_COMMON_OPTIONS}
+                       URL ${UTF8PROC_URL}
+                           FIND_PACKAGE_ARGS
+                           NAMES
+                           utf8proc
+                           CONFIG)
+  fetchcontent_makeavailable(utf8proc)
+
+  if(utf8proc_SOURCE_DIR)
+    if(NOT TARGET utf8proc::utf8proc)
+      add_library(utf8proc::utf8proc INTERFACE IMPORTED)
+      target_link_libraries(utf8proc::utf8proc INTERFACE utf8proc)
+      target_include_directories(utf8proc::utf8proc INTERFACE ${utf8proc_SOURCE_DIR})
+    endif()
+
+    set(UTF8PROC_VENDORED TRUE)
+    # utf8proc's CMake puts a raw build-tree path in INTERFACE_INCLUDE_DIRECTORIES, which
+    # install(EXPORT) rejects. Wrap it in BUILD_INTERFACE so the export is valid; utf8proc
+    # is a private dependency, so installed consumers never need its headers.
+    set_target_properties(utf8proc
+                          PROPERTIES OUTPUT_NAME "iceberg_vendored_utf8proc"
+                                     POSITION_INDEPENDENT_CODE ON
+                                     INTERFACE_INCLUDE_DIRECTORIES
+                                     "$<BUILD_INTERFACE:${utf8proc_SOURCE_DIR}>")
+    install(TARGETS utf8proc
+            EXPORT iceberg_targets
+            RUNTIME DESTINATION "${ICEBERG_INSTALL_BINDIR}"
+            ARCHIVE DESTINATION "${ICEBERG_INSTALL_LIBDIR}"
+            LIBRARY DESTINATION "${ICEBERG_INSTALL_LIBDIR}")
+  else()
+    set(UTF8PROC_VENDORED FALSE)
+    list(APPEND ICEBERG_SYSTEM_DEPENDENCIES utf8proc)
+  endif()
+
+  set(ICEBERG_SYSTEM_DEPENDENCIES
+      ${ICEBERG_SYSTEM_DEPENDENCIES}
+      PARENT_SCOPE)
+  set(UTF8PROC_VENDORED
+      ${UTF8PROC_VENDORED}
+      PARENT_SCOPE)
+endfunction()
+
+# ----------------------------------------------------------------------
 # nlohmann-json
 
 function(resolve_nlohmann_json_dependency)
@@ -719,6 +775,7 @@ endfunction()
 resolve_zlib_dependency()
 resolve_nanoarrow_dependency()
 resolve_croaring_dependency()
+resolve_utf8proc_dependency()
 resolve_nlohmann_json_dependency()
 resolve_spdlog_dependency()
 
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -72,6 +72,7 @@ endfunction()
 # ICEBERG_AVRO_GIT_URL       - Apache Avro git repository URL
 # ICEBERG_NANOARROW_URL      - Nanoarrow tarball URL
 # ICEBERG_CROARING_URL       - CRoaring tarball URL
+# ICEBERG_UTF8PROC_URL       - utf8proc tarball URL
 # ICEBERG_NLOHMANN_JSON_URL  - nlohmann-json tarball URL
 # ICEBERG_SPDLOG_URL         - spdlog tarball URL
 # ICEBERG_CPR_URL            - cpr tarball URL
@@ -106,6 +107,20 @@ else()
       "https://www.apache.org/dyn/closer.lua?action=download&filename=/arrow/apache-arrow-nanoarrow-${ICEBERG_NANOARROW_BUILD_VERSION}/apache-arrow-nanoarrow-${ICEBERG_NANOARROW_BUILD_VERSION}.tar.gz"
       "https://downloads.apache.org/arrow/apache-arrow-nanoarrow-${ICEBERG_NANOARROW_BUILD_VERSION}/apache-arrow-nanoarrow-${ICEBERG_NANOARROW_BUILD_VERSION}.tar.gz"
       "https://archive.apache.org/dist/arrow/apache-arrow-nanoarrow-${ICEBERG_NANOARROW_BUILD_VERSION}/apache-arrow-nanoarrow-${ICEBERG_NANOARROW_BUILD_VERSION}.tar.gz"
+  )
+endif()
+
+set(ICEBERG_UTF8PROC_BUILD_VERSION "2.10.0")
+set(ICEBERG_UTF8PROC_BUILD_SHA256_CHECKSUM
+    "276a37dc4d1dd24d7896826a579f4439d1e5fe33603add786bb083cab802e23e")
+
+if(DEFINED ENV{ICEBERG_UTF8PROC_URL})
+  set(UTF8PROC_SOURCE_URL "$ENV{ICEBERG_UTF8PROC_URL}")
+else()
+  # Use the release asset (stable bytes, matching subprojects/utf8proc.wrap) rather
+  # than the auto-generated tag archive, whose contents GitHub does not guarantee.
+  set(UTF8PROC_SOURCE_URL
+      "https://github.com/JuliaStrings/utf8proc/releases/download/v${ICEBERG_UTF8PROC_BUILD_VERSION}/utf8proc-${ICEBERG_UTF8PROC_BUILD_VERSION}.tar.gz"
   )
 endif()
 
@@ -427,20 +442,19 @@ endfunction()
 function(resolve_utf8proc_dependency)
   prepare_fetchcontent()
 
-  if(DEFINED ENV{ICEBERG_UTF8PROC_URL})
-    set(UTF8PROC_URL "$ENV{ICEBERG_UTF8PROC_URL}")
-  else()
-    set(UTF8PROC_URL
-        "https://github.com/JuliaStrings/utf8proc/archive/refs/tags/v2.10.0.tar.gz")
-  endif()
+  # The vendored build needs no install rules; without this, CMake < 3.28 (where
+  # FetchContent has no EXCLUDE_FROM_ALL) would install utf8proc's headers and
+  # pkg-config file into the iceberg install prefix.
+  set(UTF8PROC_INSTALL OFF)
 
   fetchcontent_declare(utf8proc
                        ${FC_DECLARE_COMMON_OPTIONS}
-                       URL ${UTF8PROC_URL}
-                           FIND_PACKAGE_ARGS
-                           NAMES
-                           utf8proc
-                           CONFIG)
+                       URL ${UTF8PROC_SOURCE_URL}
+                       URL_HASH "SHA256=${ICEBERG_UTF8PROC_BUILD_SHA256_CHECKSUM}"
+                       FIND_PACKAGE_ARGS
+                       NAMES
+                       utf8proc
+                       CONFIG)
   fetchcontent_makeavailable(utf8proc)
 
   if(utf8proc_SOURCE_DIR)
--- a/src/iceberg/util/string_util.h
+++ b/src/iceberg/util/string_util.h
@@ -22,6 +22,7 @@
 #include <algorithm>
 #include <cerrno>
 #include <charconv>
+#include <optional>
 #include <ranges>
 #include <string>
 #include <string_view>
@@ -49,7 +50,11 @@ class ICEBERG_EXPORT StringUtils {
   /// above) maps to U+0069 ("i") here, but to U+0069 U+0307 ("i" + combining dot above)
   /// in Java. For ASCII and the large majority of letters the two agree.
   ///
-  /// Invalid UTF-8 input is returned unchanged.
+  /// Pure-ASCII input takes a byte-wise fast path; utf8proc is only invoked when a
+  /// non-ASCII byte (>= 0x80) is present. The function is total: it never fails, and
+  /// input need not be valid UTF-8. A byte that does not begin a valid UTF-8 sequence
+  /// is copied through unchanged and decoding resumes at the next byte, so the valid
+  /// code points around it are still lower-cased.
   /// See https://github.com/apache/iceberg-cpp/issues/613.
   static std::string ToLower(std::string_view str);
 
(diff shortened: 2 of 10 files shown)


cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -421,6 +421,62 @@ function(resolve_croaring_dependency)
...
+  fetchcontent_makeavailable(utf8proc)
+
+  if(utf8proc_SOURCE_DIR)
+    if(NOT TARGET utf8proc::utf8proc)
+      add_library(utf8proc::utf8proc INTERFACE IMPORTED)
+      target_link_libraries(utf8proc::utf8proc INTERFACE utf8proc)
+      target_include_directories(utf8proc::utf8proc INTERFACE ${utf8proc_SOURCE_DIR})
+    endif()
+
+    set(UTF8PROC_VENDORED TRUE)
```
wgtmac, line 467: `utf8proc` is licensed under the permissive MIT License (along with some Unicode data under a similarly permissive license). We need to update the LICENSE File by adding a separator at the bottom similar to this:

```
---
This product bundles utf8proc, which is available under the MIT License:
Copyright © 2014-2021 by Steven G. Johnson, Jiahao Chen, Tony Kelman, Jonas Fonseca, and other contributors listed in the git history.
Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software... (include the rest of the utf8proc MIT license text here)
```

goel-skd: Thanks @wgtmac. Good catch, let me update it.

goel-skd: Done!

src/iceberg/util/string_util.h
```diff
@@ -41,22 +40,24 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
+  ///
+  /// Mirrors Iceberg Java's case-insensitive handling, which lower-cases names with
+  /// toLowerCase(Locale.ROOT). Invalid UTF-8 input is returned unchanged.
+  /// See https://github.com/apache/iceberg-cpp/issues/613.
+  static std::string ToLower(std::string_view str);
+
+  /// \brief Upper-case ASCII letters; non-ASCII (multibyte UTF-8) bytes pass through
+  /// unchanged.
+  ///
+  /// Unlike ToLower this is ASCII-only, since upper-casing is not used for name matching.
```
wgtmac: > Unlike ToLower this is ASCII-only, since upper-casing is not used for name matching.

This is misleading to users. We should probably provide two pairs of functions like `ToLowerAscii/ToUpperAscii` and `ToLowerUtf8/ToUpperUtf8`.

src/iceberg/util/string_util.h
```diff
@@ -41,22 +40,24 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
-  // (multibyte UTF-8) bytes, are passed through unchanged.
-  // See https://github.com/apache/iceberg-cpp/issues/613.
-  static std::string ToLower(std::string_view str) {
-    return str | std::ranges::views::transform(ToLowerAscii) |
-           std::ranges::to<std::string>();
-  }
-
+  /// \brief Lower-case a UTF-8 string using Unicode simple case mapping.
+  ///
+  /// Mirrors Iceberg Java's case-insensitive handling, which lower-cases names with
```
wgtmac: I would not say this is `Mirrors Iceberg Java's case-insensitive handling` since we are not a 100% exact matching. If we cannot guarantee that, we need to clarify where we might diverge in the comment.

src/iceberg/util/string_util.h
```diff
@@ -41,22 +40,24 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
+  /// Unlike ToLower this is ASCII-only, since upper-casing is not used for name matching.
   static std::string ToUpper(std::string_view str) {
     return str | std::ranges::views::transform(ToUpperAscii) |
            std::ranges::to<std::string>();
   }
 
   static bool EqualsIgnoreCase(std::string_view lhs, std::string_view rhs) {
-    return std::ranges::equal(
-        lhs, rhs, [](char lc, char rc) { return ToLowerAscii(lc) == ToLowerAscii(rc); });
+    return ToLower(lhs) == ToLower(rhs);
```
wgtmac: This incur performance regression (if they are ASCII) because we cannot return early if their prefixes do not equal. Perhaps we can still implement this in a streaming approach?

wgtmac (changes requested):
I haven't checked the implementation and test yet. Just post some thoughts around APIs.

src/iceberg/util/string_util.h
goel-skd: Agree, will post the differences

src/iceberg/util/string_util.h
goel-skd: ASCII is a subset of UTF-8, so the UTF-8 ```ToLower``` is already correct on ASCII input. A ```ToLowerAscii``` wouldn't be more correct, only faster on known-ASCII data, and it makes the caller decide "is my input ASCII?", which is sort of a foot-gun. One ToLower also matches Iceberg Java, which only has toLowerCase(Locale.ROOT) with no ascii/utf8 split.

I suggest that we skip ```ToUpperUtf8``` either way. Simple-mapping uppercase is just wrong for things like ß (stays ß, should be SS), and nothing here uppercases non-ASCII anyway. The two ```ToUpper``` callers just normalize codec/mode strings to match GZIP/ALL.

You're right it's misleading though. I'll rename ```ToUpper -> ToUpperAscii``` so it's in the name, and keep ```ToLower``` as the Unicode one. WDYT?

src/iceberg/util/string_util.h
goel-skd: Checked the callers, they're all short ASCII on cold paths (metrics mode, `"true"`, a couple UUIDs, a header name), nothing per row. Short strings stay in SSO so `ToLower` doesn't allocate, and the UUID compare runs once per commit, so I believe there isn't a measurable regression here.

Iceberg Java does the same thing: case-insensitive lookup just lowercases the query and hits a map
  ([`Schema.caseInsensitiveFindField`](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/Schema.java#L405-L407)), and that map is built by lowercasing every field name once ([`TypeUtil.indexByLowerCaseName`](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/TypeUtil.java#L209-L228)). No streaming `equalsIgnoreCase` anywhere; the speed comes from caching the lowercased keys, which we already do in `schema.cc`/`type.cc`. So the hot path is covered and `EqualsIgnoreCase` is just the cold fallback.

Streaming is also trickier than it looks since you can't early-out on length: simple lowercasing can change byte length (`İ` → `i` goes 2 bytes to 1). 

I suggest we keep this as is and revisit if a hot-path caller comes up.

goel-skd:
> I haven't checked the implementation and test yet. Just post some thoughts around APIs.

Thanks much @wgtmac. I responded to your comments.

src/iceberg/util/string_util.h
wgtmac: My only concern on this is the naming. I don't think it is good if one is `ToXXXAscii` but the other is `ToXXXUtf8`. It would be good to just use `ToUpper` and `ToLower` when we only have a single pair and describe their behavior clearly in the comment.

goel-skd: @wgtmac I agree. What you are suggesting will be least confusing. I will make comment more descriptive along with changing the names of the functions

src/iceberg/util/string_util.h
```diff
@@ -41,22 +40,36 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
 
+  /// \brief Case-insensitive equality; compares the ToLower forms of both operands and
+  /// therefore inherits ToLower's Unicode simple-mapping behavior.
   static bool EqualsIgnoreCase(std::string_view lhs, std::string_view rhs) {
-    return std::ranges::equal(
-        lhs, rhs, [](char lc, char rc) { return ToLowerAscii(lc) == ToLowerAscii(rc); });
+    return ToLower(lhs) == ToLower(rhs);
   }
 
   static bool StartsWithIgnoreCase(std::string_view str, std::string_view prefix) {
```
wgtmac, line 90: Now that `EqualsIgnoreCase` inherits `ToLower`'s Unicode simple-mapping behavior, this helper should not slice `str` by `prefix.size()` before lowercasing. `ToLower` can change UTF-8 byte length; for example, `İ` is two bytes but maps to `i` here. So `StartsWithIgnoreCase("\xC4\xB0x", "i")` currently slices an invalid one-byte prefix and returns false, even though the lowercased string starts with the lowercased prefix.

Could we compare `ToLower(str).starts_with(ToLower(prefix))` or otherwise compare by decoded code point, and add a test for this case?

goel-skd: Thanks for pointing this sneaky bug out @wgtmac. It will bite when num bytes change across the case toggle. 
Let me push a fix.

src/iceberg/util/string_util.h
```diff
@@ -41,29 +40,46 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
 
 class ICEBERG_EXPORT StringUtils {
  public:
-  // NOTE: These convert ASCII letters only; all other bytes, including non-ASCII
-  // (multibyte UTF-8) bytes, are passed through unchanged.
-  // See https://github.com/apache/iceberg-cpp/issues/613.
-  static std::string ToLower(std::string_view str) {
```
zhjwpku, line 47: I think there are quite a few places where ToLower is only used for ASCII strings. Should we switch those to use ToUpper? Or should we leave ToLower as it is and introduce a new ToLowerUTF8 instead?

goel-skd: Thanks for the review @zhjwpku. A few thoughts:

On switching those to `ToUpper`: it probably isn't a straight swap. The case-insensitive field maps in `schema.cc`/`type.cc` are keyed by lowercase (to match Java's `toLowerCase(Locale.ROOT)`), so flipping to upper means changing comparison targets everywhere, and the field-name path needs unicode anyway (`ToUpper` is ASCII-only). It also puts the "is this ASCII?" decision back on every caller.

On `ToLowerUTF8`: we discussed the Ascii/Utf8 split earlier in this PR with @wgtmac and landed on keeping a single `ToUpper`/`ToLower` rather than `ToXXXAscii`/`ToXXXUtf8`.

I think we can get the speedup without splitting the API by adding an ASCII fast-path inside `ToLower`, i.e.  scan once for any byte `>= 0x80`, if the input is pure ASCII, do the cheap ASCII lowercase and skip utf8proc, otherwise fall back. One function, no caller changes, still unicode-correct, and the ASCII majority gets the fast path for free.

Since it's a perf optimization and not correctness, I'd suggest doing it as a small follow-up so it doesn't hold up this PR. WDYT?

src/iceberg/util/string_util.h
```diff
@@ -41,29 +40,46 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
+  /// \brief Lower-case a UTF-8 string using Unicode simple (1:1) case mapping.
+  ///
+  /// Intended for case-insensitive name matching, similar to Iceberg Java's
+  /// toLowerCase(Locale.ROOT). The mapping is locale-independent, matching the intent
+  /// of Locale.ROOT. It uses simple (1:1) case mapping rather than Java's full case
+  /// mapping, so results differ for a few code points; e.g. U+0130 (capital I with dot
+  /// above) maps to U+0069 ("i") here, but to U+0069 U+0307 ("i" + combining dot above)
+  /// in Java. For ASCII and the large majority of letters the two agree.
+  ///
+  /// Invalid UTF-8 input is returned unchanged.
```
wgtmac: Why not using `Result<std::string>` so we can return an error for invalid UTF8 input.

src/iceberg/util/string_util.h
```diff
@@ -41,29 +40,46 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
+  /// Inherits ToLower's Unicode simple-mapping behavior. The whole strings are
+  /// lower-cased rather than byte-slicing str to prefix.size(), because ToLower can
+  /// change a string's byte length (e.g. "İ" (U+0130) is two bytes but maps to "i"),
+  /// so a byte slice could split a code point or reject a valid match.
   static bool StartsWithIgnoreCase(std::string_view str, std::string_view prefix) {
-    if (str.size() < prefix.size()) {
-      return false;
-    }
-    return EqualsIgnoreCase(str.substr(0, prefix.size()), prefix);
+    return ToLower(str).starts_with(ToLower(prefix));
```
wgtmac, line 98: ditto

src/iceberg/util/string_util.h
```diff
@@ -41,29 +40,46 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
     return str | std::ranges::views::transform(ToUpperAscii) |
            std::ranges::to<std::string>();
   }
 
+  /// \brief Case-insensitive equality; compares the ToLower forms of both operands and
+  /// therefore inherits ToLower's Unicode simple-mapping behavior.
   static bool EqualsIgnoreCase(std::string_view lhs, std::string_view rhs) {
-    return std::ranges::equal(
-        lhs, rhs, [](char lc, char rc) { return ToLowerAscii(lc) == ToLowerAscii(rc); });
+    return ToLower(lhs) == ToLower(rhs);
```
wgtmac: Since `ToLower` returns as is for invalid UTF8 input, this line is a UB instead.

src/iceberg/util/string_util.h
wgtmac: One concern I still have is that the current ASCII-only behavior was a reasonable trade-off for most existing call sites: metrics modes, UUIDs, HTTP header/property names, enum-like strings, etc. are overwhelmingly ASCII. Moving `ToLower`, `EqualsIgnoreCase`, and `StartsWithIgnoreCase` to always decode UTF-8 and allocate lower-cased strings means the common path pays for Unicode support even when it is not needed.

This feels a bit unfortunate for a minor/non-ASCII case. In SQL engines we often keep ASCII-only fast paths for identifier/string handling, because most real-world names are ASCII and these helpers can sit on hot-ish paths.

Could we think about preserving the fast path here? For example:

- keep the existing byte-wise ASCII path when both inputs are ASCII;
- only fall back to utf8proc when non-ASCII bytes are actually present;
- make `EqualsIgnoreCase` streaming/short-circuiting instead of materializing two lower-cased strings;
- avoid lower-casing the whole `str` in `StartsWithIgnoreCase` when only the prefix needs to be compared.

I’m supportive of improving Unicode behavior, but I’d prefer not to regress the common ASCII path globally if we can avoid it.

src/iceberg/util/string_util.h
```diff
@@ -41,29 +40,46 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
-    return str | std::ranges::views::transform(ToLowerAscii) |
-           std::ranges::to<std::string>();
-  }
-
+  /// \brief Lower-case a UTF-8 string using Unicode simple (1:1) case mapping.
+  ///
+  /// Intended for case-insensitive name matching, similar to Iceberg Java's
+  /// toLowerCase(Locale.ROOT). The mapping is locale-independent, matching the intent
+  /// of Locale.ROOT. It uses simple (1:1) case mapping rather than Java's full case
+  /// mapping, so results differ for a few code points; e.g. U+0130 (capital I with dot
```
manuzhang, line 49: If they don't match, how can iceberg-cpp read identifier path containing such characters written by java?

src/iceberg/util/string_util.h
```diff
@@ -41,29 +40,46 @@ concept FromChars = requires(const char* p, T& v) { std::from_chars(p, p, v); };
...
+  /// above) maps to U+0069 ("i") here, but to U+0069 U+0307 ("i" + combining dot above)
+  /// in Java. For ASCII and the large majority of letters the two agree.
+  ///
+  /// Invalid UTF-8 input is returned unchanged.
+  /// See https://github.com/apache/iceberg-cpp/issues/613.
```
manuzhang, line 58: Not sure what this comment means.

goel-skd:
Thanks for all the review here. Since this is a shared util it's picked up a bunch of good ideas, but some of them expand the scope, so I want to draw a line on what lands here vs. what we track separately.

What I'll do in this PR:

- Add ASCII fast-paths to `ToLower`, `EqualsIgnoreCase`, and `StartsWithIgnoreCase`. If the inputs are pure ASCII (no byte `>= 0x80`) we just do the cheap byte-wise thing with an early-out and no allocation, and only fall back to utf8proc when there's actually a non-ASCII byte. That keeps the common case (modes, UUIDs, header/property names, enum-ish strings) as fast as it was before, which is the regression @wgtmac and @zhjwpku registered concerns about.
- Reword the `ToLower` doc comment so the behavior and the small divergence from Java are actually clear (@manuzhang's point).

On the simple vs. full case mapping question (@manuzhang): reading Java-written tables isn't affected, since field names are stored verbatim in metadata and not lower-cased. The simple/full difference only shows up in case-insensitive name resolution, and only for a few code points like U+0130. For ASCII and the vast majority of letters they're identical. Matching Java bit-for-bit on those edge cases is a bigger change and I'd rather not block this PR on it.

Two things I'd punt to follow-up issues:

- Having `ToLower` return `Result<std::string>` so it can reject invalid UTF-8 instead of passing it through. That changes the contract for every current caller, so it deserves its own PR.
- Bit-for-bit parity with Java's `toLowerCase(Locale.ROOT)` on the code points where Java does full case mapping (e.g. U+0130 `İ` -> `i̇`). We're using simple 1:1 lowercasing here, which was a deliberate non-goal in the [design doc](https://gist.github.com/goel-skd/060348143058eee0568ad1f263771e7d): it's what makes `ß`/`GROẞE` match Java, whereas casefold would diverge there. So the remaining full-mapping cases are worth tracking on their own.

Does that split sound reasonable? If so I'll push the fast-path + doc changes and open the two follow-ups.

manuzhang:
@goel-skd Please update the PR description (supposing this no longer fully fixes #613 now) and open a new issue with your design and plan, and link all your follow-up PRs there.

wgtmac:
Thanks @goel-skd for your effort! I think 1:1 Java mapping is difficult to achieve so it's fine to be out of this PR's scope. However, it's correctness issue for invalid UTF-8 so it would be better to address them altogether in this PR. WDYT?

goel-skd:
> Thanks @goel-skd for your effort! I think 1:1 Java mapping is difficult to achieve so it's fine to be out of this PR's scope. However, it's correctness issue for invalid UTF-8 so it would be better to address them altogether in this PR. WDYT?

@wgtmac - Thanks, I agree that is fair. Will push the updated code.

goel-skd:
@wgtmac - I pushed the agreed upon changes and opened a follow-up issue. Thanks

manuzhang:
@goel-skd LGTM, can you update PR description as well?

goel-skd:
> @goel-skd LGTM, can you update PR description as well?

Done!

manuzhang (approved):

wgtmac (approved):
Sorry for the delay! Thanks @goel-skd for working on this and @manuzhang for the review!

PR #762, opened 2026-06-19
fix: allow historical sort orders with dropped fields

zhjwpku:
This fix is aligned with:

https://github.com/apache/iceberg/pull/16521

### DIFF
--- a/src/iceberg/test/metadata_serde_test.cc
+++ b/src/iceberg/test/metadata_serde_test.cc
@@ -89,6 +89,50 @@ void AssertSnapshotById(const TableMetadata& metadata, int64_t snapshot_id,
   EXPECT_EQ(*snapshot.value(), expected_snapshot);
 }
 
+nlohmann::json HistoricalSortOrderWithDroppedFieldMetadataJson(
+    int32_t default_sort_order_id) {
+  nlohmann::json metadata_json = R"({
+    "format-version": 2,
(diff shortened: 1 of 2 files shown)


src/iceberg/test/metadata_serde_test.cc
```diff
@@ -486,6 +486,54 @@ TEST(MetadataSerdeTest, DeserializeV2MissingSortOrder) {
...
+    "snapshots": [],
+    "statistics": [],
+    "partition-statistics": [],
+    "snapshot-log": [],
+    "metadata-log": []
+  })"_json;
+
+  auto metadata = TableMetadataFromJson(metadata_json);
+  ASSERT_THAT(metadata, IsOk());
+  ASSERT_EQ(metadata.value()->sort_orders.size(), 2);
```
huan233usc, line 537: Can we assert the historical sort order still has two fields, including `source_id() == 2`? Otherwise this would still pass if the dropped-field sort term were skipped. A negative case where that order is the default sort order would also help verify default sort order validation is still strict.

zhjwpku: Updated the tests to cover both preserving the dropped-field historical sort order and rejecting it as the default sort order.

wgtmac (approved):

wgtmac:
Thanks @zhjwpku for the fix and @huan233usc for the review!

PR #763, opened 2026-06-19
fix(rest): align error handlers with Java implementation

zhjwpku:
While reviewing #614, I noticed that `PlanErrorHandler::Accept` does not match the behavior of the Java implementation.

A detailed comparison with Java's `ErrorHandlers` revealed several gaps between iceberg-cpp and Iceberg Java. Some of these are oversights in the original implementation, while others correspond to improvements that were made later in Iceberg Java, including:

* https://github.com/apache/iceberg/pull/13143
* https://github.com/apache/iceberg/pull/14927
* https://github.com/apache/iceberg/pull/15051
* https://github.com/apache/iceberg/pull/16059

This PR closes those gaps and brings the error handling behavior in iceberg-cpp closer to the Java implementation.

### DIFF
--- a/src/iceberg/catalog/rest/error_handlers.cc
+++ b/src/iceberg/catalog/rest/error_handlers.cc
@@ -31,8 +31,26 @@ constexpr std::string_view kIllegalArgumentException = "IllegalArgumentException
 constexpr std::string_view kNoSuchNamespaceException = "NoSuchNamespaceException";
 constexpr std::string_view kNamespaceNotEmptyException = "NamespaceNotEmptyException";
 constexpr std::string_view kNoSuchTableException = "NoSuchTableException";
-constexpr std::string_view kNoSuchPlanIdException = "NoSuchPlanIdException";
-constexpr std::string_view kNoSuchPlanTaskException = "NoSuchPlanTaskException";
+constexpr std::string_view kNotFoundException = "NotFoundException";
+constexpr std::string_view kInvalidClient = "invalid_client";
+constexpr std::string_view kInvalidRequest = "invalid_request";
+constexpr std::string_view kInvalidGrant = "invalid_grant";
+constexpr std::string_view kUnauthorizedClient = "unauthorized_client";
+constexpr std::string_view kUnsupportedGrantType = "unsupported_grant_type";
+constexpr std::string_view kInvalidScope = "invalid_scope";
+constexpr std::string_view kNull = "null";
+
+std::string_view NullIfEmpty(const std::string& value) {
+  if (value.empty()) {
+    return kNull;
+  }
+  return value;
+}
+
+Status CreateRestError(const ErrorResponse& error) {
+  return RestError("Unable to process (code: {}, type: {}): {}", error.code,
+                   NullIfEmpty(error.type), NullIfEmpty(error.message));
+}
 
 }  // namespace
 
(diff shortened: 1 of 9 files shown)


src/iceberg/catalog/rest/error_handlers.cc
```diff
@@ -104,37 +122,34 @@ Status DropNamespaceErrorHandler::Accept(const ErrorResponse& error) const {
...
-  switch (error.code) {
-    case 404:
-      if (error.type == kNoSuchNamespaceException) {
-        return NoSuchNamespace(error.message);
-      }
-      return NoSuchTable(error.message);
-    case 409:
-      return AlreadyExists(error.message);
+Status ConfigErrorHandler::Accept(const ErrorResponse& error) const {
+  if (error.code == 404 && !error.type.empty()) {
```
wgtmac: This catches fallback errors too. If /config returns a plain 404 (bad base URI, proxy, etc.), BuildDefaultErrorResponse sets type to RESTException, so we report NoSuchWarehouse instead of a REST error. Java only maps parsed typed 404s here.

zhjwpku: Good catched, fixed alone with a RESTException test case.

src/iceberg/test/error_handlers_test.cc
```diff
@@ -0,0 +1,272 @@
...
+      .code = 404,
+      .type = "UnknownException",
+      .message = "Plan task does not exist",
+  };
+
+  EXPECT_THAT(PlanTaskErrorHandler::Instance()->Accept(error),
+              IsError(ErrorKind::kNoSuchPlanTask));
+}
+
+TEST(ErrorHandlersTest, OAuthErrorHandlerMapsInvalidClientToNotAuthorized) {
```
huan233usc, line 195: Can we add one test that goes through the OAuth error parsing path, not just `OAuthErrorHandler::Accept` with a prebuilt `ErrorResponse`? Otherwise this won't catch regressions in the new `TryParseOAuthErrorResponse` logic or accidentally using the default handler for token requests.

huan233usc (approved):
Mostly LGTM, thank you!

src/iceberg/catalog/rest/http_client.cc
```diff
@@ -151,13 +154,34 @@ Result<ErrorResponse> TryParseErrorResponse(const std::string& text) {
...
+
 /// \brief Handles failure responses by invoking the provided error handler.
 Status HandleFailureResponse(const cpr::Response& response,
                              const ErrorHandler& error_handler) {
   if (IsSuccessful(response.status_code)) {
     return {};
   }
-  auto parse_result = TryParseErrorResponse(response.text);
+  auto parse_result =
+      dynamic_cast<const OAuthErrorHandler*>(&error_handler) != nullptr
```
zhjwpku: This looks a bit ugly. Maybe we should add a `parseResponse` interface to `ErrorHandler`, similar to the Java impl.

@wgtmac WDYT?

src/iceberg/test/error_handlers_test.cc
zhjwpku: Sounds good. I'll add a test to cover this path.

src/iceberg/catalog/rest/http_client.cc
wgtmac: That sounds good to me

src/iceberg/test/error_handlers_test.cc
zhjwpku: Added a test that exercises `OAuthErrorHandler::ParseResponse` with an OAuth error body, then verifies the parsed response maps correctly through `Accept`.

src/iceberg/catalog/rest/http_client.cc
zhjwpku: Done that way.

wgtmac (approved):

wgtmac:
Thanks @zhjwpku for working on this and @huan233usc for the review!

PR #764, opened 2026-06-19
fix(rest): avoid default catalog reference cycle

lishuxu:
Store the cached default catalog view as a weak_ptr so AsCatalog() can reuse a live view without keeping it alive from the RestCatalog root.


wgtmac (approved):
Good catch!

PR #765, opened 2026-06-20
ci: cache the sccache directory across C++ test builds

abnobdoss:
## What
Turn on compiler caching (sccache) for the Linux and macOS builds in `test`, `aws_test`, `sanitizer_test`, and `sql_catalog_test`, and switch the Windows `test` build to the same setup. `main` builds once and saves the cache; pull requests reuse it without writing back.

## Why
Right now only the Windows builds reuse compiled output — every Linux and macOS build recompiles the whole bundled Arrow/Parquet/Avro/Boost stack from scratch, even though it never changes between PRs. Building it once and reusing it removes most of that repeated work. Saving the cache as a single file (instead of one upload per compiled file) also avoids the upload rate limit that causes "cache write error" spam.

## Validation
On a warm pull-request run, every build reused the cache: 99.6–99.9% of files came from cache, zero write errors. The heavy builds drop from ~10–27 min to ~1.5–5 min.

### DIFF
--- a/.github/workflows/aws_test.yml
+++ b/.github/workflows/aws_test.yml
@@ -146,6 +146,9 @@ jobs:
     name: Meson SigV4 (AMD64 Ubuntu 24.04)
     runs-on: ubuntu-24.04
     timeout-minutes: 45
+    env:
+      SCCACHE_DIR: ${{ github.workspace }}/.sccache
+      SCCACHE_CACHE_SIZE: "2G"
     steps:
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
@@ -181,9 +184,20 @@ jobs:
           echo "::error::vcpkg install failed after 3 attempts"
           exit 1
       - name: Set Ubuntu Compilers
+        # Wrap the compiler with sccache: Meson uses an explicit CC/CXX verbatim,
+        # so the launcher must be prepended here to route compiles through sccache.
         run: |
-          echo "CC=gcc-14" >> $GITHUB_ENV
-          echo "CXX=g++-14" >> $GITHUB_ENV
+          echo "CC=sccache gcc-14" >> $GITHUB_ENV
+          echo "CXX=sccache g++-14" >> $GITHUB_ENV
+      - name: Restore sccache cache
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ github.workspace }}/.sccache
+          key: sccache-meson-sigv4-${{ github.run_id }}
+          restore-keys: |
+            sccache-meson-sigv4-
+      - name: Setup sccache
+        uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
       - name: Build and test Iceberg
         shell: bash
         env:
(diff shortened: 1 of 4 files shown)


zhjwpku:
Thanks for working on this. I was just about to file an issue to track enabling sccache on all platforms after finishing `enable sccache for Meson Windows builds`.

Building third-party dependencies is quite expensive, so if this works well, it can help reducing GitHub Actions resource consumption. +1 for this direction.

.github/workflows/aws_test.yml
```diff
@@ -113,11 +115,29 @@ jobs:
         if: ${{ matrix.s3 == 'ON' }}
         shell: bash
         run: bash ci/scripts/start_minio.sh
+      - name: Restore sccache cache
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
```
zhjwpku, line 119: It looks like the Windows workflow was using `mozilla-actions/sccache-action` together with `SCCACHE_GHA_ENABLED`, while this PR changes it to use explicit `actions/cache` restore/save steps plus `mozilla-actions/sccache-action`.

I'm not sure which approach is better, but in case you missed it, `mozilla-actions/sccache-action` README documents the `SCCACHE_GHA_ENABLED` setup here:
https://github.com/mozilla-actions/sccache-action#cc-code

abnobdoss: Hey @zhjwpku, thank you for the review!

With `SCCACHE_GHA_ENABLED`, sccache uploads each compiled object as its own cache entry, which adds up to hundreds per build. On the larger Linux/macOS builds this hits GitHub's upload rate-limit throttle and surfaces as `cache write errors`. The build still passes when we hit the throttle, but the affected objects silently don't get cached. Pointing sccache at a local directory with a single `actions/cache` save/restore turns it into one archive per leg, so there's nothing to throttle. In my fork that brought write errors down to zero, and it was a bit faster overall since it uploads once instead of hundreds of small objects that each take a network round trip.

zhjwpku: Great, in that case, you may want to make the same change to the Meson Windows build as well. I believe it was missed in this PR, unless I'm overlooking.

abnobdoss: Good call out - I did leave it out of this PR. I was worried there might be too many changes in one PR for review as it already has a sizeable diff on the CI files. If you think it'd be fine here I can add it, or if you prefer I can raise it as a follow-up, wdyt?

zhjwpku: I think you can add that here in a separate commit. It will trigger another CI run, and we'll be able to see how effective the cache is in practice.

abnobdoss: Done! I've folded the Meson sccache conversion into this PR - this includes all Meson legs plus the SigV4 build.

As a heads-up, we won't see the speedup on this PR itself. PR cache uploads can only be reused by the branch that created them, so I've gated saves to main to keep PR caches from saturating the repo's cache or evicting the main branch's cache entries.

I validated the full stack warm on my fork: every build finishes in about 4 minutes or less except the Windows build at ~5 min, down from 20 to 27 min cold. Example run: https://github.com/abnobdoss/iceberg-cpp/actions/runs/27910572526

zhjwpku (approved):

wgtmac (approved):
That's awesome! Thanks @abnobdoss and @zhjwpku!

PR #766, opened 2026-06-20
ci: cache vcpkg packages on Windows builds

abnobdoss:
## What
Cache the installed vcpkg packages on the Windows builds in `test` and `sql_catalog_test`, and skip the install step when the cache is present.

## Why
Every Windows run reinstalls the same packages (zlib, nlohmann-json, nanoarrow, roaring, plus cpr / sqlite3) from scratch, costing a couple of minutes each time. Caching them removes that on every run after the first. This matches what `aws_test` already does.

## Validation
A warm run reused the cached packages and skipped the install. The cache only rebuilds when the package list changes.


zhjwpku (approved):
LGTM

wgtmac (approved):

wgtmac:
Thanks @abnobdoss and @zhjwpku!

PR #767, opened 2026-06-20
ci: cancel superseded runs in remaining workflows

abnobdoss:
## What
Add a concurrency block to the five workflows that don't have one — `cpp-linter`, `pre-commit`, `license_check`, `zizmor`, and `codeql`:

```yaml
concurrency:
  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
  cancel-in-progress: true
```

## Why
The heavier workflows already cancel outdated runs, but these five don't - so pushing again to a PR leaves the old runs going and tying up runners. Grouping on `head_ref || sha` cancels superseded PR runs while leaving `main` and scheduled runs untouched.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -28,6 +28,10 @@ on:
   schedule:
     - cron: '16 4 * * 1'
 
+concurrency:
+  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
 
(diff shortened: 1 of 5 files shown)


zhjwpku (approved):
`cpp-linter` can be expensive, I'm not sure why we weren't using `workflow concurrency` for it already. The other four finish in less than a minute, I don't expect much difference there. But, having a consistent setup across all workflows seems worthwhile.

abnobdoss:
Thank you for the review! I agree the `cpp-linter` is the one where we'd actually see it invoked. And I figured, as you said, it'd be nice to make them consistent while we're at it.

wgtmac (approved):

wgtmac:
Thanks @abnobdoss for adding this and @zhjwpku for the review!

PR #770, opened 2026-06-22
feat: parallelize update and scan processing

HuaHuaY:


src/iceberg/update/snapshot_update.cc
```diff
@@ -449,7 +455,8 @@ SnapshotSummaryBuilder SnapshotUpdate::BuildManifestCountSummary(
 std::string SnapshotUpdate::ManifestPath() {
   // Generate manifest path
   // Format: {metadata_location}/{uuid}-m{manifest_count}.avro
-  std::string filename = std::format("{}-m{}.avro", commit_uuid_, manifest_count_++);
+  auto manifest_count = manifest_count_.fetch_add(1, std::memory_order_relaxed);
```
wgtmac, line 458: ditto

src/iceberg/update/snapshot_update.cc
```diff
@@ -419,8 +424,9 @@ std::string SnapshotUpdate::ManifestListPath() {
   // Generate manifest list path
   // Format: {metadata_location}/snap-{snapshot_id}-{attempt}-{uuid}.avro
   int64_t snapshot_id = SnapshotId();
+  auto attempt = attempt_.fetch_add(1, std::memory_order_relaxed) + 1;
```
wgtmac, line 427: Why not use default serial order?

HuaHuaY: These atomic variables do not depend on other atomic variables. Therefore, there is no need to worry about memory ordering affecting instruction reordering in a way that causes issues with execution timing.

src/iceberg/update/expire_snapshots.cc
```diff
@@ -343,38 +337,38 @@ class ReachableFileCleanup : public FileCleanupStrategy {
...
-      if (!live_data_files.has_value()) {
-        continue;
-      }
-
-      data_files_to_delete.insert(live_data_files->begin(), live_data_files->end());
+    auto live_file_results = ParallelCollect(
+        plan_executor_, manifests_to_delete,
+        [this, &metadata](
+            const ManifestFile& manifest) -> Result<std::unordered_set<std::string>> {
+          auto result = ReadLiveDataFilePaths(metadata, manifest);
```
wgtmac, line 345: Can we handle delete manifests here too? `ReadLiveDataFilePaths` rejects them, so reachable cleanup removes the delete manifest but leaves the delete file behind. Java uses `ManifestFiles.readPaths`, which reads both data and delete manifests.

src/iceberg/table_scan.h
```diff
@@ -302,6 +304,12 @@ class ICEBERG_TEMPLATE_CLASS_EXPORT TableScanBuilder : public ErrorCollector {
   /// \param num_rows The minimum number of rows requested
   TableScanBuilder& MinRowsRequested(int64_t num_rows);
 
+  /// \brief Configure an executor for manifest planning.
+  ///
+  /// \param executor Executor to use while planning manifests.
+  /// \return Reference to this for method chaining.
+  TableScanBuilder& PlanWith(Executor& executor);
```
wgtmac, line 311: Please document the executor lifetime here too. The built scan stores this by reference and may use it later in `PlanFiles()`.

src/iceberg/update/expire_snapshots.h
```diff
@@ -122,6 +122,12 @@ class ICEBERG_EXPORT ExpireSnapshots : public PendingUpdate {
   /// \return Reference to this for method chaining.
   ExpireSnapshots& DeleteWith(std::function<void(const std::string&)> delete_func);
 
+  /// \brief Configure an executor for planning expired snapshot metadata.
+  ///
+  /// \param executor Executor to use while planning expired snapshot metadata.
+  /// \return Reference to this for method chaining.
+  ExpireSnapshots& PlanWith(Executor& executor);
```
wgtmac, line 129: Please add the same lifetime note as `ExecuteDeleteWith`. This stores only a reference, and planning can run later during `Apply()` or `Finalize()`.

src/iceberg/update/snapshot_update.h
```diff
@@ -81,6 +83,15 @@ class ICEBERG_EXPORT SnapshotUpdate : public PendingUpdate {
     return self;
   }
 
+  /// \brief Configure an executor for manifest planning work.
+  ///
+  /// \param executor Executor to use while planning manifests.
+  /// \return Reference to this for method chaining.
+  auto& ScanManifestsWith(this auto& self, Executor& executor) {
```
wgtmac, line 90: Please document the executor lifetime here too. This stores only a reference and `Apply()` may use it later.

src/iceberg/delete_file_index.h
```diff
@@ -356,6 +357,12 @@ class ICEBERG_EXPORT DeleteFileIndex::Builder : public ErrorCollector {
   /// \brief Ignore residual expressions after partition filtering.
   Builder& IgnoreResiduals();
 
+  /// \brief Configure an optional executor for reading delete manifests.
+  ///
+  /// \param executor Executor to use, or std::nullopt to read manifests serially.
+  /// \return Reference to this for method chaining.
+  Builder& PlanWith(OptionalExecutor executor);
```
wgtmac, line 364: Can this public API also take `Executor&`? If `OptionalExecutor` is only internal plumbing, it should live in an `_internal.h` header so users do not depend on it.

HuaHuaY: We can rename the file in future PR.

src/iceberg/update/expire_snapshots.h
HuaHuaY: I removed the comment from ExecuteDeleteWith. The existence of the executor is self-evident, guaranteed by the caller.

src/iceberg/update/expire_snapshots.cc
HuaHuaY: This isn't a problem introduced by this PR. It can be fixed in a future PR.

wgtmac (approved):

PR #771, opened 2026-06-22
ci: upgrade to ubuntu-26.04 image

HuaHuaY:


.github/workflows/asf-allowlist-check.yml
```diff
@@ -40,7 +40,7 @@ permissions:
 jobs:
   asf-allowlist-check:
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
```
zhjwpku: We can use ubuntu-slim for lightweight workflows like this. But it's still 24.04, not sure if this matters.

.github/workflows/zizmor.yml
```diff
@@ -32,7 +32,7 @@ jobs:
   zizmor:
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
     name: Run zizmor 🌈
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
```
zhjwpku, line 39: ditto

HuaHuaY:
CI failed because ubuntu 26.04 image upgrades OpenSSL which's dependency libjitterentropy.a is missing.

.github/workflows/zizmor.yml
HuaHuaY: It seems that we can't use ubuntu-slim here https://github.com/apache/iceberg-cpp/actions/runs/28011068810/job/82904111206?pr=771
And there is an issue https://github.com/actions/runner-images/issues/13583

zhjwpku: good to know

zhjwpku (approved):

wgtmac (approved):

PR #774, opened 2026-06-23
feat: add rewrite manifests

zhjwpku:


src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,499 @@
...
+    ICEBERG_RETURN_UNEXPECTED(rewrite_writer.writer->Close());
+    ICEBERG_ASSIGN_OR_RAISE(auto manifest_file, rewrite_writer.writer->ToManifestFile());
+    rewrite_writer.writer.reset();
+    return manifest_file;
+  };
+
+  auto new_writer = [this, &schema](const RewriteWriter& rewrite_writer)
+      -> Result<std::unique_ptr<ManifestWriter>> {
+    std::optional<int64_t> first_row_id = std::nullopt;
+    if (base().format_version >= 3 && rewrite_writer.content == ManifestContent::kData) {
```
manuzhang: If the table was upgraded from v2, existing live files may still have null per-file row IDs. Then `ManifestEntryAdapterV3::GetFirstRowId` falls back to the writer-level value, so those files get 0, and the manifest list skips assignment because the manifest now has a non-null first_row_id. Java leaves the rewrite writer’s manifest firstRowId null via `newManifestWriter`, allowing manifest-list assignment when needed after upgrade.

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,499 @@
...
+  cleanup_all_ = false;
+  return status;
+}
+
+bool RewriteManifests::RequiresRewrite(
+    const std::unordered_set<std::string>& current_manifest_paths) const {
+  const bool has_direct_replacements = !deleted_manifests_.empty() ||
+                                       !added_manifests_.empty() ||
+                                       !rewritten_added_manifests_.empty();
+  if (!cluster_by_func_ && !predicate_ && has_direct_replacements) {
```
manuzhang: Java does not rewrite at all when `clusterByFunc == null`, while this rewrites when  `predicate_` is set. Is it intentional?

src/iceberg/update/rewrite_manifests.cc
zhjwpku: Good catch. I removed the non-null rewrite writer first_row_id logic and added a v2 -> v3 regression test.

src/iceberg/update/rewrite_manifests.cc
zhjwpku: Not intentional. I changed RequiresRewrite to match Java.

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,488 @@
...
+  };
+
+  auto new_writer = [this, &schema](const RewriteWriter& rewrite_writer)
+      -> Result<std::unique_ptr<ManifestWriter>> {
+    return ManifestWriter::MakeWriter(base().format_version, SnapshotId(), ManifestPath(),
+                                      ctx_->table->io(), rewrite_writer.spec, schema,
+                                      rewrite_writer.content);
+  };
+
+  std::vector<ManifestFile> result;
```
manuzhang: Closed manifests are kept only in this local `result` vector until `Rewrite()` returns, and `new_manifests_` is assigned only after the call succeeds in `Apply()`. If a later `length()`, `WriteExistingEntry`, `Close()`, or
  `ToManifestFile()` returns an error, finalization/cleanup no longer has the already-created manifest paths.

Could we either append closed manifests to member state as soon as they are produced, or clean up the local partial `result` on the error path? It would also be good to add a test that injects an apply-time manifest writer failure after at least one rolled manifest has been closed.

zhjwpku: Good catch. I'll track rewritten manifests once each writer is closed so that apply-time failures can clean up already-created manifests. Will add a regression test too.

src/iceberg/update/rewrite_manifests.h
```diff
@@ -0,0 +1,149 @@
...
+/// function as well as to replace specific manifests. Manifests that are deleted
+/// or added directly are ignored during the rewrite process. The set of active
+/// files in replaced manifests must be the same as in new manifests.
+///
+/// When committing, these changes will be applied to the latest table snapshot.
+/// Commit conflicts will be resolved by applying the changes to the new latest
+/// snapshot and reattempting the commit.
+class ICEBERG_EXPORT RewriteManifests : public SnapshotUpdate {
+ public:
+  using ClusterByFunc = std::function<std::string(const DataFile&)>;
```
wgtmac, line 55: Why this function returns `string` instead of `size_t` just like what C++ hash function does?

src/iceberg/update/rewrite_manifests.h
```diff
@@ -0,0 +1,149 @@
...
+  void ResetRewriteState();
+  Status ValidateTargetBranch(const std::string& branch) const override;
+
+ private:
+  std::string table_name_;
+  ClusterByFunc cluster_by_func_;
+  RewritePredicate predicate_;
+
+  std::vector<ManifestFile> deleted_manifests_;
+  std::unordered_set<std::string> deleted_manifest_paths_;
```
wgtmac, line 133: We should combine `deleted_manifests_` and `deleted_manifest_paths_` into a single `unordered_set<ManifestFile>` by using pure `ManifestFile.path` in its hash function.

wgtmac: Same for `rewritten_manifests_` and `rewritten_manifest_paths_` below.

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+
+struct ManifestEntries {
+  ManifestFile manifest;
+  std::shared_ptr<PartitionSpec> spec;
+  std::vector<ManifestEntry> entries;
+};
+
+struct RewriteWriter {
+  std::shared_ptr<PartitionSpec> spec;
+  ManifestContent content;
```
wgtmac, line 65: Why do we need this? This class only rewrites data manifests, right?

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+  } else {
+    // The manifest must be rewritten with this update's snapshot ID. CopyManifest
+    // also validates that the manifest only contains existing entries.
+    ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto copied_manifest, CopyManifest(manifest));
+    rewritten_added_manifests_.push_back(std::move(copied_manifest));
+  }
+  return *this;
+}
+
+Status RewriteManifests::ValidateTargetBranch(const std::string& branch) const {
```
wgtmac, line 128: Why do we need this? I didn't find this in the Java code.

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+
+std::string RewriteManifests::operation() { return DataOperation::kReplace; }
+
+Result<std::vector<ManifestFile>> RewriteManifests::Apply(
+    const TableMetadata& /*metadata_to_update*/,
+    const std::shared_ptr<Snapshot>& snapshot) {
+  ICEBERG_PRECHECK(snapshot != nullptr,
+                   "Cannot rewrite manifests without a current snapshot");
+
+  SnapshotCache cached_snapshot(snapshot.get());
```
wgtmac, line 142: I have to admit that `SnapshotCache` is not a good name. Perhaps we should rename it to `SnapshotReader` or something.

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto current_manifests,
+                          cached_snapshot.Manifests(ctx_->table->io()));
+
+  std::unordered_set<std::string> current_manifest_paths;
```
wgtmac, line 146: ditto, we can use `unordered_set<ManifestFile>` here too

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+  for (const auto& manifest : current_manifests) {
+    current_manifest_paths.insert(manifest.manifest_path);
+  }
+
+  ICEBERG_RETURN_UNEXPECTED(
+      ValidateDeletedManifests(current_manifest_paths, snapshot->snapshot_id));
+
+  if (RequiresRewrite(current_manifest_paths)) {
+    ICEBERG_ASSIGN_OR_RAISE(auto rewritten, Rewrite(current_manifests));
+    new_manifests_ = std::move(rewritten);
```
wgtmac, line 157:
```suggestion
    ICEBERG_ASSIGN_OR_RAISE(new_manifests_, Rewrite(current_manifests));
```

Can we directly do this?

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+    }
+  }
+  return {};
+}
+
+Status RewriteManifests::ValidateActiveFiles() const {
+  // Compare the number of active (added + existing) files between created and
+  // replaced manifests using the persisted manifest counts, mirroring Java's
+  // BaseRewriteManifests.validateFilesCounts. This avoids re-reading manifest
+  // entries on every apply, including commit retries.
```
wgtmac, line 283:
```suggestion
  // Compare the number of active (added + existing) files between created and
  // replaced manifests using the persisted manifest counts.
```

I would rather rename it to `ValidateFilesCounts` to so we don't need these verbose comments.

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+
+  auto new_writer =
+      [this, &schema](
+          RewriteWriter& rewrite_writer) -> Result<std::unique_ptr<ManifestWriter>> {
+    auto manifest_path = ManifestPath();
+    auto writer_result = ManifestWriter::MakeWriter(
+        base().format_version, SnapshotId(), manifest_path, ctx_->table->io(),
+        rewrite_writer.spec, schema, rewrite_writer.content);
+    if (!writer_result.has_value()) {
+      failed_manifest_paths_.insert(std::move(manifest_path));
```
wgtmac, line 428: Why do we need to deal with this? It seems that other update implementations do not handle this kind of error so let's be consistent and simple?

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+      if (!rewritten_manifest_paths_.contains(manifest.manifest_path) &&
+          !deleted_manifest_paths_.contains(manifest.manifest_path)) {
+        kept_manifests_.push_back(manifest);
+      }
+    }
+  }
+
+  ICEBERG_RETURN_UNEXPECTED(ValidateActiveFiles());
+
+  std::vector<ManifestFile> manifests;
```
wgtmac, line 172: Can we use ranges/views/transform to reduce these lines?

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+    SetSnapshotId(manifest, snapshot_id);
+    manifests.push_back(manifest);
+  }
+  // Kept manifests are carried over unchanged, matching Java which adds them
+  // as-is without recomputing counts.
+  for (const auto& manifest : kept_manifests_) {
+    manifests.push_back(manifest);
+  }
+
+  manifest_count_summary_ = BuildManifestCountSummary(
```
wgtmac, line 195: Why is it produced here but not in the `Summary()` call? It is hard to understand the code when `BuildManifestCountSummary` is called. Perhaps we can just follow the Java approach to explicitly compute each key/value for the summary.

src/iceberg/manifest/rolling_manifest_writer.h
```diff
@@ -76,6 +76,9 @@ class ICEBERG_EXPORT RollingManifestWriter {
                             int64_t data_sequence_number,
                             std::optional<int64_t> file_sequence_number = std::nullopt);
 
+  /// \brief Add an existing entry while preserving snapshot and sequence fields.
+  Status WriteExistingEntry(const ManifestEntry& entry);
```
wgtmac, line 80: I'd propose removing this overload to let caller to directly pass those parameters to avoid misuse.

src/iceberg/update/pending_update.cc
```diff
@@ -35,6 +35,10 @@ Status PendingUpdate::Commit() {
   if (!ctx_->transaction) {
     // Table-created path: no transaction exists yet, create a temporary one.
     ICEBERG_ASSIGN_OR_RAISE(auto txn, Transaction::Make(ctx_));
+    auto self = weak_from_this().lock();
```
wgtmac, line 38: Why do we need this kind of change?

src/iceberg/update/snapshot_update.h
```diff
@@ -104,6 +104,11 @@ class ICEBERG_EXPORT SnapshotUpdate : public PendingUpdate {
       return self.AddError(ErrorKind::kInvalidArgument, "Branch name cannot be empty");
     }
 
+    if (auto status = static_cast<SnapshotUpdate&>(self).ValidateTargetBranch(branch);
```
wgtmac, line 107: Why do we need this?

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+  if (created_active_files != replaced_active_files) {
+    return ValidationFailed(
+        "Replaced and created manifests must have the same number of active files: {} "
+        "(new), {} (old)",
+        created_active_files, replaced_active_files);
+  }
+  return {};
+}
+
+Result<ManifestFile> RewriteManifests::CopyManifest(const ManifestFile& manifest) {
```
wgtmac, line 320: Let's move this function to `src/iceberg/manifest/manifest_util_internal.h` and follow the style of `CopyAppendManifest`?

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+      auto reader,
+      ManifestReader::Make(manifest.manifest_path, manifest.manifest_length,
+                           ctx_->table->io(), schema, spec,
+                           std::move(inheritable_metadata), manifest.first_row_id,
+                           /*is_committed=*/false));
+  ICEBERG_ASSIGN_OR_RAISE(auto entries, reader->Entries());
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto writer,
+      ManifestWriter::MakeWriter(base().format_version, SnapshotId(), ManifestPath(),
```
wgtmac, line 338: The generated path is lost after MakeWriter succeeds. If writing, closing, or ToManifestFile fails, the output is not reliably closed or tracked for cleanup. Please retain the path and manage it with RAII until the manifest is completed successfully.

Unfortunately, `CopyAppendManifest` in the manifest_util_internal.h has the same issue.

src/iceberg/update/rewrite_manifests.cc
```diff
@@ -0,0 +1,534 @@
...
+
+  // Capture a write failure so all open writers can still be closed below
+  Status write_status = [&]() -> Status {
+    for (const auto& manifest_entry : manifest_entries) {
+      for (const auto& entry : manifest_entry.entries) {
+        ICEBERG_PRECHECK(entry.data_file != nullptr,
+                         "Manifest entry in {} is missing data_file",
+                         manifest_entry.manifest.manifest_path);
+        auto key = WriterKey{
+            cluster_by_func_ ? cluster_by_func_(*entry.data_file) : std::string{},
```
wgtmac, line 443: `cluster_by_func_` is user code and may throw. An exception here bypasses the close-all loop and leaves partial manifests untracked. Please make writer cleanup RAII and add a throwing-callback test.

src/iceberg/update/pending_update.cc
wgtmac: Review comment from Codex: This registers standalone updates for commit retries, but explicit transaction updates still return from txn->Apply without calling Finalize on failure. A failed rewrite can leave generated manifests behind. Please finalize or abort transaction-owned updates on apply failure and add a Transaction::NewRewriteManifests fault-injection test.

src/iceberg/update/pending_update.cc
```diff
@@ -35,6 +35,10 @@ Status PendingUpdate::Commit() {
...
+    if (self) {
+      ICEBERG_RETURN_UNEXPECTED(txn->AddUpdate(self));
+    }
```
wgtmac, line 41: Please require shared ownership and always register the update.

```suggestion
auto self = weak_from_this().lock();
ICEBERG_PRECHECK(self != nullptr,
                 "PendingUpdate must be owned by std::shared_ptr");
ICEBERG_RETURN_UNEXPECTED(txn->AddUpdate(self));
```

src/iceberg/update/pending_update.cc
```diff
@@ -35,6 +35,10 @@ Status PendingUpdate::Commit() {
   if (!ctx_->transaction) {
     // Table-created path: no transaction exists yet, create a temporary one.
     ICEBERG_ASSIGN_OR_RAISE(auto txn, Transaction::Make(ctx_));
```
wgtmac, line 37: Please keep this temporary transaction detached. Remove `ctx->transaction = ...` from `Transaction::Make(ctx)`; only explicit transactions should set it. Otherwise the next `Commit()` sees an expired transaction.

src/iceberg/update/pending_update.cc
```diff
@@ -43,11 +47,15 @@ Status PendingUpdate::Commit() {
...
     if (!commit_result.has_value()) {
-      std::ignore = Finalize(std::unexpected(commit_result.error()));
+      if (!self) {
+        std::ignore = Finalize(std::unexpected(commit_result.error()));
+      }
       return std::unexpected(commit_result.error());
     }
 
-    std::ignore = Finalize(commit_result.value()->metadata().get());
+    if (!self) {
```
wgtmac, line 56: Please route the empty-update branch in `Transaction::Commit()` through the normal finalization block. Otherwise registered no-op updates skip `Finalize()`.

wgtmac (commented):
These comments are about the standalone update lifecycle:

- Registration must be mandatory so commit retries reapply the update.
- The temporary transaction must stay detached so its expired weak pointer does not break a later `Commit()`.
- No-op commits must still call `Finalize()` because cleanup is now delegated to `Transaction`.

Together, these keep one ownership, retry, and finalization path for standalone and explicit transactions.

zhjwpku:
> These comments are about the standalone update lifecycle:
> 
> * Registration must be mandatory so commit retries reapply the update.
> * The temporary transaction must stay detached so its expired weak pointer does not break a later `Commit()`.
> * No-op commits must still call `Finalize()` because cleanup is now delegated to `Transaction`.
> 
> Together, these keep one ownership, retry, and finalization path for standalone and explicit transactions.

I think this standalone update lifecycle logic deserves a dedicated PR, so I do that in https://github.com/apache/iceberg-cpp/pull/868, please take a look

src/iceberg/update/rewrite_manifests.cc
zhjwpku: Agreed, resolving in https://github.com/apache/iceberg-cpp/pull/945

src/iceberg/update/pending_update.cc
zhjwpku: Standalone update lifecycle issue has been resolved in a separate PR.

src/iceberg/update/rewrite_manifests.cc
zhjwpku: Fixed.

Issue #775, opened 2026-06-23
feat: implement ReplacePartitions (two-PR plan)


shangxinli:
## Overview

This issue tracks the implementation of `ReplacePartitions` in C++, assigned in #637.

`ReplacePartitions` performs dynamic partition overwrite: each file added to the operation marks its entire partition for deletion, and on commit, existing files in those partitions are atomically replaced. The resulting snapshot uses `operation="overwrite"` and summary property `"replace-partitions"="true"`.

### PR 1 — Core implementation

New files:
- `src/iceberg/update/replace_partitions.h`
- `src/iceberg/update/replace_partitions.cc`

Modified files:
- `src/iceberg/snapshot.h` — add `kReplacePartitions` constant to `SnapshotSummaryFields`
- `src/iceberg/type_fwd.h` — add `class ReplacePartitions;`
- `src/iceberg/CMakeLists.txt` — register new source

### PR 2 — Table/Transaction API surface + tests

Modified files:
- `src/iceberg/table.h` / `table.cc` — add `NewReplacePartitions()`
- `src/iceberg/transaction.h` / `transaction.cc` — add `NewReplacePartitions()`

New files:
- `src/iceberg/test/replace_partitions_test.cc`
- `src/iceberg/test/CMakeLists.txt` — register test

## Design

`ReplacePartitions` extends `SnapshotUpdate` (same pattern as `FastAppend`) and uses `ManifestFilterManager` to mark existing partition files as DELETED:

- `AddFile(file)`: calls `filter_manager_.DropPartition(spec_id, partition)`, accumulates new file in `new_data_files_by_spec_`
- `ValidateAppendOnly()`: calls `filter_manager_.FailAnyDelete()`
- `Apply()`: for unpartitioned tables calls `filter_manager_.DeleteByRowFilter(AlwaysTrue)`; calls `FilterManifests()` on existing snapshot, then prepends new manifests
- `operation()`: returns `DataOperation::kOverwrite`
- `Summary()`: sets `"replace-partitions"="true"`

Relates to: #637

PR #776, opened 2026-06-24
feat: add ReplacePartitions core class (PR1 of 2 for #637)

shangxinli:
## Summary

Adds the `iceberg::ReplacePartitions` class — the dynamic partition overwrite operation. Each `AddFile()` registers the file's partition for replacement of all existing data files in that partition. Produces an `overwrite` snapshot with `"replace-partitions"="true"` in the summary. Unpartitioned tables replace all existing data files.

**This is PR1 of 2** for #637. See tracking issue #775 for the split rationale.

### What's in PR1
- `iceberg::ReplacePartitions` class extending `MergingSnapshotUpdate`:
  - Builder API: `AddFile`, `ValidateAppendOnly`, `ValidateFromSnapshot`, `ValidateNoConflictingData`, `ValidateNoConflictingDeletes`
  - `Apply` / `Summary` / `CleanUncommitted` overrides
  - Uses `ManifestFilterManager` for partition-based manifest filtering
- `SnapshotSummaryFields::kReplacePartitions = "replace-partitions"` constant
- `ReplacePartitions` forward declaration in `type_fwd.h`
- CMake + Meson registration

### What's deferred to PR2
- `Transaction::NewReplacePartitions()` API wiring on `Table` / `Transaction`
- End-to-end test file (`replace_partitions_test.cc`)

### Design notes
- Follows the Java `ReplacePartitions` semantics: operation type is `overwrite`, summary carries `"replace-partitions"="true"`.
- Mirrors the existing C++ `FastAppend` / `MergeAppend` patterns (factory `Make()`, `MergingSnapshotUpdate` base, builder error collector).
- Unpartitioned table case treated as table-wide replace via `DeleteByRowFilter(AlwaysTrue())`, matching Java `BaseReplacePartitions`.

## Test plan
- [x] Clean build (`cmake --build build`)
- [x] `clang-format` clean
- [ ] End-to-end behavioral tests land in PR2 alongside the `Table`/`Transaction` API wiring

Closes part of #637 (PR2 will close it).
Tracking: #775


src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,112 @@
...
+    return {};
+  }
+
+  auto io = ctx_->table->io();
+  if (validate_conflicting_data_) {
+    ICEBERG_RETURN_UNEXPECTED(ValidateAddedDataFiles(
+        current_metadata, starting_snapshot_id_, replaced_partitions_, snapshot, io));
+  }
+  if (validate_conflicting_deletes_) {
+    ICEBERG_RETURN_UNEXPECTED(ValidateNoNewDeleteFiles(
```
manuzhang: Java `BaseReplacePartitions.validate` also calls `validateDeletedDataFiles`, so concurrent overwrite/delete commits in the replaced partition are rejected

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,112 @@
...
+  int32_t spec_id = file->partition_spec_id.value();
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto spec, base().PartitionSpecById(spec_id));
+
+  ICEBERG_BUILDER_RETURN_IF_ERROR(AddDataFile(file));
+  // DropPartition(spec_id, partition) registers the (spec_id, partition_values)
+  // tuple with both data and delete filter managers. For an unpartitioned spec
+  // the partition values are empty and naturally match every file under that
+  // spec — no separate AlwaysTrue path is needed, and validation stays scoped
+  // to the spec rather than the whole table.
+  ICEBERG_BUILDER_RETURN_IF_ERROR(DropPartition(spec_id, file->partition));
```
manuzhang, line 55: unpartitioned replace is spec-scoped, but Java treats it as table-wide.

shangxinli: Fixed

src/iceberg/update/replace_partitions.cc
shangxinli: Fixed

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,138 @@
...
+
+Status ReplacePartitions::Validate(const TableMetadata& current_metadata,
+                                   const std::shared_ptr<Snapshot>& snapshot) {
+  if (snapshot == nullptr) {
+    return {};
+  }
+  // No-op update: no partitions were staged and no table-wide replace was
+  // requested, so there is nothing to conflict with. Calling the validators
+  // with AlwaysTrue here would turn an empty builder into a full-table check.
+  if (!replace_by_row_filter_ && replaced_partitions_.empty()) {
```
manuzhang: Java rejects no-data-files replace. Suggest to call `DataSpec()` like Java.

manuzhang (changes requested):
It looks `replace_partitions.h` is missing from `src/iceberg/update/meson.build`.

zhjwpku:
In the PR description, should "the iceberg::ReplacePartitions class extending SnapshotUpdate" be changed to "the iceberg::ReplacePartitions class extending MergingSnapshotUpdate"?

src/iceberg/update/replace_partitions.h
```diff
@@ -0,0 +1,130 @@
...
+
+#include "iceberg/iceberg_export.h"
+#include "iceberg/result.h"
+#include "iceberg/type_fwd.h"
+#include "iceberg/update/merging_snapshot_update.h"
+#include "iceberg/util/partition_value_util.h"
+
+namespace iceberg {
+
+/// \brief Replaces partitions in a table with new data files.
```
zhjwpku: Do we need to mention `This is provided to implement SQL compatible with Hive table operations but is not recommended. Instead, use the {@link OverwriteFiles overwrite API} to explicitly overwrite data.` like java?

src/iceberg/update/replace_partitions.cc
shangxinli: Done

src/iceberg/update/replace_partitions.h
shangxinli: Changed to '/// \note This is provided to implement SQL compatible with Hive table operations but is not recommended. Instead, use OverwriteFiles to explicitly overwrite data.'

shangxinli:
> In the PR description, should "the iceberg::ReplacePartitions class extending SnapshotUpdate" be changed to "the iceberg::ReplacePartitions class extending MergingSnapshotUpdate"?

Updated the PR description — "extending SnapshotUpdate" → "extending MergingSnapshotUpdate". Thanks for catching that.

src/iceberg/update/replace_partitions.cc
manuzhang: Not sure where it's been implemented. Can you check again?

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,140 @@
...
+ReplacePartitions& ReplacePartitions::AddFile(const std::shared_ptr<DataFile>& file) {
+  ICEBERG_BUILDER_CHECK(file != nullptr, "Invalid data file: null");
+  ICEBERG_BUILDER_CHECK(file->partition_spec_id.has_value(),
+                        "Data file must have partition spec ID");
+
+  int32_t spec_id = file->partition_spec_id.value();
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto spec, base().PartitionSpecById(spec_id));
+
+  ICEBERG_BUILDER_RETURN_IF_ERROR(AddDataFile(file));
+  if (spec->fields().empty()) {
```
wgtmac: Java's `PartitionSpec.isUnpartitioned()` returns true not only for an empty field list but also for specs where all partition transforms are `void`/`alwaysNull`. With an all-void partition spec, `BaseReplacePartitions.apply` goes through the unpartitioned path and deletes by `alwaysTrue()`, replacing the whole table. This check only looks at `fields().empty()`, so C++ will treat an all-void spec as partitioned and call `DropPartition`, leaving files from other equivalent unpartitioned/all-void specs instead of doing the table-wide replace.

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,140 @@
...
+}
+
+std::string ReplacePartitions::operation() { return DataOperation::kOverwrite; }
+
+Status ReplacePartitions::Validate(const TableMetadata& current_metadata,
+                                   const std::shared_ptr<Snapshot>& snapshot) {
+  // Match Java BaseReplacePartitions: require at least one staged data file.
+  // Use the flags AddFile() sets — `DataSpec()` would also error on multi-spec
+  // stages and is not the guard we want here.
+  if (!replace_by_row_filter_ && replaced_partitions_.empty()) {
```
wgtmac: Java still calls `dataSpec()` from `validate`/`apply`, which enforces that exactly one partition spec is represented by the staged data files. This guard only checks that at least one partition was recorded, so a replace containing files from multiple spec IDs can commit in C++ even though Java rejects it. That can produce snapshot semantics that don't align with Java after spec evolution.

src/iceberg/update/replace_partitions.cc
shangxinli: Fixed. Replaced the `fields().empty()` guard with an `is_unpartitioned` lambda that matches Java's `PartitionSpec.isUnpartitioned()` — a spec is treated as unpartitioned if it has no fields or all transforms are void. All-void specs now route through `DeleteByRowFilter(AlwaysTrue())` instead of `DropPartition`.

src/iceberg/update/replace_partitions.cc
shangxinli: Fixed. `Validate()` now calls `DataSpec()` in the partitioned path, which errors if no files were staged or if staged files span more than one partition spec — matching Java's behavior. The unpartitioned path skips it because `replace_by_row_filter_` is only set after a successful `AddFile()`, so at least one file is already guaranteed.

src/iceberg/update/replace_partitions.cc
shangxinli: Sorry, the previous "Done" was premature — `DataSpec()` wasn't actually being called at that point. It's now in place: `Validate()` calls `DataSpec()` in the partitioned path (`!replace_by_row_filter_`), which throws if no files were staged or if staged files span multiple partition specs, matching Java's `BaseReplacePartitions.validate`.

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,153 @@
...
+
+std::string ReplacePartitions::operation() { return DataOperation::kOverwrite; }
+
+Status ReplacePartitions::Validate(const TableMetadata& current_metadata,
+                                   const std::shared_ptr<Snapshot>& snapshot) {
+  // Match Java BaseReplacePartitions.validate: require at least one staged data
+  // file and that all staged files share exactly one partition spec.
+  // DataSpec() enforces both invariants; ignore the returned spec here since
+  // replace_by_row_filter_ / replaced_partitions_ already scope validation.
+  if (!replace_by_row_filter_) {
```
wgtmac: Please call DataSpec() before this branch unconditionally. If an unpartitioned or all-void file is added first, replace_by_row_filter_ becomes true and the single-spec check is skipped; a later file from another spec then commits a table-wide replace, while Java rejects mixed specs.

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,153 @@
...
+}
+
+ReplacePartitions& ReplacePartitions::ValidateNoConflictingDeletes() {
+  validate_conflicting_deletes_ = true;
+  return *this;
+}
+
+std::string ReplacePartitions::operation() { return DataOperation::kOverwrite; }
+
+Status ReplacePartitions::Validate(const TableMetadata& current_metadata,
```
wgtmac, line 108: Please add direct ReplacePartitions tests before merging this core class. Cover partitioned and unpartitioned/all-void replacement, empty and mixed specs, append-only validation, same- and different-partition concurrency, retry summary retention, and DV cleanup. Existing base-class tests do not exercise this orchestration.

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,153 @@
...
+    replace_by_row_filter_ = true;
+  } else {
+    ICEBERG_BUILDER_RETURN_IF_ERROR(DropPartition(spec_id, file->partition));
+    replaced_partitions_.add(spec_id, file->partition);
+  }
+  return *this;
+}
+
+ReplacePartitions& ReplacePartitions::ValidateAppendOnly() {
+  FailAnyDelete();
```
wgtmac, line 62: Please translate the fail-any-delete error in ReplacePartitions, as BaseReplacePartitions.apply() does. C++ currently returns "Operation would delete existing data", while Java returns "Cannot commit file that conflicts with existing partition: ...".

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,153 @@
...
+    }
+  }
+  if (validate_conflicting_deletes_) {
+    // Java's BaseReplacePartitions.validate gates both ValidateNoNewDeleteFiles
+    // and ValidateDeletedDataFiles on the same validateNewDeletes flag. The
+    // second check rejects concurrent overwrite/delete commits in the replaced
+    // partitions; without it a concurrent delete in a replaced partition would
+    // commit silently.
+    if (replace_by_row_filter_) {
+      ICEBERG_RETURN_UNEXPECTED(ValidateNoNewDeleteFiles(
```
wgtmac: Please run ValidateDeletedDataFiles before ValidateNoNewDeleteFiles in both branches, matching Java. When both conflict types exist, the current order reports a different validation failure.

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,153 @@
...
+ReplacePartitions& ReplacePartitions::AddFile(const std::shared_ptr<DataFile>& file) {
+  ICEBERG_BUILDER_CHECK(file != nullptr, "Invalid data file: null");
+  ICEBERG_BUILDER_CHECK(file->partition_spec_id.has_value(),
+                        "Data file must have partition spec ID");
+
+  int32_t spec_id = file->partition_spec_id.value();
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto spec, base().PartitionSpecById(spec_id));
+
+  ICEBERG_BUILDER_RETURN_IF_ERROR(AddDataFile(file));
+  // A spec is effectively unpartitioned if it has no fields, or every field
```
wgtmac: Please keep these comments focused on the rule. For example: "Specs with no non-void transforms are unpartitioned." Remove the Java comparison and implementation-history detail here and in the validation block below.

src/iceberg/update/replace_partitions.h
```diff
@@ -0,0 +1,134 @@
...
+                  const std::shared_ptr<Snapshot>& snapshot) override;
+
+ private:
+  explicit ReplacePartitions(std::string table_name,
+                             std::shared_ptr<TransactionContext> ctx);
+
+  std::optional<int64_t> starting_snapshot_id_;
+  bool validate_conflicting_data_{false};
+  bool validate_conflicting_deletes_{false};
+  // True once an AddFile() call has staged a file whose partition spec is
```
wgtmac: Please describe this member directly, for example: "True when replacement uses an AlwaysTrue row filter." Keep the Java comparison and design rationale in the PR discussion.

wgtmac (commented):
Codex was used to review Java parity against Iceberg Java's ReplacePartitions, BaseReplacePartitions, MergingSnapshotProducer, and TestReplacePartitions. The inline comments cover the remaining behavior, API documentation, and test gaps.

src/iceberg/update/replace_partitions.h
```diff
@@ -0,0 +1,134 @@
...
+  /// \brief Enable validation that no conflicting data files were added concurrently.
+  ///
+  /// Fails the commit if a concurrent operation added a data file in any of
+  /// the partitions being replaced after the snapshot set by
+  /// ValidateFromSnapshot().
+  ///
+  /// \return Reference to this for method chaining
+  ReplacePartitions& ValidateNoConflictingData();
+
+  /// \brief Enable validation that no conflicting delete files were added concurrently.
```
wgtmac: Please align this contract with Java. This validation also rejects concurrent data-file removals, not only newly added delete files; mention that it prevents undeleting rows removed by a concurrent commit.

src/iceberg/update/replace_partitions.cc
shangxinli: Moved DataSpec() out of the row-filter branch so it runs unconditionally. A replace that stages an unpartitioned/all-void file first and then a file from another spec is now rejected instead of committing a table-wide replace. Covered by the UnpartitionedFirstThenMixedRejected test.

src/iceberg/update/replace_partitions.cc
shangxinli: Added an Apply() override that rewrites the fail-any-delete error to "Cannot commit file that conflicts with existing partition: ...", matching BaseReplacePartitions.apply(). Verified by AppendOnlyConflictTranslated.

src/iceberg/update/replace_partitions.cc
shangxinli: Swapped the order so ValidateDeletedDataFiles runs before ValidateNoNewDeleteFiles in both branches.

src/iceberg/update/replace_partitions.cc
shangxinli: Trimmed to "Specs with no non-void transforms are unpartitioned." and dropped the Java/history detail in the validation block below.

src/iceberg/update/replace_partitions.h
shangxinli: Reworded to "True when replacement uses an AlwaysTrue row filter."

src/iceberg/update/replace_partitions.h
shangxinli: Updated the doc to note this also rejects concurrent data-file removals, not just newly added delete files, so the replace can't undelete rows a concurrent commit removed.

src/iceberg/update/replace_partitions.cc
shangxinli: Added src/iceberg/test/replace_partitions_test.cc with direct tests: partitioned and unpartitioned/all-void replacement, empty and mixed specs, append-only validation (pass + translated conflict), and same/different-partition data and delete concurrency. Retry-summary retention and DV cleanup run through the base MergingSnapshotUpdate paths, which have their own coverage; happy to add ReplacePartitions-specific cases for those too if you'd prefer.

src/iceberg/update/replace_partitions.cc
wgtmac: The new `Apply()`/`Validate()` comments and the header still include Java comparisons and fix-history detail. Please keep only the contract, for example: `DataSpec()` requires non-empty, single-spec input, and deleted-data validation runs before new-delete validation.

src/iceberg/update/replace_partitions.cc
wgtmac: `CustomSummaryPropertySurvivesApplyRebuild` covers the summary hook, but the new tests still miss the distinct all-void, unpartitioned conflict-validation, concurrent data-file removal, and format-v3 DV cleanup paths. `AddUnpartitionedSpec` creates an empty spec, and `NoConflictingDeletesFails` only exercises a newly added delete file. Please add focused cases for those paths.

wgtmac (commented):
Codex was used for this Java parity review against the current Iceberg Java `ReplacePartitions`, `BaseReplacePartitions`, and tests. The remaining inline comments cover public contracts, comment cleanup, and focused test gaps.

src/iceberg/update/replace_partitions.h
```diff
@@ -0,0 +1,143 @@
...
+/// on the data files added via AddFile(). All existing data files in each
+/// touched partition are marked DELETED, and the new files are written as the
+/// sole data in those partitions. Partitions not referenced by any added file
+/// are left unchanged.
+///
+/// This operation produces a snapshot with operation="overwrite" and
+/// "replace-partitions"="true" in the summary. For unpartitioned tables, all
+/// existing files are replaced.
+///
+/// When committing, these changes are applied to the latest table snapshot.
```
wgtmac: Please include Java's default idempotent-validation contract here: when neither conflict validation is enabled, the replace is reapplied and committed despite concurrent table changes. This is part of the public API.

src/iceberg/update/replace_partitions.h
```diff
@@ -0,0 +1,143 @@
...
+  /// \brief Add a data file and mark its partition for replacement.
+  ///
+  /// Each call registers the file's partition so all existing data files in
+  /// that partition are replaced. Duplicate files (same path) are ignored.
+  ///
+  /// \param file The data file to add (must have partition_spec_id set)
+  /// \return Reference to this for method chaining
+  ReplacePartitions& AddFile(const std::shared_ptr<DataFile>& file);
+
+  /// \brief Fail the commit if any existing data file would be deleted.
```
wgtmac: `ValidateAppendOnly()` fails if any partition would be replaced, not only when an existing data file is deleted; `FailAnyDelete()` also checks delete manifests. Please use Java's contract wording here.

src/iceberg/test/replace_partitions_test.cc
```diff
@@ -0,0 +1,328 @@
...
+  CommitFastAppend(file_a_);
+  CommitFastAppend(file_b_);
+
+  ICEBERG_UNWRAP_OR_FAIL(auto op, NewReplace());
+  op->AddFile(MakeDataFile("/data/file_a_new.parquet", /*partition_x=*/1L));
+  EXPECT_THAT(op->Commit(), IsOk());
+
+  EXPECT_THAT(table_->Refresh(), IsOk());
+  ICEBERG_UNWRAP_OR_FAIL(auto snapshot, table_->current_snapshot());
+  EXPECT_EQ(snapshot->summary.at(SnapshotSummaryFields::kDeletedDataFiles), "1");
```
wgtmac, line 283: Summary counts do not prove that the other partition was preserved; deleting `file_b_` and leaving `file_a_` would produce the same counts. Assert the live file paths or manifest entries include `file_b_` and the replacement and exclude `file_a_`.

src/iceberg/test/replace_partitions_test.cc
```diff
@@ -0,0 +1,328 @@
...
+// the error is translated into Java's partition-conflict wording.
+TEST_F(ReplacePartitionsTest, AppendOnlyConflictTranslated) {
+  CommitFastAppend(file_a_);
+
+  ICEBERG_UNWRAP_OR_FAIL(auto op, NewReplace());
+  op->AddFile(MakeDataFile("/data/file_a_new.parquet", /*partition_x=*/1L));
+  op->ValidateAppendOnly();
+  auto result = op->Commit();
+  EXPECT_THAT(result, IsError(ErrorKind::kValidationFailed));
+  EXPECT_THAT(result, HasErrorMessage(
```
wgtmac: Please assert the full translated message, including the partition path. This substring would pass if `Apply()` dropped or corrupted the value extracted from the base error.

src/iceberg/test/replace_partitions_test.cc
```diff
@@ -0,0 +1,328 @@
...
+}
+
+// A concurrent append in the replaced partition is a conflict.
+TEST_F(ReplacePartitionsTest, NoConflictingDataSamePartitionFails) {
+  const int64_t first_id = CommitFastAppend(file_a_);
+  CommitFastAppend(MakeDataFile("/data/concurrent_x1.parquet", /*partition_x=*/1L));
+
+  ICEBERG_UNWRAP_OR_FAIL(auto op, NewReplace());
+  op->AddFile(MakeDataFile("/data/replacement_x1.parquet", /*partition_x=*/1L));
+  op->ValidateFromSnapshot(first_id);
```
wgtmac, line 378: Please add the no-validation counterpart from Java: the same-partition concurrent replace must commit when only `ValidateFromSnapshot()` is set. This locks the API's default idempotent behavior, which the current conflict tests do not cover.

src/iceberg/update/replace_partitions.cc
shangxinli: Done — trimmed the Java comparisons and history out of Apply(), Validate() and the header, kept just the rule.

src/iceberg/update/replace_partitions.cc
shangxinli: Added them: AllVoidSpecReplacesWholeTable (distinct all-void spec), UnpartitionedNoConflictingDataFails (conflict validation on the unpartitioned/row-filter path), NoConflictingDeletesConcurrentDataRemovalFails (concurrent data-file removal, not just a new delete file), and ReplacePartitionsV3Test.ReplaceCleansUpDeletionVectors (format-v3 DV cleanup).

src/iceberg/update/replace_partitions.h
shangxinli: Added the default idempotent behavior to the class doc: with neither conflict validation enabled the replace is reapplied and committed despite concurrent changes.

src/iceberg/update/replace_partitions.h
shangxinli: Reworded to Java's contract: fails if any partition would be replaced (any existing data file deleted) or if a delete file already exists in those partitions, not only on data-file deletion.

src/iceberg/test/replace_partitions_test.cc
shangxinli: Done — it now asserts the live file paths are exactly file_b and the replacement, with file_a gone, instead of relying on the counts.

src/iceberg/test/replace_partitions_test.cc
shangxinli: Done — now asserts the full message including the partition path: "Cannot commit file that conflicts with existing partition: x=1".

src/iceberg/test/replace_partitions_test.cc
shangxinli: Added NoValidationSamePartitionConcurrentReplaceCommits: a same-partition concurrent replace commits when only ValidateFromSnapshot() is set.

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,168 @@
...
+  ICEBERG_BUILDER_CHECK(file != nullptr, "Invalid data file: null");
+  ICEBERG_BUILDER_CHECK(file->partition_spec_id.has_value(),
+                        "Data file must have partition spec ID");
+
+  int32_t spec_id = file->partition_spec_id.value();
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN(auto spec, base().PartitionSpecById(spec_id));
+
+  ICEBERG_BUILDER_RETURN_IF_ERROR(AddDataFile(file));
+  // Specs with no non-void transforms are unpartitioned.
+  auto is_unpartitioned = [](const PartitionSpec& s) {
```
zhjwpku: nit: I think it's no harm to make this a method of PartitionSpec?

src/iceberg/update/replace_partitions.cc
```diff
@@ -0,0 +1,168 @@
...
+}
+
+ReplacePartitions::ReplacePartitions(std::string table_name,
+                                     std::shared_ptr<TransactionContext> ctx)
+    : MergingSnapshotUpdate(std::move(table_name), std::move(ctx)) {
+  SetSummaryProperty(SnapshotSummaryFields::kReplacePartitions, "true");
+}
+
+ReplacePartitions& ReplacePartitions::AddFile(const std::shared_ptr<DataFile>& file) {
+  ICEBERG_BUILDER_CHECK(file != nullptr, "Invalid data file: null");
```
zhjwpku, line 50: Should we check the file-content to be kData, since our DataFile can be kPositionDeletes/kEqualityDeletes while java impl's DataFile is for data only?

zhjwpku (approved):
Generally LGTM, except for two minor nits from me.

wgtmac (approved):
Thanks @shangxinli for working on this! I took the liberty to address open comments and align with the Java behavior as much as possible.

PR #777, opened 2026-06-24
feat(puffin): support deletion-vector-v1 blob read/write

zhaoxuan1994:
feat(puffin): support deletion-vector-v1 blob read/write

Add end-to-end support for the `deletion-vector-v1` Puffin blob type on top
of the existing PuffinReader/PuffinWriter and RoaringPositionBitmap.

Encoding (puffin/deletion_vector.{h,cc}):
- SerializeDeletionVectorBlob / DeserializeDeletionVectorBlob implement the
  DV blob framing: 4-byte big-endian length, 0xD1D33964 magic, the portable
  Roaring vector, and a trailing big-endian CRC-32. Reads validate the magic
  and checksum.
- MakeDeletionVectorBlob builds a spec-compliant Blob: fields set to the
  row-position metadata column id, snapshot-id/sequence-number = -1, no
  compression, and the required referenced-data-file/cardinality properties.

Writing (data/deletion_vector_writer.{h,cc}):
- DeletionVectorWriter accumulates deleted positions per data file, run-length
  encodes each bitmap before serializing, writes one blob per data file, and
  produces a DataFile per blob carrying content_offset/content_size_in_bytes,
  referenced_data_file and record_count for manifest registration.

Reading (data/delete_loader.cc):
- Implement DeleteLoader::LoadDV: read the blob bytes referenced by
  content_offset/content_size_in_bytes, validate referenced_data_file, the 2GB
  limit and cardinality == record_count, then apply positions to the index.
  This wires DV deletes into DeleteFilter::ComputeAliveRows.

Puffin writer convenience:
- PuffinWriter::Make now fills a default `created-by` property when the caller
  does not provide one, matching the Java writer.

Behavior matches the Java implementation (BaseDVFileWriter / BaseDeleteLoader /
BitmapPositionDeleteIndex): magic, byte order, CRC coverage, blob fields, RLE
and the read-side validations are all aligned.

Tests:
- DV blob framing round-trip and error cases (bad magic, corrupted CRC,
  truncated blob, size mismatch).
- DeletionVectorWriter -> DeleteLoader write/read round trip and guard cases.
- delete_loader: load DV, referenced-file filtering, cardinality mismatch,
  mixed DV + position-delete loading.
- delete_filter: end-to-end ComputeAliveRows filtering with a real Puffin DV
  file over Arrow FileIO.
- PuffinWriter created-by default and caller-precedence.

Not included (deletion-vector upper layer, follow-up): merging with previously
written DVs (compaction), the bulk delete(PositionDeleteIndex) API, and
per-data-file partition/spec in a single DV file.

### DIFF
--- a/src/iceberg/data/delete_loader.cc
+++ b/src/iceberg/data/delete_loader.cc
@@ -31,6 +31,7 @@
 
 #include "iceberg/arrow/nanoarrow_status_internal.h"
 #include "iceberg/arrow_c_data_guard_internal.h"
+#include "iceberg/data/dv_util_internal.h"
 #include "iceberg/deletes/position_delete_index.h"
 #include "iceberg/deletes/position_delete_range_consumer.h"
 #include "iceberg/file_io.h"
@@ -88,17 +89,31 @@ bool StringEquals(const ArrowArrayView* view, int64_t row_idx, std::string_view 
   return sv.data != nullptr && std::memcmp(sv.data, target.data(), target.size()) == 0;
 }
 
+Status ValidateDV(const DataFile& dv, std::string_view data_file_path) {
+  ICEBERG_PRECHECK(dv.content_offset.has_value(), "Invalid DV, offset cannot be null: {}",
+                   ContentFileUtil::DVDesc(dv));
+  ICEBERG_PRECHECK(dv.content_size_in_bytes.has_value(), "Invalid DV, length is null: {}",
+                   ContentFileUtil::DVDesc(dv));
+  ICEBERG_PRECHECK(
+      dv.content_size_in_bytes.value() <= std::numeric_limits<int32_t>::max(),
+      "Can't read DV larger than 2GB: {}", dv.content_size_in_bytes.value());
+  ICEBERG_PRECHECK(dv.referenced_data_file.has_value() &&
+                       dv.referenced_data_file.value() == data_file_path,
+                   "DV is expected to reference {}, not {}", data_file_path,
+                   dv.referenced_data_file.value_or(""));
+  return {};
+}
+
 }  // namespace
 
 DeleteLoader::DeleteLoader(std::shared_ptr<FileIO> io) : io_(std::move(io)) {}
 
 DeleteLoader::~DeleteLoader() = default;
 
-Status DeleteLoader::LoadPositionDelete(const std::shared_ptr<DataFile>& file,
-                                        PositionDeleteIndex& index,
+Status DeleteLoader::LoadPositionDelete(const DataFile& file, PositionDeleteIndex& index,
                                         std::string_view data_file_path) const {
-  // TODO(gangwu): push down path filter to open the file.
-  ICEBERG_ASSIGN_OR_RAISE(auto reader, OpenDeleteFile(*file, PosDeleteSchema(), io_));
+  // TODO(gangwu): Add cache hooks, worker pool, and filter pushdown.
+  ICEBERG_ASSIGN_OR_RAISE(auto reader, OpenDeleteFile(file, PosDeleteSchema(), io_));
 
   ICEBERG_ASSIGN_OR_RAISE(auto arrow_schema, reader->Schema());
   internal::ArrowSchemaGuard schema_guard(&arrow_schema);
@@ -116,18 +131,14 @@ Status DeleteLoader::LoadPositionDelete(const std::shared_ptr<DataFile>& file,
   // `ForEachPositionDelete`. Trusts the hint -- spec-compliant writers
   // only set it when all rows share one data file.
   const bool use_referenced_data_file_fast_path =
-      file->referenced_data_file.has_value() &&
-      file->referenced_data_file.value() == data_file_path;
+      file.referenced_data_file.has_value() &&
+      file.referenced_data_file.value() == data_file_path;
 
   // Filter-path staging buffer; reused across batches via `clear()`.
   std::vector<int64_t> positions;
   // Scratch buffer for `ForEachPositionDelete`'s bulk dispatch path;
   // reused across batches and across both routing branches.
   std::vector<uint32_t> bulk_scratch;
-
-  // Whether any position for the target data file came from this file, so the
-  // source delete file is recorded once.
-  bool contributed = false;
 
   while (true) {
     ICEBERG_ASSIGN_OR_RAISE(auto batch_opt, reader->Next());
@@ -160,9 +171,6 @@ Status DeleteLoader::LoadPositionDelete(const std::shared_ptr<DataFile>& file,
     const int64_t* pos_data = Int64ValuesBuffer(pos_view);
 
     if (use_referenced_data_file_fast_path) {
-      // Reaching here means length > 0 (empty batches are skipped above), so
-      // this file contributes positions for the target data file.
-      contributed = true;
       ForEachPositionDelete(std::span<const int64_t>(pos_data, length), index,
                             bulk_scratch);
       continue;
@@ -177,67 +185,10 @@ Status DeleteLoader::LoadPositionDelete(const std::shared_ptr<DataFile>& file,
         positions.push_back(pos_data[i]);
       }
     }
-    if (!positions.empty()) {
-      contributed = true;
-    }
     ForEachPositionDelete(positions, index, bulk_scratch);
   }
 
-  ICEBERG_RETURN_UNEXPECTED(reader->Close());
-
-  // Record the source delete file so callers (e.g. DeletionVectorWriter) can
-  // report file-scoped position deletes as rewritten.
-  if (contributed) {
-    index.AddDeleteFile(file);
-  }
-  return {};
-}
-
-Status DeleteLoader::LoadDV(const std::shared_ptr<DataFile>& file,
-                            PositionDeleteIndex& index,
-                            std::string_view data_file_path) const {
-  // A deletion vector must reference exactly one data file; without it the
-  // caller cannot know which data file the positions apply to.
-  ICEBERG_PRECHECK(file->referenced_data_file.has_value(),
-                   "Deletion vector requires referenced_data_file: {}", file->file_path);
-
-  // The DV must reference the data file being read. A mismatch means the caller
-  // grouped delete files incorrectly; failing here avoids silently applying the
-  // wrong DV or hiding bad metadata behind an empty index.
-  ICEBERG_PRECHECK(file->referenced_data_file.value() == data_file_path,
-                   "Deletion vector references {}, not the requested data file {}",
-                   file->referenced_data_file.value(), data_file_path);
-
-  // For deletion vectors, content_offset and content_size_in_bytes point directly
-  // at the DV blob bytes within the Puffin file and are required by the spec.
-  ICEBERG_PRECHECK(
-      file->content_offset.has_value() && file->content_size_in_bytes.has_value(),
-      "Deletion vector requires content_offset and content_size_in_bytes: {}",
-      file->file_path);
-
-  const int64_t offset = file->content_offset.value();
-  const int64_t length = file->content_size_in_bytes.value();
-  ICEBERG_PRECHECK(offset >= 0 && length >= 0,
-                   "Invalid deletion vector offset/length: offset={}, length={}", offset,
-                   length);
-  ICEBERG_PRECHECK(length <= std::numeric_limits<int32_t>::max(),
-                   "Cannot read deletion vector larger than 2GB: {}", length);
-
-  ICEBERG_ASSIGN_OR_RAISE(auto input_file, io_->NewInputFile(file->file_path));
-  ICEBERG_ASSIGN_OR_RAISE(auto stream, input_file->Open());
-
-  std::vector<std::byte> bytes(static_cast<size_t>(length));
-  ICEBERG_RETURN_UNEXPECTED(stream->ReadFully(offset, bytes));
-  ICEBERG_RETURN_UNEXPECTED(stream->Close());
-
-  std::span<const uint8_t> blob(reinterpret_cast<const uint8_t*>(bytes.data()),
-                                bytes.size());
-  // Deserialize validates the blob length and cardinality against `file` and
-  // retains it as the source delete file.
-  ICEBERG_ASSIGN_OR_RAISE(auto dv, PositionDeleteIndex::Deserialize(blob, file));
-
-  index.Merge(dv);
-  return {};
+  return reader->Close();
 }
 
 Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
@@ -247,19 +198,14 @@ Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
     ICEBERG_PRECHECK(file != nullptr, "Delete file must not be null");
   }
 
-  PositionDeleteIndex index;
-
-  // A single deletion vector replaces all other deletes for a data file, so it
-  // is read through the dedicated DV path and validated against the requested
-  // data file.
   if (ContentFileUtil::ContainsSingleDV(delete_files)) {
-    ICEBERG_RETURN_UNEXPECTED(LoadDV(delete_files.front(), index, data_file_path));
-    return index;
+    const auto& dv = delete_files.front();
+    ICEBERG_RETURN_UNEXPECTED(ValidateDV(*dv, data_file_path));
+    return DVUtil::ReadDV(dv, io_);
   }
 
-  // Otherwise all entries must be position delete files. A DV must not be mixed
-  // with position deletes: once a DV applies, readers ignore matching position
-  // deletes, so a mixed list means the caller grouped delete files incorrectly.
+  std::vector<std::shared_ptr<DataFile>> position_delete_files;
+
   for (const auto& file : delete_files) {
     ICEBERG_PRECHECK(!file->IsDeletionVector(),
                      "Deletion vector cannot be mixed with position delete files: {}",
@@ -268,14 +214,18 @@ Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
                      "Expected position delete file but got content type {}",
                      ToString(file->content));
 
-    // A file-scoped position delete for another data file has nothing for the
-    // target path; skip it as an optimization.
     if (file->referenced_data_file.has_value() &&
         file->referenced_data_file.value() != data_file_path) {
       continue;
     }
 
-    ICEBERG_RETURN_UNEXPECTED(LoadPositionDelete(file, index, data_file_path));
+    position_delete_files.push_back(file);
+  }
+
+  PositionDeleteIndex index(position_delete_files);
+
+  for (const auto& file : position_delete_files) {
+    ICEBERG_RETURN_UNEXPECTED(LoadPositionDelete(*file, index, data_file_path));
   }
 
   return index;
--- a/src/iceberg/data/delete_loader.cc
+++ b/src/iceberg/data/delete_loader.cc
@@ -19,9 +19,12 @@
 
 #include "iceberg/data/delete_loader.h"
 
+#include <cstdint>
 #include <cstring>
+#include <limits>
 #include <span>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include <nanoarrow/nanoarrow.h>
@@ -30,12 +33,14 @@
 #include "iceberg/arrow_c_data_guard_internal.h"
 #include "iceberg/deletes/position_delete_index.h"
 #include "iceberg/deletes/position_delete_range_consumer.h"
+#include "iceberg/file_io.h"
 #include "iceberg/file_reader.h"
 #include "iceberg/manifest/manifest_entry.h"
 #include "iceberg/metadata_columns.h"
 #include "iceberg/result.h"
 #include "iceberg/row/arrow_array_wrapper.h"
 #include "iceberg/schema.h"
+#include "iceberg/util/content_file_util.h"
 #include "iceberg/util/macros.h"
 #include "iceberg/util/struct_like_set.h"
 
@@ -89,10 +94,11 @@ DeleteLoader::DeleteLoader(std::shared_ptr<FileIO> io) : io_(std::move(io)) {}
 
 DeleteLoader::~DeleteLoader() = default;
 
-Status DeleteLoader::LoadPositionDelete(const DataFile& file, PositionDeleteIndex& index,
+Status DeleteLoader::LoadPositionDelete(const std::shared_ptr<DataFile>& file,
+                                        PositionDeleteIndex& index,
                                         std::string_view data_file_path) const {
   // TODO(gangwu): push down path filter to open the file.
-  ICEBERG_ASSIGN_OR_RAISE(auto reader, OpenDeleteFile(file, PosDeleteSchema(), io_));
+  ICEBERG_ASSIGN_OR_RAISE(auto reader, OpenDeleteFile(*file, PosDeleteSchema(), io_));
 
   ICEBERG_ASSIGN_OR_RAISE(auto arrow_schema, reader->Schema());
   internal::ArrowSchemaGuard schema_guard(&arrow_schema);
@@ -110,14 +116,18 @@ Status DeleteLoader::LoadPositionDelete(const DataFile& file, PositionDeleteInde
   // `ForEachPositionDelete`. Trusts the hint -- spec-compliant writers
   // only set it when all rows share one data file.
   const bool use_referenced_data_file_fast_path =
-      file.referenced_data_file.has_value() &&
-      file.referenced_data_file.value() == data_file_path;
+      file->referenced_data_file.has_value() &&
+      file->referenced_data_file.value() == data_file_path;
 
   // Filter-path staging buffer; reused across batches via `clear()`.
   std::vector<int64_t> positions;
   // Scratch buffer for `ForEachPositionDelete`'s bulk dispatch path;
   // reused across batches and across both routing branches.
   std::vector<uint32_t> bulk_scratch;
+
+  // Whether any position for the target data file came from this file, so the
+  // source delete file is recorded once.
+  bool contributed = false;
 
   while (true) {
     ICEBERG_ASSIGN_OR_RAISE(auto batch_opt, reader->Next());
@@ -150,6 +160,9 @@ Status DeleteLoader::LoadPositionDelete(const DataFile& file, PositionDeleteInde
     const int64_t* pos_data = Int64ValuesBuffer(pos_view);
 
     if (use_referenced_data_file_fast_path) {
+      // Reaching here means length > 0 (empty batches are skipped above), so
+      // this file contributes positions for the target data file.
+      contributed = true;
       ForEachPositionDelete(std::span<const int64_t>(pos_data, length), index,
                             bulk_scratch);
       continue;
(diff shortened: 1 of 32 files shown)


src/iceberg/data/deletion_vector_writer.cc
```diff
@@ -0,0 +1,157 @@
...
+    // One blob per referenced data file, in deterministic (sorted) order.
+    struct Entry {
+      std::string referenced_data_file;
+      int64_t offset;
+      int64_t length;
+      int64_t cardinality;
+    };
+    std::vector<Entry> entries;
+    entries.reserve(bitmaps_.size());
+    for (auto& [referenced_data_file, bitmap] : bitmaps_) {
```
wgtmac: This should merge existing deletes before writing a DV for a path. The spec requires at most one DV per data file, and Java BaseDVFileWriter loads previous DVs/position deletes and reports rewritten delete files. As-is, adding another DV can leave stale delete files in table state.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,92 @@
...
+/// \brief Options for creating a DeletionVectorWriter.
+struct ICEBERG_DATA_EXPORT DeletionVectorWriterOptions {
+  /// Output Puffin file location.
+  std::string path;
+  /// FileIO used to create the Puffin file.
+  std::shared_ptr<FileIO> io;
+  /// Partition spec the referenced data files belong to (optional).
+  std::shared_ptr<PartitionSpec> spec;
+  /// Partition the referenced data files belong to.
+  PartitionValues partition;
```
wgtmac: Please store the partition spec and partition per referenced data file, not per writer. Java DVFileWriter passes them on each delete call, and the spec allows one Puffin file to hold DVs for files from different partitions. With this API, one writer can emit the wrong delete-file metadata.

src/iceberg/puffin/puffin_writer.cc
```diff
@@ -46,6 +48,14 @@ Result<std::unique_ptr<PuffinWriter>> PuffinWriter::Make(
     PuffinCompressionCodec default_codec, bool compress_footer) {
   ICEBERG_PRECHECK(output_file, "Output file must not be null");
   ICEBERG_ASSIGN_OR_RAISE(auto stream, output_file->Create());
+  // Identify the writer in the footer unless the caller already set it. Only
+  // format the default value when it is actually needed.
+  const std::string created_by_key(StandardPuffinProperties::kCreatedBy);
+  if (!properties.contains(created_by_key)) {
```
wgtmac: Please keep this default in the DV writer path instead of the generic PuffinWriter. Java does not add created-by in Puffin.write(...).build(); BaseDVFileWriter sets it when creating DV files. This changes footer metadata for all Puffin writers.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,92 @@
...
+  DeletionVectorWriter& operator=(const DeletionVectorWriter&) = delete;
+
+  /// \brief Create a new DeletionVectorWriter.
+  static Result<std::unique_ptr<DeletionVectorWriter>> Make(
+      DeletionVectorWriterOptions options);
+
+  /// \brief Mark a row position as deleted for the given data file.
+  ///
+  /// Positions are accumulated per data file and serialized on Close().
+  Status Delete(std::string_view referenced_data_file, int64_t pos);
```
wgtmac: Please make this match Java DVFileWriter more closely. Delete should take the data file spec and partition, and there should be an overload that accepts a PositionDeleteIndex. Otherwise the writer cannot represent per-file metadata or bulk deletes.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,92 @@
...
+
+  /// \brief Write all accumulated deletion vectors to the Puffin file and close it.
+  Status Close();
+
+  /// \brief Metadata for the DataFiles produced (one per referenced data file).
+  /// \note Must be called after Close().
+  Result<FileWriter::WriteResult> Metadata();
```
wgtmac: Please return a delete-specific result here, not FileWriter::WriteResult. Java returns DeleteWriteResult with delete files, referenced data files, and rewritten delete files. Row delta needs those sets for conflict checks and cleanup.

src/iceberg/puffin/deletion_vector.h
```diff
@@ -0,0 +1,97 @@
...
+  static constexpr int32_t kMagicBytes = 4;
+  /// Length of the trailing big-endian CRC-32 checksum, in bytes.
+  static constexpr int32_t kCrcBytes = 4;
+};
+
+/// \brief Serialize a position bitmap into a `deletion-vector-v1` blob.
+///
+/// The returned bytes include the length prefix, magic sequence, the Roaring
+/// "portable" serialization of the bitmap, and the trailing CRC-32 checksum.
+ICEBERG_DATA_EXPORT Result<std::vector<uint8_t>> SerializeDeletionVectorBlob(
```
wgtmac: Please avoid exposing DV blob serialization as a public Puffin API. Java keeps this behind PositionDeleteIndex and BitmapPositionDeleteIndex. Exposing RoaringPositionBitmap here makes the bitmap layout part of the public surface.

src/iceberg/data/deletion_vector_writer.cc
```diff
@@ -0,0 +1,157 @@
...
+    ICEBERG_CHECK(closed_, "Cannot get metadata before closing the writer");
+    FileWriter::WriteResult result;
+    result.data_files = data_files_;
+    return result;
+  }
+
+ private:
+  DeletionVectorWriterOptions options_;
+  // Ordered by referenced data file path for deterministic blob layout.
+  std::map<std::string, RoaringPositionBitmap> bitmaps_;
```
wgtmac: Please model the private state like Java Deletes: path, positions, spec, and partition. A map from path to bitmap loses per-file metadata, so the wrong manifest metadata can be produced.

src/iceberg/puffin/puffin_writer.cc
wgtmac: Please do not set created-by in the generic PuffinWriter. Java only sets it in BaseDVFileWriter when writing DVs. Keeping it here changes footer metadata for every Puffin writer.

src/iceberg/puffin/deletion_vector.h
```diff
@@ -0,0 +1,97 @@
...
+struct StandardDeletionVectorProperties {
+  /// Location of the data file the deletion vector applies to.
+  static constexpr std::string_view kReferencedDataFile = "referenced-data-file";
+  /// Number of deleted rows (set positions) in the deletion vector.
+  static constexpr std::string_view kCardinality = "cardinality";
+};
+
+/// \brief Constants describing the `deletion-vector-v1` blob framing.
+///
+/// The serialized blob has the following layout (see the Puffin spec):
```
wgtmac: This is too much spec text for a public header. Please keep the header brief and point to the Puffin spec. The byte layout is better covered in the implementation or tests.

src/iceberg/data/deletion_vector_writer.cc
```diff
@@ -0,0 +1,157 @@
...
+      std::string referenced_data_file;
+      int64_t offset;
+      int64_t length;
+      int64_t cardinality;
+    };
+    std::vector<Entry> entries;
+    entries.reserve(bitmaps_.size());
+    for (auto& [referenced_data_file, bitmap] : bitmaps_) {
+      // Run-length encode before serializing for space efficiency, matching the
+      // Java DV writer.
```
wgtmac: Please avoid comments that justify the code by saying it matches Java. A short local comment is enough, or the test can cover the parity. This reads like review explanation.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/// \brief Writes one or more deletion vectors into a single Puffin file.
+///
+/// Each referenced data file gets its own `deletion-vector-v1` blob. After
+/// Close(), Metadata() returns one DataFile per blob, each carrying the
+/// content_offset/content_size_in_bytes and referenced_data_file required to
+/// register the deletion vector in a manifest.
+///
+/// \note All referenced data files are assumed to belong to the single
```
wgtmac: Please remove this note once the API is fixed. It documents a limitation that Java does not have and that the spec does not require.

src/iceberg/puffin/deletion_vector.cc
```diff
@@ -0,0 +1,173 @@
...
+  T value;
+  std::memcpy(&value, buf, sizeof(value));
+  return FromBigEndian(value);
+}
+
+}  // namespace
+
+Result<std::vector<uint8_t>> SerializeDeletionVectorBlob(
+    const RoaringPositionBitmap& bitmap) {
+  ICEBERG_ASSIGN_OR_RAISE(auto serialized, bitmap.Serialize());
```
wgtmac: If this helper stays public, please run-length encode before serializing here, not only in DeletionVectorWriter. Java PositionDeleteIndex.serialize() does it on every serialization path. Direct calls to this helper currently skip that step.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,92 @@
...
+
+  /// \brief Metadata for the DataFiles produced (one per referenced data file).
+  /// \note Must be called after Close().
+  Result<FileWriter::WriteResult> Metadata();
+
+ private:
+  class Impl;
+  std::unique_ptr<Impl> impl_;
+
+  explicit DeletionVectorWriter(std::unique_ptr<Impl> impl);
```
wgtmac, line 80: Please return a delete-specific result here. Java returns DeleteWriteResult with delete files, referenced data files, and rewritten delete files. RowDelta needs those sets for validation and cleanup.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,92 @@
...
+  /// \brief Mark a row position as deleted for the given data file.
+  ///
+  /// Positions are accumulated per data file and serialized on Close().
+  Status Delete(std::string_view referenced_data_file, int64_t pos);
+
+  /// \brief Write all accumulated deletion vectors to the Puffin file and close it.
+  Status Close();
+
+  /// \brief Metadata for the DataFiles produced (one per referenced data file).
+  /// \note Must be called after Close().
```
wgtmac: Please align this with Java DVFileWriter. Delete should take the data file spec and partition, and there should be an overload that accepts a PositionDeleteIndex. A writer-level partition cannot model DVs for files in different partitions.

src/iceberg/data/deletion_vector_writer.cc
```diff
@@ -0,0 +1,163 @@
...
+    ICEBERG_CHECK(closed_, "Cannot get metadata before closing the writer");
+    FileWriter::WriteResult result;
+    result.data_files = data_files_;
+    return result;
+  }
+
+ private:
+  DeletionVectorWriterOptions options_;
+  // Ordered by referenced data file path for deterministic blob layout.
+  std::map<std::string, RoaringPositionBitmap> bitmaps_;
```
wgtmac: Please store a Deletes object per path, not just a bitmap. Java keeps path, positions, spec, and partition together. This map loses the per-file metadata needed to build correct delete-file entries.

src/iceberg/data/deletion_vector_writer.cc
```diff
@@ -0,0 +1,163 @@
...
+    // One blob per referenced data file, in deterministic (sorted) order.
+    struct Entry {
+      std::string referenced_data_file;
+      int64_t offset;
+      int64_t length;
+      int64_t cardinality;
+    };
+    std::vector<Entry> entries;
+    entries.reserve(bitmaps_.size());
+    for (auto& [referenced_data_file, bitmap] : bitmaps_) {
```
wgtmac: Please merge existing deletes before writing each DV. The spec allows only one DV per data file in a snapshot, and Java BaseDVFileWriter loads previous DVs or file-scoped position deletes before writing the replacement.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,92 @@
...
+#include "iceberg/data/writer.h"
+#include "iceberg/iceberg_data_export.h"
+#include "iceberg/result.h"
+#include "iceberg/row/partition_values.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief Options for creating a DeletionVectorWriter.
+struct ICEBERG_DATA_EXPORT DeletionVectorWriterOptions {
```
wgtmac, line 42: Please add a previous-delete loader to these options. Java BaseDVFileWriter takes Function<String, PositionDeleteIndex>; without that callback, Close cannot merge old DVs or file-scoped position deletes before writing the replacement.

src/iceberg/data/delete_loader.cc
```diff
@@ -171,7 +177,44 @@ Status DeleteLoader::LoadPositionDelete(const DataFile& file, PositionDeleteInde
...
+  ICEBERG_ASSIGN_OR_RAISE(auto input_file, io_->NewInputFile(file.file_path));
+  ICEBERG_ASSIGN_OR_RAISE(auto stream, input_file->Open());
+
+  std::vector<std::byte> bytes(static_cast<size_t>(length));
+  ICEBERG_RETURN_UNEXPECTED(stream->ReadFully(offset, bytes));
+  ICEBERG_RETURN_UNEXPECTED(stream->Close());
+
+  std::span<const uint8_t> blob(reinterpret_cast<const uint8_t*>(bytes.data()),
+                                bytes.size());
+  ICEBERG_ASSIGN_OR_RAISE(auto bitmap, puffin::DeserializeDeletionVectorBlob(blob));
```
wgtmac: Please keep the source delete file with the loaded index. Java PositionDeleteIndex.deserialize(bytes, deleteFile) preserves it, and BaseDVFileWriter uses deleteFiles() to fill rewrittenDeleteFiles after merging old deletes.

src/iceberg/puffin/deletion_vector.cc
```diff
@@ -0,0 +1,173 @@
...
+Result<Blob> MakeDeletionVectorBlob(const RoaringPositionBitmap& bitmap,
+                                    std::string referenced_data_file) {
+  ICEBERG_PRECHECK(!referenced_data_file.empty(),
+                   "Deletion vector requires a non-empty referenced data file");
+
+  ICEBERG_ASSIGN_OR_RAISE(auto data, SerializeDeletionVectorBlob(bitmap));
+
+  Blob blob{
+      .type = std::string(StandardBlobTypes::kDeletionVectorV1),
+      // The deletion vector is computed over row positions, matching the Java
```
wgtmac: Please drop the Java reference from this comment. A shorter local comment like "input_fields records the row-position metadata column" is enough here.

src/iceberg/data/deletion_vector_writer.cc
```diff
@@ -0,0 +1,238 @@
...
+    }
+    return it->second;
+  }
+
+  Status Delete(std::string_view referenced_data_file, int64_t pos,
+                std::shared_ptr<PartitionSpec> spec, PartitionValues partition) {
+    ICEBERG_CHECK(!closed_, "Cannot delete after the writer is closed");
+    ICEBERG_PRECHECK(!referenced_data_file.empty(),
+                     "Deletion vector requires a non-empty referenced data file");
+    DeletesFor(referenced_data_file, std::move(spec), std::move(partition))
```
wgtmac: Please validate `pos` before adding it to the index. `PositionDeleteIndex::Delete` still silently ignores out-of-range values, while Java fails in `RoaringPositionBitmap.set`. As-is, `Delete(..., -1, ...)` returns OK and writes a DV without that delete.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,114 @@
...
+  /// Data files referenced by the produced deletion vectors.
+  std::vector<std::string> referenced_data_files;
+  /// Previously written, file-scoped delete files that were merged and should be
+  /// removed from table state.
+  std::vector<std::shared_ptr<DataFile>> rewritten_delete_files;
+};
+
+/// \brief Previously written deletes for a data file, used to merge before
+/// writing a new deletion vector.
+struct ICEBERG_DATA_EXPORT PreviousDeletes {
```
wgtmac: Please keep source delete files on `PositionDeleteIndex`, like Java `deleteFiles()`, and let the previous-delete hook return the index directly. This wrapper keeps the writer API different from `BaseDVFileWriter`.

src/iceberg/deletes/position_delete_index.h
```diff
@@ -66,7 +67,20 @@ class ICEBERG_DATA_EXPORT PositionDeleteIndex {
...
+  /// \brief Serialize the index into a `deletion-vector-v1` blob.
+  ///
+  /// The positions are run-length encoded, then framed per the Puffin spec:
+  /// https://iceberg.apache.org/puffin-spec/#deletion-vector-v1-blob-type
+  Result<std::vector<uint8_t>> Serialize();
+
+  /// \brief Deserialize a `deletion-vector-v1` blob into an index.
+  ///
+  /// Validates the length prefix, magic sequence, and CRC-32 checksum.
+  static Result<PositionDeleteIndex> Deserialize(std::span<const uint8_t> blob);
```
wgtmac: Please pass delete-file metadata to `Deserialize`, like Java `PositionDeleteIndex.deserialize(bytes, deleteFile)`. Deserialization should validate `content_size_in_bytes` and `record_count`, so public callers cannot bypass those DV checks.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,114 @@
...
+#include "iceberg/deletes/position_delete_index.h"
+#include "iceberg/iceberg_data_export.h"
+#include "iceberg/result.h"
+#include "iceberg/row/partition_values.h"
+#include "iceberg/type_fwd.h"
+
+namespace iceberg {
+
+/// \brief The result of writing deletion vectors.
+struct ICEBERG_DATA_EXPORT DeleteWriteResult {
```
wgtmac: Please move this to a shared delete-writer API instead of the DV writer header. Java uses DeleteWriteResult across position, equality, and DV writers; keeping it here makes the public API DV-specific.

src/iceberg/data/deletion_vector_writer.h
```diff
@@ -0,0 +1,114 @@
...
+  /// Output Puffin file location.
+  std::string path;
+  /// FileIO used to create the Puffin file.
+  std::shared_ptr<FileIO> io;
+  /// File-level Puffin properties (e.g. "created-by").
+  std::unordered_map<std::string, std::string> properties;
+  /// Optional hook to load previously written deletes for a data file so they
+  /// can be merged into the new deletion vector, with the old delete files
+  /// reported as rewritten.
+  std::function<Result<PreviousDeletes>(std::string_view)> load_previous_deletes;
```
wgtmac: Please make this loader explicit instead of optional. Java BaseDVFileWriter always receives loadPreviousDeletes, even when it returns null. A default no-merge path makes it easy to write a DV that does not replace existing deletes.

src/iceberg/test/deletion_vector_writer_test.cc
```diff
@@ -0,0 +1,303 @@
...
+                             .path = "memory://empty.puffin", .io = io}));
+  ASSERT_THAT(writer->Close(), IsOk());
+  ICEBERG_UNWRAP_OR_FAIL(auto result, writer->Metadata());
+  EXPECT_TRUE(result.delete_files.empty());
+
+  // No blobs were written, so no (orphan) Puffin file should have been created.
+  EXPECT_THAT(io->NewInputFile("memory://empty.puffin"), IsError(ErrorKind::kNotFound));
+}
+
+TEST(DeletionVectorWriterTest, DeleteRejectsInvalidPosition) {
```
wgtmac: Please rename this test or add real invalid-position cases. It only checks an empty referenced path, while negative and out-of-range positions are the Java-parity cases.

src/iceberg/test/deletion_vector_writer_test.cc
```diff
@@ -0,0 +1,303 @@
...
+      .file_format = FileFormatType::kPuffin,
+      .referenced_data_file = "data.parquet",
+  });
+
+  ICEBERG_UNWRAP_OR_FAIL(
+      auto writer,
+      DeletionVectorWriter::Make(DeletionVectorWriterOptions{
+          .path = "memory://merged.puffin",
+          .io = io,
+          .load_previous_deletes = [&](std::string_view path) -> Result<PreviousDeletes> {
```
wgtmac: Please add a test where load_previous_deletes returns an error. Close should propagate it and should not return partial delete metadata.

src/iceberg/test/position_delete_index_test.cc
```diff
@@ -200,4 +205,89 @@ TEST(PositionDeleteIndexTest, TestMergeIdempotence) {
...
+  PositionDeleteIndex index;
+  index.Delete(1);
+  ICEBERG_UNWRAP_OR_FAIL(auto blob, index.Serialize());
+
+  blob[4] = 0x00;
+  EXPECT_THAT(PositionDeleteIndex::Deserialize(blob),
+              IsError(ErrorKind::kInvalidArgument));
+}
+
+TEST(PositionDeleteIndexTest, DeserializeRejectsTruncatedBlob) {
```
wgtmac, line 313: Please add a blob-size mismatch case, not only a too-short blob. Java validates the length prefix against the DV content size, and this path should stay covered.

src/iceberg/test/delete_loader_test.cc
```diff
@@ -239,15 +256,73 @@ TEST_F(DeleteLoaderTest, LoadPositionDeletesFastPathHonorsReferencedDataFile) {
...
+  auto dv_file = WriteDeletionVector("dv-card.puffin", "data.parquet", {0, 1, 2});
+  // Corrupt the recorded cardinality so it no longer matches the bitmap.
+  dv_file->record_count = 99;
+
+  std::vector<std::shared_ptr<DataFile>> files = {dv_file};
+  auto result = loader_->LoadPositionDeletes(files, "data.parquet");
+  ASSERT_THAT(result, IsError(ErrorKind::kInvalidArgument));
+}
+
+// Iceberg uses either a deletion vector or position deletes for a data file, not
```
wgtmac: Please remove this mixed DV + position delete case or move it to DeleteFileIndex if needed. A real scan should not pass matching position deletes once a DV applies, so this loader test teaches the wrong contract.

src/iceberg/test/deletion_vector_writer_test.cc
```diff
@@ -0,0 +1,303 @@
...
+  return nullptr;
+}
+
+std::shared_ptr<PartitionSpec> UnpartitionedSpec() {
+  return PartitionSpec::Unpartitioned();
+}
+
+}  // namespace
+
+// Full write -> read round trip: write deletion vectors with the writer, then
```
wgtmac: Please trim these narrative comments. The test name and assertions already explain the flow, so this adds noise.

wgtmac:
The PR description is out of date. It still says previous-delete merging and delete(PositionDeleteIndex) are follow-ups, but this version includes both. Please update the description.

src/iceberg/data/delete_loader.cc
```diff
@@ -180,21 +237,23 @@ Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
   PositionDeleteIndex index;
 
   for (const auto& file : delete_files) {
+    ICEBERG_PRECHECK(file != nullptr, "Delete file must not be null");
+
     if (file->referenced_data_file.has_value() &&
```
wgtmac, line 217: Please don't skip a mismatched DV here. Java validates the DV against the requested data file and fails when it references another file. Returning an empty index hides bad metadata.

src/iceberg/data/delete_loader.cc
```diff
@@ -180,21 +237,23 @@ Result<PositionDeleteIndex> DeleteLoader::LoadPositionDeletes(
...
         file->referenced_data_file.value() != data_file_path) {
       continue;
     }
 
     if (file->IsDeletionVector()) {
```
wgtmac: Please don't merge DVs with regular position delete files in this loader path. Java only reads a DV through the single-DV path, and Iceberg readers should ignore matching position deletes once a DV applies.

src/iceberg/deletes/position_delete_index.cc
```diff
@@ -37,6 +84,106 @@ int64_t PositionDeleteIndex::Cardinality() const {
...
+      ComputeCrc32(std::span<const uint8_t>(checksum_begin, magic_and_vector_size)), buf);
+  return blob;
+}
+
+Result<PositionDeleteIndex> PositionDeleteIndex::Deserialize(
+    std::span<const uint8_t> blob, std::shared_ptr<DataFile> delete_file) {
+  ICEBERG_PRECHECK(delete_file != nullptr,
+                   "Deletion vector requires a source delete file");
+  // The blob bytes must match the length recorded in the delete file metadata.
+  if (delete_file->content_size_in_bytes.has_value()) {
```
wgtmac: Please require content_size_in_bytes here instead of treating it as optional. DV metadata requires this field, and deserialization should not allow callers to bypass the size check.

src/iceberg/data/writer.h
```diff
@@ -33,6 +34,20 @@
 
 namespace iceberg {
 
+/// \brief The result of writing delete files.
+///
+/// Shared by the position, equality, and deletion-vector delete writers: the
```
wgtmac: Please either wire this result through the other delete writers or keep the comment scoped to the DV writer. Position and equality delete writers still return FileWriter::WriteResult today.

src/iceberg/test/deletion_vector_writer_test.cc
```diff
@@ -0,0 +1,381 @@
...
+
+  ICEBERG_UNWRAP_OR_FAIL(
+      auto writer,
+      DeletionVectorWriter::Make(DeletionVectorWriterOptions{
+          .path = "memory://merged-partition.puffin",
+          .io = io,
+          .load_previous_deletes =
+              [&](std::string_view path) -> Result<std::shared_ptr<PositionDeleteIndex>> {
+            ICEBERG_ASSIGN_OR_RAISE(
+                auto index, PositionDeleteIndex::Deserialize(previous_blob,
```
wgtmac: Please don't create this previous position-delete index by deserializing a DV blob. This test is meant to model a Parquet position delete, so it should get the index through DeleteLoader or add the source file directly.

src/iceberg/test/delete_loader_test.cc
```diff
@@ -239,15 +302,90 @@ TEST_F(DeleteLoaderTest, LoadPositionDeletesFastPathHonorsReferencedDataFile) {
...
+  auto& index = result.value();
+  ASSERT_EQ(index.Cardinality(), 4);
+  ASSERT_TRUE(index.IsDeleted(0));
+  ASSERT_TRUE(index.IsDeleted(5));
+  ASSERT_TRUE(index.IsDeleted(10));
+  ASSERT_TRUE(index.IsDeleted(4'000'000'000LL));
+  ASSERT_FALSE(index.IsDeleted(1));
+}
+
+TEST_F(DeleteLoaderTest, LoadDeletionVectorSkipsMismatchedReferencedDataFile) {
```
wgtmac: This test should expect an error for a mismatched DV, not an empty index. Java validates referencedDataFile against the requested path before reading the DV.

src/iceberg/test/deletion_vector_writer_test.cc
```diff
@@ -0,0 +1,381 @@
...
+
+  // One DataFile per referenced data file.
+  ASSERT_EQ(delete_files.size(), 2u);
+
+  auto dv_a = FindByReferencedFile(delete_files, "data-a.parquet");
+  auto dv_b = FindByReferencedFile(delete_files, "data-b.parquet");
+  ASSERT_NE(dv_a, nullptr);
+  ASSERT_NE(dv_b, nullptr);
+
+  // Metadata is spec-compliant for a deletion vector.
```
wgtmac: Please remove this narration. The assertions already show the DV metadata being checked, and the comment adds noise.

wgtmac:
Two test comments still read like implementation notes but are on unchanged lines, so I can't leave inline comments there. Please trim the routing note in LoadPositionDeletesFiltersByDataFilePath and the fast-path threshold note in LoadPositionDeletesFastPathHonorsReferencedDataFile.

wgtmac (approved):
Thanks for working on this! I've slightly polished the API design and extracted some functions so they can be reused by other features (like merging v3 DVs during update).

PR #780, opened 2026-06-25
feat: add ArrowRowBuilder for materializing Arrow batches

WZhuo:
## Summary

Adds `ArrowRowBuilder` (`arrow_row_builder_internal.h` / `arrow_row_builder.cc`), a schema-driven RAII helper that materializes in-memory rows into an Arrow `ArrowArray` (a struct batch) for an arbitrary Iceberg schema. It wraps the nanoarrow boilerplate and exposes per-column access plus typed append free functions, so metadata tables (snapshots, history, manifests, …) can emit rows without re-implementing it.

This is the first of a series splitting metadata-table support into focused PRs; the `InMemoryBatchReader` and the `SnapshotsTable::Scan` integration are intended to follow in separate PRs that build on this.

## What's included
- **`ArrowRowBuilder`** — a single RAII class (move-only) with `Make(const Schema&)` and `Make(const ArrowSchema*)` overloads. Handles the full nanoarrow lifecycle: `InitFromSchema` → `StartAppending` → … append values … → `FinishBuilding` → `Release`. The `ArrowArray` is guarded immediately after `InitFromSchema` so a failure in `StartAppending` releases it automatically.
- **`ArrowArrayGuard::Release()`** — added to the existing guard so other call sites (`position_delete_writer`, `manifest_adapter`) can reuse the RAII-release pattern instead of manually managing nanoarrow resources.
- **Free functions** in the `iceberg` namespace: `AppendNull`, `AppendBoolean`, `AppendInt` (covers int32/int64/timestamp via nanoarrow's int64), `AppendString`, `AppendStringMap`.
- The implementation lives at the **core** `iceberg` library level — it only needs nanoarrow + `ToArrowSchema` (no Apache Arrow), matching peers like `manifest_adapter` and `arrow_c_data_util`.
- Unit tests in `arrow_row_builder_test.cc` covering typed appends (int32/string/int64/boolean/map), null handling for optional columns, multi-entry/empty string maps, zero-row batches, and column-index bounds. Compiled into the `iceberg-data-test` test target.

## Testing
- CMake (Ninja): `cmake --build build --target iceberg-data-test` then ran the test binary — 5/5 `ArrowRowBuilderTest` tests pass. `ctest` green.
- The test verifies output by importing the produced C-data into Apache Arrow (`arrow::ImportRecordBatch`), so its target is under `USE_BUNDLE`.

## Notes
- The test is registered under CMake's bundle build only. The meson build (which has no Apache Arrow/bundle layer) is left unchanged; the core-only test target continues to build there.
- Developed with AI-assisted tooling, reviewed by the author.

### DIFF
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -45,6 +45,7 @@ set(ICEBERG_SOURCES
     file_writer.cc
     inspect/history_table.cc
     inspect/metadata_table.cc
+    inspect/row_builder_internal.cc
     inspect/snapshots_table.cc
     inheritable_metadata.cc
     json_serde.cc
--- a/src/iceberg/CMakeLists.txt
+++ b/src/iceberg/CMakeLists.txt
@@ -18,8 +18,9 @@
 set(ICEBERG_INCLUDES "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/src>"
                      "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>")
 set(ICEBERG_SOURCES
-    arrow_c_data_util.cc
     arrow_c_data_guard_internal.cc
+    arrow_c_data_util.cc
+    arrow_row_builder.cc
     catalog/memory/in_memory_catalog.cc
     catalog/session_catalog.cc
     catalog/session_context.cc
@@ -45,7 +46,6 @@ set(ICEBERG_SOURCES
     file_writer.cc
     inspect/history_table.cc
     inspect/metadata_table.cc
-    inspect/row_builder_internal.cc
     inspect/snapshots_table.cc
     inheritable_metadata.cc
     json_serde.cc
--- a/src/iceberg/inspect/row_builder_internal.cc
+++ b/src/iceberg/inspect/row_builder_internal.cc
@@ -1,0 +1,134 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/inspect/row_builder_internal.h"
+
+#include <utility>
+
+#include <nanoarrow/nanoarrow.h>
+
+#include "iceberg/arrow/nanoarrow_status_internal.h"
+#include "iceberg/arrow_c_data_guard_internal.h"
+#include "iceberg/schema.h"
+#include "iceberg/schema_internal.h"
+
+namespace iceberg {
+
+Result<ArrowRowBuilder> ArrowRowBuilder::Make(const Schema& schema) {
+  ArrowSchema arrow_schema;
+  ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(schema, &arrow_schema));
+  internal::ArrowSchemaGuard schema_guard(&arrow_schema);
+
+  auto array = std::make_unique<ArrowArray>();
+  ArrowError error;
+  ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+      ArrowArrayInitFromSchema(array.get(), &arrow_schema, &error), error);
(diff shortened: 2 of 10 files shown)


src/iceberg/CMakeLists.txt
```diff
@@ -45,11 +45,11 @@ set(ICEBERG_SOURCES
     file_writer.cc
     inspect/history_table.cc
     inspect/metadata_table.cc
+    inspect/row_builder_internal.cc
```
wgtmac: We can just name it to `row_builder.cc` because `_internal` suffix is used for marking header files that do not need to be installed.

src/iceberg/inspect/row_builder_internal.h
```diff
@@ -0,0 +1,114 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+/// \file iceberg/inspect/row_builder_internal.h
```
wgtmac: Why putting it here instead of treating it as a general utility for building arrow arrays?

src/iceberg/inspect/row_builder_internal.cc
```diff
@@ -0,0 +1,134 @@
...
+Result<ArrowRowBuilder> ArrowRowBuilder::Make(const Schema& schema) {
+  ArrowSchema arrow_schema;
+  ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(schema, &arrow_schema));
+  internal::ArrowSchemaGuard schema_guard(&arrow_schema);
+
+  auto array = std::make_unique<ArrowArray>();
+  ArrowError error;
+  ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+      ArrowArrayInitFromSchema(array.get(), &arrow_schema, &error), error);
+  ICEBERG_NANOARROW_RETURN_UNEXPECTED(ArrowArrayStartAppending(array.get()));
```
wgtmac: If this call fails, the `ArrowArray` initialized above is destroyed without `ArrowArrayRelease`, so nanoarrow-owned state leaks. Please guard it right after `ArrowArrayInitFromSchema` and only disarm the guard once ownership has moved into `ArrowRowBuilder`.

wgtmac:
I think this is a good point to factor out a small internal nanoarrow builder/helper layer instead of adding another local copy of the same lifecycle code.

We already have the same pattern in a few production paths: initialize from schema, start appending, append child values, finish each row/list/map element, then finish building. For example, `position_delete_writer.cc` and `arrow_c_data_util.cc` both guard the initialized `ArrowArray` immediately after `ArrowArrayInitFromSchema`, while this new builder currently misses that and can leak if `ArrowArrayStartAppending` fails. `manifest_adapter.cc` also has local primitive/list/map append helpers, including map-shape validation, and this PR adds another set of primitive/string-map helpers.

A concrete way to keep this scoped would be:
1. Add an internal move-only nanoarrow array builder/owner that handles init/start/finish/release/ownership transfer.
2. Move the simple append helpers (`null`, signed/unsigned int, double, string, bytes, and maybe map entry validation) behind that internal helper.
3. Keep higher-level code like `ManifestAdapter`, batch projection, and metadata-table row materialization separate, since those have different domain semantics.

That should fix the leak path here and avoid repeating this fragile nanoarrow lifecycle in the next metadata-table PRs without turning this into a broad refactor.

src/iceberg/CMakeLists.txt
WZhuo: Renamed to arrow_row_builder.cc -- internal suffix removed from .cc.

src/iceberg/inspect/row_builder_internal.h
WZhuo: Moved to core level as arrow_row_builder_internal.h / arrow_row_builder.cc. ArrowRowBuilder now has two Make overloads: Make(const Schema&) for Iceberg callers and Make(const ArrowSchema*) for lower-level callers like position_delete_writer and manifest_adapter.

src/iceberg/inspect/row_builder_internal.cc
WZhuo: Fixed. ArrowArrayGuard now has Release(). Make(const ArrowSchema*) guards immediately after InitFromSchema, disarms after StartAppending succeeds. If StartAppending fails, the guard calls ArrowArrayRelease. If it succeeds, ArrowRowBuilder owns the array and releases it in the destructor.

WZhuo:
Thanks for the thorough review. All three points addressed:

1. ArrowRowBuilder is now a single RAII class (merged the two classes), moved to core level, with `Make(const Schema&)` and `Make(const ArrowSchema*)` overloads so both metadata tables and lower-level code like position_delete_writer/manifest_adapter can use it.

2. Append* helpers are free functions alongside ArrowRowBuilder: null, int (int32/int64/timestamp), boolean, string, and string_map.

3. The lifecycle is properly guarded: `ArrowArrayGuard` right after `InitFromSchema`, `Release()` after `StartAppending` succeeds. Also added `Release()` to `ArrowArrayGuard` for other call sites to reuse this pattern.

wgtmac (approved):

PR #783, opened 2026-06-26
feat(rest): add scan plan endpoint support to REST catalog client

gsandeep1241:
When a table is loaded from a REST catalog that advertises the PlanTableScan endpoint, NewScan() now returns a RestTableScanBuilder whose Build() produces a RestTableScan. PlanFiles() on that scan delegates manifest resolution to the server via POST /plan, GET /plan/{id} (with exponential backoff), POST /tasks/{id}, and DELETE /plan/{id} (best-effort cancel), instead of reading manifests locally.

- Add RestTable, RestTableScanBuilder, RestTableScan and RestScanContext
- Promote DataTableScan::PlanFiles and TableScanBuilder::Build to virtual
- Convert RestCatalog::client_ and paths_ to shared_ptr so RestScanContext can share ownership with live scans

### DIFF
--- a/src/iceberg/catalog/rest/rest_catalog.cc
+++ b/src/iceberg/catalog/rest/rest_catalog.cc
@@ -39,6 +39,7 @@
 #include "iceberg/catalog/rest/json_serde_internal.h"
 #include "iceberg/catalog/rest/resource_paths.h"
 #include "iceberg/catalog/rest/rest_file_io.h"
+#include "iceberg/catalog/rest/rest_table.h"
 #include "iceberg/catalog/rest/rest_util.h"
 #include "iceberg/catalog/rest/types.h"
 #include "iceberg/json_serde_internal.h"
@@ -451,7 +452,7 @@ Result<std::shared_ptr<RestCatalog>> RestCatalog::Make(
   // Get snapshot loading mode
   ICEBERG_ASSIGN_OR_RAISE(auto snapshot_mode, final_config.SnapshotLoadingMode());
 
-  auto client = std::make_unique<HttpClient>(final_config.ExtractHeaders());
+  auto client = std::make_shared<HttpClient>(final_config.ExtractHeaders());
   ICEBERG_ASSIGN_OR_RAISE(auto catalog_session,
                           auth_manager->CatalogSession(*client, final_config.configs()));
 
@@ -466,8 +467,8 @@ Result<std::shared_ptr<RestCatalog>> RestCatalog::Make(
 }
 
 RestCatalog::RestCatalog(RestCatalogProperties config, std::shared_ptr<FileIO> file_io,
-                         std::unique_ptr<HttpClient> client,
-                         std::unique_ptr<ResourcePaths> paths,
+                         std::shared_ptr<HttpClient> client,
+                         std::shared_ptr<ResourcePaths> paths,
                          std::unordered_set<Endpoint> endpoints,
                          std::unique_ptr<auth::AuthManager> auth_manager,
                          std::shared_ptr<auth::AuthSession> catalog_session,
@@ -873,6 +874,20 @@ Result<std::shared_ptr<Table>> RestCatalog::MakeTableFromLoadResult(
       TableAuthSession(identifier, table_config, std::move(contextual_session)));
   auto table_catalog = std::make_shared<TableScopedCatalog>(
       shared_from_this(), context, identifier, table_config, table_session);
+
+  if (supported_endpoints_.contains(Endpoint::PlanTableScan())) {
+    RestScanContext rest_ctx{
+        .client = client_,
+        .paths = paths_,
+        .session = table_session,
+        .supported_endpoints = supported_endpoints_,
+        .identifier = identifier,
+    };
+    return RestTable::Make(identifier, std::move(result.metadata),
+                           std::move(result.metadata_location), std::move(table_io),
+                           std::move(table_catalog), std::move(rest_ctx));
+  }
+
   return Table::Make(identifier, std::move(result.metadata),
                      std::move(result.metadata_location), std::move(table_io),
                      std::move(table_catalog));
--- a/src/iceberg/catalog/rest/rest_table.h
+++ b/src/iceberg/catalog/rest/rest_table.h
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "iceberg/catalog/rest/iceberg_rest_export.h"
+#include "iceberg/catalog/rest/rest_table_scan.h"
+#include "iceberg/result.h"
+#include "iceberg/table.h"
+#include "iceberg/type_fwd.h"
+
+/// \file iceberg/catalog/rest/rest_table.h
+/// A Table subclass that uses server-side distributed scan planning via the REST catalog.
+
+namespace iceberg::rest {
+
+/// \brief A Table whose NewScan() returns a RestTableScanBuilder, delegating
+/// PlanFiles() to the REST catalog server's scan planning endpoints.
+class ICEBERG_REST_EXPORT RestTable final : public Table {
+ public:
+  static Result<std::shared_ptr<RestTable>> Make(TableIdentifier identifier,
+                                                 std::shared_ptr<TableMetadata> metadata,
+                                                 std::string metadata_location,
+                                                 std::shared_ptr<FileIO> io,
+                                                 std::shared_ptr<Catalog> catalog,
+                                                 RestScanContext rest_context);
+
+  ~RestTable() override;
+
+  /// \brief Returns a RestTableScanBuilder that will delegate PlanFiles() to the
+  /// REST catalog server.
+  Result<std::unique_ptr<DataTableScanBuilder>> NewScan() const override;
+
+ private:
+  RestTable(TableIdentifier identifier, std::shared_ptr<TableMetadata> metadata,
+            std::string metadata_location, std::shared_ptr<FileIO> io,
+            std::shared_ptr<Catalog> catalog, RestScanContext rest_context);
+
+  RestScanContext rest_context_;
+};
+
+}  // namespace iceberg::rest
--- a/src/iceberg/catalog/rest/rest_table_scan.cc
+++ b/src/iceberg/catalog/rest/rest_table_scan.cc
@@ -1,0 +1,269 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/catalog/rest/rest_table_scan.h"
+
+#include <chrono>
+#include <thread>
+
+#include <nlohmann/json.hpp>
+
+#include "iceberg/catalog/rest/endpoint.h"
+#include "iceberg/catalog/rest/error_handlers.h"
+#include "iceberg/catalog/rest/http_client.h"
+#include "iceberg/catalog/rest/json_serde_internal.h"
+#include "iceberg/catalog/rest/resource_paths.h"
+#include "iceberg/catalog/rest/types.h"
+#include "iceberg/json_serde_internal.h"
+#include "iceberg/partition_spec.h"
+#include "iceberg/result.h"
+#include "iceberg/schema.h"
+#include "iceberg/table_metadata.h"
+#include "iceberg/util/macros.h"
+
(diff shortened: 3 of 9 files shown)


src/iceberg/catalog/rest/rest_catalog.cc
```diff
@@ -873,6 +874,20 @@ Result<std::shared_ptr<Table>> RestCatalog::MakeTableFromLoadResult(
       TableAuthSession(identifier, table_config, std::move(contextual_session)));
   auto table_catalog = std::make_shared<TableScopedCatalog>(
       shared_from_this(), context, identifier, table_config, table_session);
+
+  if (supported_endpoints_.contains(Endpoint::PlanTableScan())) {
```
wgtmac, line 878: This should also gate on the effective `scan-planning-mode`, not only endpoint support. Java defaults to client-side planning and lets the table config override the client config, so a table can otherwise be forced into REST planning even when the server says `client`, or silently fall back when the server says `server` but the endpoint is missing.

src/iceberg/catalog/rest/rest_table_scan.cc
```diff
@@ -0,0 +1,269 @@
...
+  ICEBERG_ENDPOINT_CHECK(rest_context_.supported_endpoints, Endpoint::PlanTableScan());
+
+  // Build request from scan context
+  PlanTableScanRequest request;
+  request.select = context_.selected_columns;
+  request.filter = context_.filter;
+  request.case_sensitive = context_.case_sensitive;
+  request.min_rows_requested = context_.min_rows_requested;
+
+  if (context_.from_snapshot_id.has_value() && context_.to_snapshot_id.has_value()) {
```
wgtmac, line 100: We need to set `use-snapshot-schema` for snapshot/time-travel and incremental scans. Java sends it for `useSnapshot` and start/end snapshot scans, and the REST spec says time travel should use the snapshot schema. Without it, schema-evolved tables can be planned against the current schema.

src/iceberg/catalog/rest/rest_table_scan.cc
```diff
@@ -0,0 +1,269 @@
...
+  ICEBERG_ASSIGN_OR_RAISE(auto path,
+                          rest_context_.paths->Plan(rest_context_.identifier));
+  ICEBERG_ASSIGN_OR_RAISE(auto request_json, ToJson(request));
+  ICEBERG_ASSIGN_OR_RAISE(auto json_request, ToJsonString(request_json));
+  ICEBERG_ASSIGN_OR_RAISE(
+      const auto response,
+      rest_context_.client->Post(path, json_request, /*headers=*/{},
+                                 *PlanErrorHandler::Instance(), *rest_context_.session));
+  ICEBERG_ASSIGN_OR_RAISE(auto json, FromJsonString(response.body()));
+  ICEBERG_ASSIGN_OR_RAISE(auto result,
```
wgtmac, line 125: Planning responses can include `storage-credentials`. Java switches to a scan-scoped `FileIO` built from those credentials, and the spec expects clients to use them for the returned tasks. Ignoring them means servers that vend temporary storage credentials can plan successfully but reads may fail.

src/iceberg/catalog/rest/rest_table_scan.cc
```diff
@@ -0,0 +1,269 @@
...
+                          PlanTableScanResponseFromJson(json, specs, *schema_));
+  ICEBERG_RETURN_UNEXPECTED(result.Validate());
+
+  plan_id = result.plan_id;
+
+  switch (result.plan_status) {
+    case PlanStatus::kCompleted:
+      return ResolveScanTasks(result.plan_tasks, result.file_scan_tasks, specs);
```
wgtmac, line 133: Once a `plan-id` is returned, the server may hold resources until all plan tasks are fetched or the plan is cancelled. If resolving paginated tasks fails partway through, this returns without cancelling the remaining plan; Java cancels from the scan-task iterable cleanup path.

wgtmac:
Hey @gsandeep1241, do you want to revive this?

gsandeep1241:
@wgtmac Thanks for re-opening it. Apologies for the long delay, I'll get back on it this week - next set of changes should be out for review in the next couple of days!

src/iceberg/catalog/rest/rest_catalog.cc
gsandeep1241: Done!

src/iceberg/catalog/rest/rest_table_scan.cc
gsandeep1241: Done!

src/iceberg/catalog/rest/rest_table_scan.cc
gsandeep1241: Updated here. Are you also suggesting that we change the return signature of PlanFiles to have an iterable and let the blocking happen at the caller instead of here?

src/iceberg/catalog/rest/rest_table_scan.cc
gsandeep1241: Done!

src/iceberg/catalog/rest/rest_table_scan.cc
```diff
@@ -0,0 +1,298 @@
...
+      if (!tasks.has_value()) CancelPlanning(plan_id);
+      return tasks;
+    }
+    case PlanStatus::kSubmitted:
+      return FetchPlanningResult(plan_id, specs);
+    case PlanStatus::kFailed:
+      return IOError("Scan planning failed: {}",
+                      result.error ? result.error->message : "unknown error");
+    case PlanStatus::kCancelled:
+      return IOError("Scan planning was cancelled for plan_id={}", plan_id);
```
gsandeep1241, line 146: Do we want to introduce a new error type? Or use InvalidArgument instead?

gsandeep1241:
> Hey @gsandeep1241, do you want to revive this?

Thanks @wgtmac for reviving this!  This is now ready for review. Please take a look when you can :)

src/iceberg/catalog/rest/rest_table_scan.h
```diff
@@ -0,0 +1,137 @@
...
+ public:
+  ~RestTableScan() override = default;
+
+  static Result<std::unique_ptr<DataTableScan>> Make(
+      std::shared_ptr<TableMetadata> metadata, std::shared_ptr<Schema> schema,
+      std::shared_ptr<FileIO> io, internal::TableScanContext context,
+      RestScanContext rest_context);
+
+  /// \brief Plans files via the REST scan planning endpoints.
+  Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles() const override;
```
wgtmac, line 76: After #873, callers can use `PlanFilesStream()`. This class only overrides `PlanFiles()`, so the stream path still reads manifests locally and never calls `/plan`. This bypasses server-side planning.

src/iceberg/catalog/rest/rest_table_scan.h
```diff
@@ -0,0 +1,137 @@
...
+
+  /// \brief Returns the FileIO to use when reading scan results.
+  ///
+  /// If the server vended storage credentials during planning, returns a FileIO
+  /// initialised with those credentials; otherwise returns the table's FileIO.
+  /// Must be called after PlanFiles().
+  const std::shared_ptr<FileIO>& effective_io() const;
```
wgtmac, line 83: The scan is returned as `std::unique_ptr<DataTableScan>`, but `DataTableScan::io()` is not virtual and still returns the table IO. A normal caller cannot reach `effective_io()` without a downcast, so the vended credentials are not used by the normal read path.

src/iceberg/catalog/rest/rest_table_scan.cc
```diff
@@ -0,0 +1,294 @@
...
+                         Endpoint::FetchPlanningResult());
+
+  ICEBERG_ASSIGN_OR_RAISE(auto path,
+                          rest_context_.paths->Plan(rest_context_.identifier, plan_id));
+
+  auto delay_ms = kMinSleepMs;
+  auto start = std::chrono::steady_clock::now();
+
+  for (int retry = 0; retry <= kMaxRetries; ++retry) {
+    ICEBERG_ASSIGN_OR_RAISE(
```
wgtmac, line 164: After a `plan-id` exists, any GET, JSON parse, or response validation error here returns without calling `CancelPlanning()`. The server may keep the plan resources.

src/iceberg/catalog/rest/rest_table_scan.cc
```diff
@@ -0,0 +1,294 @@
...
+                                   *PlanErrorHandler::Instance(), *rest_context_.session);
+}
+
+const std::shared_ptr<FileIO>& RestTableScan::effective_io() const {
+  return scan_io_ ? scan_io_ : io_;
+}
+
+Status RestTableScan::ApplyStorageCredentials(
+    const std::vector<StorageCredential>& credentials) const {
+  if (credentials.empty()) return {};
```
wgtmac, line 268: An empty credential list leaves `scan_io_` unchanged. If this scan is planned twice and only the first response has credentials, the second result still uses the old credentials.

src/iceberg/catalog/rest/rest_table.h
```diff
@@ -0,0 +1,62 @@
...
+      TableIdentifier identifier, std::shared_ptr<TableMetadata> metadata,
+      std::string metadata_location, std::shared_ptr<FileIO> io,
+      std::shared_ptr<Catalog> catalog, std::string full_name,
+      std::shared_ptr<MetricsReporter> reporter, RestScanContext rest_context);
+
+  ~RestTable() override;
+
+  /// \brief Returns a RestTableScanBuilder that will delegate PlanFiles() to the
+  /// REST catalog server.
+  Result<std::unique_ptr<DataTableScanBuilder>> NewScan() const override;
```
wgtmac, line 51: This only overrides `NewScan()`. `NewIncrementalAppendScan()` and `NewIncrementalChangelogScan()` still use the base table path and plan manifests locally, even when `scan-planning-mode=server`.

src/iceberg/test/rest_table_scan_test.cc
```diff
@@ -0,0 +1,467 @@
...
+// --------------------------------------------------------------------------
+// PlanFiles: server returns COMPLETED immediately, no file scan tasks.
+// --------------------------------------------------------------------------
+TEST_F(RestTableScanTest, PlanFilesCompleted) {
+  constexpr std::string_view kResponseBody = R"({"status":"completed"})";
+  EXPECT_CALL(*mock_client_, Post(_, _, _, _, _))
+      .WillOnce(Return(HttpResponse::MakeForTesting(200, std::string(kResponseBody))));
+
+  ICEBERG_UNWRAP_OR_FAIL(auto scan, MakeScan(MakeContext()));
+  ICEBERG_UNWRAP_OR_FAIL(auto tasks, scan->PlanFiles());
```
wgtmac, line 187: These tests only call `PlanFiles()`. After #873, `PlanFilesStream()` is a separate caller path, so this suite does not catch the local-planning bypass.

src/iceberg/test/rest_table_scan_test.cc
```diff
@@ -0,0 +1,467 @@
...
+// PlanFiles: server returns SUBMITTED → poll returns COMPLETED.
+// --------------------------------------------------------------------------
+TEST_F(RestTableScanTest, PlanFilesSubmittedThenCompleted) {
+  constexpr std::string_view kSubmittedBody =
+      R"({"status":"submitted","plan-id":"plan-poll-1"})";
+  constexpr std::string_view kCompletedBody = R"({"status":"completed"})";
+
+  EXPECT_CALL(*mock_client_, Post(_, _, _, _, _))
+      .WillOnce(Return(HttpResponse::MakeForTesting(200, std::string(kSubmittedBody))));
+  EXPECT_CALL(*mock_client_, Get(_, _, _, _, _))
```
wgtmac, line 215: There is no case where a `plan-id` is returned and the poll GET, JSON parsing, or response validation fails. Those errors currently skip cancellation.

src/iceberg/test/rest_table_scan_test.cc
```diff
@@ -0,0 +1,467 @@
...
+// use_snapshot_schema: UseSnapshot() sets it to true in the builder context.
+// RestTableScanBuilder propagates context from DataTableScanBuilder.
+// --------------------------------------------------------------------------
+TEST_F(RestTableScanTest, UseSnapshotPropagatesUseSnapshotSchemaInContext) {
+  constexpr int64_t kSnapshotId = 1000L;
+  RestTableScanBuilder builder(metadata_, file_io_, "test.my_table", nullptr,
+                               MakeContext(std::nullopt));
+  builder.UseSnapshot(kSnapshotId);
+  ICEBERG_UNWRAP_OR_FAIL(auto scan, builder.Build());
+  EXPECT_TRUE(scan->context().use_snapshot_schema);
```
wgtmac, line 373: This checks the builder context, not the JSON sent to the server. None of the POST expectations inspect the body, so request fields can be missing and these tests still pass.

src/iceberg/test/rest_table_scan_test.cc
```diff
@@ -0,0 +1,467 @@
...
+    ]
+  })";
+  EXPECT_CALL(*mock_client_, Post(_, _, _, _, _))
+      .WillOnce(Return(HttpResponse::MakeForTesting(200, std::string(kResponseBody))));
+
+  ICEBERG_UNWRAP_OR_FAIL(auto scan, MakeScan(MakeContext()));
+  ICEBERG_UNWRAP_OR_FAIL(auto tasks, scan->PlanFiles());
+  EXPECT_TRUE(tasks.empty());
+
+  auto* rest_scan = dynamic_cast<RestTableScan*>(scan.get());
```
wgtmac, line 404: This downcast hides the public API problem. Real callers hold a `DataTableScan` and use `io()`, which still returns the original table IO.

src/iceberg/test/rest_table_scan_test.cc
```diff
@@ -0,0 +1,467 @@
...
+  ASSERT_NE(rest_scan, nullptr);
+  // effective_io() must return a credential-scoped IO, not the original file_io_.
+  EXPECT_NE(rest_scan->effective_io().get(), file_io_.get());
+}
+
+// --------------------------------------------------------------------------
+// No storage credentials: effective_io() falls back to the table's FileIO.
+// --------------------------------------------------------------------------
+TEST_F(RestTableScanTest, NoStorageCredentialsEffectiveIoFallsBackToTableIO) {
```
wgtmac, line 413: This uses a fresh scan. It does not catch stale credentials when the same scan is planned twice, first with credentials and then without them.

src/iceberg/test/rest_table_scan_test.cc
```diff
@@ -0,0 +1,467 @@
...
+
+  auto* rest_scan = dynamic_cast<RestTableScan*>(scan.get());
+  ASSERT_NE(rest_scan, nullptr);
+  EXPECT_NE(rest_scan->effective_io().get(), file_io_.get());
+}
+
+// --------------------------------------------------------------------------
+// RestTable::NewScan returns a RestTableScanBuilder (not a plain builder).
+// --------------------------------------------------------------------------
+TEST_F(RestTableScanTest, RestTableNewScanReturnsRestTableScanBuilder) {
```
wgtmac, line 457: This builds `RestTable` directly. It does not cover `RestCatalog::LoadTable()` choosing the scan type from client and table `scan-planning-mode`, or the missing-endpoint error.

PR #791, opened 2026-06-29
feat: write snapshot v3 row lineage fields at top level

wgtmac:


blackmwk (approved):

zhjwpku (approved):

src/iceberg/json_serde.cc
```diff
@@ -808,12 +822,54 @@ Result<std::unique_ptr<Snapshot>> SnapshotFromJson(const nlohmann::json& json) {
...
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(auto added_rows,
+                          GetJsonValueOptional<int64_t>(json, kAddedRows));
+  if (!added_rows.has_value()) {
+    ICEBERG_ASSIGN_OR_RAISE(added_rows,
+                            parse_summary_int64(SnapshotSummaryFields::kAddedRows));
+  }
+
+  summary.erase(SnapshotSummaryFields::kFirstRowId);
```
manuzhang: Will custom summary keys "first-row-id" and "added-rows" from v1/v2 tables be erased as well?

src/iceberg/json_serde.cc
```diff
@@ -808,12 +822,54 @@ Result<std::unique_ptr<Snapshot>> SnapshotFromJson(const nlohmann::json& json) {
     }
   }
 
+  auto parse_summary_int64 =
```
manuzhang: Will this parse nested `first-row-id` and `added-rows` as well?

src/iceberg/json_serde.cc
zhjwpku: I think these two fields are dedicated to v3, see https://iceberg.apache.org/spec/#snapshots

manuzhang: Yes, but this doesn't preclude them from existing v1 and v2 tables.

wgtmac:
Thanks @manuzhang for the comment. Good catch!

src/iceberg/snapshot.cc
```diff
@@ -186,7 +172,8 @@ bool Snapshot::Equals(const Snapshot& other) const {
   return snapshot_id == other.snapshot_id &&
          parent_snapshot_id == other.parent_snapshot_id &&
          sequence_number == other.sequence_number && timestamp_ms == other.timestamp_ms &&
-         schema_id == other.schema_id;
+         schema_id == other.schema_id && first_row_id == other.first_row_id &&
```
manuzhang, line 175: Not sure these fields are required in `equals`. Java implementation doesn't have them.

https://github.com/apache/iceberg/blob/8c1ee9d9df4184a7889f85f6a9e1f82f959e0eaf/core/src/main/java/org/apache/iceberg/BaseSnapshot.java#L332-L347

wgtmac: It looks odd if two snapshots with different row lineage attributes are deemed identical. I would rather think this is an oversight from https://github.com/apache/iceberg/pull/11948. Both iceberg-rust and iceberg-python enforce all fields to be equal so I think it is fine to include them here.

manuzhang (approved):

src/iceberg/snapshot.cc
manuzhang: Agreed. Opened https://github.com/apache/iceberg/pull/17015 to fix at Java side.

PR #792, opened 2026-06-29
feat(parquet): apply column default values when reading missing fields (2/4)

huan233usc:
## What

Part 2 of 4 of Iceberg v3 column default-value support (POC #731), built on the
schema-layer support merged in #746.

When a column is present in the read (table) schema but absent from a Parquet
data file — because the column was added after those rows were written — fill it
with the column's v3 `initial-default` instead of `null`.

## Changes

- **`arrow/literal_util`** (new): a shared Arrow materializer that converts a
  `Literal` to an Arrow scalar (`ToArrowScalar`), to a constant-filled array
  (`MakeDefaultArray`), or appends it once to a builder
  (`AppendDefaultToBuilder`). Covers all primitive types, including
  decimal/uuid/fixed and the four timestamp variants.
- **Parquet projection** (`parquet_schema_util.cc`): when a field is missing
  from the file and carries an `initial-default`, project it as
  `FieldProjection::Kind::kDefault`, mirroring the generic `schema_util.cc` path
  already merged in #746.
- **Parquet data** (`parquet_data_util.cc`): materialize the `kDefault` branch
  via `MakeDefaultArray`.

## Tests

- `literal_util_test`: scalar/array/builder conversion across primitive types,
  null and narrowing-sentinel handling, and casting to the target Arrow type.
- `parquet_data_test`: `ProjectRecordBatch` fills missing required and optional
  default fields, including a nested struct.
- `parquet_test`: end-to-end — write a file with an old schema, then read it
  through `ReaderFactoryRegistry` with an evolved schema carrying defaults
  (`ReadMissingFieldsWithDefaults`).

## Stack

2. **this PR** — read path: Parquet
3. read path: Avro (follows)
4. schema evolution: `addColumn` / `updateColumnDefault` (follows)

### DIFF
--- a/src/iceberg/arrow/literal_util.cc
+++ b/src/iceberg/arrow/literal_util.cc
@@ -1,0 +1,198 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <cstring>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <arrow/array/util.h>
+#include <arrow/buffer.h>
+#include <arrow/compute/api.h>
+#include <arrow/extension_type.h>
+#include <arrow/scalar.h>
+#include <arrow/type.h>
+
+#include "iceberg/arrow/arrow_status_internal.h"
+#include "iceberg/arrow/literal_util_internal.h"
+#include "iceberg/type.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/formatter.h"  // IWYU pragma: keep
+#include "iceberg/util/macros.h"
+
+namespace iceberg::arrow {
+
+namespace {
+
+Result<std::shared_ptr<::arrow::DataType>> ToArrowType(const PrimitiveType& type) {
+  switch (type.type_id()) {
+    case TypeId::kBoolean:
+      return ::arrow::boolean();
+    case TypeId::kInt:
+      return ::arrow::int32();
+    case TypeId::kLong:
+      return ::arrow::int64();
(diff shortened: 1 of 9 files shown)


src/iceberg/arrow/literal_util.cc
```diff
@@ -0,0 +1,208 @@
...
+    }
+    case TypeId::kUuid:
+      return ::arrow::fixed_size_binary(16);
+    default:
+      return NotSupported("Cannot convert {} to an Arrow type", type);
+  }
+}
+
+Result<std::shared_ptr<::arrow::Buffer>> ToArrowBuffer(
+    const std::vector<uint8_t>& bytes) {
```
wgtmac: add `arrow::MemoryPool*` as an input parameter.

src/iceberg/arrow/literal_util.cc
```diff
@@ -0,0 +1,208 @@
...
+                                   ::arrow::compute::Cast(array, type));
+    return cast_result.make_array();
+  }
+  return array;
+}
+
+Status AppendDefaultToBuilder(const Literal& literal, ::arrow::ArrayBuilder* builder) {
+  ICEBERG_ASSIGN_OR_RAISE(std::shared_ptr<::arrow::Scalar> scalar,
+                          ToArrowScalar(literal));
+  if (!scalar->type->Equals(*builder->type())) {
```
wgtmac: Could we mirror `MakeDefaultArray`'s extension handling here? `ToArrowScalar(UUID)` returns `fixed_size_binary(16)`, and `Scalar::CastTo(arrow.uuid())` is not implemented in Arrow, so using this helper with an `arrow.uuid` builder would fail.

huan233usc: Thanks for catching this, I took a second look and this method is not used in parquet(but avro in the poc). It seems that it got accidentally added to this pr. I will include the fix in avro pr later.

src/iceberg/arrow/literal_util_internal.h
```diff
@@ -0,0 +1,47 @@
...
+#include "iceberg/result.h"
+
+namespace iceberg::arrow {
+
+/// \brief Convert a primitive literal to an Arrow scalar of its canonical Arrow type.
+///
+/// A null literal converts to a null scalar of the corresponding Arrow type. `pool` is
+/// used for the backing buffer of binary/fixed/uuid scalars.
+Result<std::shared_ptr<::arrow::Scalar>> ToArrowScalar(
+    const Literal& literal, ::arrow::MemoryPool* pool = ::arrow::default_memory_pool());
```
wgtmac:
```suggestion
Result<std::shared_ptr<::arrow::Scalar>> ToArrowScalar(const Literal& literal,
                                                       ::arrow::MemoryPool* pool);
```

nit: don't provide default value here

wgtmac (approved):

src/iceberg/arrow/literal_util_internal.h
huan233usc: Applied suggestion, thanks

wgtmac:
Thank you, @huan233usc!

PR #793, opened 2026-06-29
feat: support v3 column default values in UpdateSchema (3/4)

huan233usc:
## What

Part 3 of 4 of Iceberg v3 column default-value support (POC #731), built on the
schema-layer support merged in #746. Independent of the read-path PRs (#792 and
the Avro follow-up).

Adds default-value handling to `UpdateSchema` (schema evolution).

## Changes

- **`AddColumn` / `AddRequiredColumn`** take an optional `default_value`. When
  provided it is set as both the column's `initial-default` and `write-default`.
  A non-null default also lets a required column be added without
  `AllowIncompatibleChanges()` — rows written before the change read the default
  instead of null.
- **`UpdateColumnDefault(name, default)`** (new) sets, or clears with
  `std::nullopt`, a column's `write-default`; the `initial-default` is fixed when
  the column is added.
- Defaults are **cast to the column type** (rejecting uncastable or out-of-range
  values) and **preserved** across rename / doc / type-promotion updates and
  nested field-id reassignment.
- `RequireColumn` may now mark a column that was added with a default required.

The `SchemaField` constructor stores defaults verbatim (it does not coerce
them), so the cast/promotion is performed explicitly at each evolution site —
the same effect as Java, where `NestedField`'s constructor runs `castDefault`.
Same-scale decimal precision widening is handled directly (the unscaled value is
unchanged), since `Literal::CastTo` does not cast between decimal types.

## Tests

13 cases in `update_schema_test.cc`: add optional/required/nested column with a
default, mismatched/narrowing rejection, `UpdateColumnDefault`
(set / clear / cast-to-type / pre-existing column), require-after-default, and
preservation across doc updates and type promotion — including same-scale
decimal precision promotion.

## Stack

2. #792 — read path: Parquet
3. read path: Avro (follows)
4. **this PR** — schema evolution: `addColumn` / `updateColumnDefault`


src/iceberg/update/update_schema.cc
```diff
@@ -474,9 +555,10 @@ UpdateSchema& UpdateSchema::UpdateColumnRequirementInternal(std::string_view nam
     return *this;
   }
 
-  // TODO(GuotaoYu): support added column with default value
-  // bool is_defaulted_add = IsAdded(name) && field.initial_default() != null;
-  bool is_defaulted_add = false;
+  // A column added in this update with a default value can be made required: rows
+  // written before the change read the initial-default instead of null.
+  bool is_defaulted_add = added_name_to_id_.contains(CaseSensitivityAwareName(name)) &&
```
manuzhang, line 536: For map/list nested adds, `AddColumnInternal` stores the new field in `added_name_to_id_` using the canonical internal path:

```
  added_name_to_id_[CaseSensitivityAwareName(full_name)] = new_id;
```

  where full_name becomes something like:

```
  locations.value.alt
  points.element.z
```

  But `RequireColumn` receives and checks the user-facing shorthand path:

```
  added_name_to_id_.contains(CaseSensitivityAwareName(name))
```

  where name is typically:

```
  locations.alt
  points.z
```

  So the lookup misses, even though it refers to the same added column

huan233usc: Thanks for catching this -- yes it stored only the canonical full_name but callers use the short name. Fixed by also indexing the short name (matching how the schema name index handles map value / list element paths), with a regression test for requiring a defaulted nested column by short name in the same transaction. 

It seems that same bug exists in Java; working on a fix apache/iceberg#17034.

src/iceberg/update/update_schema.cc
```diff
@@ -420,9 +459,52 @@ UpdateSchema& UpdateSchema::UpdateColumnDoc(std::string_view name,
...
+    updates_[field_id] = rebuild(nullptr);
+    return *this;
+  }
+
+  ICEBERG_BUILDER_CHECK(field.type()->is_primitive(),
+                        "Invalid default value for {}: {} (must be null)", *field.type(),
+                        *new_default);
+  ICEBERG_BUILDER_ASSIGN_OR_RETURN_WITH_ERROR(
+      Literal typed_default,
+      new_default->CastTo(internal::checked_pointer_cast<PrimitiveType>(field.type())),
```
manuzhang: Could we share the decimal default promotion logic used above in `UpdateColumn` here too? `Literal::CastTo` only returns decimal values when source and target `DecimalType` are exactly equal; it does not handle same-scale precision widening. That means adding `decimal(18,2)` with `Literal::Decimal(..., 9, 2)` is rejected even though this is the same allowed promotion handled above. The same issue appears in `AddColumnInternal`.

src/iceberg/update/update_schema.cc
```diff
@@ -282,6 +289,24 @@ std::vector<SchemaField> ApplyChangesVisitor::MoveFields(
...
+/// promotion leaves the unscaled value unchanged, so the literal is rebuilt with the
+/// target type instead of going through `CastTo`. All other conversions delegate to
+/// `CastTo`, which reports narrowing via AboveMax/BelowMin sentinels that callers reject.
+Result<Literal> CastDefaultToType(const Literal& value,
+                                  const std::shared_ptr<PrimitiveType>& target_type) {
+  if (value.type()->type_id() == TypeId::kDecimal &&
+      target_type->type_id() == TypeId::kDecimal) {
+    const auto& decimal_type = internal::checked_cast<const DecimalType&>(*target_type);
+    return Literal::Decimal(std::get<Decimal>(value.value()).value(),
+                            decimal_type.precision(), decimal_type.scale());
```
wgtmac: Could we guard this decimal fast path with a same-scale check before rebuilding the literal? `Decimal` stores only the unscaled value, so accepting `decimal(9,3)` as a default for `decimal(18,2)` would reinterpret `1234` from `1.234` to `12.34` instead of rejecting it. Java default parsing/serialization requires the decimal default scale to match the column type, so only same-scale precision widening should be accepted here.

src/iceberg/update/update_schema.cc
huan233usc: Done — shared it into a `CastDefaultToType` helper used in all three spots, with tests. Thanks!

src/iceberg/update/update_schema.cc
huan233usc: Good catch — added a same-scale guard so different-scale defaults fall through to `CastTo` and get rejected. Tested both paths. Thanks!

src/iceberg/update/update_schema.cc
```diff
@@ -282,6 +289,34 @@ std::vector<SchemaField> ApplyChangesVisitor::MoveFields(
...
+/// the caller's dereferenced default, so a missing default must be filtered out before
+/// this is reached.
+Result<Literal> CastDefaultToType(const Literal& value,
+                                  const std::shared_ptr<PrimitiveType>& target_type) {
+  if (value.type()->type_id() == TypeId::kDecimal &&
+      target_type->type_id() == TypeId::kDecimal) {
+    const auto& source_type = internal::checked_cast<const DecimalType&>(*value.type());
+    const auto& decimal_type = internal::checked_cast<const DecimalType&>(*target_type);
+    if (source_type.scale() == decimal_type.scale()) {
+      return Literal::Decimal(std::get<Decimal>(value.value()).value(),
```
wgtmac: Could we handle null/sentinel literals before this decimal fast path? A typed null default like `Literal::Null(decimal(18, 2))` has decimal type metadata but stores `std::monostate`, so this `std::get<Decimal>` would throw `bad_variant_access` instead of treating the null default as absent/rejecting it cleanly. The same early check would also avoid bypassing the existing sentinel rejection path.

src/iceberg/update/update_schema.cc
```diff
@@ -282,6 +289,34 @@ std::vector<SchemaField> ApplyChangesVisitor::MoveFields(
...
+                              decimal_type.precision(), decimal_type.scale());
+    }
+  }
+  return value.CastTo(target_type);
```
wgtmac: Could this path also support the numeric-to-decimal casts that Java default handling allows? Java `Literal.to(decimal)` converts int/long/float/double defaults to a decimal at the target scale, but the current fallback to `Literal::CastTo` rejects those cases, so a valid default like `Literal::Int(12)` for a `decimal(9, 2)` column cannot be used through `AddColumn` or `UpdateColumnDefault`.

src/iceberg/update/update_schema.cc
wgtmac: Could we also verify that the decimal value actually fits the target precision before relabeling it with the target type? This fast path preserves the unscaled value but changes the literal type, so an out-of-range value can pass `SchemaField::Validate` and then serialize to metadata that the C++ JSON parser later rejects because the parsed decimal precision exceeds the field precision. `Decimal::FitsInPrecision(decimal_type.precision())` looks like the check needed here.

huan233usc: Good catch — fixed. The fast path now only fires when the literal actually holds a `Decimal` value (`std::holds_alternative<Decimal>`), so a typed null or sentinel falls through to `CastTo` and is rejected by the same null/sentinel check the callers already apply. Added a regression test with `Literal::Null(decimal(18,2))`.

huan233usc: Right — added a `FitsInPrecision(target precision)` check before relabeling, so an out-of-range value is rejected here instead of passing `Validate` and later failing the JSON parser. Regression test added (decimal(4,2) column with a value needing more digits).

src/iceberg/update/update_schema.cc
huan233usc: Good point — this is a pre-existing gap in `Literal::CastTo` itself: `CastFromInt/Long/Float/Double` have no decimal target branch (only `CastFromString` produces a decimal), so numeric→decimal defaults were never supported, independent of this PR. 

I raise https://github.com/apache/iceberg-cpp/pull/805 for the gap

wgtmac (approved):
This PR overall looks great. I just found a few issues while reviewing this PR and pushed fixes directly to avoid more back-and-forth:

- Removed the added `TableMetadataV3Valid.json` test fixture and generate v3 metadata in the unit test instead. We don't want to give the confusion that this json file is copied from other repo.
- Moved default literal casting out of `update_schema.cc` into `SchemaField`, closer to Java’s `NestedField` handling.
- Added coverage for default casting, decimal widening, null/range sentinel failures, and `UpdateColumnDefault` no-op behavior.
- Fixed `UpdateColumnDefault` to match Java behavior when the new write default is already set.
- Added small `SchemaField::With*` helpers to make schema field updates less error-prone.
- Cleaned up overly verbose comments.

Thanks @huan233usc for working on this and @manuzhang for the great review feedback!

PR #794, opened 2026-06-30
feat: retry stale v3 snapshot row-lineage validation

wgtmac:
Add a retryable validation error kind and use it for add-snapshot stale sequence-number and stale first-row-id checks, matching Java Iceberg's RetryableValidationException behavior.

Include the new retryable validation kind in commit retry policy, while preserving normal validation failures for mixed/non-retryable builder errors. Add focused v3 row-lineage tests for multi-file assignment, branch commits, retry reassignment, stale snapshot validation, and delete-manifest null first_row_id handling.

### DIFF
--- a/src/iceberg/util/error_collector.h
+++ b/src/iceberg/util/error_collector.h
@@ -159,13 +159,19 @@ class ICEBERG_EXPORT ErrorCollector {
   /// in Build(), Apply(), or Commit() methods) to validate that no errors
   /// were accumulated during the builder method calls.
   ///
-  /// \return Status indicating success if no errors, or a ValidationFailed
-  ///         error with all accumulated error messages
+  /// \return Status indicating success if no errors, a RetryableValidationFailed if
+  ///         all accumulated errors are retryable validations, or a ValidationFailed
+  ///         error with all accumulated error messages otherwise
   [[nodiscard]] Status CheckErrors() const {
     if (!errors_.empty()) {
       std::string error_msg = "Validation failed due to the following errors:\n";
+      bool all_retryable = true;
       for (const auto& [kind, message] : errors_) {
+        all_retryable &= kind == ErrorKind::kRetryableValidationFailed;
         error_msg += "  - " + message + "\n";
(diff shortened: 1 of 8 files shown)


huan233usc (approved):
LGTM

src/iceberg/util/error_collector.h
```diff
@@ -159,14 +159,20 @@ class ICEBERG_EXPORT ErrorCollector {
...
   ///
-  /// \return Status indicating success if no errors, or a ValidationFailed
-  ///         error with all accumulated error messages
+  /// \return Status indicating success if no errors, a RetryableValidationFailed if
+  ///         all accumulated errors are retryable validations, or a ValidationFailed
+  ///         error with all accumulated error messages otherwise
   [[nodiscard]] Status CheckErrors() const {
     if (!errors_.empty()) {
       std::string error_msg = "Validation failed due to the following errors:\n";
+      bool all_retryable = true;
```
zhjwpku, line 168: I'm wondering why we require **all** the errors to be retryable before treating the result as `RetryableValidationFailed`. Java impl throws a `RetryableValidationException` immediately when it encounters a retryable validation error, will this affect the behavior of lib users?

wgtmac: The `all_retryable` check is intentional because C++ builders accumulate errors, while Java is fail-fast. In Java, a `RetryableValidationException` stops evaluation immediately, so there is no mixed-error result to classify.

In C++, if we treated the result as retryable when only the first error is retryable, a chained builder call could retry even when another collected error is deterministic and cannot be fixed by refreshing metadata. Requiring all collected errors to be retryable keeps retry limited to cases where refreshing can plausibly fix the whole update.

zhjwpku: Ok, I get it.

zhjwpku (approved):

blackmwk (approved):

wgtmac:
Thanks all for the review!

PR #796, opened 2026-07-01
feat(hive): add HmsClient connection lifecycle and URI parsing

MisterRaindrop:
- **`ParseHmsUris()`** — tolerant HMS URI parser: `thrift://host:port`,  bare `host:port`, default port 9083, and comma-separated HA lists. Rejects empty hosts and invalid ports.
- **`HmsClient::Connect()`** — sets up the `TSocket → TTransport → TBinaryProtocol → ThriftHiveMetastoreClient` chain (transport and timeouts from `HiveCatalogProperties`), translating Thrift exceptions  to `kIOError`. Thrift types stay out of the public header via a pImpl.
- **CI** — new `Ubuntu Hive` job builds with `-DICEBERG_BUILD_HIVE=ON` (bundled Thrift) and runs `hive_catalog_test`.

### DIFF
--- a/src/iceberg/catalog/hive/hms_client.cc
+++ b/src/iceberg/catalog/hive/hms_client.cc
@@ -155,8 +155,8 @@ Result<std::unique_ptr<HmsClient>> HmsClient::Connect(
         endpoints.size());
   }
   const HmsEndpoint& endpoint = endpoints.front();
-  ICEBERG_ASSIGN_OR_RAISE(auto connect_timeout_ms, config.ConnectTimeoutMs());
-  ICEBERG_ASSIGN_OR_RAISE(auto socket_timeout_ms, config.SocketTimeoutMs());
+  const int connect_timeout_ms = config.Get(HiveCatalogProperties::kConnectTimeoutMs);
+  const int socket_timeout_ms = config.Get(HiveCatalogProperties::kSocketTimeoutMs);
 
   auto socket =
       std::make_shared<apache::thrift::transport::TSocket>(endpoint.host, endpoint.port);
--- a/src/iceberg/catalog/hive/hms_client.cc
+++ b/src/iceberg/catalog/hive/hms_client.cc
@@ -1,0 +1,199 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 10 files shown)


src/iceberg/catalog/hive/hms_client.cc
```diff
@@ -0,0 +1,194 @@
...
+}
+
+Result<std::unique_ptr<HmsClient>> HmsClient::Connect(
+    const HiveCatalogProperties& config) {
+  ICEBERG_ASSIGN_OR_RAISE(auto uri, config.Uri());
+  ICEBERG_ASSIGN_OR_RAISE(auto endpoints, ParseHmsUris(uri));
+  ICEBERG_ASSIGN_OR_RAISE(auto transport_mode, config.ThriftTransport());
+
+  // HA failover beyond the first endpoint is left for a later commit.
+  const HmsEndpoint& endpoint = endpoints.front();
```
wgtmac, line 160: This silently ignores the rest of a HA URI list. If the first metastore is down, a valid fallback endpoint is never tried, while Hive's client iterates the configured URIs. Could we either try each endpoint here or reject multi-endpoint configs until failover is implemented?

src/iceberg/catalog/hive/hms_client.cc
```diff
@@ -0,0 +1,194 @@
...
+  const int connect_timeout_ms = config.Get(HiveCatalogProperties::kConnectTimeoutMs);
```
wgtmac, line 161: These Get<int>() calls can throw for malformed values, for example connect-timeout-ms=abc, so Connect() can leak an exception instead of returning InvalidArgument like the other config errors. Can we parse and validate these through Result before building the socket?

src/iceberg/catalog/hive/hms_client.cc
MisterRaindrop: I have disabled failover for now, multi endpoint will throw exception, and only a single endpoint is being processed. I plan to implement failover in last commit.

src/iceberg/catalog/hive/hms_client.cc
MisterRaindrop: It's processed.

MisterRaindrop:
@wgtmac It is ready, please review it

wgtmac (approved):

PR #800, opened 2026-07-02
feat(avro): apply column default values when reading missing fields

huan233usc:
## What

Part 3 of 4 of Iceberg v3 column default-value support (POC #731), built on the
schema layer (#746) and the Parquet read path (#792).

When a column is present in the read (table) schema but absent from an Avro data
file — because the column was added after those rows were written — fill it with
the column's v3 `initial-default` instead of `null`.

## Changes

- **Avro projection** (`avro_schema_util.cc`): when a field is missing from the
  file and carries an `initial-default`, project it as
  `FieldProjection::Kind::kDefault`, mirroring the generic / Parquet paths.
- **Avro decode** (`avro_data_util.cc`, `avro_direct_decoder.cc`): materialize
  the `kDefault` branch through an Avro-local `AppendDefaultToBuilder` helper.
  It reuses the shared `ToArrowScalar` conversion, while keeping Avro's
  row-by-row `ArrayBuilder` append behavior out of the shared Arrow utility.

## Tests

- `avro_data_test`: `AppendDefaultToBuilder` appends a value and casts to the
  builder type; `AppendDatumToBuilder` fills missing required and optional
  default fields.
- `avro_test`: end-to-end — write an Avro file with an old schema, then read it
  through `ReaderFactoryRegistry` with an evolved schema carrying defaults
  (`ReadMissingFieldsWithDefaults`).

## Stack

2. #792 — read path: Parquet (merged)
3. **this PR** — read path: Avro
4. schema evolution: `addColumn` / `updateColumnDefault` (#793)

Draft while the earlier PRs in the stack settle.

### DIFF
--- a/src/iceberg/avro/avro_data_util.cc
+++ b/src/iceberg/avro/avro_data_util.cc
@@ -23,6 +23,7 @@
 #include <arrow/array/builder_primitive.h>
 #include <arrow/extension_type.h>
 #include <arrow/json/from_string.h>
+#include <arrow/scalar.h>
 #include <arrow/type.h>
 #include <arrow/util/decimal.h>
 #include <avro/Generic.hh>
@@ -90,7 +91,7 @@ Status AppendStructToBuilder(const ::avro::NodePtr& avro_node,
     } else if (field_projection.kind == FieldProjection::Kind::kNull) {
       ICEBERG_ARROW_RETURN_NOT_OK(field_builder->AppendNull());
     } else if (field_projection.kind == FieldProjection::Kind::kDefault) {
-      ICEBERG_RETURN_UNEXPECTED(arrow::AppendDefaultToBuilder(
+      ICEBERG_RETURN_UNEXPECTED(AppendDefaultToBuilder(
           std::get<Literal>(field_projection.from), field_builder));
     } else if (field_projection.kind == FieldProjection::Kind::kMetadata) {
       int32_t field_id = expected_field.field_id();
@@ -471,8 +472,7 @@ Status AppendFieldToBuilder(const ::avro::NodePtr& avro_node,
   }
 
   if (projection.kind == FieldProjection::Kind::kDefault) {
-    return arrow::AppendDefaultToBuilder(std::get<Literal>(projection.from),
-                                         array_builder);
+    return AppendDefaultToBuilder(std::get<Literal>(projection.from), array_builder);
   }
 
   const bool is_row_lineage =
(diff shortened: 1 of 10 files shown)


src/iceberg/avro/avro_data_util.cc
```diff
@@ -497,6 +506,29 @@ Status AppendFieldToBuilder(const ::avro::NodePtr& avro_node,
 
 }  // namespace
 
+Status AppendDefaultToBuilder(const Literal& literal, ::arrow::ArrayBuilder* builder) {
+  // The builder's own memory pool is not exposed, so the small scalar buffer uses the
+  // default pool.
+  ICEBERG_ASSIGN_OR_RAISE(std::shared_ptr<::arrow::Scalar> scalar,
+                          arrow::ToArrowScalar(literal, ::arrow::default_memory_pool()));
+
```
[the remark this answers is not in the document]
huan233usc: Fixed in 0e4c6a40: added `PrepareDefaultScalars` to cache a cast Arrow scalar in `FieldProjection::attributes` (`AvroDefaultAttributes`) when the Avro reader initializes; per-row default fills now only call `AppendScalar`.

src/iceberg/avro/avro_data_util.cc
```diff
@@ -497,6 +507,128 @@ Status AppendFieldToBuilder(const ::avro::NodePtr& avro_node,
...
+      case ::arrow::Type::STRUCT:
+        ICEBERG_RETURN_UNEXPECTED(
+            PrepareDefaultScalars(field_projection.children, field_builder));
+        break;
+      case ::arrow::Type::LIST: {
+        // List projections store a single child for the element. Defaults only appear
+        // on nested struct fields of that element.
+        auto* list_builder = internal::checked_cast<::arrow::ListBuilder*>(field_builder);
+        auto& element_projection = field_projection.children[0];
+        if (element_projection.kind == FieldProjection::Kind::kProjected &&
```
wgtmac: Please recurse through nested list and map builders, not just an immediate struct. For example, a default under `list<list<struct<...>>>` is not cached and is rebuilt for every element. Please add a nested-collection test as well.

huan233usc: Done — the recursion now dispatches on the builder type (struct/list/large_list/map) instead of only descending when the collection child is an immediate struct, so a default under `list<list<struct<...>>>` gets its scalar prepared once at any depth. Added a nested-collection test asserting the scalar is prepared, plus an end-to-end `list<list<struct>>` decode test (the former fails without the change).

src/iceberg/avro/avro_data_util_internal.h
```diff
@@ -19,14 +19,48 @@
...
+#include "iceberg/expression/literal.h"
 #include "iceberg/schema_util.h"
 
 namespace iceberg::avro {
 
+/// \brief Cached Arrow scalar for a `kDefault` field projection.
+///
+/// Materialized once (see `PrepareDefaultScalars`) so row-by-row Avro decode only
+/// needs `AppendScalar` instead of repeating `ToArrowScalar` / `CastTo` per row.
+struct AvroDefaultAttributes : FieldProjection::ExtraAttributes {
```
wgtmac: Could this follow the `ParquetExtraAttributes` pattern and become `AvroExtraAttributes` with a `default_scalar` member? `FieldProjection` has a single attributes slot, so separate subclasses for future Avro attributes would not compose. A single Avro-specific container would also let the append path use `checked_cast` instead of a per-row `dynamic_cast`.

huan233usc: Done — renamed to `AvroExtraAttributes` with a `default_scalar` member, mirroring `ParquetExtraAttributes`. Since Avro now has a single attributes container, the append path uses `checked_cast` instead of a per-row `dynamic_cast`, and future Avro attributes can be added to the same struct rather than as competing subclasses.

src/iceberg/avro/avro_data_util.cc
```diff
@@ -497,6 +507,135 @@ Status AppendFieldToBuilder(const ::avro::NodePtr& avro_node,
...
+        "Inconsistent number of struct builder fields ({}) and projections ({})",
+        struct_builder->num_fields(), projections.size());
+  }
+
+  for (size_t i = 0; i < projections.size(); ++i) {
+    auto& field_projection = projections[i];
+    auto* field_builder = struct_builder->field_builder(static_cast<int>(i));
+
+    if (field_projection.kind == FieldProjection::Kind::kDefault) {
+      if (field_projection.attributes != nullptr) {
```
wgtmac: `attributes != nullptr` does not guarantee that `default_scalar` is prepared now that this container may hold other Avro attributes. If another attribute creates the container first, this skips preparation and the append path falls back to rebuilding the scalar per row. Please reuse/cast the existing `AvroExtraAttributes`, check `default_scalar` instead, and add a test with a pre-existing empty container.

wgtmac (approved):

wgtmac:
Thanks for taking feedback quickly!

PR #801, opened 2026-07-02
feat(inspect): implement streaming SnapshotsTable scans

WZhuo:
## Summary

Implement streaming scans for `SnapshotsTable` and refine the metadata-table scan APIs so snapshot metadata can be consumed safely in bounded Arrow batches.

### Metadata table APIs

- Add a typed `MetadataTable::Make<T>()` factory that preserves the concrete metadata-table type.
- Keep the base `MetadataTable` interface limited to non-time-travel `Scan()` calls.
- Add `TimeTravelMetadataTable` as the capability-specific interface for scans using `SnapshotSelection`.
- Model snapshot ID and as-of timestamp as mutually exclusive values with `std::variant<std::monostate, int64_t, TimePointMs>`.
- Return `ArrowArrayStream` from scan APIs and document stream ownership.
- Provide shared metadata-table schema and source-table accessors.

### Snapshots table

- Implement `SnapshotsTable::Scan()` using a stateful Arrow stream.
- Convert the Iceberg schema to `ArrowSchema` once when creating the stream and release owned resources when the stream closes.
- Emit at most `MetadataTable::kBatchSize` rows per batch.
- Skip null snapshot entries safely.
- Materialize `committed_at`, `snapshot_id`, `parent_id`, `operation`, `manifest_list`, and `summary` using `ArrowRowBuilder`.
- Remove `SnapshotSummaryFields::kOperation` from the summary map and emit null when the summary is empty before remove `kOperation`.

### Arrow row builder

- Add `ArrowRowBuilder::num_rows()` so streaming producers can enforce batch-size limits without maintaining duplicate row counters.

### Tests

- Add shared metadata-table fixtures and Arrow stream import helpers.
- Cover snapshots schema and row values, summary filtering, null snapshots, empty summaries, empty tables, and multi-batch scans.
- Cover metadata-table construction, time-travel capability reporting, and history-table schema behavior.

## Testing

```shell
cmake --build build --target metadata_table_test -j2
build/src/iceberg/test/metadata_table_test
```

### DIFF
--- a/src/iceberg/inspect/history_table.cc
+++ b/src/iceberg/inspect/history_table.cc
@@ -26,37 +26,32 @@
 #include "iceberg/schema.h"
 #include "iceberg/schema_field.h"
 #include "iceberg/table.h"
-#include "iceberg/table_identifier.h"
 #include "iceberg/type.h"
+#include "iceberg/util/macros.h"
 
 namespace iceberg {
-namespace {
 
-std::shared_ptr<Schema> MakeHistoryTableSchema() {
-  return std::make_shared<Schema>(std::vector<SchemaField>{
+HistoryTable::HistoryTable(std::shared_ptr<Table> table)
+    : MetadataTable(std::move(table)) {}
+
+HistoryTable::~HistoryTable() = default;
+
+const std::shared_ptr<Schema>& HistoryTable::schema() const {
+  static const auto schema = std::make_shared<Schema>(std::vector<SchemaField>{
       SchemaField::MakeRequired(1, "made_current_at", timestamp_tz()),
       SchemaField::MakeRequired(2, "snapshot_id", int64()),
       SchemaField::MakeOptional(3, "parent_id", int64()),
       SchemaField::MakeRequired(4, "is_current_ancestor", boolean())});
+  return schema;
 }
-
-TableIdentifier MakeHistoryTableName(const TableIdentifier& source_name) {
-  return TableIdentifier{.ns = source_name.ns, .name = source_name.name + ".history"};
-}
-
-}  // namespace
-
-HistoryTable::HistoryTable(std::shared_ptr<Table> table)
-    : MetadataTable(table, MakeHistoryTableName(table->name()),
-                    MakeHistoryTableSchema()) {}
-
-HistoryTable::~HistoryTable() = default;
 
 Result<std::unique_ptr<HistoryTable>> HistoryTable::Make(std::shared_ptr<Table> table) {
-  if (table == nullptr) [[unlikely]] {
-    return InvalidArgument("Table cannot be null");
-  }
+  ICEBERG_PRECHECK(table != nullptr, "Table cannot be null");
   return std::unique_ptr<HistoryTable>(new HistoryTable(std::move(table)));
+}
+
+Result<ArrowArrayStream> HistoryTable::Scan() {
+  return NotSupported("Scan is not supported for the history table");
 }
 
 }  // namespace iceberg
--- a/src/iceberg/inspect/metadata_table.h
+++ b/src/iceberg/inspect/metadata_table.h
@@ -23,13 +23,27 @@
 /// \brief Define base APIs for metadata tables.
 
 #include <memory>
+#include <optional>
+#include <string>
 
+#include "iceberg/arrow_c_data.h"
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
 #include "iceberg/table_identifier.h"
 #include "iceberg/type_fwd.h"
+#include "iceberg/util/timepoint.h"
 
 namespace iceberg {
+
+/// \brief Parameters for snapshot selection (time travel).
+struct SnapshotSelection {
+  /// \brief The snapshot ID to read.
+  std::optional<int64_t> snapshot_id;
+  /// \brief Read the snapshot that was current at this timestamp.
+  std::optional<TimePointMs> as_of_timestamp;
+  /// \brief Read the snapshot referenced by this named ref (branch or tag).
+  std::optional<std::string> ref_name;
+};
 
 /// \brief Base class for Iceberg metadata tables.
 class ICEBERG_EXPORT MetadataTable {
@@ -45,6 +59,28 @@ class ICEBERG_EXPORT MetadataTable {
   virtual ~MetadataTable();
 
   virtual Kind kind() const noexcept = 0;
+
+  /// \brief Whether this metadata table supports time-travel queries.
+  ///
+  /// Time travel is supported for tables that read from a single snapshot's
+  /// manifests (e.g., Entries, Files, Manifests, Partitions). Tables that
+  /// scan all snapshots (All*) or return in-memory history (Snapshots,
+  /// History, Refs) do not support time travel.
+  bool supports_time_travel() const noexcept;
+
+  /// \brief Scan the metadata table using the current snapshot.
+  ///
+  /// Convenience overload — delegates to Scan(std::nullopt).
+  Result<ArrowArray> Scan() { return Scan(std::nullopt); }
+
+  /// \brief Scan the metadata table and return all rows as an Arrow struct array.
+  ///
+  /// The returned ArrowArray is a struct array where each element is one row.
+  /// The caller takes ownership and must call ArrowArrayRelease when done.
+  ///
+  /// The default implementation returns NotSupported. Subclasses override this
+  /// to materialize their data.
+  virtual Result<ArrowArray> Scan(std::optional<SnapshotSelection> snapshot_selection);
 
   const TableIdentifier& name() const { return identifier_; }
 
--- a/src/iceberg/inspect/metadata_table.h
+++ b/src/iceberg/inspect/metadata_table.h
@@ -62,10 +62,8 @@ class ICEBERG_EXPORT MetadataTable {
 
   /// \brief Whether this metadata table supports time-travel queries.
   ///
-  /// Time travel is supported for tables that read from a single snapshot's
-  /// manifests (e.g., Entries, Files, Manifests, Partitions). Tables that
-  /// scan all snapshots (All*) or return in-memory history (Snapshots,
-  /// History, Refs) do not support time travel.
+  /// The currently supported snapshots and history metadata tables do not
+  /// support time travel.
   bool supports_time_travel() const noexcept;
 
   /// \brief Scan the metadata table using the current snapshot.
@@ -80,7 +78,8 @@ class ICEBERG_EXPORT MetadataTable {
   ///
   /// The default implementation returns NotSupported. Subclasses override this
   /// to materialize their data.
-  virtual Result<ArrowArray> Scan(std::optional<SnapshotSelection> snapshot_selection);
+  virtual Result<ArrowArray> Scan(
+      const std::optional<SnapshotSelection>& snapshot_selection);
 
   const TableIdentifier& name() const { return identifier_; }
 
--- a/src/iceberg/inspect/metadata_table.h
+++ b/src/iceberg/inspect/metadata_table.h
@@ -20,81 +20,108 @@
 #pragma once
 
 /// \file iceberg/inspect/metadata_table.h
-/// \brief Define base APIs for metadata tables.
+/// \brief Base APIs for inspecting Iceberg metadata tables.
 
+#include <concepts>
 #include <memory>
-#include <optional>
 #include <string>
+#include <utility>
+#include <variant>
 
 #include "iceberg/arrow_c_data.h"
 #include "iceberg/iceberg_export.h"
 #include "iceberg/result.h"
-#include "iceberg/table_identifier.h"
 #include "iceberg/type_fwd.h"
 #include "iceberg/util/timepoint.h"
 
 namespace iceberg {
 
-/// \brief Parameters for snapshot selection (time travel).
-struct SnapshotSelection {
-  /// \brief The snapshot ID to read.
-  std::optional<int64_t> snapshot_id;
-  /// \brief Read the snapshot that was current at this timestamp.
-  std::optional<TimePointMs> as_of_timestamp;
-  /// \brief Read the snapshot referenced by this named ref (branch or tag).
-  std::optional<std::string> ref_name;
-};
-
-/// \brief Base class for Iceberg metadata tables.
+/// \brief Base interface for an Iceberg metadata table.
 class ICEBERG_EXPORT MetadataTable {
  public:
+  /// \brief Supported metadata table kinds.
   enum class Kind {
     kSnapshots,
     kHistory,
   };
 
-  static Result<std::unique_ptr<MetadataTable>> Make(std::shared_ptr<Table> table,
-                                                     Kind kind);
+  /// \brief Maximum number of rows emitted in each Arrow batch.
+  static constexpr int64_t kBatchSize = 1024;
+
+  /// \brief Create a metadata table of the requested concrete type.
+  ///
+  /// \tparam MetadataTableType Concrete class derived from MetadataTable.
+  /// \param table Source table whose metadata will be exposed.
+  /// \return The constructed metadata table, or an error.
+  template <typename MetadataTableType>
+    requires std::derived_from<MetadataTableType, MetadataTable>
+  static Result<std::unique_ptr<MetadataTableType>> Make(std::shared_ptr<Table> table) {
+    return MetadataTableType::Make(std::move(table));
+  }
 
   virtual ~MetadataTable();
 
+  /// \brief Return this metadata table's kind.
   virtual Kind kind() const noexcept = 0;
 
-  /// \brief Whether this metadata table supports time-travel queries.
+  /// \brief Return the schema of rows emitted by scans.
+  virtual const std::shared_ptr<Schema>& schema() const = 0;
+
+  /// \brief Return the source table whose metadata is exposed.
+  const std::shared_ptr<Table>& source_table() const;
+
+  /// \brief Return whether this metadata table supports time travel.
+  virtual bool supports_time_travel() const noexcept;
+
+  /// \brief Scan the metadata table without time travel.
   ///
-  /// The currently supported snapshots and history metadata tables do not
-  /// support time travel.
-  bool supports_time_travel() const noexcept;
-
-  /// \brief Scan the metadata table using the current snapshot.
-  ///
-  /// Convenience overload — delegates to Scan(std::nullopt).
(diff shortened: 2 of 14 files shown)


src/iceberg/inspect/metadata_table.h
```diff
@@ -23,14 +23,28 @@
...
+#include "iceberg/util/timepoint.h"
 
 namespace iceberg {
 
+/// \brief Parameters for snapshot selection (time travel).
+struct SnapshotSelection {
+  /// \brief The snapshot ID to read.
+  std::optional<int64_t> snapshot_id;
+  /// \brief Read the snapshot that was current at this timestamp.
+  std::optional<TimePointMs> as_of_timestamp;
```
wgtmac: Should it be `std::variant<std::monostate, int64_t, TimePointMs>`? Otherwise we need to define the priority if both `snapshot_id` and `as_of_timestamp` are set.

src/iceberg/inspect/metadata_table.h
```diff
@@ -23,14 +23,28 @@
...
+  /// \brief Read the snapshot referenced by this named ref (branch or tag).
+  std::optional<std::string> ref_name;
```
wgtmac: Is `std::string` enough? An empty one is unset.

src/iceberg/inspect/metadata_table.h
```diff
@@ -46,6 +60,27 @@ class ICEBERG_EXPORT MetadataTable {
...
+
+  /// \brief Scan the metadata table and return all rows as an Arrow struct array.
+  ///
+  /// The returned ArrowArray is a struct array where each element is one row.
+  /// The caller takes ownership and must call ArrowArrayRelease when done.
+  ///
+  /// The default implementation returns NotSupported. Subclasses override this
+  /// to materialize their data.
+  virtual Result<ArrowArray> Scan(
+      const std::optional<SnapshotSelection>& snapshot_selection);
```
wgtmac: It doesn't seem necessary to add a std::optional on top of SnapshotSelection if all its values are optional.

src/iceberg/inspect/metadata_table.h
```diff
@@ -46,6 +60,27 @@ class ICEBERG_EXPORT MetadataTable {
...
+  Result<ArrowArray> Scan() { return Scan(std::nullopt); }
+
+  /// \brief Scan the metadata table and return all rows as an Arrow struct array.
+  ///
+  /// The returned ArrowArray is a struct array where each element is one row.
+  /// The caller takes ownership and must call ArrowArrayRelease when done.
+  ///
+  /// The default implementation returns NotSupported. Subclasses override this
+  /// to materialize their data.
+  virtual Result<ArrowArray> Scan(
```
wgtmac: Should we return `ArrowArrayStream` instead? My concern is that a single ArrowArray may lead to OOM if the result is too large. `ArrowArrayStream` is an iterator of `ArrowArray`s which is more scalable.

src/iceberg/inspect/snapshots_table.cc
```diff
@@ -65,4 +68,50 @@ Result<std::unique_ptr<SnapshotsTable>> SnapshotsTable::Make(
...
+      ICEBERG_RETURN_UNEXPECTED(AppendString(builder.column(3), *op));
+    } else {
+      ICEBERG_RETURN_UNEXPECTED(AppendNull(builder.column(3)));
+    }
+
+    // column 4: manifest_list (string, optional)
+    ICEBERG_RETURN_UNEXPECTED(AppendString(builder.column(4), snapshot->manifest_list));
+
+    // column 5: summary (map<string,string>)
+    ICEBERG_RETURN_UNEXPECTED(AppendStringMap(builder.column(5), snapshot->summary));
```
wgtmac: `Snapshot::summary` is not equivalent to Java’s `Snapshot.summary()` here. In C++, the map still contains the `operation` entry, while Java stores the operation separately and `SnapshotsTable.snapshotToRow` emits the remaining summary map. As a result, every normal row duplicates `operation` in both columns (the copied Java fixture has 10 summary entries, not 11). Please omit `SnapshotSummaryFields::kOperation` when materializing this map, while preserving null versus empty-map semantics.

src/iceberg/inspect/snapshots_table.cc
```diff
@@ -65,4 +68,50 @@ Result<std::unique_ptr<SnapshotsTable>> SnapshotsTable::Make(
   return std::unique_ptr<SnapshotsTable>(new SnapshotsTable(std::move(table)));
 }
 
+Result<ArrowArray> SnapshotsTable::Scan(
+    const std::optional<SnapshotSelection>& /*snapshot_selection*/) {
```
wgtmac: Silently ignoring an invalid selection diverges from Java. `SnapshotScan.useSnapshot`, `useRef`, and `asOfTime` validate or resolve the requested snapshot before `SnapshotsTable` returns all known snapshots. With this implementation, `Scan({.snapshot_id = 999})` succeeds and returns unrelated rows. Please validate the selector (or reject a non-null selection as unsupported) instead of treating invalid input as an unqualified scan.

src/iceberg/inspect/snapshots_table.h
```diff
@@ -40,6 +40,13 @@ class ICEBERG_EXPORT SnapshotsTable : public MetadataTable {
 
   Kind kind() const noexcept override { return Kind::kSnapshots; }
 
+  /// \brief Scan all snapshots as rows.
+  ///
+  /// The snapshots table always returns every known snapshot, so the
+  /// snapshot_selection parameter is ignored.
+  Result<ArrowArray> Scan(
```
wgtmac: This override hides the zero-argument `MetadataTable::Scan()` overload under C++ name lookup. A caller using the public `SnapshotsTable::Make()` result cannot write `snapshots_table->Scan()`; only callers typed as `MetadataTable` compile, which is why the current tests miss it. Please add `using MetadataTable::Scan;` (or an explicit zero-argument overload).

src/iceberg/inspect/metadata_table.h
WZhuo: Done. `SnapshotSelection` now uses `std::variant<std::monostate, int64_t, TimePointMs>`, so snapshot ID and as-of timestamp are mutually exclusive.

src/iceberg/inspect/metadata_table.h
WZhuo: Yes. `ref_name` is now a `std::string`; an empty value selects the main branch.

src/iceberg/inspect/metadata_table.h
WZhuo: Done. The outer `std::optional` has been removed. A default `SnapshotSelection` uses `std::monostate` and an empty ref to represent the current snapshot on the main branch.

src/iceberg/inspect/metadata_table.h
WZhuo: Done. `Scan()` now returns `ArrowArrayStream`, and the snapshots stream emits batches capped at `MetadataTable::kBatchSize` (1024 rows).

src/iceberg/inspect/snapshots_table.cc
WZhuo: Done. `SnapshotSummaryFields::kOperation` is removed before materializing `summary`. The current implementation emits null when the remaining map is empty, and tests cover both operation omission and empty-summary handling.

src/iceberg/inspect/snapshots_table.cc
WZhuo: Done. Snapshot selection is now exposed only by `TimeTravelMetadataTable`. `SnapshotsTable` inherits `MetadataTable` and exposes only the zero-argument `Scan()`, so a selector cannot be silently ignored.

src/iceberg/inspect/snapshots_table.h
WZhuo: Done. `SnapshotsTable` now explicitly overrides the zero-argument `Scan()`, so callers using `SnapshotsTable::Make()` or `MetadataTable::Make<SnapshotsTable>()` can call it directly.

src/iceberg/inspect/metadata_table.h
```diff
@@ -20,46 +20,109 @@
...
+  /// \brief Create a metadata table of the requested concrete type.
+  ///
+  /// \tparam MetadataTableType Concrete class derived from MetadataTable.
+  /// \param table Source table whose metadata will be exposed.
+  /// \return The constructed metadata table, or an error.
+  template <typename MetadataTableType>
+    requires std::derived_from<MetadataTableType, MetadataTable>
+  static Result<std::unique_ptr<MetadataTableType>> Make(std::shared_ptr<Table> table) {
+    return MetadataTableType::Make(std::move(table));
+  }
```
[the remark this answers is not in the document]
WZhuo: `MetadataTable` has not been released as a public API yet, so there are no downstream compatibility guarantees to preserve. This is an intentional pre-release API refinement: the typed factory preserves the concrete capability-specific type, and the redundant identifier state was removed before the API is released. No deprecation path is needed at this stage.

src/iceberg/inspect/metadata_table.h
```diff
@@ -20,46 +20,108 @@
...
+  bool supports_time_travel() const noexcept final;
+
+  /// \brief Scan using the current snapshot on the main branch.
+  Result<ArrowArrayStream> Scan() final;
+
+  /// \brief Scan using the requested snapshot selection.
+  ///
+  /// \param snapshot_selection Snapshot ID, timestamp, and optional ref selection.
+  /// \return An Arrow stream containing the metadata table rows, or an error.
+  Result<ArrowArrayStream> Scan(const SnapshotSelection& snapshot_selection);
```
wgtmac, line 120: Is it better to remove this extra layer of `TimeTravelMetadataTable`? We can check `supports_time_travel()` and by default `Scan(const SnapshotSelection& snapshot_selection)` can just return `NotSupported`.

src/iceberg/inspect/metadata_table.h
```diff
@@ -20,46 +20,108 @@
...
 
-  const std::shared_ptr<Schema>& schema() const { return schema_; }
+  /// \brief Return whether this metadata table supports time travel.
+  virtual bool supports_time_travel() const noexcept;
 
-  const std::shared_ptr<Table>& source_table() const { return source_table_; }
+  /// \brief Scan the metadata table without time travel.
+  ///
+  /// The caller owns the returned stream and must release it with
+  /// ArrowArrayStreamRelease.
```
wgtmac: ArrowArrayStreamRelease is not declared by this repository's Arrow C Data API; callers must invoke the stream's release callback. Please correct the ownership documentation.

src/iceberg/inspect/snapshots_table.cc
```diff
@@ -19,21 +19,133 @@
...
+  Status Close() {
+    table_.reset();
+    if (arrow_schema_.release != nullptr) {
+      ArrowSchemaRelease(&arrow_schema_);
+    }
+    return {};
+  }
+
+  Result<std::optional<ArrowArray>> Next() {
+    const auto& snapshots = table_->snapshots();
```
wgtmac: This stream re-reads the mutable Table on every batch. If Table::Refresh() runs between get_next calls, later batches can come from a different snapshot list, skipping or truncating rows. Please capture the metadata or snapshot list when Scan() creates the stream.

src/iceberg/inspect/history_table.cc
```diff
@@ -26,37 +26,32 @@
...
-
 Result<std::unique_ptr<HistoryTable>> HistoryTable::Make(std::shared_ptr<Table> table) {
-  if (table == nullptr) [[unlikely]] {
-    return InvalidArgument("Table cannot be null");
-  }
+  ICEBERG_PRECHECK(table != nullptr, "Table cannot be null");
   return std::unique_ptr<HistoryTable>(new HistoryTable(std::move(table)));
 }
 
+Result<ArrowArrayStream> HistoryTable::Scan() {
```
wgtmac, line 53: HistoryTable remains constructible, but the new common Scan() API always returns NotSupported here. Java HistoryTable scans table.history() into rows. Please implement a streaming history scan, or keep non-scannable tables out of this interface.

src/iceberg/inspect/snapshots_table.cc
```diff
@@ -19,21 +19,133 @@
...
+ public:
+  static Result<std::unique_ptr<SnapshotsTableStream>> Make(
+      std::shared_ptr<Table> table, const iceberg::Schema& schema) {
+    ArrowSchema arrow_schema{};
+    ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(schema, &arrow_schema));
+    return std::unique_ptr<SnapshotsTableStream>(
+        new SnapshotsTableStream(std::move(table), std::move(arrow_schema)));
+  }
+
+  ~SnapshotsTableStream() { std::ignore = Close(); }
```
wgtmac: std::ignore is used here without a direct <tuple> include. Please add <tuple> or replace std::ignore so this file does not rely on transitive includes.

src/iceberg/inspect/metadata_table.h
WZhuo: Keeping `TimeTravelMetadataTable` is intentional. It keeps `Scan(const SnapshotSelection&)` out of `MetadataTable` and `SnapshotsTable`, so metadata tables without time-travel support cannot expose or accept snapshot selectors. `supports_time_travel()` remains available for capability introspection.

src/iceberg/inspect/metadata_table.h
WZhuo: Updated the ownership documentation to state that callers must invoke the returned stream's `release` callback. `ArrowArrayStreamRelease` is not part of this repository's Arrow C Data API.

src/iceberg/inspect/snapshots_table.cc
WZhuo: Fixed by capturing a `shared_ptr<const TableMetadata>` when `Scan()` creates the stream. Subsequent batches now use the same metadata snapshot even if the source `Table` is refreshed. Added a test covering refresh between batches.

src/iceberg/inspect/history_table.cc
WZhuo: HistoryTable scanning is intentionally deferred to the next PR, as requested. It remains constructible but returns `NotSupported` for `Scan()` in this change.

src/iceberg/inspect/snapshots_table.cc
WZhuo: Kept `std::ignore = Close()` and added the direct `<tuple>` include.

src/iceberg/test/CMakeLists.txt
```diff
@@ -179,7 +186,12 @@ if(ICEBERG_BUILD_BUNDLE)
 
   add_iceberg_test(catalog_test USE_BUNDLE SOURCES in_memory_catalog_test.cc)
 
-  add_iceberg_test(metadata_table_test USE_BUNDLE SOURCES metadata_table_test.cc)
+  add_iceberg_test(metadata_table_test
```
manuzhang, line 189: these tests need to be added to the meson build file as well.

src/iceberg/inspect/snapshots_table.cc
```diff
@@ -19,21 +19,139 @@
...
+  } else {
+    ICEBERG_RETURN_UNEXPECTED(AppendNull(builder.column(3)));
+  }
+
+  ICEBERG_RETURN_UNEXPECTED(AppendString(builder.column(4), snapshot.manifest_list));
+
+  auto summary = snapshot.summary;
+  summary.erase(SnapshotSummaryFields::kOperation);
+  if (summary.empty()) {
+    ICEBERG_RETURN_UNEXPECTED(AppendNull(builder.column(5)));
```
manuzhang, line 73: When the summary is operation only, `null` will be appended, same as when the summary is absent.

manuzhang:
@WZhuo Can you check latest review comments? I think it's getting close to merge and I have a follow-up PR #876 that depends on this.

src/iceberg/inspect/metadata_table.h
```diff
@@ -20,46 +20,111 @@
...
-  const std::shared_ptr<Schema>& schema() const { return schema_; }
+  /// \brief Return whether this metadata table supports time travel.
+  virtual bool supports_time_travel() const noexcept;
 
-  const std::shared_ptr<Table>& source_table() const { return source_table_; }
+  /// \brief Scan the metadata table without time travel.
+  ///
+  /// The caller owns the returned stream and must invoke its `release` callback
+  /// when the stream is no longer needed.
+  virtual Result<ArrowArrayStream> Scan() = 0;
```
manuzhang, line 80: It would be better to decouple with Arrow types.

WZhuo:
> @WZhuo Can you check latest review comments? I think it's getting close to merge and I have a follow-up PR #876 that depends on this.

Ok, I'll fix the comments

src/iceberg/inspect/metadata_table.h
WZhuo: Agreed. Decoupling the metadata table API from Arrow C types would be cleaner. This likely deserves a metadata-specific batch reader abstraction rather than reusing the file-oriented `Reader` interface. Since that is a broader API change, I would prefer to address it in a follow-up PR and keep `ArrowArrayStream` for this PR.

manuzhang: Well, this defines the basic interface for all metadata tables.  I think it will be better to get it right first, maybe in a separate PR. @wgtmac What do you think?

src/iceberg/test/snapshots_table_test.cc
```diff
@@ -0,0 +1,230 @@
...
+                         MetadataTable::Make<SnapshotsTable>(table));
+
+  ICEBERG_UNWRAP_OR_FAIL(auto stream, snapshots_table->Scan());
+  ICEBERG_UNWRAP_OR_FAIL(auto batches, ReadAllBatches(std::move(stream)));
+  ASSERT_EQ(batches.size(), 1);
+  auto summaries =
+      std::static_pointer_cast<::arrow::MapArray>(batches.front()->column(5));
+  EXPECT_TRUE(summaries->IsNull(0));
+  EXPECT_FALSE(summaries->IsNull(1));
+  EXPECT_EQ(summaries->value_length(1), 0);
```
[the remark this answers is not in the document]
WZhuo: The PR description has been modified

wgtmac (approved):
Thanks @WZhuo for working on this and @manuzhang for the feedback! I think metadata table support is still in its initial phase. Let me merge this to unblock followup PRs. I also have some ideas about its design and will work on that after we have sufficient implementations.

Issue #799, opened 2026-07-02
Cutting CI build times: prebuilt Arrow and build configuration


abnobdoss:
I have some CI changes I'd like to propose. They go a bit deeper than the recent caching work, so I wanted to open an issue first to align on direction before pursuing any of them. The pieces below are independent; taking forward only some of them works fine.

## Resolving Arrow from conda-forge

Eleven jobs build the vendored Arrow and Parquet from source, but the toolchain already prefers a prebuilt one (the FetchContent declaration uses FIND_PACKAGE_ARGS); CI just never installs one. Installing libarrow and libparquet from conda-forge with setup-miniconda (ASF allowlisted) takes that compile out of each leg:

| Test leg | Build targets | Cold build step |
|---|---|---|
| Ubuntu | 951 to 606 | 21 to 16 min |
| macOS | 946 to 602 | 11 to 8 min |
| Windows | 900 to 578 | 47 to 35 min |

(fork measurements; times vary with runner load, target counts don't; the AWS, SQL catalog, sanitizer, and linter legs shrink similarly)

It is also most of our cache pressure: one push to main saves around 10 GB of sccache entries, more than the repository's 10 GB limit by itself, so eviction ends up deleting entries that have no newer replacement. Over the past week about every third push to main rebuilt at least one leg from scratch. With prebuilt Arrow, and without debug info that nothing in CI reads, saves drop under 3 GB.

Coverage holds: the AWS leg keeps building the bundled AWS SDK from source, the Meson legs don't use Arrow, and the sanitizer leg passes against the non-instrumented Arrow. The conda pin would track the version in the toolchain file and bump in the same PR.

## Building one library flavor per leg

CI builds with ICEBERG_BUILD_STATIC and ICEBERG_BUILD_SHARED both ON, and the two targets compile the same sources twice (the shared build adds the export define and hidden visibility, so objects can't be reused). The tests link one flavor, so building one roughly halves what a leg compiles of our own code. A static-only fork run passes the full build and test matrix; one leg could keep both ON to keep both exercised.

## Windows: build Debug like the other legs

The Unix test legs build Debug; Windows builds Release and has been the slowest leg fairly consistently. MSVC Debug needs embedded debug info (/Z7 via CMP0141) for sccache to cache the objects, and that combination is green on a fork. Windows is currently the only Release build in CI, though, so this is partly a question of what the matrix should cover.

## Two smaller cleanups

[...]

PR #802, opened 2026-07-03
ci: scope push trigger to main to drop redundant PR runs

abnobdoss:
## Change
Make push triggers more consistent by having the test workflows run branch pushes only on `main`. Pull requests still run the same checks, and tag builds are unchanged.

The duplicate-run case is narrow: it only happens for branches pushed directly to the Apache repo, so it mostly affects committers and bots rather than normal fork-based contributors. Pull request checks cover those branch updates; direct `main` runs still give us post-merge CI and warm caches.

Also cleans up nearby `on:` sections so the triggers are easier to compare, including removing `zizmor`'s redundant all-branches pull request filter.

## Validation
Checked that all workflow YAML files still parse.

## Remaining Trigger Inconsistencies
These are intentional or out of scope, so this PR leaves them alone:

- Test workflows still run on all tag pushes to preserve release behavior.
- `rc` only runs on release-candidate tags.
- `pre-commit` still runs on regular branch pushes because it is cheap.
- `docs`, `cpp-linter`, `license_check`, and `codeql` keep their narrower rules.

Related discussion: apache/iceberg-cpp#799.

### DIFF
--- a/.github/workflows/asf-allowlist-check.yml
+++ b/.github/workflows/asf-allowlist-check.yml
@@ -24,11 +24,11 @@
 name: "ASF Allowlist Check"
 
 on:
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
   push:
     branches:
       - main
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 permissions:
   contents: read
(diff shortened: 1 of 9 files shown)


zhjwpku (approved):
Reasonable, thanks for keeping improve the CI workflows :-)

wgtmac (approved):
Thanks @abnobdoss and @zhjwpku!

PR #804, opened 2026-07-03
fix: generate public API docs correctly

zhjwpku:


zhjwpku:
This patch resolves the following doc issues:

image

image

src/iceberg/arrow/s3/s3_properties.h
```diff
@@ -19,6 +19,9 @@
 
 #pragma once
 
+/// \file iceberg/arrow/s3/s3_properties.h
```
manuzhang, line 22: are these comments required for this fix?

zhjwpku: I guess so. If not for this comment on the header files, the header would be listed but wouldn't be clickable， see https://cpp.iceberg.apache.org/api/files.html

manuzhang (changes requested):
The Doxygen config still scans implementation `.cc` files, so internal implementation helpers can remain in the public API docs.

zhjwpku:
> The Doxygen config still scans implementation `.cc` files, so internal implementation helpers can remain in the public API docs.

Good catch, I have changed `FILE_PATTERNS` to include only *.h *.md *.dox, hope this works.

manuzhang (approved):

wgtmac (approved):

wgtmac:
Nice! Thanks @zhjwpku for fixing this and @manuzhang for the careful review!

PR #805, opened 2026-07-03
feat(expression): cast numeric literals to decimal type

huan233usc:
## What

Add casting of numeric literals (`int`, `long`, `float`, `double`) to a decimal target type in `Literal::CastTo`, so a numeric value can be used as a default for a decimal column.

Previously `CastFromInt` / `CastFromLong` / `CastFromFloat` / `CastFromDouble` had no `kDecimal` case and fell through to `NotSupported`, so a default like `Literal::Int(12)` or `Literal::Double(9.99)` for a `decimal(9, 2)` column was rejected. Java allows these (`IntegerLiteral.to`, `DoubleLiteral.to`, etc. scale the value to the target scale), so this brings the C++ literal cast layer to parity for numeric sources.

## How

- **Integer → decimal**: `CastIntegerToDecimal` treats the integer as scale 0 and rescales it to the target scale via `RescaleHalfUp` (exact when increasing scale, HALF_UP when decreasing it), then verifies the result fits the target precision (`FitsInPrecision`). Example: `12` → `decimal(9,2)` yields unscaled `1200` (`12.00`).
- **Float/double → decimal**: `CastRealToDecimal` parses the value's shortest round-tripping decimal representation (matching Java's `BigDecimal.valueOf(double)` via `Double.toString`) into an integer coefficient and exponent-derived scale without expanding scientific notation, then rounds to the target scale with **HALF_UP** rounding (round half away from zero, as Java does — `2.5` → `3`, `-2.5` → `-3`) and checks precision. Non-finite values are rejected.
- Both paths reject an out-of-range decimal scale before indexing the powers-of-ten table (`DecimalType` does not bound its scale on construction, so `decimal(9, 40)` would otherwise read past the table).

## Scope

Follow-up split out from the v3 default-value work (see the `CastDefaultToType` discussion on #793). It only extends the shared `Literal::CastTo` layer for numeric sources.

## Testing

`LiteralTest.IntegerCastToDecimal` (int/long scaling, out-of-precision rejection, out-of-range scale rejection) and `LiteralTest.RealCastToDecimal` (float/double scaling, HALF_UP rounding incl. negative, round-down, out-of-precision and non-finite rejection, scientific-notation overflow rejection and negative-scale acceptance), all verified fail-without / pass-with. Full `expression_test` passes (495 tests).

### DIFF
--- a/src/iceberg/expression/literal.cc
+++ b/src/iceberg/expression/literal.cc
@@ -22,6 +22,7 @@
 #include <cmath>
 #include <concepts>
 #include <cstdint>
+#include <format>
 #include <string>
 #include <vector>
 
@@ -89,6 +90,10 @@ class LiteralCaster {
   /// Cast an integer value (from Int or Long) to a decimal target type.
   static Result<Literal> CastIntegerToDecimal(
       int64_t value, const std::shared_ptr<PrimitiveType>& target_type);
+
+  /// Cast a floating-point value (from Float or Double) to a decimal target type.
+  static Result<Literal> CastRealToDecimal(
+      double value, const std::shared_ptr<PrimitiveType>& target_type);
 };
 
 Literal LiteralCaster::BelowMinLiteral(std::shared_ptr<PrimitiveType> type) {
@@ -118,6 +123,50 @@ Result<Literal> LiteralCaster::CastIntegerToDecimal(
   }
   return Literal::Decimal(decimal.value(), decimal_type.precision(),
                           decimal_type.scale());
+}
+
+Result<Literal> LiteralCaster::CastRealToDecimal(
+    double value, const std::shared_ptr<PrimitiveType>& target_type) {
+  const auto& decimal_type = internal::checked_cast<const DecimalType&>(*target_type);
+  const int32_t target_scale = decimal_type.scale();
+  if (target_scale < 0 || target_scale > Decimal::kMaxScale) {
+    return InvalidArgument("Cannot cast {} as a {} value", value,
+                           target_type->ToString());
+  }
+  if (!std::isfinite(value)) {
+    return InvalidArgument("Cannot cast {} as a {} value", value,
+                           target_type->ToString());
+  }
+
+  // Parse the shortest round-tripping decimal representation of the value (matching
+  // Java's BigDecimal.valueOf(double), which goes through Double.toString) into a
+  // full-precision decimal, then round to the target scale below.
+  int32_t parsed_scale = 0;
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto parsed, Decimal::FromString(std::format("{}", value), nullptr, &parsed_scale));
+
+  Decimal unscaled = parsed;
+  if (parsed_scale > target_scale) {
+    // Drop excess fractional digits with HALF_UP rounding (round half away from zero, as
+    // Java does), since Rescale itself only truncates and rejects any dropped remainder.
+    const int32_t drop = parsed_scale - target_scale;
+    ICEBERG_ASSIGN_OR_RAISE(auto divisor, Decimal(1).Rescale(0, drop));
+    ICEBERG_ASSIGN_OR_RAISE(auto divmod, parsed.Divide(divisor));
+    Decimal quotient = divmod.first;
+    Decimal remainder = Decimal::Abs(divmod.second);
+    if (remainder * Decimal(2) >= divisor) {
+      quotient += (value < 0) ? Decimal(-1) : Decimal(1);
+    }
+    unscaled = quotient;
+  } else if (parsed_scale < target_scale) {
+    ICEBERG_ASSIGN_OR_RAISE(unscaled, parsed.Rescale(parsed_scale, target_scale));
+  }
+
+  if (!unscaled.FitsInPrecision(decimal_type.precision())) {
+    return InvalidArgument("Cannot cast {} as a {} value", value,
+                           target_type->ToString());
+  }
+  return Literal::Decimal(unscaled.value(), decimal_type.precision(), target_scale);
 }
 
 Result<Literal> LiteralCaster::CastFromInt(
@@ -195,6 +244,8 @@ Result<Literal> LiteralCaster::CastFromFloat(
   switch (target_type->type_id()) {
     case TypeId::kDouble:
       return Literal::Double(static_cast<double>(float_val));
+    case TypeId::kDecimal:
+      return CastRealToDecimal(static_cast<double>(float_val), target_type);
     default:
       return NotSupported("Cast from Float to {} is not supported",
                           target_type->ToString());
@@ -215,6 +266,8 @@ Result<Literal> LiteralCaster::CastFromDouble(
       }
       return Literal::Float(static_cast<float>(double_val));
     }
+    case TypeId::kDecimal:
+      return CastRealToDecimal(double_val, target_type);
     default:
       return NotSupported("Cast from Double to {} is not supported",
                           target_type->ToString());
--- a/src/iceberg/expression/literal.cc
+++ b/src/iceberg/expression/literal.cc
@@ -102,9 +102,16 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
 Result<Literal> LiteralCaster::CastIntegerToDecimal(
     int64_t value, const std::shared_ptr<PrimitiveType>& target_type) {
   const auto& decimal_type = internal::checked_cast<const DecimalType&>(*target_type);
+  const int32_t scale = decimal_type.scale();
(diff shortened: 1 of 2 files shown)


src/iceberg/expression/literal.cc
```diff
@@ -95,6 +104,71 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
...
+  int32_t parsed_scale = 0;
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto parsed, Decimal::FromString(std::format("{}", value), nullptr, &parsed_scale));
+
+  Decimal unscaled = parsed;
+  if (parsed_scale > target_scale) {
+    // Drop excess fractional digits with HALF_UP rounding (round half away from zero, as
+    // Java does), since Rescale itself only truncates and rejects any dropped remainder.
+    const int32_t drop = parsed_scale - target_scale;
+    ICEBERG_ASSIGN_OR_RAISE(auto divisor, Decimal(1).Rescale(0, drop));
```
wgtmac: Please guard `drop` before calling `Rescale`. For values like `1e-100` cast to `decimal(9,2)`, `drop` is 98 and indexes past the powers-of-ten table; Java rounds this to `0.00`.

src/iceberg/expression/literal.cc
```diff
@@ -95,6 +104,71 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
   return Literal(Literal::AboveMax{}, std::move(type));
 }
 
+Result<Literal> LiteralCaster::CastIntegerToDecimal(
+    int64_t value, const std::shared_ptr<PrimitiveType>& target_type) {
+  const auto& decimal_type = internal::checked_cast<const DecimalType&>(*target_type);
+  const int32_t scale = decimal_type.scale();
+  // DecimalType does not bound its scale, but Rescale indexes a powers-of-ten table sized
+  // for [0, kMaxScale]; reject an out-of-range scale here rather than reading past it.
+  if (scale < 0 || scale > Decimal::kMaxScale) {
```
wgtmac: Please allow negative decimal scales here and in the real-value path. Iceberg decimal(P,S) supports negative S, and Java uses `setScale(scale, HALF_UP)`, so `149 -> decimal(9,-2)` should round to `1E+2` instead of failing.

src/iceberg/expression/literal.cc
huan233usc: Fixed. Rounding now goes through a shared `RescaleHalfUp` helper that returns `0` when the drop exceeds the max scale (so `1e-100` -> `0.00` instead of indexing past the table). Added a regression test.

src/iceberg/expression/literal.cc
huan233usc: Done — both the integer and real paths now allow negative scales and round HALF_UP, so `149 -> decimal(9,-2)` gives `1E+2`. Tested for both int and double sources.

src/iceberg/expression/literal.cc
```diff
@@ -95,6 +105,85 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
...
+  if (drop > Decimal::kMaxScale) {
+    return Decimal(0);
+  }
+  ICEBERG_ASSIGN_OR_RAISE(auto divisor, Decimal(1).Rescale(0, drop));
+  ICEBERG_ASSIGN_OR_RAISE(auto divmod, unscaled.Divide(divisor));
+  Decimal quotient = divmod.first;
+  Decimal remainder = Decimal::Abs(divmod.second);
+  if (remainder * Decimal(2) >= divisor) {
+    quotient += negative ? Decimal(-1) : Decimal(1);
+  }
```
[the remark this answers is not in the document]
huan233usc: Fixed in cf1dd9e4: HALF_UP now compares `remainder >= divisor / 2` so we don't overflow `int128` when `drop` is near `kMaxScale`.

src/iceberg/expression/literal.cc
```diff
@@ -95,6 +105,85 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
...
+
+}  // namespace
+
+Result<Literal> LiteralCaster::CastIntegerToDecimal(
+    int64_t value, const std::shared_ptr<PrimitiveType>& target_type) {
+  const auto& decimal_type = internal::checked_cast<const DecimalType&>(*target_type);
+  // An integer has scale 0; rescale it to the target scale, rounding HALF_UP when the
+  // target scale is negative (matching Java's numeric-to-decimal default handling).
+  ICEBERG_ASSIGN_OR_RAISE(auto unscaled, RescaleHalfUp(Decimal(value), /*from_scale=*/0,
+                                                       decimal_type.scale(), value < 0));
```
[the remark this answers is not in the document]
huan233usc: Fixed in cf1dd9e4: both integer and real paths reject a target scale outside `[-kMaxScale, kMaxScale]` via `ValidateDecimalScale` (with regression tests for `-40`).

src/iceberg/expression/literal.cc
```diff
@@ -95,6 +105,85 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
...
+    return InvalidArgument("Cannot cast {} as a {} value", value,
+                           target_type->ToString());
+  }
+
+  // Parse the shortest round-tripping decimal representation of the value (matching
+  // Java's BigDecimal.valueOf(double), which goes through Double.toString) into a
+  // full-precision decimal, then round to the target scale.
+  int32_t parsed_scale = 0;
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto parsed, Decimal::FromString(std::format("{}", value), nullptr, &parsed_scale));
```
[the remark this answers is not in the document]
huan233usc: Fixed in cf1dd9e4: switched to `std::to_chars` (no format specifier) for the shortest round-trip decimal string.

On float→decimal we still widen to `double` first, which matches Java Iceberg: `FloatLiteral.to(DECIMAL)` calls `BigDecimal.valueOf(value())`, and that overload takes `double` (float is promoted), so it uses `Double.toString` of the widened value rather than `Float.toString`. Splitting float/`Float.toString` would diverge from Java.

Added a regression for a value that needs more than 6 significant digits (`1.23456789 → decimal(20,8)`).

src/iceberg/expression/literal.cc
```diff
@@ -95,6 +105,85 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
...
+
+  ICEBERG_ASSIGN_OR_RAISE(auto unscaled, RescaleHalfUp(parsed, parsed_scale,
+                                                       decimal_type.scale(), value < 0));
```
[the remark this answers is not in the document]
huan233usc: Fixed in cf1dd9e4: same `ValidateDecimalScale` guard as the integer path.

src/iceberg/expression/literal.cc
```diff
@@ -95,6 +109,105 @@ Literal LiteralCaster::AboveMaxLiteral(std::shared_ptr<PrimitiveType> type) {
...
+  std::array<char, 64> buf{};
+  auto [ptr, ec] = std::to_chars(buf.data(), buf.data() + buf.size(), value);
+  if (ec != std::errc{}) {
+    return InvalidArgument("Cannot cast {} as a {} value", value,
+                           target_type->ToString());
+  }
+  int32_t parsed_scale = 0;
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto parsed,
+      Decimal::FromString(std::string_view(buf.data(), ptr), nullptr, &parsed_scale));
```
wgtmac: `Decimal::FromString` normalizes negative scales to 0 before this rescale, which can overflow its int128 intermediate. For example, `Literal::Double(4e38).CastTo(decimal(38, 0))` may wrap `4 * 10^38` to a 38-digit value and accept the wrong literal instead of rejecting it. It also rejects representable negative-scale cases such as `1e39 -> decimal(2, -38)`. Could we retain the parsed coefficient and exponent scale, rescale directly to the target scale, and add regressions for both cases?

huan233usc: Good catch. Now I parse the coefficient and its scale directly from the `to_chars` output (via a small `ParseRealCoefficient` helper) instead of going through `Decimal::FromString`, which was normalizing the negative scale by multiplying the coefficient by 10^-scale and overflowing int128. The coefficient itself always fits, and `RescaleHalfUp` combines the exponent with the target scale and rejects true overflow. Added regressions for both `4e38 -> decimal(38, 0)` (rejected) and `1e39 -> decimal(2, -38)` (accepted as 10).

wgtmac (approved):

PR #807, opened 2026-07-03
ci: retry build dir cleanup on Windows

abnobdoss:
## What
Retry `rm -rf` cleanup in the CMake CI helper scripts. Windows can briefly keep a just-built executable or DLL open, which can make cleanup fail even after the build and tests have passed.

`build_example.sh` retries before configuring and still lets `mkdir` fail if the directory remains, so it will not build in a half-deleted tree. `build_iceberg.sh` retries after the build and does not fail an otherwise successful job if final cleanup is still blocked.

### DIFF
--- a/ci/scripts/build_example.sh
+++ b/ci/scripts/build_example.sh
@@ -31,6 +31,8 @@ for attempt in 1 2 3; do
     fi
     if [[ "${attempt}" != "3" ]]; then
         sleep 2
+    else
+        echo "Failed to remove build directory after 3 attempts: ${build_dir}" >&2
     fi
 done
 mkdir "${build_dir}"
--- a/ci/scripts/build_example.sh
+++ b/ci/scripts/build_example.sh
@@ -23,7 +23,16 @@ source_dir=${1}
 build_dir=${1}/build
 run_example=${ICEBERG_RUN_EXAMPLE:-OFF}
 
-rm -rf "${build_dir}"
+# Clean up before configuring. If Windows still holds a just-built exe/dll
+# after the retries, let mkdir fail rather than reuse a half-deleted tree.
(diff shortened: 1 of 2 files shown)


ci/scripts/build_example.sh
```diff
@@ -23,7 +23,16 @@ source_dir=${1}
...
+# Clean up before configuring. If Windows still holds a just-built exe/dll
+# after the retries, let mkdir fail rather than reuse a half-deleted tree.
+for attempt in 1 2 3; do
+    if rm -rf "${build_dir}"; then
+        break
+    fi
+    if [[ "${attempt}" != "3" ]]; then
+        sleep 2
+    fi
+done
```
evindj, line 37: It would be nice to print a message in case all three attempts fail to remove the folder, that way it is easier to debug.

ci/scripts/build_iceberg.sh
```diff
@@ -99,5 +99,13 @@ fi
...
-rm -rf ${build_dir}
+# Clean up after the build. Windows can briefly hold a just-built exe/dll,
+# so retry but do not fail an otherwise successful CI job.
+for attempt in 1 2 3; do
+    if rm -rf "${build_dir}"; then
+        break
+    fi
+    if [[ "${attempt}" != "3" ]]; then
+        sleep 2
+    fi
```
evindj, line 112: ditto.

evindj (changes requested):

abnobdoss:
Thank you for reviewing @evindj; I've added the failure message as suggested!

evindj (approved):

wgtmac (approved):

wgtmac:
Thank you, @abnobdoss and @evindj!

PR #810, opened 2026-07-03
fix: reject non-finite floating-point default values

huan233usc:
## What

A `float` / `double` column default of `NaN` or `±Infinity` passes `SchemaField::Validate` but cannot survive a metadata round-trip: the JSON serializer emits a non-finite number as `null`, which reads back as an *absent* default. The declared default is silently lost (and, because the key is present but null, the reader then errors on it).

## How

`ValidateDefault` now rejects a non-finite floating default with a clear "value must be finite" error, so an unrepresentable default is caught up front rather than silently dropped on serialization.

The check uses `std::isfinite`, which is false for **both** `NaN` and `±Infinity` — so a single condition covers all three cases (there is no separate `IsNaN` branch because `isfinite` already rejects NaN).

## Testing

`SchemaFieldTest.ValidateRejectsNonFiniteFloatingDefault` explicitly covers `NaN`, `+Inf` and `-Inf`, and `ValidateAcceptsFiniteFloatingDefault` covers a normal value; verified fail-without / pass-with. Full `schema_test` passes (551 tests).

### DIFF
--- a/src/iceberg/schema_field.cc
+++ b/src/iceberg/schema_field.cc
@@ -19,6 +19,7 @@
 
 #include "iceberg/schema_field.h"
 
+#include <cmath>
 #include <format>
 #include <string_view>
 #include <utility>
@@ -167,6 +168,21 @@ Status ValidateDefault(const SchemaField& field, const Literal& value,
   if (*value.type() != *field.type()) {
     return InvalidSchema("{} of field {} has type {} but expected {}", kind, field.name(),
                          *value.type(), *field.type());
+  }
+  // A non-finite float/double default cannot be represented in JSON: the serializer emits
+  // it as `null`, which reads back as an absent default. Reject it so the default is not
+  // silently lost when the metadata round-trips. The literal type already matches the
+  // field type here, so this only inspects the value for actual floating fields and other
+  // types pay nothing.
+  if (field.type()->type_id() == TypeId::kFloat &&
+      !std::isfinite(std::get<float>(value.value()))) {
+    return InvalidSchema("Invalid {} value for {}: value must be finite", kind,
+                         field.name());
+  }
+  if (field.type()->type_id() == TypeId::kDouble &&
+      !std::isfinite(std::get<double>(value.value()))) {
+    return InvalidSchema("Invalid {} value for {}: value must be finite", kind,
+                         field.name());
   }
   return {};
 }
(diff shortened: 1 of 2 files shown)


evindj (approved):
This looks good to me. For my own understanding how is the caller supposed to handle this error?

src/iceberg/schema_field.cc
```diff
@@ -124,6 +126,18 @@ Status ValidateDefault(const SchemaField& field, const Literal& value,
     return InvalidSchema("{} of field {} has type {} but expected {}", kind, field.name(),
                          *value.type(), *field.type());
   }
+  // A non-finite float/double default cannot be represented in JSON: the serializer emits
+  // it as `null`, which reads back as an absent default. Reject it so the default is not
+  // silently lost when the metadata round-trips.
+  const auto& literal_value = value.value();
+  const bool non_finite = (std::holds_alternative<float>(literal_value) &&
```
zhjwpku: You mentioned `NaN` in the PR description but doesn't seem to handle that, is that intentional?

zhjwpku:
I think java impl has the same issue, created a PR to strict the validation https://github.com/apache/iceberg/pull/17106

src/iceberg/schema_field.cc
huan233usc: It is handled by following `std::isfinite` returns false for both NaN and infinity, so a NaN default hits the same `non_finite` branch and is rejected. The test `ValidateRejectsNonFiniteFloatingDefault` covers NaN, +Inf and -Inf explicitly.

huan233usc:
@evindj thanks for the review! `Validate()` returns a `Status`, so the error propagates up through the schema / `UpdateSchema` construction path the same way other invalid-schema errors do — the caller gets a failed `Result` and never builds a schema carrying a default that could not be round-tripped through metadata.

zhjwpku (approved):

src/iceberg/schema_field.cc
wgtmac, line 152: Why not move the check here so we don't pay the cost for other types?

wgtmac (approved):

src/iceberg/schema_field.cc
huan233usc: Good call — moved it into the float/double branches (gated on the field type id), so other types no longer touch the literal for this check. The literal type already matches the field type at that point, so `std::get` is safe there.

huan233usc:
Hi @wgtmac, do you mind taking another look at this PR? Thanks

wgtmac (approved):

wgtmac:
Sorry for the delay! Thanks for fixing this!

PR #811, opened 2026-07-04
ci: share sccache setup across workflows

abnobdoss:
## What
Add two composite actions for sccache:

- `.github/actions/setup-sccache` sets the sccache environment, restores the cache, resolves the MSVC compiler version on Windows, and starts `mozilla-actions/sccache-action`.
- `.github/actions/save-sccache` shows sccache stats for every run and saves the cache only on main branch runs.

The Test, SQL Catalog, sanitizer, C++ linter, and AWS workflows now pass only their cache key prefix at each call site.

This is part of the CI work discussed in https://github.com/apache/iceberg-cpp/issues/799.

## Notes
The existing cache key shapes are preserved for the current CMake workflow legs. The Meson Windows leg now also gets the MSVC-version key suffix, matching the other Windows sccache legs and avoids cache misses after runner image compiler updates.

The MSVC version source also changes for the existing Windows legs (test and SQL catalog): the inline `cl.exe` banner parse is replaced by reading `VCToolsVersion.default.txt` via vswhere, which is more reliable and does not require `cl.exe` on PATH. This produces a different version format (e.g. `14.44.35217` vs `19.44.35217.1`), so all Windows caches will miss once and reseed on the first post-merge main run.

Third-party actions stay pinned inside composite `action.yml` files, which are covered by the ASF allowlist checker glob.

## Validation
- Validated builds in fork: Test, AWS, SQL Catalog, ASAN/UBSAN, and pre-commit.
- The C++ linter workflow ignores `.github/**`, so this PR does not exercise that workflow until a source-touching run or post-merge.

### DIFF
--- a/.github/actions/save-sccache/action.yml
+++ b/.github/actions/save-sccache/action.yml
@@ -1,0 +1,44 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Save sccache
+description: Show sccache stats and save the cache on main branch runs after setup-sccache.
+
+inputs:
+  key-prefix:
+    description: Cache key prefix without the run id or Windows compiler suffix.
+    required: true
+  job-status:
+    description: >-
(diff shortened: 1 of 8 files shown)


.github/actions/save-sccache/action.yml
```diff
@@ -0,0 +1,39 @@
...
+inputs:
+  key-prefix:
+    description: Cache key prefix without the run id or Windows compiler suffix.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Show sccache stats
+      if: always()
```
wgtmac, line 35: `always()` only applies after the composite action starts. The call sites still default to `success()`, so this action is skipped after a failed build and stats are not shown. Could we add `if: always()` to the call sites and keep cache saving success-gated?

wgtmac (commented):
Looks good overall. Could you please rebase this onto the latest `main`? It currently conflicts with the cpp-linter changes from #838.

.github/actions/save-sccache/action.yml
abnobdoss: Definitely! I've updated the jobs to propagate the status of the job from the call sites then gate the save on that status but leave the show sccache stats ungated.

abnobdoss:
Hi @wgtmac, thank you for reviewing! I've rebased and updated the composite action gating per the feedback.

wgtmac (approved):

wgtmac:
Thank you, @abnobdoss!

PR #812, opened 2026-07-04
ci: add RelWithDebInfo build and fix free-nonheap-object error

zhjwpku:
Adding a RelWithDebInfo build revealed a build error. This PR fixes the issue.

```
error: ‘void operator delete(void*, std::size_t)’ called on a pointer to an unallocated object ‘1’ [-Werror=free-nonheap-object]
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
     name: AMD64 Ubuntu 26.04 (${{ matrix.cmake_build_type }})
     runs-on: ubuntu-26.04
-    timeout-minutes: 30
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
@@ -64,9 +64,9 @@ jobs:
         uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: ${{ github.workspace }}/.sccache
-          key: sccache-test-ubuntu-${{ github.run_id }}
+          key: sccache-test-ubuntu-${{ matrix.cmake_build_type }}-${{ github.run_id }}
           restore-keys: |
-            sccache-test-ubuntu-
+            sccache-test-ubuntu-${{ matrix.cmake_build_type }}-
       - name: Setup sccache
         uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
       - name: Build Iceberg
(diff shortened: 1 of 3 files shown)


src/iceberg/json_serde.cc
```diff
@@ -1377,9 +1377,9 @@ Result<std::unique_ptr<TableMetadata>> TableMetadataFromJson(const nlohmann::jso
       TimePointMs{std::chrono::milliseconds(last_updated_ms)};
 
   if (json.contains(kRefs)) {
-    ICEBERG_ASSIGN_OR_RAISE(
-        table_metadata->refs,
-        FromJsonMap<std::shared_ptr<SnapshotRef>>(json, kRefs, SnapshotRefFromJson));
+    ICEBERG_ASSIGN_OR_RAISE(auto refs, FromJsonMap<std::shared_ptr<SnapshotRef>>(
+                                           json, kRefs, SnapshotRefFromJson));
+    table_metadata->refs = std::move(refs);
```
evindj, line 1382: seems like a more readable version of previous version, however it seems disconnected from the change discussed in this PR?

.github/workflows/test.yml
```diff
@@ -39,11 +40,15 @@ env:
 jobs:
   ubuntu:
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
-    name: AMD64 Ubuntu 26.04
+    name: AMD64 Ubuntu 26.04 (${{ matrix.cmake_build_type }})
     runs-on: ubuntu-26.04
-    timeout-minutes: 30
+    timeout-minutes: 60
```
evindj, line 45: Is my understanding correct that the timeout was doubled becuase now for ubuntu 2 build configurations are going to be executed?

evindj (approved):

.github/workflows/test.yml
zhjwpku: Not really. Each build runs as its own job. I doubled that because this is a cold build, no cache can be utilized. Also, `RelWithDebInfo` may take longer since it's an -O2 build. It shouldn't take this long once it's merged into main and sccache starts taking effect.

src/iceberg/json_serde.cc
zhjwpku: Yeah, this is related to the new RelWithDebInfo coverage: that build exposed the `-Werror=free-nonheap-object` failure, so this change avoids assigning directly through `ICEBERG_ASSIGN_OR_RAISE` and fixes the issue surfaced by the new CI job. I'll update the PR title to make that relationship clearer.

wgtmac (approved):

wgtmac:
Thanks @zhjwpku for improving this and @evindj for the review!

src/iceberg/json_serde.cc
wgtmac: This seems like a GCC bug?

PR #817, opened 2026-07-06
fix: return error for reserved property keys in UpdateProperties::Set()

WZhuo:
Previously, reserved table properties (except `format-version`) were silently ignored when set via `UpdateProperties::Set()`. This caused undetected configuration drift compared to Java, which throws exceptions for such attempts.

Now attempts to set reserved properties accumulate a validation error via `ICEBERGER_BUILDER_CHECK`, surfaced at `Apply()` time. This matches Java's loud-failure behavior.


zhjwpku (approved):
LGTM

wgtmac (approved):

wgtmac:
Thanks @WZhuo for fixing this and @zhjwpku for the review!

PR #818, opened 2026-07-06
fix: correct URI-vs-path handling in PlanFiles() for Java-written file: paths

### LINKED ISSUE #341: Failed to plan files: Invalid: Expected a local filesystem path, got a URI in PlanFiles() call in demo_example.cc
There is a "file:" prefix getting appended to the avro file path which leads to this error inside arrow localfs.cc (RETURN_NOT_OK(ValidatePath(path));) while running demo_example.cc on WIndows while making the PlanFiles() call.

iceberg-reader.exe  "C:\iceberg_warehouse\db" sample "C:\iceberg_warehouse\db\sample\metadata\v5.metadata.json"
Failed to plan files: Invalid: Expected a local filesystem path, got a URI: 'file:/C:/iceberg_warehouse/db/sample/metadata/snap-4130091257426129961-1-0b7261a2-d123-4635-82a6-0a5a30e19a06.avro'

yadavay-amzn:
## What

Fixes #341. `PlanFiles()` (via `ResolvePath()` / `DetectBuiltinFileIO()`) mis-handled `file:` locations written by Java Iceberg in the short form `file:/C:/warehouse/...` (fewer than three slashes): the value was not recognized as a URI and failed to resolve, so planning broke on Windows-style paths.

## How
- Detect the URI scheme per RFC 3986 (a valid scheme token followed by `:`) instead of searching for `://`.
- Normalize authority-less `file:` forms (`file:/path`, `file:/C:/...`) to the canonical `file:///` form before handing them to Arrow's `PathFromUri`.
- Authority-bearing URIs (`file://host/path`, `s3://bucket/...`) and bare local paths pass through unchanged.

## Tests
Regression tests added in `arrow_io_test.cc` and `rest_file_io_test.cc` covering the Java `file:/C:/...` form, the canonical `file:///` form, and `file://host/path` pass-through. Build 788/788, tests 18/18 green.

---

### DIFF
--- a/src/iceberg/arrow/arrow_io.cc
+++ b/src/iceberg/arrow/arrow_io.cc
@@ -18,7 +18,6 @@
  */
 
 #include <algorithm>
-#include <cctype>
 #include <chrono>
 #include <limits>
 #include <mutex>
@@ -37,6 +36,7 @@
 #include "iceberg/arrow/arrow_io_util.h"
 #include "iceberg/arrow/arrow_status_internal.h"
 #include "iceberg/util/macros.h"
+#include "iceberg/util/uri.h"
 
 namespace iceberg::arrow {
 
@@ -486,25 +486,14 @@ class ArrowOutputFile : public OutputFile {
 
 Result<std::string> ArrowFileSystemFileIO::ResolvePath(const std::string& file_location) {
   // Detect whether the location is a URI by looking for a scheme component.
-  // A URI scheme (RFC 3986 §3.1) starts with a letter and is followed by any
-  // combination of letters, digits, '+', '-', or '.', ending at the first ':'.
-  // A single character before ':' is a Windows drive letter (e.g., "C:\..."),
-  // not a URI scheme.
-  auto colon_pos = file_location.find(':');
-  bool is_uri =
-      colon_pos != std::string::npos && colon_pos > 1 &&
-      std::isalpha(static_cast<unsigned char>(file_location[0])) &&
-      std::all_of(file_location.begin(),
-                  file_location.begin() + static_cast<std::ptrdiff_t>(colon_pos),
-                  [](char c) {
-                    return std::isalpha(static_cast<unsigned char>(c)) ||
-                           std::isdigit(static_cast<unsigned char>(c)) || c == '+' ||
-                           c == '-' || c == '.';
-                  });
+  // See iceberg/util/uri.h for the RFC 3986 scheme grammar.
+  bool is_uri = IsUriScheme(file_location);
 
   if (!is_uri) {
     return file_location;  // Bare local path (Unix or Windows drive letter)
   }
+
+  auto colon_pos = file_location.find(':');
 
   // Normalize authority-less file: URI short forms to the canonical three-slash
   // form so that Arrow's PathFromUri can parse them.  Java Iceberg may write
--- a/src/iceberg/arrow/arrow_io.cc
+++ b/src/iceberg/arrow/arrow_io.cc
@@ -18,6 +18,7 @@
  */
 
 #include <algorithm>
+#include <cctype>
 #include <chrono>
 #include <limits>
 #include <mutex>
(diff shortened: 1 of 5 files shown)


WZhuo (commented):
Thanks for this fix — the `file:/path` interop issue is a real problem and the test coverage is solid.

One observation: the RFC 3986 URI scheme detection logic (`colon_pos > 1 && std::isalpha(...) && std::all_of(...)`) is duplicated verbatim in both `arrow_io.cc` (`ArrowFileSystemFileIO::ResolvePath`) and `rest_file_io.cc` (`DetectBuiltinFileIO`). If the detection logic ever needs to change, both sites must be updated.

Consider extracting a small shared helper like `bool IsUriScheme(std::string_view)` into e.g. `iceberg/util/uri.h`. Not blocking — if the duplication is intentional to keep the `rest_file_io` layer independent, a comment at each site noting the duplication would suffice.

Otherwise looks good.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -484,24 +485,63 @@ class ArrowOutputFile : public OutputFile {
...
+  bool is_uri =
+      colon_pos != std::string::npos && colon_pos > 1 &&
+      std::isalpha(static_cast<unsigned char>(file_location[0])) &&
+      std::all_of(file_location.begin(),
+                  file_location.begin() + static_cast<std::ptrdiff_t>(colon_pos),
+                  [](char c) {
+                    return std::isalpha(static_cast<unsigned char>(c)) ||
+                           std::isdigit(static_cast<unsigned char>(c)) || c == '+' ||
+                           c == '-' || c == '.';
+                  });
```
WZhuo: The RFC 3986 URI scheme detection logic here is duplicated verbatim in `rest_file_io.cc` `DetectBuiltinFileIO`. Consider extracting a shared `bool IsUriScheme(std::string_view)` helper into e.g. `iceberg/util/uri.h`. If the duplication is intentional to keep the `rest_file_io` layer independent, a comment at each site noting the duplication would suffice.

src/iceberg/catalog/rest/rest_file_io.cc
```diff
@@ -51,12 +53,25 @@ std::unordered_map<std::string, std::string> MergeFileIOProperties(
...
+  auto colon_pos = location.find(':');
+  bool is_uri =
+      colon_pos != std::string_view::npos && colon_pos > 1 &&
+      std::isalpha(static_cast<unsigned char>(location[0])) &&
+      std::all_of(location.begin(),
+                  location.begin() + static_cast<std::ptrdiff_t>(colon_pos), [](char c) {
+                    return std::isalpha(static_cast<unsigned char>(c)) ||
+                           std::isdigit(static_cast<unsigned char>(c)) || c == '+' ||
+                           c == '-' || c == '.';
+                  });
```
WZhuo: This RFC 3986 scheme detection is the same logic as in `arrow_io.cc` `ArrowFileSystemFileIO::ResolvePath`. See the comment there about extracting a shared helper.

src/iceberg/arrow/arrow_io.cc
yadavay-amzn: Good call - extracted the RFC 3986 scheme detection into a shared header-only helper IsUriScheme(std::string_view) in src/iceberg/util/uri.h, and both ResolvePath() here and DetectBuiltinFileIO() in rest_file_io.cc now call it, so the logic is no longer duplicated.

src/iceberg/catalog/rest/rest_file_io.cc
yadavay-amzn: Done - this and the arrow_io.cc site now both call the shared IsUriScheme() helper in src/iceberg/util/uri.h; the duplicated detection block is removed.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -484,24 +485,52 @@ class ArrowOutputFile : public OutputFile {
...
-    return file_location;
+  // Detect whether the location is a URI by looking for a scheme component.
+  // See iceberg/util/uri.h for the RFC 3986 scheme grammar.
+  bool is_uri = IsUriScheme(file_location);
+
+  if (!is_uri) {
+    return file_location;  // Bare local path (Unix or Windows drive letter)
+  }
+
+  auto colon_pos = file_location.find(':');
```
zhjwpku: Now we have two calls to `find`. That's probably acceptable, but we could also add an output parameter to return `colon_pos`. Not a strong opinion.

yadavay-amzn: Done - IsUriScheme now takes an optional std::size_t* scheme_colon_pos out-param, and both ResolvePath here and DetectBuiltinFileIO in rest_file_io.cc reuse it instead of calling find(':') again.

src/iceberg/util/uri.h
```diff
@@ -0,0 +1,68 @@
...
+/// \param scheme_colon_pos If non-null and the function returns true, set to the
+///   position of the scheme-delimiting ':' so callers can reuse it without a
+///   redundant search.
+/// \return true if \p value starts with a valid URI scheme followed by ':'.
+inline bool IsUriScheme(std::string_view value, std::size_t* scheme_colon_pos = nullptr) {
+  auto colon_pos = value.find(':');
+  // Reject if there is no ':', an empty scheme (colon_pos == 0), or only a
+  // single character before ':' (colon_pos == 1), which is a Windows drive
+  // letter (e.g. "C:\path"), not a URI scheme.
+  if (colon_pos == std::string_view::npos || colon_pos <= 1) {
```
wgtmac: Please separate Windows-drive detection from RFC scheme validation. RFC 3986 and Java accept one-character schemes such as a:; rejecting every colon_pos == 1 breaks custom schemes and conflicts with LocationUtil::ParseScheme. Treat only C:/ or C:\\ as a path at the call site.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -484,24 +485,51 @@ class ArrowOutputFile : public OutputFile {
...
+    return file_location;  // Bare local path (Unix or Windows drive letter)
+  }
+
+  // Normalize authority-less file: URI short forms to the canonical three-slash
+  // form so that Arrow's PathFromUri can parse them.  Java Iceberg may write
+  // "file:/path" (one slash) which lacks the "://" that Arrow expects.
+  // Authority-bearing URIs like "file://host/path" (char after "file://" is not
+  // '/') are already valid RFC 3986 and pass through unchanged.
+  std::string normalized = file_location;
+  if (normalized.starts_with("file:/") && !normalized.starts_with("file://")) {
```
wgtmac: URI schemes are case-insensitive. Please normalize the scheme before checking file:/ and before passing it to Arrow; otherwise FILE:/... can fall through and be treated as a local path.

src/iceberg/util/uri.h
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
```
wgtmac: Please add uri.h to src/iceberg/util/meson.build, or avoid adding a new public header. CMake discovers it automatically, but Meson's explicit install list does not.

src/iceberg/arrow/arrow_io.cc
```diff
@@ -484,24 +485,51 @@ class ArrowOutputFile : public OutputFile {
...
+  const auto sep_pos = file_location.find("://");
+  if (sep_pos == std::string::npos) {
+    // URI without "://" that is not file: — attempt best-effort strip of scheme
+    auto scheme_end = colon_pos + 1;
+    while (scheme_end < file_location.size() && file_location[scheme_end] == '/') {
+      ++scheme_end;
+    }
+    // Keep one leading slash for absolute paths
+    return std::string(
+        file_location.substr(scheme_end > colon_pos + 1 ? scheme_end - 1 : scheme_end));
```
wgtmac, line 524: This fallback is too broad. `file:relative/path` and `s3:/bucket/key` are stripped to local paths after `PathFromUri` fails, so a URI can be read from the wrong local file. Please keep this fallback only for supported foreign aliases with `://` and return the original parse error otherwise.

src/iceberg/util/uri.h
wgtmac, line 51: `a:` is a valid RFC 3986 scheme and Java accepts it. Please move the Windows-drive exception out of this helper and only treat `C:/` or `C:\\` as paths at the call sites; otherwise custom one-character schemes are rejected.

src/iceberg/util/uri.h
```diff
@@ -0,0 +1,68 @@
...
+/// \return true if \p value starts with a valid URI scheme followed by ':'.
+inline bool IsUriScheme(std::string_view value, std::size_t* scheme_colon_pos = nullptr) {
+  auto colon_pos = value.find(':');
+  // Reject if there is no ':', an empty scheme (colon_pos == 0), or only a
+  // single character before ':' (colon_pos == 1), which is a Windows drive
+  // letter (e.g. "C:\path"), not a URI scheme.
+  if (colon_pos == std::string_view::npos || colon_pos <= 1) {
+    return false;
+  }
+  if (!std::isalpha(static_cast<unsigned char>(value[0]))) {
```
wgtmac, line 54: Use explicit ASCII predicates instead of `std::isalpha` and `std::isdigit`. RFC 3986 defines ALPHA and DIGIT as ASCII, while these ctype calls are locale-sensitive.

src/iceberg/arrow/arrow_io.cc
wgtmac, line 503: URI schemes are case-insensitive. Please normalize the scheme before checking `file:/`; otherwise `FILE:/...` can fall through and be treated as a local path.

src/iceberg/catalog/rest/rest_file_io.cc
```diff
@@ -51,12 +52,16 @@ std::unordered_map<std::string, std::string> MergeFileIOProperties(
 }  // namespace
 
 Result<BuiltinFileIOKind> DetectBuiltinFileIO(std::string_view location) {
```
wgtmac, line 54: This code no longer matches the current FileIO resolution flow. Please rebase this change onto the current resolver design and keep the Arrow `file:/` fix separately.

wgtmac (changes requested):

PR #819, opened 2026-07-06
build: scope warnings-as-error to project targets, not FetchContent subprojects

### LINKED ISSUE #246: Werror in CMake Build propogates warnings from subprojects as errors
When using the CMake configuration, dependencies that need to be built may throw warnings. With the current CMake configuration, these warnings cause the build to fail:

```sh
[ 62%] Building CXX object src/iceberg/CMakeFiles/iceberg_static.dir/manifest_entry.cc.o
/home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.cc: In member function ‘const std::shared_ptr<arrow::Array>& arrow::StructArray::field(int) const’:
/home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.cc:1080:51: error: ‘std::shared_ptr<_Tp> std::atomic_load(const shared_ptr<_Tp>*) [with _Tp = arrow::Array]’ is deprecated: use 'std::atomic<std::shared_ptr<T>>' instead [-Werror=deprecated-declarations]
 1080 |   std::shared_ptr<Array> result = std::atomic_load(&boxed_fields_[i]);
      |                                   ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/14/memory:81,
                 from /home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.h:24,
                 from /home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.cc:18:
/usr/include/c++/14/bits/shared_ptr_atomic.h:142:5: note: declared here
  142 |     atomic_load(const shared_ptr<_Tp>* __p)
      |     ^~~~~~~~~~~
/home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.cc:1089:22: error: ‘void std::atomic_store(shared_ptr<_Tp>*, shared_ptr<_Tp>) [with _Tp = arrow::Array]’ is deprecated: use 'std::atomic<std::shared_ptr<T>>' instead [-Werror=deprecated-declarations]
 1089 |     std::atomic_store(&boxed_fields_[i], std::move(result));
      |     ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/bits/shared_ptr_atomic.h:183:5: note: declared here
  183 |     atomic_store(shared_ptr<_Tp>* __p, shared_ptr<_Tp> __r)
      |     ^~~~~~~~~~~~
/home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.cc: In member function ‘std::shared_ptr<arrow::Array> arrow::UnionArray::field(int) const’:
/home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.cc:1360:51: error: ‘std::shared_ptr<_Tp> std::atomic_load(const shared_ptr<_Tp>*) [with _Tp = arrow::Array]’ is deprecated: use 'std::atomic<std::shared_ptr<T>>' instead [-Werror=deprecated-declarations]
 1360 |   std::shared_ptr<Array> result = std::atomic_load(&boxed_fields_[i]);
      |                                   ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/bits/shared_ptr_atomic.h:142:5: note: declared here
  142 |     atomic_load(const shared_ptr<_Tp>* __p)
      |     ^~~~~~~~~~~
/home/will/clones/iceberg-cpp/build/_deps/vendoredarrow-src/cpp/src/arrow/array/array_nested.cc:1372:22: error: ‘void std::atomic_store(shared_ptr<_Tp>*, shared_ptr<_Tp>) [with _Tp = arrow::Array]’ is deprecated: use 'std::atomic<std::shared_ptr<T>>' instead [-Werror=deprecated-declarations]
 1372 |     std::atomic_store(&boxed_fields_[i], result);
      |     ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/bits/shared_ptr_atomic.h:183:5: note: declared here
  183 |     atomic_store(shared_ptr<_Tp>* __p, shared_ptr<_Tp> __r)
      |     ^~~~~~~~~~~~

```

yadavay-amzn:
## What

Fixes #246. `-Werror` (via a global `CMAKE_COMPILE_WARNING_AS_ERROR`) leaked into FetchContent-built dependencies, so a warning in a third-party subproject (e.g. Arrow) could fail the whole build. As noted on the issue, the global flag does not reliably isolate subprojects.

## How
- Removed the global `CMAKE_COMPILE_WARNING_AS_ERROR ON`.
- Set `COMPILE_WARNING_AS_ERROR ON` as a per-target property on Iceberg's own shared/static libraries, so the project stays strict.
- Set `BUILD_WARNING_LEVEL PRODUCTION` for Arrow to suppress its independent `-Werror` injection.
- Kept a defensive guard in `prepare_fetchcontent()` for the case where `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON` is passed on the command line.

## Verified
Iceberg targets still compile with `-Werror`; FetchContent subprojects do not. Build + tests green (18/18).

---


zhjwpku:
Could you please configure pre-commit locally? It seems the patch contains some unnecessary changes.

yadavay-amzn:
Done - configured pre-commit locally with the pinned cmake-format (v0.6.10), reverted the unrelated cosmetic reformattings, and re-pushed. The diff now contains only the warnings-as-error scoping change and passes pre-commit.

zhjwpku (approved):

evindj (commented):
The change looks good to me. However, I wonder if this comes from an hypothetical or an actual occurence where the dependencies triggering the issue was not able/willing to address. I generally agree that we should do this but I wonder if relaxing the requirement causes side effects.

yadavay-amzn:
Thanks @evindj - it's a real occurrence, not hypothetical. Arrow injects its own -Werror during its build, and as noted on the issue by @wgtmac/@zhjwpku the global CMAKE_COMPILE_WARNING_AS_ERROR does not reliably stay off for FetchContent subprojects, so a warning in a bundled dependency can fail our build. On side effects: this does not relax our own strictness - iceberg's own shared/static targets keep -Werror via a per-target COMPILE_WARNING_AS_ERROR property. It only stops us from imposing -Werror on third-party subprojects we don't control, which is the intended behavior (a dependency's warnings shouldn't fail our build, and we can't fix their code). So our code stays exactly as strict as before; only the vendored dependencies are no longer forced warning-clean under our flags.

wgtmac (approved):

wgtmac:
Nice! Thanks @yadavay-amzn for fixing it and @zhjwpku @evindj for the review!

PR #822, opened 2026-07-06
feat: read v3 row lineage metadata columns

wgtmac:
Synthesize _row_id and _last_updated_sequence_number when reading data files that omit row lineage columns, and preserve physical non-null lineage values when present.

Carry inherited data sequence numbers on DataFile metadata so scan planning and readers follow the same model as Java.

### DIFF
--- a/src/iceberg/avro/avro_data_util.cc
+++ b/src/iceberg/avro/avro_data_util.cc
@@ -54,6 +54,22 @@ Status AppendFieldToBuilder(const ::avro::NodePtr& avro_node,
                             const arrow::MetadataColumnContext& metadata_context,
                             ::arrow::ArrayBuilder* array_builder);
 
+Status AppendRowLineageValue(int32_t field_id,
+                             const arrow::MetadataColumnContext& metadata_context,
+                             ::arrow::ArrayBuilder* array_builder) {
+  auto* int_builder = internal::checked_cast<::arrow::Int64Builder*>(array_builder);
+  if (!arrow::HasRowLineageValue(field_id, metadata_context)) {
+    ICEBERG_ARROW_RETURN_NOT_OK(int_builder->AppendNull());
+  } else if (field_id == MetadataColumns::kRowIdColumnId) {
+    ICEBERG_ARROW_RETURN_NOT_OK(int_builder->Append(
+        metadata_context.first_row_id.value() + metadata_context.next_file_pos));
+  } else {
+    ICEBERG_ARROW_RETURN_NOT_OK(
+        int_builder->Append(metadata_context.data_sequence_number.value()));
+  }
+  return {};
+}
+
 /// \brief Append Avro record data to Arrow struct builder.
 Status AppendStructToBuilder(const ::avro::NodePtr& avro_node,
                              const ::avro::GenericDatum& avro_datum,
@@ -97,6 +113,9 @@ Status AppendStructToBuilder(const ::avro::NodePtr& avro_node,
       } else if (field_id == MetadataColumns::kFilePositionColumnId) {
         auto int_builder = internal::checked_cast<::arrow::Int64Builder*>(field_builder);
         ICEBERG_ARROW_RETURN_NOT_OK(int_builder->Append(metadata_context.next_file_pos));
+      } else if (MetadataColumns::IsRowLineageColumn(field_id)) {
+        ICEBERG_RETURN_UNEXPECTED(
+            AppendRowLineageValue(field_id, metadata_context, field_builder));
       } else {
         return NotSupported("Unsupported metadata column field id: {}", field_id);
       }
@@ -463,9 +482,21 @@ Status AppendFieldToBuilder(const ::avro::NodePtr& avro_node,
     return {};
   }
 
+  const bool is_row_lineage =
+      MetadataColumns::IsRowLineageColumn(projected_field.field_id());
+  if (is_row_lineage &&
+      !arrow::HasRowLineageValue(projected_field.field_id(), metadata_context)) {
+    ICEBERG_ARROW_RETURN_NOT_OK(array_builder->AppendNull());
+    return {};
+  }
+
   if (avro_node->type() == ::avro::AVRO_UNION) {
     size_t branch = avro_datum.unionBranch();
     if (avro_node->leafAt(branch)->type() == ::avro::AVRO_NULL) {
+      if (is_row_lineage) {
+        return AppendRowLineageValue(projected_field.field_id(), metadata_context,
+                                     array_builder);
+      }
       ICEBERG_ARROW_RETURN_NOT_OK(array_builder->AppendNull());
       return {};
     } else {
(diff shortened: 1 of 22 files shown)


src/iceberg/data/file_scan_task_reader.cc
```diff
@@ -161,9 +165,9 @@ class FileScanTaskReader::Impl {
                      data_file->file_size_in_bytes);
 
     if (task.delete_files().empty()) {
-      auto options =
-          MakeReaderOptions(*data_file, io_, projected_schema_, task.residual_filter(),
-                            name_mapping_, properties_);
+      auto options = MakeReaderOptions(
```
manuzhang, line 168: REST scan-task JSON appears to round-trip only `first-row-id` for `DataFile`; `FileScanTasksFromJson` constructs the `FileScanTask` without any data sequence number. That means this call can pass `data_file->first_row_id` but an empty `data_file->data_sequence_number`, and `MakeLastUpdatedSequenceNumberArray` will return nulls for `_last_updated_sequence_number`.

The row-lineage spec says missing/null `_last_updated_sequence_number` should be replaced with the data file’s data sequence number.

*Disclosure: I used an LLM to help review this PR and draft this comment.*

src/iceberg/avro/avro_data_util.cc
```diff
@@ -54,6 +54,22 @@ Status AppendFieldToBuilder(const ::avro::NodePtr& avro_node,
                             const arrow::MetadataColumnContext& metadata_context,
                             ::arrow::ArrayBuilder* array_builder);
 
+Status AppendRowLineageValue(int32_t field_id,
```
WZhuo: `AppendRowLineageValue` is defined verbatim in both `avro_data_util.cc:57` and `avro_direct_decoder.cc:147` with identical logic.

src/iceberg/inheritable_metadata.cc
```diff
@@ -64,6 +64,7 @@ Status BaseInheritableMetadata::Apply(ManifestEntry& entry) {
 
   if (entry.data_file) {
     entry.data_file->partition_spec_id = spec_id_;
+    entry.data_file->data_sequence_number = entry.sequence_number;
```
WZhuo, line 67: `BaseInheritableMetadata::Apply()` only propagates `data_sequence_number` to the `DataFile`, but the Java equivalent (`InheritableMetadataFactory.BaseInheritableMetadata.apply()`) sets four fields:

```java
file.setSpecId(specId);
file.setDataSequenceNumber(manifestEntry.dataSequenceNumber());
file.setFileSequenceNumber(manifestEntry.fileSequenceNumber());
file.setManifestLocation(manifestLocation);
```

Suggestion: either add the fields and propagate them now, or leave an explicit `// TODO: file_sequence_number and manifest_location` comment so the gap is visible.

src/iceberg/avro/avro_data_util.cc
wgtmac: Good catch! Fixed

src/iceberg/inheritable_metadata.cc
wgtmac: Added a comment. I would prefer adding them later when they are really needed.

src/iceberg/data/file_scan_task_reader.cc
wgtmac: I looked into the Java REST implementation and OpenAPI. This is not C++-specific: `RESTFileScanTaskParser` serializes the task file through `ContentFileParser`, and `ContentFileParser` only round-trips `first-row-id`, not the manifest-entry data sequence number. The OpenAPI `FileScanTask`/`DataFile` schema also has `first-row-id` but no field for the data sequence number.

So the failing case is REST-planned scan tasks reading row lineage metadata, where `_last_updated_sequence_number` needs the manifest entry sequence number but that value is not part of the REST contract. I think the correct fix should start in the REST API / Java parser, then C++ can implement the same field. I don’t think C++ should add a non-standard private field here.

src/iceberg/parquet/parquet_data_util.cc
```diff
@@ -70,6 +70,53 @@ Result<std::shared_ptr<::arrow::Array>> ProjectPrimitiveArray(
...
+        metadata_context.first_row_id, metadata_context.data_sequence_number, num_rows,
+        pool);
+  }
+  return NotSupported("Unsupported row lineage field id: {}", field_id);
+}
+
+Result<std::shared_ptr<::arrow::Array>> ProjectRowLineageArray(
+    const std::shared_ptr<::arrow::Array>& physical_array, int32_t field_id,
+    const arrow::MetadataColumnContext& metadata_context, ::arrow::MemoryPool* pool) {
+  if (!arrow::HasRowLineageValue(field_id, metadata_context)) {
```
manuzhang: It looks physical array is not examined and its non-null value might not be preserved with `first_row_id` present but `data_sequence_number` absent. Both Avro generic decoding and direct decoding have similar issues.

src/iceberg/data/file_scan_task_reader.cc
manuzhang: Agreed, the fix is beyond this PR. We can create a follow-up issue to track.

WZhuo (approved):
LGTM

src/iceberg/data/file_scan_task_reader.cc
wgtmac: I opened apache/iceberg#17234 to address this on the Java/REST side. It preserves `data-sequence-number` in `ContentFile` JSON and updates the OpenAPI/generated model, so REST-planned scan tasks can carry the manifest-entry data sequence number needed for `_last_updated_sequence_number` inheritance.

src/iceberg/parquet/parquet_data_util.cc
wgtmac: I opened apache/iceberg#17235 for the corresponding Java reader behavior. It updates Avro, Parquet, ORC, and Arrow readers to preserve physical row-lineage values when inheritance constants are missing, and to inherit `_last_updated_sequence_number` from the data sequence number without depending on `first_row_id`. That matches the spec behavior this C++ path is implementing.

src/iceberg/data/file_scan_task_reader.cc
manuzhang: Ah ha, you beat me to it ;)

wgtmac: I just let Codex create these to provide me with more detail so I marked them as `[DO NOT REVIEW]`. Feel free to create your own PRs if you have time to work on them :)

manuzhang: I [opened a PR at my own fork](https://github.com/manuzhang/iceberg/pull/201) but am still reviewing the change. Let me check out your PR and see where we can go from there.

src/iceberg/catalog/rest/json_serde.cc
```diff
@@ -323,6 +323,9 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> FileScanTasksFromJson(
                             GetJsonValue<nlohmann::json>(task_json, kDataFile));
     ICEBERG_ASSIGN_OR_RAISE(
         auto data_file, DataFileFromJson(data_file_json, partition_spec_by_id, schema));
+    // FIXME: REST scan-task DataFile JSON currently carries first-row-id,
```
manuzhang, line 326: @wgtmac Can we add a test that exposes this gap and we skip for now?

manuzhang: Let's also create a follow-up issue and refer it in the comments for the test.

manuzhang:
LGTM

blackmwk (approved):

wgtmac:
Thanks all for the review!

Issue #690, opened 2026-05-29
Add a minimal micro-benchmark suite to validate filtering hot-path optimizations


sentomk:
While reading the scan-planning filtering path, I found a small optimization in the metrics evaluators. Using it as a concrete example to raise a broader question about how to validate this kind of change.

## Proposed change

The metrics evaluators run per data file. Each predicate currently calls `expr->reference()` repeatedly, and `reference()` returns a `shared_ptr` via `shared_from_this()` — an atomic refcount bump every time. The `StrictMetricsEvaluator` macro even discards a `dynamic_cast` result only to re-fetch the same reference:

```diff
- #define RETURN_IF_NOT_REFERENCE(expr)                                         \
-   if (auto ref = dynamic_cast<BoundReference*>(expr.get()); ref == nullptr) { \
-     return kRowsMightNotMatch;                                                \
-   }
+ #define BIND_REFERENCE_OR_RETURN(ref, expr)                            \
+   const auto* ref = dynamic_cast<const BoundReference*>((expr).get()); \
+   if (ref == nullptr) {                                                \
+     return kRowsMightNotMatch;                                         \
+   }

    Result<bool> IsNull(const std::shared_ptr<Bound>& expr) override {
-     RETURN_IF_NOT_REFERENCE(expr);
-     int32_t id = expr->reference()->field().field_id();
+     BIND_REFERENCE_OR_RETURN(ref, expr);
+     int32_t id = ref->field().field_id();
      ...
```

Reusing the cast result drops the repeated virtual `reference()` calls (and their atomic ops) across every predicate, with no behavior change.

## Expected benefit

The win is on the CPU-bound filtering step, evaluated in isolation. Scan planning as a whole is IO-bound, so on an e2e scan this kind of change is almost certainly unmeasurable — which is exactly why it needs to be measured on the filtering step alone.

## Which raises the question: do we need a benchmark suite?

This is exactly the kind of change that's hard to justify without one. The repo has no benchmark infrastructure today, only the gtest suite. A minimal benchmark on the filtering path would let us measure such changes on the CPU-bound step alone, rather than guessing or claiming a win against IO-dominated planning.

So before going further:

1. **How** should we add it — Google Benchmark fetched the same way googletest already is, behind an off-by-default CMake option?
2. **Where** should it live — a top-level `benchmark/`, or co-located under `src/iceberg/**/`?

I'm happy to put up a draft PR for a minimal suite + the filtering benchmark above once there's agreement on direction.

[...]

PR #825, opened 2026-07-08
build: add benchmark build scaffolding

timothyw553:
## Summary

Adds an opt-in benchmark build path so future C++ microbenchmarks can live in-tree without affecting normal builds. Benchmarks stay disabled by default, and this PR keeps the executable intentionally small so reviewers can focus on CMake/Meson wiring and Google Benchmark integration.

- Adds `ICEBERG_BUILD_BENCHMARKS` for CMake and a disabled-by-default `benchmarks` Meson feature.
- Wires `src/iceberg/benchmark` into both build systems.
- Adds Google Benchmark dependency setup for CMake and Meson.
- Adds a tiny smoke benchmark so the new target is buildable in isolation.

Part of #690.

## Stack

This is PR 1 of 2.

PR 2 adds the actual metrics evaluator benchmark cases on top of this scaffolding: timothyw553/iceberg-cpp#1.

## Test Plan

- [x] Configured benchmark-enabled CMake build with `-DICEBERG_BUILD_BENCHMARKS=ON`.
- [x] Built `metrics_evaluator_benchmark` on this branch.
- [x] Ran the smoke benchmark with `--benchmark_min_time=0.01s`.
- [x] Re-ran the default build and test suite on the top of the stack: 18/18 tests passed.
- [ ] Meson wiring is syntax-reviewed but not locally executed because `meson` is not installed in this shell.

## Verification Commands

```bash
cmake -S . -B build-bench-stack -G Ninja \
  -DFETCHCONTENT_TRY_FIND_PACKAGE_MODE=NEVER \
  -DCMAKE_IGNORE_PREFIX_PATH=/opt/homebrew/anaconda3 \
  -DCMAKE_PREFIX_PATH='/opt/homebrew/opt/snappy;/opt/homebrew/opt/zstd' \
  -DICEBERG_BUILD_BENCHMARKS=ON
cmake --build build-bench-stack --target metrics_evaluator_benchmark
build-bench-stack/src/iceberg/benchmark/metrics_evaluator_benchmark --benchmark_min_time=0.01s
cmake --build build
ctest --test-dir build --output-on-failure
```


src/iceberg/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+set(BENCHMARK_ENABLE_INSTALL
+    OFF
+    CACHE BOOL "" FORCE)
+set(BENCHMARK_ENABLE_TESTING
+    OFF
+    CACHE BOOL "" FORCE)
+
+fetchcontent_declare(google_benchmark
+                     GIT_REPOSITORY https://github.com/google/benchmark.git
+                     GIT_TAG v1.8.4
```
zhjwpku: nit: should we use the commit SHA like gtest?

zhjwpku (approved):

src/iceberg/benchmark/CMakeLists.txt
timothyw553: done. I pinned Google Benchmark to the v1.8.4 release commit SHA (`a4cf155615c63e019ae549e31703bf367df5b471`) and kept the version as an inline comment, matching the gtest style.

src/iceberg/benchmark/benchmark_smoke.cc
```diff
@@ -0,0 +1,46 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <benchmark/benchmark.h>
+
+namespace iceberg {
+namespace {
+
+void BM_BenchmarkSmoke(benchmark::State& state) {
```
wgtmac, line 25: nit: the file name looks odd since it's unrelated to metrics evaluator.

src/iceberg/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+    OFF
+    CACHE BOOL "" FORCE)
+set(BENCHMARK_ENABLE_INSTALL
+    OFF
+    CACHE BOOL "" FORCE)
+set(BENCHMARK_ENABLE_TESTING
+    OFF
+    CACHE BOOL "" FORCE)
+
+fetchcontent_declare(google_benchmark
```
wgtmac, line 28: Should we move this to `IcebergThirdpartyToolchain.cmake`? I think we should move GTest to there, too.

.github/workflows/cpp-linter.yml
```diff
@@ -77,6 +77,7 @@ jobs:
             -DCMAKE_C_COMPILER_LAUNCHER=sccache \
             -DCMAKE_CXX_COMPILER_LAUNCHER=sccache \
             -DICEBERG_BUILD_SQL_CATALOG=ON \
+            -DICEBERG_BUILD_BENCHMARKS=ON \
```
wgtmac, line 80: Why do we need this but do not change other workflows?

src/iceberg/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+                     GIT_REPOSITORY https://github.com/google/benchmark.git
+                     GIT_TAG a4cf155615c63e019ae549e31703bf367df5b471 # v1.8.4
+                     FIND_PACKAGE_ARGS
+                     NAMES
+                     benchmark
+                     CONFIG)
+fetchcontent_makeavailable(google_benchmark)
+
+add_executable(metrics_evaluator_benchmark metrics_evaluator_benchmark.cc)
```
wgtmac: Should we add `EXCLUDE_FROM_ALL` so it does not build by default?

wgtmac (commented):
Thanks for adding this! I just left some comments.

src/iceberg/benchmark/CMakeLists.txt
zhjwpku: I think this PR follows the gtest way ;) Maybe we can change that together, I have the same feeling at the first look.

src/iceberg/benchmark/benchmark_smoke.cc
timothyw553: renamed the smoke benchmark so it isn't tied to metrics evaluator yet.

src/iceberg/benchmark/CMakeLists.txt
timothyw553: opened #829 to track moving both GTest and Google Benchmark into IcebergThirdpartyToolchain.cmake. I’ll keep this PR scoped and follow the current GTest pattern here.

src/iceberg/benchmark/CMakeLists.txt
timothyw553: ah I don’t think we need it. I set benchmarks to be off by default (see my [CMakeLists.txt](https://github.com/apache/iceberg-cpp/pull/825/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a) change), and when users set `ICEBERG_BUILD_BENCHMARKS=ON`, the benchmark target should build normally.

timothyw553: hmm... if we did add `EXCLUDE_FROM_ALL` then users would have to do 
```
cmake --build build --target benchmark_a benchmark_b benchmark_c benchmark_d
```
to me a better pattern seems to be:
```
cmake -S . -B build -DICEBERG_BUILD_BENCHMARKS=ON
cmake --build build
```
and if user did want to just build specific targets then they can still do:
```
cmake --build build --target metrics_evaluator_benchmark
```
what do you think?

.github/workflows/cpp-linter.yml
timothyw553: this is just so the benchmarks get linted. I previously did not have this and cpp-linter failed because the benchmark source wasn’t in compile_commands.json.

wgtmac: Fair enough. Let's keep it for now.

wgtmac (approved):

src/iceberg/benchmark/CMakeLists.txt
wgtmac: Sounds good!

PR #828, opened 2026-07-11
feat(io): add ResolvingFileIO to resolve FileIO by location scheme and forward vended credentials

plusplusjiajia:
Follow-up to #719: replaces the temporary warehouse-based FileIO detection
  with a scheme-based `ResolvingFileIO`, mirroring Java.

  ### Why

  The FileIO impl was picked from `DetectBuiltinFileIO(warehouse)`, so any
  non-`s3://` warehouse resolved to `arrow-fs-local` and binding vended
  credentials failed with `Configured FileIO does not support vended storage
  credentials`. Credential-vending catalogs typically use a logical warehouse
  identifier (bucket ARN / catalog name), not a storage URI, so the feature
  only worked when `warehouse` was literally `s3://…`.

  ### What

  1. **`ResolvingFileIO` (core)** — picks the impl per file-path scheme
     (`s3`/`s3a`/`s3n`/`oss` → `arrow-fs-s3`; `file`/no scheme → local; else
     `NotSupported`), lazily loaded via `FileIORegistry` and cached; forwards
     the full credential list to impls that support it (Java
     `ResolvingFileIO` semantics). Registered as `resolving-file-io`.
  2. **REST default** — `MakeCatalogFileIO` defaults to `resolving-file-io`
     when `io-impl` is unset (Java `DEFAULT_FILE_IO_IMPL` likewise); the
     warehouse detection and the `io-impl`-vs-warehouse check are removed.
     Explicit `io-impl` is honored as before.
  3. **`ArrowS3FileIO`** — `SetStorageCredentials` skips non-S3 credential
     prefixes instead of rejecting the set (Java `S3FileIO` filters by `"s3"`);
     servers may vend credentials for several storage systems at once.


MisterRaindrop:
Could we add an end-to-end test using the actual Arrow FileIO implementations? The current tests use mocks and don’t cover the full `REST → ResolvingFileIO → FileIORegistry → Arrow FileIO` path.

src/iceberg/resolving_file_io.cc
```diff
@@ -0,0 +1,126 @@
...
+    return FileIORegistry::kArrowLocalFileIO;
+  }
+
+  const auto scheme = location.substr(0, pos);
+  if (scheme == "file") {
+    return FileIORegistry::kArrowLocalFileIO;
+  }
+  // S3-compatible schemes served by the S3 FileIO (Java: SCHEME_TO_FILE_IO).
+  // Keep in sync with CanonicalizeS3Scheme in arrow_s3_file_io.cc.
+  if (scheme == "s3" || scheme == "s3a" || scheme == "s3n" || scheme == "oss") {
```
wgtmac, line 47: Please either remove `oss` from this mapping or make `ArrowS3FileIO` accept and canonicalize `oss://` credentials. REST credentials are scoped by storage-location prefix, but the S3 delegate currently drops `oss://` credentials and falls back to its defaults. A catalog that vends only `oss://bucket/table` credentials will therefore fail at I/O time. Java avoids this mismatch by not mapping `oss` to `S3FileIO`.

src/iceberg/resolving_file_io.cc
```diff
@@ -0,0 +1,126 @@
...
+  for (auto& [io, locations] : locations_by_io) {
+    ICEBERG_RETURN_UNEXPECTED(io->DeleteFiles(locations));
+  }
+  return {};
+}
+
+Status ResolvingFileIO::SetStorageCredentials(
+    const std::vector<StorageCredential>& storage_credentials) {
+  std::lock_guard lock(mutex_);
+  storage_credentials_ = storage_credentials;
```
wgtmac, line 115: Please do not update `storage_credentials_` until the cached delegates have accepted the new credentials. If one delegate rejects the update, this method returns an error while `credentials()` reports the new values and the cached delegate keeps using the old ones; later cache hits never retry. Invalidating and rebuilding delegates on credential changes may be simpler than rolling back partial updates.

src/iceberg/resolving_file_io.h
```diff
@@ -0,0 +1,87 @@
...
+/// credentials are forwarded in full to every resolved FileIO that supports
+/// them; each applies the prefixes it understands and ignores the rest.
+class ICEBERG_EXPORT ResolvingFileIO final : public FileIO,
+                                             public SupportsStorageCredentials {
+ public:
+  explicit ResolvingFileIO(std::unordered_map<std::string, std::string> properties);
+  ~ResolvingFileIO() override;
+
+  /// \brief The FileIORegistry name of the implementation serving `location`.
+  static Result<std::string_view> ResolveFileIOName(std::string_view location);
```
wgtmac: Can this be private or internal? The resolver and its tests are the only callers, while exposing registry implementation names from an installed header makes the current scheme mapping part of the public API. Java keeps the equivalent `implFromLocation` package-private.

src/iceberg/test/rest_file_io_test.cc
```diff
@@ -158,16 +103,29 @@ TEST(RestFileIOTest, MakeCatalogFileIOUnregisteredCustomImplReturnsNotFound) {
...
-  auto result = MakeCatalogFileIO(config);
+  std::vector<StorageCredential> credentials = {
+      {.prefix = "oss", .config = {{"k1", "v1"}}},
+      {.prefix = "s3", .config = {{"k2", "v2"}}}};
+  auto result = MakeTableFileIO({{"warehouse", "logical_warehouse_name"}},
+                                /*table_config=*/{}, credentials);
   ASSERT_THAT(result, IsOk());
+
+  // Reaching a data file routes to the S3 FileIO with the full credential list.
+  (void)result.value()->NewInputFile("oss://bucket/db/table/data/file.parquet");
```
wgtmac, line 127: Could we cover this path with the real registered Arrow S3 FileIO? The mock only proves that the full credential list reaches the delegate; it cannot catch that production `ArrowS3FileIO` drops the `oss://` credential before opening this `oss://` location. A focused `REST -> ResolvingFileIO -> FileIORegistry -> ArrowS3FileIO` test would catch this regression.

src/iceberg/resolving_file_io.cc
plusplusjiajia: @wgtmac   Fixed via your second option: `ArrowS3FileIO` now accepts and canonicalizes `oss://` credential prefixes. Removing `oss` from the mapping isn't viable here — unlike Java, `arrow-fs-s3` is the only C++ impl that can serve `oss://`, so that would turn those locations into `NotSupported`.

src/iceberg/resolving_file_io.cc
plusplusjiajia: @wgtmac   Went with your simpler suggestion: `SetStorageCredentials` now invalidates the delegate cache and rebuilds lazily, so there's no window where a rejecting delegate leaves `credentials()` and the live delegates disagreeing. The test asserts the rebuild (factory invoked again, new delegate gets the new creds).

src/iceberg/resolving_file_io.h
plusplusjiajia: @wgtmac Done — now a free function in `resolving_file_io_internal.h` (not installed), so  the scheme mapping is no longer public API. It keeps `ICEBERG_EXPORT` only so tests link in shared builds, as `arrow/arrow_io_internal.h` does.

src/iceberg/test/rest_file_io_test.cc
plusplusjiajia: @wgtmac Added `test/rest_arrow_file_io_test.cc`, covering the chain against the real registered Arrow implementations on both legs: a local write/read round-trip with real bytes, and an `oss://` case where any break in routing/forwarding/prefix handling shows up as the warning, which the test asserts is absent. Also fixed `add_rest_iceberg_test`, which declared `USE_BUNDLE` but never used it, so REST tests couldn't link the bundle.

plusplusjiajia:
> Could we add an end-to-end test using the actual Arrow FileIO implementations? The current tests use mocks and don’t cover the full `REST → ResolvingFileIO → FileIORegistry → Arrow FileIO` path.

 @MisterRaindrop Done in `test/rest_arrow_file_io_test.cc` — full REST -> ResolvingFileIO -> registry -> Arrow FileIO chain against the real implementations: a local round-trip with real bytes, plus an S3 case asserting a vended `oss://` credential is applied. Good suggestion — the mocks couldn't catch either.

wgtmac (approved):
Thanks for adding ResolvingFileIO! I think current PR is good to merge. I have a followup refactoring work so let me merge it first.

PR #838, opened 2026-07-22
ci: replace cpp-linter-action with cpp-linter CLI

### LINKED ISSUE #336: Find alternative to cpp-linter/cpp-linter-action
cpp-linter/cpp-linter-action has introduced untrusted dependency since v2.16 and has been blocked by the Apache Infra team. See https://github.com/apache/infrastructure-actions/pull/325. We need to find a good alternative to help lint and format the code.

LuciferYang:
## What

`cpp-linter/cpp-linter-action` is frozen at v2.15.1. Its v2.16 release pulled in an untrusted dependency and was blocked by ASF Infra (apache/infrastructure-actions#325), and the ASF gateway ignores newer versions. The pinned action still runs, but it can no longer be updated.

## How

The action is a thin wrapper around the `cpp-linter` PyPI package, and the block does not touch that package. This installs it with pip and calls the `cpp-linter` CLI to run clang-tidy. Because it is a `run:` step and not a `uses:` reference, it adds no new third-party action to the ASF allowlist. The `checks-failed` output and the Fail fast step stay as they were.

Formatting is left to the pre-commit workflow, which already runs the pinned clang-format hook on every PR. So cpp-linter runs tidy-only: `--style=''`, and clang-format is not installed here. This keeps the clang-format version in one place instead of tracking it in two. `--version` points at the setup-python bin directory so clang-tidy resolves to the pip-installed version rather than a system clang-tidy on PATH.

I also looked at emitting SARIF and uploading it with `github/codeql-action/upload-sarif`, which is allowed since it is a `github/*` action. That route loses the thread comment and needs an extra conversion step, so calling the CLI directly stays closer to what we have now.

## Verified

CI ran the job on a temporary probe file seeded with a `NULL` and a mis-named member. clang-tidy resolved to the pip-installed 22.1.8 (not a system clang-tidy-22), flagged both issues, and set `checks-failed`, so the Fail fast step failed as expected. The probe has since been removed.

---

Closes #336


src/iceberg/temp_linter_probe.cc
```diff
@@ -0,0 +1,16 @@
+// TEMP verification file — will be dropped before merge.
```
LuciferYang: This is a temporary test for lint validation and will be deleted after testing.

src/iceberg/temp_linter_probe.cc
```diff
@@ -0,0 +1,16 @@
...
+// Intentionally violates .clang-tidy to confirm the new cpp-linter CLI runs.
```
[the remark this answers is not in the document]
LuciferYang: This is a temporary test for lint validation and will be deleted after testing.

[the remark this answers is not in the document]
LuciferYang: image

It looks like the check is taking effect. I will revert this file and run the tests again.

.github/workflows/cpp-linter.yml
```diff
@@ -91,24 +91,36 @@ jobs:
         with:
           path: ${{ github.workspace }}/.sccache
           key: sccache-cpp-linter-ubuntu-${{ github.run_id }}
-      - uses: cpp-linter/cpp-linter-action@0f6d1b8d7e38b584cbee606eb23d850c217d54f8 # v2.15.1
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        if: github.event_name == 'pull_request'
+      - name: Install cpp-linter and clang tools
+        if: github.event_name == 'pull_request'
+        run: pip install "cpp-linter==1.13.0" "clang-format==22.1.8" "clang-tidy==22.1.8"
```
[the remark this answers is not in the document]
LuciferYang: Good catch, done. Pinned setup-python to 3.13 and switched to `python -m pip install` so it uses the interpreter the action set up. I kept the run step as `cpp-linter` (its console entry point) since the package has no `__main__`, so it can't be called with `python -m`; it's installed into the same interpreter, so they stay in sync.

.github/workflows/cpp-linter.yml
```diff
@@ -91,24 +91,38 @@ jobs:
         with:
           path: ${{ github.workspace }}/.sccache
           key: sccache-cpp-linter-ubuntu-${{ github.run_id }}
-      - uses: cpp-linter/cpp-linter-action@0f6d1b8d7e38b584cbee606eb23d850c217d54f8 # v2.15.1
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        if: github.event_name == 'pull_request'
+        with:
+          python-version: '3.13'
+      - name: Install cpp-linter and clang tools
```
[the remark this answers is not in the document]
LuciferYang: https://github.com/apache/iceberg-cpp/pull/838#discussion_r3628066752

have previously tested this by introducing a faulty temporary file.

LuciferYang:
cc @wgtmac

.github/workflows/cpp-linter.yml
```diff
@@ -91,24 +91,38 @@ jobs:
...
+        if: github.event_name == 'pull_request'
+        run: python -m pip install "cpp-linter==1.13.0" "clang-format==22.1.8" "clang-tidy==22.1.8"
```
wgtmac: Should we parameterize these versions somewhere?

wgtmac (approved):
Nice job! Thanks a lot, @LuciferYang!

.github/workflows/cpp-linter.yml
LuciferYang: Done. The versions now sit in the job's `env` block (`CPP_LINTER_VERSION`, `CLANG_TOOLS_VERSION`), and `--version` takes its major from `${CLANG_TOOLS_VERSION%%.*}`, so the clang version is only written once.

I left one thing out: `.pre-commit-config.yaml` pins its own clang-format (v22.1.5), so it and the CI job aren't sharing a source. Actions has no way to read a version across files without extra tooling, so I didn't want to widen this PR for it. I can open a separate issue if aligning them is worth it.

wgtmac: Since `.pre-commit-config.yaml` pins clang-format to 22.1.5, shouldn't we use the same version here?

.github/workflows/cpp-linter.yml
```diff
@@ -48,6 +48,8 @@ jobs:
     env:
       SCCACHE_DIR: ${{ github.workspace }}/.sccache
       SCCACHE_CACHE_SIZE: "2G"
+      CPP_LINTER_VERSION: "1.13.0"
+      CLANG_TOOLS_VERSION: "22.1.8"
```
wgtmac:
```suggestion
      CLANG_TOOLS_VERSION: "22.1.5"  # Don't forget to update .pre-commit-config.yaml
```

.github/workflows/cpp-linter.yml
LuciferYang: Let me test this out.

LuciferYang: Good point on aligning. One wrinkle: `clang-format` 22.1.5 is on PyPI, but `clang-tidy` skips it (only 22.1.0, 22.1.0.1, 22.1.7, and 22.1.8 are published), so I can't pin both to 22.1.5 from a single variable without breaking the tidy install.

Two ways to handle it: split into `CLANG_FORMAT_VERSION` (22.1.5, matching pre-commit) and `CLANG_TIDY_VERSION` (nearest available, 22.1.8); or bump pre-commit's clang-format to 22.1.8 so all three land on one version. pre-commit only runs clang-format here, not clang-tidy, so the alignment really only concerns the format version. Which do you prefer?

.github/workflows/cpp-linter.yml
```diff
@@ -48,6 +48,8 @@ jobs:
     env:
       SCCACHE_DIR: ${{ github.workspace }}/.sccache
       SCCACHE_CACHE_SIZE: "2G"
+      CPP_LINTER_VERSION: "1.13.0"
+      CLANG_TOOLS_VERSION: "22.1.5"  # Don't forget to update .pre-commit-config.yaml
```
wgtmac:
```suggestion
      CLANG_TOOLS_VERSION: "22.1.8"
```

.github/workflows/cpp-linter.yml
wgtmac: Sorry I clicked the suggestion by accident, now I've reverted it.

wgtmac: I would prefer decoupling them into CLANG_FORMAT_VERSION and CLANG_TIDY_VERSION first because they may diverge in the future anyway. I'm fine to choose whatever version. We need to add a comment on either side to notify they should be in sync. WDYT?

LuciferYang: Split them as you suggested: `CLANG_FORMAT_VERSION` (22.1.5, matching pre-commit) and `CLANG_TIDY_VERSION` (22.1.8, since clang-tidy has no 22.1.5 on PyPI). Added a comment on the format version pointing at `.pre-commit-config.yaml` as the sync reminder.

LuciferYang: Before we settle on 22.1.5, one thing I hit while testing the pin. Running clang-format 22.1.5 (the version pre-commit pins) over the current tree, 4 files come back as non-conforming: `catalog/rest/error_handlers.cc`, `error_handlers.h`, `test/error_handlers_test.cc`, and `util/error_collector.h`. They pass under 22.1.8 (what CI runs today), but 22.1.5 wants to reformat them.

So the tree is effectively formatted to 22.1.8 right now, while pre-commit declares 22.1.5. The two are already out of sync. If CI's clang-format goes to 22.1.5, any later PR touching those files would fail the linter on formatting its author never changed.

I'd lean toward putting both CI and pre-commit on 22.1.8 instead: `CLANG_FORMAT_VERSION`, `CLANG_TIDY_VERSION`, and pre-commit all land on one version, and nothing needs reformatting. Same "keep them in sync" outcome you asked for, without the churn. WDYT?

.github/workflows/cpp-linter.yml
```diff
@@ -91,24 +96,40 @@ jobs:
...
-      - uses: cpp-linter/cpp-linter-action@0f6d1b8d7e38b584cbee606eb23d850c217d54f8 # v2.15.1
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        if: github.event_name == 'pull_request'
+        with:
+          python-version: '3.13'
+      - name: Install cpp-linter and clang tools
+        if: github.event_name == 'pull_request'
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install "cpp-linter==${CPP_LINTER_VERSION}" "clang-format==${CLANG_FORMAT_VERSION}" "clang-tidy==${CLANG_TIDY_VERSION}"
```
wgtmac: Do we need clang-format in this job at all? The pre-commit workflow already runs the pinned clang-format hook over all files on every PR. Keeping formatting there and making cpp-linter tidy-only would avoid duplicate checks and the version-sync problem. Could we drop `CLANG_FORMAT_VERSION` and the `clang-format` package here, and set `--style` to an empty string to disable format checks?

.github/workflows/cpp-linter.yml
```diff
@@ -91,24 +96,40 @@ jobs:
...
-          database: build
-          verbosity: 'debug'
-          # need '-fno-builtin-std-forward_like', see https://github.com/llvm/llvm-project/issues/101614
-          extra-args: '-std=c++23 -I$PWD/src -I$PWD/build/src -I$PWD/build/_deps/sqlpp23-src/include -I/usr/include/postgresql -I/usr/include/mysql -fno-builtin-std-forward_like'
+        # need '-fno-builtin-std-forward_like', see https://github.com/llvm/llvm-project/issues/101614
+        run: |
+          cpp-linter \
+            --style=file \
+            --tidy-checks='' \
+            --version="${CLANG_FORMAT_VERSION%%.*}" \
```
wgtmac: Even with a tidy-only setup, deriving `--version=22` would still bypass the pip pin. cpp-linter looks for `clang-tidy-22` first, and the probe run selected `/usr/bin/clang-tidy-22` at 22.1.2 instead of the installed 22.1.8. Could we point this to `${pythonLocation}/bin`, or omit `--version`, so the pip-installed binary is actually used?

wgtmac (commented):
Thinking more about this, I just realized that we do not need to duplicate the format check in cpp-linter. That should make our lives easier.

wgtmac:
Nice work! Thanks @LuciferYang!

LuciferYang:
@wgtmac Thanks for your thorough review and help.

PR #840, opened 2026-07-23
fix: compare transform function parameters

WZhuo:
## Summary
- include bucket counts in BucketTransform equality
- include truncate widths in TruncateTransform equality
- cover parameter and source-type differences in transform tests

## Testing
- build/src/iceberg/test/schema_test
- pre-commit hooks

### DIFF
--- a/src/iceberg/transform.h
+++ b/src/iceberg/transform.h
@@ -275,10 +275,11 @@ class ICEBERG_EXPORT TransformFunction {
     return lhs.Equals(rhs);
   }
 
- private:
-  /// \brief Compare two partition specs for equality.
+ protected:
+  /// \brief Compare the common properties of two transform functions for equality.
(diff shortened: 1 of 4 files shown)


src/iceberg/transform.h
```diff
@@ -275,10 +275,11 @@ class ICEBERG_EXPORT TransformFunction {
     return lhs.Equals(rhs);
   }
 
- private:
-  /// \brief Compare two partition specs for equality.
+ protected:
+  /// \brief Compare the common properties of two transform functions for equality.
   [[nodiscard]] virtual bool Equals(const TransformFunction& other) const;
```
zhjwpku, line 280: Nit: This is only used by the `==` overload, so maybe we can remove the `[[nodiscard]]` attribute? Same for the below.

zhjwpku (approved):

wgtmac (approved):
Make sense. Thanks @WZhuo for fixing it!

PR #841, opened 2026-07-24
REST: Support data sequence numbers in scan tasks

### LINKED ISSUE #834: Support data sequence numbers in REST scan tasks
REST-planned `FileScanTask` currently parses `first-row-id` from the embedded `data-file` JSON, but the REST scan-task payload does not expose the manifest-entry data sequence number.

For v3 row lineage readers, `_last_updated_sequence_number` inheritance needs the data file data sequence number. Without it, REST-planned tasks can pass `first_row_id` but cannot inherit `_last_updated_sequence_number` for files that omit or store null row-lineage columns.

Follow-up from apache/iceberg-cpp#822. The PR adds a skipped test, `FileScanTasksFromJsonTest.RowLineageSequence`, that should be enabled once REST scan-task JSON can carry this value.

u70b3:
## Summary

- parse the optional task-level `data-sequence-number` in REST file scan tasks
- serialize the field when a scan task's data file has a sequence number
- enable the row-lineage sequence test and cover absent/null compatibility

## Motivation

REST-planned scans can carry `first-row-id`, but C++ currently cannot receive the
manifest-entry data sequence number needed to inherit
`_last_updated_sequence_number` for v3 tables.

Closes #834

## Compatibility and dependency

This is the C++ side only. The task-level `data-sequence-number` field is not yet
present in the upstream REST OpenAPI, so this PR remains draft pending agreement
and publication of that protocol field. Older servers that omit the field, and
explicit `null`, continue to produce an unset data sequence number.

## Testing

- `pre-commit run --all-files`
- full CMake Debug build with GCC 14 and the repository-pinned CRoaring 4.4.3
- full CTest suite (18/18 passed)
- `rest_catalog_test` (317/317 passed)

## AI assistance

This change was implemented with assistance from OpenAI Codex. I reviewed the
implementation and tests, verified the JSON compatibility behavior, and ran the
full lint/build/test checks above. The remaining uncertainty is the final
upstream REST OpenAPI field definition; this draft intentionally does not modify
the Java SDK or specification.

### DIFF
--- a/src/iceberg/catalog/rest/json_serde.cc
+++ b/src/iceberg/catalog/rest/json_serde.cc
@@ -130,6 +130,7 @@ constexpr std::string_view kReferencedDataFile = "referenced-data-file";
 constexpr std::string_view kContentOffset = "content-offset";
 constexpr std::string_view kContentSizeInBytes = "content-size-in-bytes";
 constexpr std::string_view kDataFile = "data-file";
+constexpr std::string_view kDataSequenceNumber = "data-sequence-number";
 constexpr std::string_view kDeleteFileReferences = "delete-file-references";
 constexpr std::string_view kResidualFilter = "residual-filter";
 constexpr std::string_view kMapKeys = "keys";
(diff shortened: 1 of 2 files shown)


src/iceberg/catalog/rest/json_serde.cc
```diff
@@ -323,10 +324,11 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> FileScanTasksFromJson(
                             GetJsonValue<nlohmann::json>(task_json, kDataFile));
     ICEBERG_ASSIGN_OR_RAISE(
         auto data_file, DataFileFromJson(data_file_json, partition_spec_by_id, schema));
-    // FIXME: REST scan-task DataFile JSON currently carries first-row-id,
-    // but not the manifest-entry data sequence number. Until the REST API exposes
-    // it, REST-planned tasks cannot inherit _last_updated_sequence_number.
-    // See https://github.com/apache/iceberg-cpp/issues/834.
+    if (task_json.contains(kDataSequenceNumber) &&
```
wgtmac, line 327: Unfortunately we can't do this until the rest spec has adopted data sequence number to the FileScanTask object.

u70b3: Understood — I'll keep this as a draft blocked on the upstream REST spec adopting the field. I don't plan to file the spec proposal right now; happy to revisit once the protocol supports it. Feel free to close this PR if you'd rather not keep it open in the meantime.

PR #842, opened 2026-07-24
fix: track batch position delete references

u70b3:
## Summary

- collect referenced data-file paths when writing position deletes as an Arrow batch
- update referenced-path metadata only after the underlying write succeeds
- reject null batch inputs, null file paths, and sliced (non-zero offset) batches
- cover single-file, multi-file, mixed, empty, invalid, and failed-then-successful batches

## Testing

- `pre-commit run --files src/iceberg/data/position_delete_writer.cc src/iceberg/test/data_writer_test.cc`
- `build-gcc14/src/iceberg/test/data_test` (159 tests passed)
- `git diff --check`

### DIFF
--- a/src/iceberg/data/position_delete_writer.cc
+++ b/src/iceberg/data/position_delete_writer.cc
@@ -62,10 +62,40 @@ class PositionDeleteWriter::Impl {
   }
 
   Status Write(ArrowArray* data) {
+    ICEBERG_PRECHECK(data != nullptr, "Position delete data must not be null");
     ICEBERG_PRECHECK(buffered_paths_.empty(),
                      "Cannot write batch data when there are buffered deletes.");
-    // TODO(anyone): Extract file paths from ArrowArray to update referenced_paths_.
-    return writer_->Write(data);
+
+    ArrowSchema arrow_schema;
+    ICEBERG_RETURN_UNEXPECTED(ToArrowSchema(*delete_schema_, &arrow_schema));
+    internal::ArrowSchemaGuard schema_guard(&arrow_schema);
+
+    ArrowArrayView array_view;
+    internal::ArrowArrayViewGuard view_guard(&array_view);
+    ArrowError error;
+    ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+        ArrowArrayViewInitFromSchema(&array_view, &arrow_schema, &error), error);
+    ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+        ArrowArrayViewSetArray(&array_view, data, &error), error);
+
+    const auto* path_view = array_view.children[0];
+    if (ArrowArrayViewComputeNullCount(path_view) != 0) {
+      return InvalidArrowData("Position delete file paths must not contain null values");
+    }
+
+    std::set<std::string> pending_paths;
+    for (int64_t i = 0; i < data->length; ++i) {
+      auto path = ArrowArrayViewGetStringUnsafe(path_view, i);
+      if (path.size_bytes == 0) {
+        pending_paths.emplace();
+      } else {
+        pending_paths.emplace(path.data, static_cast<size_t>(path.size_bytes));
+      }
+    }
+
+    ICEBERG_RETURN_UNEXPECTED(writer_->Write(data));
+    referenced_paths_.merge(pending_paths);
+    return {};
   }
 
   Status WriteDelete(std::string_view file_path, int64_t pos) {
--- a/src/iceberg/data/position_delete_writer.cc
+++ b/src/iceberg/data/position_delete_writer.cc
@@ -63,6 +63,7 @@ class PositionDeleteWriter::Impl {
 
   Status Write(ArrowArray* data) {
     ICEBERG_PRECHECK(data != nullptr, "Position delete data must not be null");
+    internal::ArrowArrayGuard data_guard(data);
     ICEBERG_PRECHECK(data->offset == 0,
                      "Position delete data with a non-zero offset is not supported");
     ICEBERG_PRECHECK(buffered_paths_.empty(),
--- a/src/iceberg/data/position_delete_writer.cc
+++ b/src/iceberg/data/position_delete_writer.cc
@@ -63,6 +63,8 @@ class PositionDeleteWriter::Impl {
 
   Status Write(ArrowArray* data) {
     ICEBERG_PRECHECK(data != nullptr, "Position delete data must not be null");
+    ICEBERG_PRECHECK(data->offset == 0,
+                     "Position delete data with a non-zero offset is not supported");
     ICEBERG_PRECHECK(buffered_paths_.empty(),
                      "Cannot write batch data when there are buffered deletes.");
 
@@ -71,10 +73,10 @@ class PositionDeleteWriter::Impl {
     internal::ArrowSchemaGuard schema_guard(&arrow_schema);
 
     ArrowArrayView array_view;
-    internal::ArrowArrayViewGuard view_guard(&array_view);
     ArrowError error;
     ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
         ArrowArrayViewInitFromSchema(&array_view, &arrow_schema, &error), error);
+    internal::ArrowArrayViewGuard view_guard(&array_view);
     ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
         ArrowArrayViewSetArray(&array_view, data, &error), error);
 
--- a/src/iceberg/data/position_delete_writer.cc
+++ b/src/iceberg/data/position_delete_writer.cc
@@ -19,8 +19,11 @@
 
 #include "iceberg/data/position_delete_writer.h"
 
+#include <functional>
 #include <map>
 #include <set>
+#include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
@@ -57,8 +60,17 @@ class PositionDeleteWriter::Impl {
     ICEBERG_ASSIGN_OR_RAISE(auto writer,
                             WriterFactoryRegistry::Open(options.format, writer_options));
 
-    return std::unique_ptr<Impl>(
+    auto impl = std::unique_ptr<Impl>(
         new Impl(std::move(options), std::move(delete_schema), std::move(writer)));
+    ICEBERG_RETURN_UNEXPECTED(impl->InitSchema());
+    return impl;
+  }
+
+  ~Impl() {
+    ArrowArrayViewReset(&array_view_);
+    if (arrow_schema_.release != nullptr) {
+      ArrowSchemaRelease(&arrow_schema_);
+    }
   }
 
   Status Write(ArrowArray* data) {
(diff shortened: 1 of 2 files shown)


u70b3:
Hi @wgtmac, gentle ping for a review when you have a moment. This has been sitting for about two weeks; all CI checks are green and it merges cleanly against the latest main. The change tracks referenced data-file paths for batch position-delete writes so `referenced_data_file` and bounds metadata match the `WriteDelete` path, and hardens ownership/validation for rejected batches. Thanks!

src/iceberg/data/position_delete_writer.cc
```diff
@@ -62,10 +62,43 @@ class PositionDeleteWriter::Impl {
...
+    ArrowArrayView array_view;
+    ArrowError error;
+    ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+        ArrowArrayViewInitFromSchema(&array_view, &arrow_schema, &error), error);
+    internal::ArrowArrayViewGuard view_guard(&array_view);
+    ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+        ArrowArrayViewSetArray(&array_view, data, &error), error);
+
+    const auto* path_view = array_view.children[0];
+    if (ArrowArrayViewComputeNullCount(path_view) != 0) {
```
wgtmac, line 85: `ArrowArrayViewComputeNullCount(path_view)` scans the child view length, not necessarily the parent batch length. Please check it only for i less than data->length, and apply the same check to `pos` field.

src/iceberg/data/position_delete_writer.cc
```diff
@@ -62,10 +62,43 @@ class PositionDeleteWriter::Impl {
...
+    ICEBERG_PRECHECK(data != nullptr, "Position delete data must not be null");
+    internal::ArrowArrayGuard data_guard(data);
+    ICEBERG_PRECHECK(data->offset == 0,
+                     "Position delete data with a non-zero offset is not supported");
     ICEBERG_PRECHECK(buffered_paths_.empty(),
                      "Cannot write batch data when there are buffered deletes.");
-    // TODO(anyone): Extract file paths from ArrowArray to update referenced_paths_.
-    return writer_->Write(data);
+
+    ArrowSchema arrow_schema;
```
wgtmac, line 72: Every Write rebuilds the same schema and allocates a fresh ArrowArrayView. The schema is immutable for this writer, so this adds allocator work to every batch. Perhaps we can initialize the schema and view once in Impl and only rebind the incoming array here.

src/iceberg/data/position_delete_writer.cc
```diff
@@ -62,10 +62,43 @@ class PositionDeleteWriter::Impl {
...
+    internal::ArrowArrayViewGuard view_guard(&array_view);
+    ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+        ArrowArrayViewSetArray(&array_view, data, &error), error);
+
+    const auto* path_view = array_view.children[0];
+    if (ArrowArrayViewComputeNullCount(path_view) != 0) {
+      return InvalidArrowData("Position delete file paths must not contain null values");
+    }
+
+    std::set<std::string> pending_paths;
```
wgtmac, line 89: `pending_paths` allocates a tree node and copies each unique path for every batch, then merges into another set. Batch writes can be frequent so a reusable scratch set or a vector plus post-write insertion would avoid much of this churn while keeping the failure-safe delayed merge.

src/iceberg/data/position_delete_writer.cc
```diff
@@ -62,10 +62,43 @@ class PositionDeleteWriter::Impl {
...
+
+    ArrowArrayView array_view;
+    ArrowError error;
+    ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+        ArrowArrayViewInitFromSchema(&array_view, &arrow_schema, &error), error);
+    internal::ArrowArrayViewGuard view_guard(&array_view);
+    ICEBERG_NANOARROW_RETURN_UNEXPECTED_WITH_ERROR(
+        ArrowArrayViewSetArray(&array_view, data, &error), error);
+
+    const auto* path_view = array_view.children[0];
```
wgtmac: children[0] is correct only because this class privately fixes the schema to file_path, pos. Write accepts a raw ArrowArray, so callers get no visible schema contract. Add a short API note or validate the schema at the boundary if arbitrary batches are expected.

src/iceberg/data/position_delete_writer.cc
```diff
@@ -62,10 +62,43 @@ class PositionDeleteWriter::Impl {
...
+    if (ArrowArrayViewComputeNullCount(path_view) != 0) {
+      return InvalidArrowData("Position delete file paths must not contain null values");
+    }
+
+    std::set<std::string> pending_paths;
+    for (int64_t i = 0; i < data->length; ++i) {
+      auto path = ArrowArrayViewGetStringUnsafe(path_view, i);
+      if (path.size_bytes == 0) {
```
wgtmac: An empty non-null path is inserted into pending_paths and can become an empty referenced_data_file. The spec requires a fully-qualified data-file URI. Reject empty paths here and in WriteDelete, or make the contract explicit.

src/iceberg/test/data_writer_test.cc
```diff
@@ -458,6 +453,145 @@ TEST_F(PositionDeleteWriterTest, WriteBatchData) {
   const auto& data_file = metadata_result.value().data_files[0];
   EXPECT_EQ(data_file->content, DataFile::Content::kPositionDeletes);
   EXPECT_GT(data_file->file_size_in_bytes, 0);
+  ASSERT_TRUE(data_file->referenced_data_file.has_value());
+  EXPECT_EQ(data_file->referenced_data_file.value(), "data_file_1.parquet");
+  // Bounds for delete metadata columns are kept when referencing a single file.
```
wgtmac: Checking only that bound keys exist does not prove the batch was written correctly. A wrong serialized bound, missing row, or incorrect counts can still pass. Read back the delete records and assert exact row_count, bounds, and filtered count maps, matching the Java tests.

src/iceberg/test/data_writer_test.cc
```diff
@@ -458,6 +453,145 @@ TEST_F(PositionDeleteWriterTest, WriteBatchData) {
...
+  EXPECT_TRUE(data_file->lower_bounds.contains(MetadataColumns::kDeleteFilePathColumnId));
+  EXPECT_TRUE(data_file->lower_bounds.contains(MetadataColumns::kDeleteFilePosColumnId));
+  EXPECT_TRUE(data_file->upper_bounds.contains(MetadataColumns::kDeleteFilePathColumnId));
+  EXPECT_TRUE(data_file->upper_bounds.contains(MetadataColumns::kDeleteFilePosColumnId));
+}
+
+TEST_F(PositionDeleteWriterTest, WriteBatchRejectsSlicedData) {
+  auto writer_result = PositionDeleteWriter::Make(MakeDeleteOptions());
```
wgtmac: All of the new PositionDeleteWriter cases use MakeDeleteOptions, which hard-codes Parquet. The batch path is shared, but Avro has different metrics expectations, so this leaves the Avro branch untested. Parameterize the relevant cases for both formats.

src/iceberg/test/data_writer_test.cc
```diff
@@ -458,6 +453,145 @@ TEST_F(PositionDeleteWriterTest, WriteBatchData) {
...
+
+  auto metadata_result = writer->Metadata();
+  ASSERT_THAT(metadata_result, IsOk());
+
+  const auto& data_file = metadata_result.value().data_files[0];
+  ASSERT_TRUE(data_file->referenced_data_file.has_value());
+  EXPECT_EQ(data_file->referenced_data_file.value(), "data_file_1.parquet");
+}
+
+TEST_F(PositionDeleteWriterTest, WriteBatchDataForMultipleFiles) {
```
wgtmac, line 512: This is one batch containing two paths, not multiple successful Write calls. A bug that replaces referenced_paths_ instead of unioning across batches would still pass. Add two successful batches with disjoint paths.

src/iceberg/test/data_writer_test.cc
```diff
@@ -458,6 +453,145 @@ TEST_F(PositionDeleteWriterTest, WriteBatchData) {
...
+  ASSERT_THAT(writer_result, IsOk());
+  auto writer = std::move(writer_result.value());
+
+  // A rejected batch must not contribute referenced paths.
+  auto bad_data =
+      CreatePositionDeleteData(R"([[null, 0], ["data_file_bad.parquet", 1]])");
+  ArrowArray bad_array;
+  ASSERT_TRUE(::arrow::ExportArray(*bad_data, &bad_array).ok());
+  internal::ArrowArrayGuard bad_array_guard(&bad_array);
+  ASSERT_THAT(writer->Write(&bad_array), IsError(ErrorKind::kInvalidArrowData));
```
wgtmac, line 496: This looks odd to me because we continue to use a failed writer which should not happen in production. And this does actually verify the case name `FailedBatchWriteDoesNotTrackReferencedFiles`.

src/iceberg/test/data_writer_test.cc
```diff
@@ -458,6 +453,145 @@ TEST_F(PositionDeleteWriterTest, WriteBatchData) {
...
+  auto test_data = CreatePositionDeleteData(R"([["data_file_1.parquet", 0]])");
+  ArrowArray arrow_array;
+  ASSERT_TRUE(::arrow::ExportArray(*test_data, &arrow_array).ok());
+  ASSERT_THAT(writer->Write(&arrow_array), IsOk());
+  ASSERT_THAT(writer->WriteDelete("data_file_2.parquet", 5), IsOk());
+  ASSERT_THAT(writer->Close(), IsOk());
+
+  auto metadata_result = writer->Metadata();
+  ASSERT_THAT(metadata_result, IsOk());
+  EXPECT_FALSE(metadata_result.value().data_files[0]->referenced_data_file.has_value());
```
wgtmac, line 551: This only checks the per-file hint. WriteResult also exposes referenced_data_files; assert that public result as well if this writer is meant to satisfy the FileWriter contract.

src/iceberg/test/data_writer_test.cc
```diff
@@ -458,6 +453,145 @@ TEST_F(PositionDeleteWriterTest, WriteBatchData) {
...
+
+  auto result = writer->Write(&arrow_array);
+  EXPECT_EQ(arrow_array.release, nullptr);
+  internal::ArrowArrayGuard array_guard(&arrow_array);
+  ASSERT_THAT(result, IsError(ErrorKind::kInvalidArrowData));
+  EXPECT_THAT(result,
+              HasErrorMessage("Position delete file paths must not contain null values"));
+}
+
+TEST_F(PositionDeleteWriterTest, WriteBatchRejectsNullData) {
```
wgtmac, line 571: This only exercises a one-line null precondition and does not touch batch paths or metadata. It is low-value so please consider dropping it or folding it into a broader invalid-input test.

src/iceberg/data/position_delete_writer.cc
wgtmac, line 92: Should we error out in this case?

wgtmac (changes requested):
Thanks @u70b3 for fixing this and sorry for the delay! I have left some comments.

src/iceberg/data/position_delete_writer.cc
u70b3: Done in ce7184a. The null check is now per row over `i < data->length` using `ArrowArrayViewIsNull`, applied to both the `file_path` and the `pos` child views, so it no longer depends on the child view length.

src/iceberg/data/position_delete_writer.cc
u70b3: Done in ce7184a. The delete schema is immutable, so it is converted to Arrow once in `Impl::InitSchema()` and the `ArrowArrayView` is initialized from it there. `Write` now only rebinds the existing view with `ArrowArrayViewSetArray`, and `FlushBuffer` reuses the same Arrow schema instead of rebuilding it. The view and the schema are released in `Impl`'s destructor.

src/iceberg/data/position_delete_writer.cc
u70b3: Done in ce7184a, with a slightly different approach. A reusable scratch set does not actually avoid the churn, since `clear()` frees every node, so the allocations would keep coming for paths that are already tracked.

Instead the paths are recorded optimistically: `referenced_paths_` is now `std::set<std::string, std::less<>>`, so a `string_view` that is already tracked costs only a lookup and no allocation at all. The iterators of the entries inserted by the current batch are kept in a small reused vector and erased again if the batch is rejected, which keeps the failure-safe delayed merge: a rejected batch still leaves no trace in the metadata.

src/iceberg/test/data_writer_test.cc
u70b3: Done in ce7184a. `WriteBatchDataForMultipleFiles` now performs two successful writes with disjoint paths, so a bug that replaced `referenced_paths_` instead of unioning across batches would fail the test. It also asserts the public `WriteResult::referenced_data_files`.

src/iceberg/test/data_writer_test.cc
u70b3: Agreed, reworked in ce7184a. The test now writes a successful batch first and a rejected batch last, so the writer is never used after a failure. The rejected batch references a valid path and is then rejected by the null path, and after `Close()` the test verifies that only `data_file_1.parquet` is tracked. Because the bad path is inserted before the batch fails, this also covers the rollback path.

src/iceberg/test/data_writer_test.cc
u70b3: Done in ce7184a. `Metadata()` now populates `WriteResult::referenced_data_files`, and both `WriteBatchThenDeleteTracksAllReferencedFiles` and `WriteBatchDataForMultipleFiles` assert it in addition to the per-file hint.

src/iceberg/test/data_writer_test.cc
u70b3: Done in ce7184a. `WriteBatchRejectsNullData` and `WriteBatchRejectsNullFilePath` are folded into a single `WriteBatchRejectsInvalidInput` covering a null array, a null file path, a null position, and an empty file path.

src/iceberg/data/position_delete_writer.cc
u70b3: Good catch, it is now an error in ce7184a: `InvalidArrowData("Position delete file paths must not be empty")`. Silently tracking an empty path would have been wrong, and it is covered by `WriteBatchRejectsInvalidInput`.

wgtmac (approved):
Thanks, @u70b3!

PR #843, opened 2026-07-25
fix: reject malformed commit property integers

fallintoplace:
## Summary

- require commit property integers to parse completely, matching Java's `Integer.parseInt`
- reject values with alphanumeric or decimal suffixes instead of accepting their numeric prefix
- preserve the existing validation errors for invalid and out-of-range values

## Testing

- `ctest --test-dir build -R ^table_test$ --output-on-failure`

### DIFF
--- a/src/iceberg/util/property_util.cc
+++ b/src/iceberg/util/property_util.cc
@@ -30,14 +30,17 @@ Status PropertyUtil::ValidateCommitProperties(
   for (const auto& property : TableProperties::commit_properties()) {
     if (auto it = properties.find(property); it != properties.end()) {
       int32_t parsed;
-      auto [ptr, ec] = std::from_chars(it->second.data(),
-                                       it->second.data() + it->second.size(), parsed);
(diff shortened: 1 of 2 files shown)


evindj (approved):

src/iceberg/util/property_util.cc
```diff
@@ -30,14 +30,17 @@ Status PropertyUtil::ValidateCommitProperties(
   for (const auto& property : TableProperties::commit_properties()) {
     if (auto it = properties.find(property); it != properties.end()) {
       int32_t parsed;
-      auto [ptr, ec] = std::from_chars(it->second.data(),
-                                       it->second.data() + it->second.size(), parsed);
+      const auto* end = it->second.data() + it->second.size();
```
wgtmac: I would suggest to replace it with `StringUtils::ParseNumber<int32_t>`.

wgtmac (approved):

wgtmac:
Thanks @fallintoplace and @evindj!

PR #845, opened 2026-07-25
fix: reject malformed Arrow field IDs

fallintoplace:
## Summary

- reject malformed `PARQUET:field_id` metadata instead of accepting a valid numeric prefix or returning the unknown-field sentinel
- preserve the unknown-field sentinel when field-ID metadata is absent
- add coverage for trailing characters, overflow, and an empty value

## Testing

- `arrow_test --gtest_filter=FromArrowSchemaTest.RejectMalformedFieldIdMetadata`

### DIFF
--- a/src/iceberg/schema_internal.cc
+++ b/src/iceberg/schema_internal.cc
@@ -227,7 +227,7 @@ Status ToArrowSchema(const Schema& schema, ArrowSchema* out) {
 
 namespace {
 
-int32_t GetFieldId(const ArrowSchema& schema) {
+Result<int32_t> GetFieldId(const ArrowSchema& schema) {
   if (schema.metadata == nullptr) {
     return kUnknownFieldId;
   }
@@ -240,9 +240,14 @@ int32_t GetFieldId(const ArrowSchema& schema) {
     return kUnknownFieldId;
   }
 
-  int32_t field_id = kUnknownFieldId;
-  std::from_chars(field_id_value.data, field_id_value.data + field_id_value.size_bytes,
-                  field_id);
+  int32_t field_id = 0;
+  const auto* end = field_id_value.data + field_id_value.size_bytes;
+  const auto [ptr, ec] = std::from_chars(field_id_value.data, end, field_id);
+  if (ec != std::errc{} || ptr != end) {
+    return InvalidSchema(
+        "Invalid Arrow field ID: '{}'",
+        std::string_view(field_id_value.data, field_id_value.size_bytes));
+  }
 
   return field_id;
 }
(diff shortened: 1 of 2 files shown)


src/iceberg/schema_internal.cc
```diff
@@ -227,7 +227,7 @@ Status ToArrowSchema(const Schema& schema, ArrowSchema* out) {
 
 namespace {
 
-int32_t GetFieldId(const ArrowSchema& schema) {
+Result<int32_t> GetFieldId(const ArrowSchema& schema) {
   if (schema.metadata == nullptr) {
     return kUnknownFieldId;
```
evindj, line 231: Should this be changed? It is not clear to me what is the difference between unknowFieldId (returning the constant) and InvalidFieldId (returning in a error)

fallintoplace: They are intentionally different. If the `PARQUET:field_id` metadata is missing, we keep the existing behavior and use the unknown field ID sentinel. If the key is present but the value is malformed or outside `int32_t`, we return an invalid schema error instead.

src/iceberg/schema_internal.cc
```diff
@@ -240,9 +240,14 @@ int32_t GetFieldId(const ArrowSchema& schema) {
...
   }
 
-  int32_t field_id = kUnknownFieldId;
-  std::from_chars(field_id_value.data, field_id_value.data + field_id_value.size_bytes,
-                  field_id);
+  int32_t field_id = 0;
+  const auto* end = field_id_value.data + field_id_value.size_bytes;
+  const auto [ptr, ec] = std::from_chars(field_id_value.data, end, field_id);
```
wgtmac: What about using `ParseNumber<int32_t>` to simplify this?

src/iceberg/schema_internal.cc
```diff
@@ -240,19 +239,22 @@ int32_t GetFieldId(const ArrowSchema& schema) {
...
   }
 
-  int32_t field_id = kUnknownFieldId;
-  std::from_chars(field_id_value.data, field_id_value.data + field_id_value.size_bytes,
-                  field_id);
+  std::string_view field_id(field_id_value.data, field_id_value.size_bytes);
+  auto field_id_result = StringUtils::ParseNumber<int32_t>(field_id);
+  if (!field_id_result.has_value()) {
+    return InvalidSchema(
+        "Invalid Arrow field ID: '{}'", field_id);
```
wgtmac:
```suggestion
    return InvalidSchema("Invalid Arrow field ID: '{}'", field_id);
```

CI failed to due this unformatted line.

wgtmac (approved):

wgtmac:
Sorry for the late review. Thanks @fallintoplace and @evindj!

PR #846, opened 2026-07-25
fix: reject malformed metadata file versions

fallintoplace:
## Summary

- reject malformed and out-of-range version segments in metadata filenames instead of accepting a valid numeric prefix
- start generated metadata filenames at version 0 when the current version cannot be parsed, matching Java
- add coverage for trailing characters, overflow, and an empty version segment through the metadata write path

## Testing

- `arrow_test --gtest_filter=MetadataIOTest.WriteMetadataWithBase`

### DIFF
--- a/src/iceberg/table_metadata.cc
+++ b/src/iceberg/table_metadata.cc
@@ -517,11 +517,18 @@ int32_t TableMetadataUtil::ParseVersionFromLocation(std::string_view metadata_lo
   size_t version_start = metadata_location.find_last_of('/') + 1;
   size_t version_end = metadata_location.find('-', version_start);
 
-  int32_t version = -1;
-  if (version_end != std::string::npos) {
-    std::from_chars(metadata_location.data() + version_start,
-                    metadata_location.data() + version_end, version);
+  if (version_end == std::string::npos) {
+    return -1;
   }
+
+  int32_t version = -1;
+  const auto* version_begin = metadata_location.data() + version_start;
+  const auto* version_limit = metadata_location.data() + version_end;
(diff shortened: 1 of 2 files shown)


src/iceberg/table_metadata.cc
```diff
@@ -517,11 +517,18 @@ int32_t TableMetadataUtil::ParseVersionFromLocation(std::string_view metadata_lo
...
+
   int32_t version = -1;
-  if (version_end != std::string::npos) {
-    std::from_chars(metadata_location.data() + version_start,
-                    metadata_location.data() + version_end, version);
+  const auto* version_begin = metadata_location.data() + version_start;
+  const auto* version_limit = metadata_location.data() + version_end;
+  const auto [ptr, ec] = std::from_chars(version_begin, version_limit, version);
+  if (ec != std::errc{} || ptr != version_limit) {
+    return -1;
```
evindj, line 529: This is a non blocking comment (especially given that this is a private method) but I wonder if we should change ```ParseVersionFromLocation()``` to return ```Result<>``` to ensure that callers explicit have to check for errors and act accordingly.

evindj (approved):

evindj (approved):

src/iceberg/table_metadata.cc
fallintoplace: I considered that, but a parse failure is intentionally handled by starting at version 0, which matches the Java behavior. Returning the error from `Write` would change that fallback. Since this is a private helper with one caller and the `-1` behavior is documented, I think keeping the sentinel is simpler here.

wgtmac (approved):
Sorry for the late review. Thanks @fallintoplace and @evindj!

PR #847, opened 2026-07-25
fix: parse boolean properties case-insensitively

fallintoplace:
## Summary

- parse boolean table and catalog properties case-insensitively, matching Java's `Boolean.parseBoolean`
- continue treating values other than a case-insensitive `true` as `false`
- add coverage for lowercase, uppercase, mixed-case, false, and unrecognized values

## Testing

- `util_test --gtest_filter=ConfigTest.ParseBooleanIgnoringCase`
- pre-commit hooks for the changed files

### DIFF
--- a/src/iceberg/util/config.h
+++ b/src/iceberg/util/config.h
@@ -54,7 +54,7 @@ U DefaultFromString(const std::string& val) {
   if constexpr (std::is_same_v<U, std::string>) {
     return val;
   } else if constexpr (std::is_same_v<U, bool>) {
-    return val == "true";
+    return StringUtils::EqualsIgnoreCase(val, "true");
   } else if constexpr ((std::is_signed_v<U> && std::is_integral_v<U>) ||
                        std::is_floating_point_v<U>) {
     ICEBERG_ASSIGN_OR_THROW(auto res, StringUtils::ParseNumber<U>(val));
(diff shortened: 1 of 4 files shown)


src/iceberg/util/config.h
```diff
@@ -54,7 +54,7 @@ U DefaultFromString(const std::string& val) {
   if constexpr (std::is_same_v<U, std::string>) {
     return val;
   } else if constexpr (std::is_same_v<U, bool>) {
-    return val == "true";
+    return StringUtils::EqualsIgnoreCase(val, "true");
```
evindj, line 57: pretty sure you did but it would ne nice to check for other places in the code base where similar comparison with literals is made.

evindj (approved):

src/iceberg/util/config.h
fallintoplace: Good call. I found the same case-sensitive parsing in the S3 properties and pushed a follow-up for `s3.ssl.enabled` and `s3.path-style-access`. The other literal comparisons I found are Avro schema attributes rather than configuration properties, so I left those unchanged.

zhjwpku (approved):

wgtmac (approved):
Thanks @fallintoplace for fixing it and @evindj @zhjwpku for the review!

PR #857, opened 2026-07-27
fix: guard equality delete bounds against missing fields

### LINKED ISSUE #856: bug: CanContainEqDeletesForFile crashes with std::bad_optional_access when an equality delete file has bounds for only some fields
**Summary**

`internal::CanContainEqDeletesForFile` (`src/iceberg/delete_file_index.cc`) throws `std::bad_optional_access` and crashes the process when an equality delete file carries lower/upper bounds for some of its equality fields but not others. The throw is an uncaught exception escaping the `Result`-based (non-throwing) call chain, so it terminates the caller rather than surfacing as an error.

Two reachable entry points, both consuming files written by arbitrary engines:

- Scan planning: `ManifestGroup` → `DeleteFileIndex::ForEntry` (`manifest_group.cc:214`).
- Commit validation: `MergingSnapshotUpdate::ValidateNoNewDeletesForDataFiles` → `DeleteFileIndex::ForDataFile` (`merging_snapshot_update.cc:1196`/`:1256`), reached from `RowDelta`/`OverwriteFiles`/`RewriteFiles`.

Equality delete files with bounds for only some fields are produced legitimately by per-column metrics modes (`counts`/`none`/`truncate`) and by cross-engine writers; `ContentFileUtil::DropUnselectedStats` drops non-equality column stats without back-filling missing bounds, so the asymmetry is preserved on read. The crash is reachable in normal reader use, not just a synthetic input.

**Root Cause**

`EqualityDeleteFile::LowerBound`/`UpperBound` return `Result<std::optional<std::reference_wrapper<const Literal>>>`, i.e. `std::expected<std::optional<...>, Error>`. A field that has no bound yields an *engaged* `expected` wrapping a *disengaged* `optional`. The guard checks the wrong layer:

```cpp
auto delete_lower = delete_file.LowerBound(field_id);   // expected<optional<...>>
auto delete_upper = delete_file.UpperBound(field_id);
if (!delete_lower.has_value() || !delete_upper.has_value()) {  // expected::has_value(): only the error state
  continue;  // Missing bounds, assume may match
}
...
RangesOverlap(..., delete_lower->value().get(), delete_upper->value().get());  // optional::value() on nullopt -> throws
```

`has_value()` here calls `std::expected::has_value()`, which only reports whether an error is present; it does not detect the empty inner `optional`. So a missing bound passes the guard and `->value()` throws. The data-side guard a few lines above (`data_file.lower_bounds.find(...) == cend() || ...->second.empty()`) checks field-level presence correctly, so the two sides are asymmetric. Java's `DeleteFileIndex.canContainEqDeletesForFile` (checked against apache/iceberg `main`) guards all four bounds uniformly (`dataLower == null || dataUpper == null || deleteLower == null || deleteUpper == null` → `continue`, "assume the delete file may match"), and only prunes when all four are present and the ranges do not overlap.

**Reproduction**

An equality delete over fields `{1, 2}` whose `lower_bounds`/`upper_bounds` contain only field 1, evaluated against a data file that has bounds for both fields (field 1's range overlapping so evaluation proceeds to field 2), throws `std::bad_optional_access` out of `EqualityDeletes::Filter`.

**Proposed Fix**

Unwrap the `expected` first with `ICEBERG_ASSIGN_OR_RAISE` so `has_value()` tests the inner `optional`, matching the data-side guard and Java:

```cpp
ICEBERG_ASSIGN_OR_RAISE(auto delete_lower, delete_file.LowerBound(field_id));
ICEBERG_ASSIGN_OR_RAISE(auto delete_upper, delete_file.UpperBound(field_id));
if (!delete_lower.has_value() || !delete_upper.has_value()) {
  continue;
}
...
RangesOverlap(..., delete_lower->get(), delete_upper->get());
```

This also propagates a genuine bound-deserialization error instead of silently swallowing it as "may match", consistent with the data-side handling in the same function.

I have a fix and a regression test ready and will open a PR.

LuciferYang:
## What

`internal::CanContainEqDeletesForFile` crashes with `std::bad_optional_access` when an equality delete file has lower/upper bounds for some of its equality fields but not others. `EqualityDeleteFile::LowerBound`/`UpperBound` return `Result<std::optional<...>>` (i.e. `std::expected<std::optional<...>, Error>`), and a field with no bound yields an engaged `expected` wrapping a disengaged `optional`. The guard checked `std::expected::has_value()`, which only reports the error state and does not detect the empty inner `optional`, so a missing bound slipped past and `->value()` threw at the range check.

The throw is an uncaught exception escaping the `Result`-based call chain, so it terminates the caller instead of surfacing as an error. It is reachable during scan planning (`ManifestGroup` → `DeleteFileIndex::ForEntry`) and commit validation (`MergingSnapshotUpdate::ValidateNoNewDeletesForDataFiles` → `ForDataFile`). Equality delete files with bounds for only some fields are produced legitimately by per-column metrics modes (`counts`/`none`/`truncate`) and by cross-engine writers, so this is reachable in normal reader use.

Fixes #856.

## How

Unwrap the `expected` with `ICEBERG_ASSIGN_OR_RAISE` before the guard, so `has_value()` tests the inner `optional`. This matches the data-side guard already used a few lines above in the same function, and matches Java's `DeleteFileIndex.canContainEqDeletesForFile`, which guards all four bounds uniformly and assumes may-match on any missing bound (it never prunes on missing statistics).

As a side effect this propagates a genuine bound-deserialization error upward instead of silently treating it as may-match. That is consistent with the data-side `Literal::Deserialize` calls in the same function, which already propagate via `ICEBERG_ASSIGN_OR_RAISE`.

## Testing

Added `DeleteFileIndexTest.TestEqualityDeletePartialFieldBounds`: an equality delete over fields `{1, 2}` with bounds for field 1 only, evaluated against a data file that has bounds for both (field 1's range overlapping so evaluation reaches field 2). Verified fail-without (throws `std::bad_optional_access`) / pass-with (returns the delete file as may-match). `null_value_counts` is pinned to 0 for both fields so the null short-circuits cannot mask the range-check path if the schema fields later change nullability. Full `manifest_test` passes.

### DIFF
--- a/src/iceberg/delete_file_index.cc
+++ b/src/iceberg/delete_file_index.cc
@@ -145,8 +145,8 @@ Result<bool> CanContainEqDeletesForFile(const DataFile& data_file,
       continue;  // Missing bounds, assume may match
     }
 
-    auto delete_lower = delete_file.LowerBound(field_id);
-    auto delete_upper = delete_file.UpperBound(field_id);
+    ICEBERG_ASSIGN_OR_RAISE(auto delete_lower, delete_file.LowerBound(field_id));
+    ICEBERG_ASSIGN_OR_RAISE(auto delete_upper, delete_file.UpperBound(field_id));
     if (!delete_lower.has_value() || !delete_upper.has_value()) {
       continue;  // Missing bounds, assume may match
     }
@@ -158,8 +158,8 @@ Result<bool> CanContainEqDeletesForFile(const DataFile& data_file,
     ICEBERG_ASSIGN_OR_RAISE(auto data_upper,
                             Literal::Deserialize(data_upper_it->second, primitive_type));
 
-    if (!RangesOverlap(data_lower, data_upper, delete_lower->value().get(),
-                       delete_upper->value().get())) {
+    if (!RangesOverlap(data_lower, data_upper, delete_lower->get(),
+                       delete_upper->get())) {
       return false;  // Ranges don't overlap - cannot match
     }
   }
--- a/src/iceberg/test/delete_file_index_test.cc
+++ b/src/iceberg/test/delete_file_index_test.cc
@@ -840,6 +840,71 @@ TEST_P(DeleteFileIndexTest, TestEqualityDeletesGroup) {
   }
 }
 
+// Regression test: an equality delete file that carries bounds for some equality
+// fields but not others must not crash. CanContainEqDeletesForFile guarded the
+// missing-bound case with std::expected::has_value() (which only reports the error
+// state), so a field whose bound was absent produced an engaged expected wrapping a
+// disengaged optional, and dereferencing it threw std::bad_optional_access. Such
+// files are produced legitimately by per-column metrics modes (counts/none/truncate)
+// and by cross-engine writers.
+TEST_P(DeleteFileIndexTest, TestEqualityDeletePartialFieldBounds) {
+  auto partition_a = PartitionValues({Literal::Int(0)});
+
+  auto serialize = [](const Literal& literal) {
+    auto result = literal.Serialize();
+    EXPECT_THAT(result, IsOk());
+    return result.value();
+  };
+
+  // Equality delete over fields {1 (int id), 2 (string data)} but with bounds only
+  // for field 1. Field 1's bounds must deserialize cleanly, otherwise
+  // ConvertBoundsIfNeeded fails first and masks the missing-bound path.
+  //
+  // null_value_counts is set to 0 for both fields so the null short-circuits in
+  // CanContainEqDeletesForFile (ContainsNull/AllNull) never fire, which keeps the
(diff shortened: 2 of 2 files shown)


LuciferYang:
@wgtmac when you have a moment, could you take a look? This fixes a `bad_optional_access` crash in `CanContainEqDeletesForFile` (the delete file index you authored in #435) — the guard checked the outer `expected` instead of the inner `optional`, so an equality delete file with bounds for only some fields threw and terminated planning/validation. Aligned the behavior with the Java `DeleteFileIndex` (assume may-match on any missing bound). Thanks!

wgtmac (approved):
Good catch! Thanks for fixing it.

LuciferYang:
Thank you @wgtmac

PR #862, opened 2026-07-31
fix(arrow): release schema after conversion failure

WZhuo:
Release partially initialized ArrowSchema output when internal Iceberg-to-Arrow conversion fails. Add coverage using fixed(0), which passes compatibility validation but is rejected by nanoarrow, and verify the output release callback is cleared. Tests: arrow_test.


wgtmac (approved):

wgtmac:
Thanks for fixing this! @WZhuo

PR #867, opened 2026-08-02
feat(rest): support OAuth2 token exchange sessions

lishuxu:
Add RFC 8693 token exchange support, including OAuth2 endpoint handling and response parsing. Create contextual and table child sessions from direct tokens, credentials, or typed tokens.

Keep OAuth2 metadata in one synchronized session snapshot. Use shared `HttpClient` ownership so refresh tasks can safely keep the client alive. Align token expiry with Java by preferring JWT `exp` and otherwise using the request start time plus `expires_in`.

Move OAuth2 helpers into `OAuth2Util` and update unit and integration tests. Leave catalog token refresh, token exchange during refresh, and child session caching as follow-up work.

### DIFF
--- a/src/iceberg/catalog/rest/auth/auth_manager.cc
+++ b/src/iceberg/catalog/rest/auth/auth_manager.cc
@@ -19,11 +19,16 @@
 
 #include "iceberg/catalog/rest/auth/auth_manager.h"
 
+#include <array>
+#include <chrono>
 #include <optional>
+#include <string_view>
+#include <utility>
 
 #include "iceberg/catalog/rest/auth/auth_manager_internal.h"
 #include "iceberg/catalog/rest/auth/auth_properties.h"
 #include "iceberg/catalog/rest/auth/auth_session.h"
+#include "iceberg/catalog/rest/auth/auth_session_internal.h"
 #include "iceberg/catalog/rest/auth/oauth2_util.h"
 #include "iceberg/catalog/session_context.h"
 #include "iceberg/util/base64.h"
@@ -31,11 +36,50 @@
 
 namespace iceberg::rest::auth {
 
+namespace {
+
+constexpr std::string_view kAuthorizationHeader = "Authorization";
+
+const std::array<std::string_view, 5> kTokenPreferenceOrder = {
+    AuthProperties::kIdTokenType,    AuthProperties::kAccessTokenType,
+    AuthProperties::kJwtTokenType,   AuthProperties::kSaml2TokenType,
+    AuthProperties::kSaml1TokenType,
+};
+
+std::optional<std::pair<std::string, std::string>> FindPreferredTypedToken(
+    const std::unordered_map<std::string, std::string>& credentials) {
+  for (std::string_view token_type : kTokenPreferenceOrder) {
+    auto token_it = credentials.find(std::string(token_type));
+    if (token_it != credentials.end()) {
+      return std::pair{token_it->first, token_it->second};
+    }
+  }
+  return std::nullopt;
+}
+
+std::unordered_map<std::string, std::string> FilterTableSessionProperties(
+    const std::unordered_map<std::string, std::string>& properties) {
+  std::unordered_map<std::string, std::string> filtered;
+  if (auto token_it = properties.find(AuthProperties::kToken.key());
+      token_it != properties.end()) {
+    filtered.emplace(token_it->first, token_it->second);
+  }
+  for (std::string_view token_type : kTokenPreferenceOrder) {
+    auto token_it = properties.find(std::string(token_type));
+    if (token_it != properties.end()) {
+      filtered.emplace(token_it->first, token_it->second);
+    }
+  }
+  return filtered;
+}
+
+}  // namespace
+
 Result<std::shared_ptr<AuthSession>> AuthManager::InitSession(
-    HttpClient& init_client,
+    std::shared_ptr<HttpClient> init_client,
     const std::unordered_map<std::string, std::string>& properties) {
   // By default, use the catalog session for initialization
-  return CatalogSession(init_client, properties);
+  return CatalogSession(std::move(init_client), properties);
 }
 
 Result<std::shared_ptr<AuthSession>> AuthManager::ContextualSession(
@@ -56,7 +100,7 @@ Result<std::shared_ptr<AuthSession>> AuthManager::TableSession(
 class NoopAuthManager : public AuthManager {
  public:
   Result<std::shared_ptr<AuthSession>> CatalogSession(
-      [[maybe_unused]] HttpClient& client,
+      [[maybe_unused]] std::shared_ptr<HttpClient> client,
       [[maybe_unused]] const std::unordered_map<std::string, std::string>& properties)
       override {
     return AuthSession::MakeDefault({});
@@ -73,7 +117,7 @@ Result<std::unique_ptr<AuthManager>> MakeNoopAuthManager(
 class BasicAuthManager : public AuthManager {
  public:
   Result<std::shared_ptr<AuthSession>> CatalogSession(
-      [[maybe_unused]] HttpClient& client,
+      [[maybe_unused]] std::shared_ptr<HttpClient> client,
       const std::unordered_map<std::string, std::string>& properties) override {
     auto username_it = properties.find(AuthProperties::kBasicUsername);
     ICEBERG_PRECHECK(username_it != properties.end() && !username_it->second.empty(),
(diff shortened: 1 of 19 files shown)


src/iceberg/catalog/rest/auth/auth_manager.cc
```diff
@@ -148,15 +158,109 @@ class OAuth2Manager : public AuthManager {
...
+    ICEBERG_PRECHECK(shared_client_ != nullptr,
+                     "OAuth2 catalog session must be initialized before child sessions");
+    auto parent_info = parent->OAuth2Info();
+    ICEBERG_PRECHECK(parent_info.has_value(),
+                     "OAuth2 child session requires OAuth2 parent metadata");
+
+    if (token_it != credentials.end()) {
+      ICEBERG_ASSIGN_OR_RAISE(auto config,
+                              ChildConfig(*parent_info, parent_info->credential));
+      return MakeSession(AccessTokenResponse(token_it->second), config,
```
wgtmac, line 289: Child sessions are always created with keep_refreshed=false, so a token-exchange response with expires_in or JWT exp is never refreshed. A long-lived table/context session will keep sending an expired token. Can we either schedule child refresh or explicitly limit expiring child tokens here?

src/iceberg/catalog/rest/auth/auth_manager.cc
```diff
@@ -148,15 +158,109 @@ class OAuth2Manager : public AuthManager {
                                      config.optional_oauth_params(), client);
     }
 
-    return AuthSession::MakeDefault({});
+    return MakeSession(AccessTokenResponse(""), config, /*keep_refreshed=*/false);
+  }
+
+  Result<std::shared_ptr<AuthSession>> ContextualSession(
+      const SessionContext& context, std::shared_ptr<AuthSession> parent) override {
+    return MaybeCreateChildSession(context.credentials, /*allow_credential=*/true,
```
wgtmac, line 222: This path creates a new child session for every contextual operation. Without a cache, repeated operations with the same context re-exchange or fetch a token each time, which can hit OAuth rate limits. Can we cache child sessions by context/credential before enabling this path?

src/iceberg/catalog/rest/auth/auth_manager.cc
```diff
@@ -121,6 +122,7 @@ class OAuth2Manager : public AuthManager {
       HttpClient& client,
       const std::unordered_map<std::string, std::string>& properties) override {
     ICEBERG_ASSIGN_OR_RAISE(auto config, AuthProperties::FromProperties(properties));
+    shared_client_ = &client;
```
wgtmac: shared_client_ is a borrowed pointer, but it is dereferenced after CatalogSession returns. A standalone manager can outlive the HttpClient passed by reference, and a later CatalogSession call can rebind the pointer used by existing child sessions. Can we retain shared ownership of this client, or document and enforce that it outlives the manager and all child sessions?

src/iceberg/catalog/rest/auth/auth_manager.cc
lishuxu: This patch intentionally supports initial token exchange only.
Child-session caching and refresh are deferred to a follow-up change, because enabling refresh without cache/session ownership would create unmanaged refresh tasks.

src/iceberg/catalog/rest/auth/auth_manager.cc
lishuxu: Agreed. This is part of the session lifecycle limitation.
This patch only supports initial child-session creation. Session caching and child-session refresh will be implemented in a follow-up PR.

src/iceberg/catalog/rest/auth/auth_manager.cc
wgtmac: shared_client_ is a borrowed `HttpClient*`, but child sessions use it after this call. Please pass a `std::shared_ptr<HttpClient>` through the manager/session API so the client lifetime is explicit.

src/iceberg/catalog/rest/auth/auth_session.cc
```diff
@@ -141,11 +156,14 @@ class OAuth2AuthSession : public AuthSession,
...
-    headers_ = {{std::string(kAuthorizationHeader), std::string(kBearerPrefix) + token_}};
+    issued_token_type_ = token_response.issued_token_type.empty()
+                             ? AuthProperties::kAccessTokenType
+                             : token_response.issued_token_type;
+    headers_ = AuthHeaders(token_);
 
-    // Determine expiration time
+    expires_at_ = std::chrono::steady_clock::time_point{};
     if (token_response.expires_in_secs.has_value()) {
       expires_at_ = std::chrono::steady_clock::now() +
```
wgtmac: `expires_in` is measured from the token request, but the session starts the clock later. A slow config request can consume part of the token lifetime; please carry the init fetch start time into the session.

src/iceberg/catalog/rest/auth/auth_properties.cc
```diff
@@ -35,6 +36,28 @@ std::pair<std::string, std::string> ParseCredential(const std::string& credentia
...
+  auto uri_it = properties.find(RestCatalogProperties::kUri.key());
+  if (uri_it == properties.end() || uri_it->second.empty()) {
+    return endpoint;
+  }
+
+  auto base_uri = std::string(TrimTrailingSlash(uri_it->second));
+  if (endpoint.starts_with('/')) {
+    return base_uri + endpoint;
+  }
+  return base_uri + "/" + std::string(TrimTrailingSlash(endpoint));
```
wgtmac: Java preserves the relative endpoint suffix. Trimming `oauth/token/` here changes the resolved URI; please keep the trailing slash and update the test.

src/iceberg/catalog/rest/auth/auth_properties.cc
```diff
@@ -35,6 +36,28 @@ std::pair<std::string, std::string> ParseCredential(const std::string& credentia
   return {credential.substr(0, colon_pos), credential.substr(colon_pos + 1)};
 }
 
+Result<std::string> ResolveOAuth2ServerUri(
+    const std::unordered_map<std::string, std::string>& properties) {
+  auto endpoint_it = properties.find(AuthProperties::kOAuth2ServerUri.key());
+  std::string endpoint = endpoint_it == properties.end() || endpoint_it->second.empty()
```
wgtmac: Java defaults only when `oauth2-server-uri` is absent. This also defaults an explicitly empty value; please preserve that distinction or reject empty explicitly.

src/iceberg/test/auth_manager_test.cc
```diff
@@ -266,6 +576,11 @@ TEST_F(AuthManagerTest, OAuth2StaticToken) {
   std::unordered_map<std::string, std::string> properties = {
       {AuthProperties::kAuthType, "oauth2"},
       {AuthProperties::kToken.key(), "my-static-token"},
+      {AuthProperties::kCredential.key(), "client-id:client-secret"},
```
wgtmac, line 404: Please add a test with a short `expires_in` and delayed config fetch to verify elapsed init time is deducted from the session lifetime.

src/iceberg/test/rest_catalog_integration_test.cc
```diff
@@ -211,6 +215,105 @@ TEST_F(RestCatalogIntegrationTest, MakeCatalogSuccess) {
   EXPECT_THAT(root->WithContext(SessionContext{}), IsError(ErrorKind::kInvalidArgument));
 }
 
+TEST_F(RestCatalogIntegrationTest, OAuthContextCredentialEndToEnd) {
```
wgtmac, line 218: These tests call `AuthManager` directly, so they do not cover `RestCatalog` context/table wiring. Please add one real catalog path or move them to the unit suite.

wgtmac:
As discussed offline, I'll directly address these minor issues to move forward.

wgtmac:
I've made some changes to this PR. The intention is to be more Java-like so future feature catch-up and review will be much easier.

- `HttpClient` now uses `std::shared_ptr` through the auth flow. The session can refresh in the background, so it must keep the client alive.
- `OAuth2Session` is now a separate internal class.
- Token expiry now follows Java. JWT `exp` takes precedence. Otherwise, expiry is based on the token request start time plus `expires_in`.
- `OAuth2Util` is now a class so we are able to have internal functions.
- I updated the related unit and integration tests by removing meaningless cases.

wgtmac (approved):
Let me merge this to proceed. Thanks @lishuxu!

PR #868, opened 2026-08-02
fix: make table updates retryable and cleanup-safe

zhjwpku:
Table updates currently have inconsistent retry and cleanup behavior between
standalone updates and explicit transactions.

This change:

- Registers standalone updates in their temporary transaction before Apply, so
  commit retries can replay the update instead of losing it.
- Replays explicit transaction updates when refreshed metadata changes, while
  keeping standalone retries replayable.
- Makes Apply failures terminal for explicit transactions and adds explicit
  Abort cleanup.
- Separates known failure cleanup from `CommitStateUnknown`, preserving staged
  files when the catalog outcome is uncertain.
- Tracks staged snapshot files before writing them and retries failed cleanup
  best effort.
- Cleans no-op snapshot updates without reporting a commit.
- Defers `ExpireSnapshots` physical cleanup until the final transaction commit
  and uses reachable cleanup for explicit transactions.
- Keeps the existing `Apply()` APIs and removes unnecessary freeze/deep-copy
  lifecycle machinery.

The implementation also adds coverage for standalone retry, transaction retry,
no-op updates, Apply failures, Abort, unknown commit outcomes, staged-file
cleanup, ExpireSnapshots retry, and finalization/reporting behavior.

### DIFF
--- a/src/iceberg/test/expire_snapshots_test.cc
+++ b/src/iceberg/test/expire_snapshots_test.cc
@@ -40,6 +40,9 @@
 #include "iceberg/test/executor.h"
 #include "iceberg/test/matchers.h"
 #include "iceberg/test/update_test_base.h"
+#include "iceberg/transaction.h"
+#include "iceberg/update/fast_append.h"
+#include "iceberg/update/set_snapshot.h"
 
 namespace iceberg {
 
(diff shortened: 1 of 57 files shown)


wgtmac (approved):
I simplified the patch and kept the changes needed for correct retry and cleanup:
- Standalone updates are registered in their temporary transaction before Apply. This prevents a retry from replaying an empty update list and returning success without the update.
- Explicit transactions still allow only one pending operation. An Apply failure marks the transaction failed, cleans staged files best effort, and rejects further use.
- Explicit transactions replay updates only when refreshed metadata has changed. Standalone updates still re-apply on each retry. A replay failure stops the outer retry loop.
- Known commit failures and Abort clean staged files. `CommitStateUnknown` preserves all files because the catalog may already have committed them.
- Snapshot updates register file paths before writing. Failed deletions remain tracked, so Abort can retry them later.
- Snapshot updates that become no-ops clean their temporary files and do not emit a commit report.
- `ExpireSnapshots` performs physical deletion only after the final transaction succeeds. Explicit transactions use reachable cleanup against the final metadata, so files referenced by later updates are preserved.
- I removed the freeze/deep-copy machinery, `EnsureMutable()`, and redundant lifecycle state. Callers are instead required not to modify an update or its inputs after the first Commit.
- I kept the existing `Apply()` names. Cleanup and finalization are internal transaction hooks.
- I also removed tests for unsupported callback re-entry, restored identity assertions, and fixed validation tests that created duplicate snapshot metadata.

PR #869, opened 2026-08-02
test: complete v3 row lineage test coverage

wgtmac:
- enable RewriteFiles V3 coverage and verify DV cleanup
- cover V2-to-V3 assignment and incremental/changelog lineage
- verify delete-aware readers preserve lineage

### DIFF
--- a/src/iceberg/test/rewrite_files_test.cc
+++ b/src/iceberg/test/rewrite_files_test.cc
@@ -84,19 +84,23 @@ class RewriteFilesTest : public MinimalUpdateTestBase {
     return f;
   }
 
-  std::shared_ptr<DataFile> MakePositionDeleteFile(
-      const std::string& path, int64_t partition_x,
-      const std::string& referenced_data_file) {
-    auto file = MakeDataFile(path, partition_x);
-    file->content = DataFile::Content::kPositionDeletes;
-    if (table_->metadata()->format_version >= 3) {
-      file->file_format = FileFormatType::kPuffin;
-      file->referenced_data_file = referenced_data_file;
-      file->content_offset = 0;
-      file->content_size_in_bytes = 10;
-    }
-    return file;
+std::shared_ptr<DataFile> MakePositionDeleteFile(
+    const std::string& path, int64_t partition_x,
+    const std::string& referenced_data_file) {
+  std::string effective_path = path;
+  if (table_->metadata()->format_version >= 3 && effective_path.ends_with(".parquet")) {
+    effective_path = effective_path.substr(0, effective_path.size() - 7) + ".puffin";
   }
+  auto file = MakeDataFile(effective_path, partition_x);
+  file->content = DataFile::Content::kPositionDeletes;
+  if (table_->metadata()->format_version >= 3) {
+    file->file_format = FileFormatType::kPuffin;
+    file->referenced_data_file = referenced_data_file;
+    file->content_offset = 0;
+    file->content_size_in_bytes = 10;
+  }
+  return file;
+}
 
   std::shared_ptr<DataFile> MakeEqualityDeleteFile(const std::string& path,
                                                    int64_t partition_x) {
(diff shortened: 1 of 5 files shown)


blackmwk (approved):

zhjwpku (approved):

src/iceberg/test/rewrite_files_test.cc
```diff
@@ -83,18 +84,30 @@ class RewriteFilesTest : public MinimalUpdateTestBase {
...
-  std::shared_ptr<DataFile> MakeDeleteFile(const std::string& path, int64_t partition_x) {
-    auto f = MakeDataFile(path, partition_x);
-    f->content = DataFile::Content::kPositionDeletes;
-    return f;
+  std::shared_ptr<DataFile> MakePositionDeleteFile(
+      const std::string& path, int64_t partition_x,
+      const std::string& referenced_data_file) {
+    std::string effective_path = path;
+    if (table_->metadata()->format_version >= 3 && effective_path.ends_with(".parquet")) {
+      effective_path = effective_path.substr(0, effective_path.size() - 7) + ".puffin";
```
manuzhang: `.parquet` is 8 chars, not 7

src/iceberg/test/rewrite_files_test.cc
```diff
@@ -83,18 +84,30 @@ class RewriteFilesTest : public MinimalUpdateTestBase {
...
+    }
+    auto file = MakeDataFile(effective_path, partition_x);
+    file->content = DataFile::Content::kPositionDeletes;
+    if (table_->metadata()->format_version >= 3) {
```
manuzhang, line 93: This is the same check as above and can be collapsed into one.

src/iceberg/test/rewrite_files_test.cc
```diff
@@ -544,9 +564,10 @@ TEST_P(RewriteFilesFormatVersionTest, DataSequenceNumber) {
     GTEST_SKIP() << "Requires format version >= 2";
   }
   CommitFileA();
+  const auto data_sequence_number = table_->metadata()->last_sequence_number;
 
   ICEBERG_UNWRAP_OR_FAIL(auto rw, NewRewriteFiles());
-  rw->SetDataSequenceNumber(5);
+  rw->SetDataSequenceNumber(data_sequence_number);
```
manuzhang, line 572: why is this changed?

wgtmac: This aligns the test with Java `TestRewriteFiles`: the replacement entry should retain the pre-rewrite data sequence number, while the new manifest receives the rewrite commit sequence. The previous hard-coded value 5 was unrelated to the table state, so the test now captures `last_sequence_number` after `CommitFileA()` and verifies both values.

manuzhang (commented):
The ~35-line `snapshot_b` + `metadata` construction is byte-identical between                                                                           `incremental_append_scan_test.cc:PlanRowLineage` and `incremental_changelog_scan_test.cc:PlanAddedRowLineage`. Both include `scan_test_base.h`; adding optional `first_row_id`/`added_rows` params to `MakeAppendSnapshot` (or a `MakeAppendSnapshotWithRowLineage`) would remove both copies.
 
Similarly, the 4-line v3 upgrade block is inlined twice in `rewrite_files_test.cc` while `merging_snapshot_update_test.cc:350` already has `UpgradeTableToV3()` — worth hoisting to `update_test_base.h` now that two files need it.
 
The three new reader lineage tests are also ~95% identical to each other and could take a delete-file factory.

*This is discovered by LLM and could be a follow-up*.

wgtmac:
I was thinking of refactoring this repo with a larger change to eliminate similar/duplicate lines across all test files. I agree it can be a followup.

manuzhang (approved):

wgtmac:
Thanks all for the review!

PR #872, opened 2026-08-06
fix(logging): run FatalHandler on all kFatal paths

kamcheungting-db:
Follow-up to #725 (merged): the registered `FatalHandler` runs only for the fixed-severity `ICEBERG_LOG_FATAL`. Reaching `kFatal` through the runtime-level forms aborts **without** invoking it:

- `ICEBERG_LOG(LogLevel::kFatal, ...)` → `LogToCurrentRuntime` flushes and aborts, no handler.
- `ICEBERG_LOG_TO(sink, LogLevel::kFatal, ...)` → `LogToExplicitRuntime`, same.

Those paths also format the message only when `ShouldLog(kFatal)` passes, so a handler (once wired) would miss the text for a filtered record. This was raised in review on #726; the fix didn't make it into the #725 merge, so it is submitted separately here.

**Change:** extract the fatal sequence into a shared `internal::DispatchFatal` — format the message once, emit it if the logger is enabled, flush that same logger, run any registered `FatalHandler`, then `std::abort()` — and route all three fatal paths (`LogFatal`, `LogToCurrentRuntime`, `LogToExplicitRuntime`) through it. A handler that itself throws cannot prevent the abort.

**Tests:** two death tests asserting the handler fires for the generic runtime-level `kFatal` and for `ICEBERG_LOG_TO(..., kFatal)`.

Behavior for non-fatal levels and for `ICEBERG_LOG_FATAL` is unchanged.

### DIFF
--- a/src/iceberg/logging/log_macros.h
+++ b/src/iceberg/logging/log_macros.h
@@ -79,53 +79,24 @@ void LogToCurrent(LogLevel level, const std::source_location& location,
   }
 }
 
-/// \brief Runtime-level variant against the current logger: emit if enabled, then
-/// flush + abort when level == kFatal (using the same acquired logger).
-template <typename MakeMessage>
-void LogToCurrentRuntime(LogLevel level, const std::source_location& location,
-                         MakeMessage&& make_message) noexcept {
-  const std::shared_ptr<Logger>& logger = CurrentLogger();
-  if (logger) {
-    EmitIfEnabled(*logger, level, location, std::forward<MakeMessage>(make_message));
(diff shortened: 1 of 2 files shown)


wgtmac:
I've incorporated this patch into https://github.com/apache/iceberg-cpp/commit/31274f5fd1b5f6b34ca18f833e4b25ea76faf1cd. So let me close this PR.

I just created https://github.com/apache/iceberg-cpp/issues/887 while reviewing this. Let me know if you have any idea.

PR #873, opened 2026-08-06
feat: add lazy scan planning streams

manuzhang:
Scan planning currently materializes manifest entries and file scan tasks before returning. This change adds fallible, single-pass streams so callers can consume entries and tasks incrementally or stop early.

- Add `DataTableScan::PlanFilesStream()` and `ManifestGroup::PlanFilesStream()`. The eager `PlanFiles()` methods collect the stream into a vector.
- Add `ManifestReader::EntriesStream()` and `LiveEntriesStream()` as the reader implementation extension points. `Entries()` and `LiveEntries()` collect these streams for eager callers.
- Rename the pull utility from `Iterator<T>` to `Stream<T>`, with `ManifestEntryStream` and `FileScanTaskStream` aliases and corresponding owning `*Ptr` aliases. `Next()` returns `Result<std::optional<T>>`; errors and exhaustion are terminal, and `ToVector()` collects the remaining values.
- Distinguish explicit empty selections from default projections: `Select({})` selects no user columns, while `Select({"*"})` or omitting `Select()` selects all columns. Table projections still include fields required by filters, and an explicit empty selection conflicts with an explicit projection schema.
- Preserve partition values needed for delete matching and residual evaluation under narrow projections. Read statistics needed for equality-delete matching, then drop statistics the caller did not request.
- Report scan metrics on exhaustion or early stream destruction, without emitting a successful report after a planning error.

Streams own their reader or planning resources and may outlive the object that created them. A configured executor is borrowed and must remain alive until the planning stream is destroyed. Serial planning opens one data manifest at a time; executor-backed planning opens at most 32 matching manifest streams per batch and consumes their entries incrementally. `DataTableScan` uses the configured executor when there is more than one data manifest or more than one delete manifest.

Snapshot and manifest-list metadata remain materialized, and delete manifests are read eagerly to build the delete-file index before returning the planning stream.

API changes: include `iceberg/util/stream.h` and use `Stream<T>` in place of `Iterator<T>`. `ManifestGroup::PlanFiles()` and `PlanFilesStream()` consume the group and require an rvalue, such as `std::move(*group).PlanFilesStream()`. Custom manifest readers implement `EntriesStream()` and `LiveEntriesStream()` and return streams that remain valid after reader destruction. `DataTableScan::PlanFiles()` retains its existing caller syntax and eager result type.

Validation: built `manifest_test` and `scan_test` with CMake and ran `ctest --test-dir build --output-on-failure -R '^(manifest_test|scan_test)$'`; both suites passed. Coverage includes empty and wildcard projections, equality and position deletes, partition isolation, executor-backed planning, stream lifetimes, and scan reporting. `git diff --check` passed.


src/iceberg/manifest/manifest_group.cc
```diff
@@ -252,6 +475,12 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> ManifestGroup::PlanFiles() {
   return file_tasks;
 }
 
+Result<std::unique_ptr<Iterator<std::shared_ptr<FileScanTask>>>>
+ManifestGroup::PlanFilesIterator() {
+  auto group = std::make_unique<ManifestGroup>(std::move(*this));
```
wgtmac, line 562: Can we make this `PlanFilesIterator() &&`? It moves from `*this`, but callers can invoke it on an lvalue and accidentally reuse a consumed `ManifestGroup`.

src/iceberg/util/iterator.h
```diff
@@ -0,0 +1,67 @@
...
+
+  /// \brief Consume the remaining values into a vector.
+  Result<std::vector<T>> ToVector() {
+    std::vector<T> values;
+    while (true) {
+      ICEBERG_ASSIGN_OR_RAISE(auto value, Next());
+      if (!value.has_value()) {
+        return values;
+      }
+      values.push_back(std::move(value).value());
```
wgtmac: This fails to compile for a copy-only `T`, because `push_back` selects the deleted move constructor. Please either constrain `Iterator<T>` to movable types or use a copy fallback, and test the chosen contract.

src/iceberg/table_scan.cc
```diff
@@ -64,6 +65,93 @@ const std::vector<std::string> kScanColumnsWithStats = [] {
...
+        report_(std::move(report)) {}
+
+  ~ReportingFileTaskIterator() override { Finalize(); }
+
+  Result<std::optional<std::shared_ptr<FileScanTask>>> Next() override {
+    auto start = std::chrono::steady_clock::now();
+    auto result = iterator_->Next();
+    planning_duration_ += std::chrono::duration_cast<std::chrono::nanoseconds>(
+        std::chrono::steady_clock::now() - start);
+    if (result.has_value() && !result.value().has_value()) {
```
wgtmac: Please handle `Next()` errors here. Right now reporting waits until destruction and records the failed plan as normal partial consumption. Finalize immediately or skip the report, and cover both error and early-destruction paths.

src/iceberg/manifest/manifest_reader.h
```diff
@@ -134,6 +145,19 @@ class ICEBERG_EXPORT ManifestReader {
...
 };
 
+/// \brief Optional mix-in for ManifestReader implementations that support lazy entry
+/// iteration.
+class ICEBERG_EXPORT SupportsManifestEntryIteration {
+ public:
+  virtual ~SupportsManifestEntryIteration() = default;
+
+  /// \brief Lazily read manifest entries.
+  virtual Result<std::unique_ptr<Iterator<ManifestEntry>>> EntriesIterator() = 0;
```
wgtmac: Please document that these methods must return self-contained iterators. The planning path destroys the `ManifestReader` immediately after obtaining the iterator, so a third-party implementation that keeps a pointer to `this` will dangle.

src/iceberg/manifest/manifest_group.h
```diff
@@ -136,6 +137,15 @@ class ICEBERG_EXPORT ManifestGroup : public ErrorCollector {
   /// \brief Plan scan tasks for all matching data files.
   Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles();
 
+  /// \brief Lazily plan scan tasks for matching data files.
+  ///
+  /// The returned iterator owns the planning state and may outlive this ManifestGroup.
+  /// It reads one manifest batch at a time instead of materializing all manifest entries
+  /// and scan tasks. Creating the iterator consumes this group's configuration. Streaming
+  /// planning is pull-based and does not eagerly submit manifests to the executor set by
```
wgtmac: `PlanFilesIterator()` still receives `plan_executor`, but this path ignores it and reads data manifests serially. Was that slowdown measured? Please preserve executor-backed planning or add benchmark data and document the trade-off.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -132,6 +132,308 @@ ManifestGroup::~ManifestGroup() = default;
...
+  Result<bool> LoadNextManifestBatch() {
+    std::vector<const ManifestFile*> manifests;
+    manifests.reserve(kManifestReadBatchSize);
+    while (next_manifest_ < group_->data_manifests_.size() &&
+           manifests.size() < kManifestReadBatchSize) {
+      const auto& manifest = group_->data_manifests_[next_manifest_++];
+      ICEBERG_ASSIGN_OR_RAISE(bool should_read, ShouldReadManifest(manifest));
+      if (should_read) {
+        manifests.push_back(&manifest);
+      }
```
wgtmac: With an executor, this can stop before the end. If the next 32 manifests are all filtered out, `manifests.empty()` returns false even when more manifests remain, so a later matching manifest is never read. Please keep loading batches until one is non-empty or all manifests are consumed, and add a test for 32 skipped plus 1 matching manifest.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -374,6 +681,18 @@ Result<std::unique_ptr<ManifestReader>> ManifestGroup::MakeReader(
   return reader;
 }
 
+bool ManifestGroup::PrepareStatsProjection(bool has_equality_deletes) {
+  // The caller's projection records whether stats were requested. Equality-delete
+  // matching may add stats temporarily, but they should still be dropped from the
+  // result when the original projection did not request them. Keeping this decision
+  // here ensures eager and iterator planning use identical semantics.
+  const bool drop_stats = ManifestReader::ShouldDropStats(columns_);
```
wgtmac: Please keep the original projection separate from the temporary stats projection. `PrepareStatsProjection()` mutates `columns_`; calling `PlanFiles()` or `Plan()` again after an equality-delete plan can retain or duplicate stats columns and change the result.

src/iceberg/manifest/manifest_group.h
```diff
@@ -136,6 +137,16 @@ class ICEBERG_EXPORT ManifestGroup : public ErrorCollector {
...
 
+  /// \brief Lazily plan scan tasks for matching data files.
+  ///
+  /// The returned iterator owns the planning state and may outlive this ManifestGroup.
+  /// It reads one bounded manifest batch at a time instead of materializing all manifest
+  /// entries and scan tasks. When PlanWith() configures an executor, entry iterators for
+  /// manifests in each batch are opened in parallel, while entries are consumed one
+  /// manifest at a time. Creating the iterator consumes this group's configuration, so
+  /// this method may only be called on an rvalue.
+  Result<FileScanTaskIterator> PlanFilesIterator() &&;
```
wgtmac: The iterator keeps `OptionalExecutor` as a non-owning reference, but it uses it on later `Next()` calls. If the executor is destroyed after this method returns, the iterator dereferences a dangling reference. Please document that lifetime requirement or make the iterator own/share the executor.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -132,6 +132,308 @@ ManifestGroup::~ManifestGroup() = default;
...
+  std::unordered_map<int32_t, std::shared_ptr<ResidualEvaluator>> residual_evaluators_;
+  std::unique_ptr<Iterator<ManifestEntry>> entry_iterator_;
+  std::vector<TaggedIterator> batch_iterators_;
+  size_t next_manifest_ = 0;
+  size_t next_batch_iterator_ = 0;
+  int32_t current_spec_id_ = 0;
+  bool drop_stats_;
+
+  // Limit the number of manifest readers and iterators retained by executor-backed
+  // planning. The executor still controls actual task concurrency, while this fixed
```
wgtmac: Please make the batch size configurable or explain why it is fixed at 32. It controls both the number of concurrent readers and the iterator's memory footprint, so it is a tuning knob for large scans.

manuzhang:
@wgtmac I split the iterator util into #905 and will rebase this once that PR is merged.

src/iceberg/type_fwd.h
```diff
@@ -24,6 +24,8 @@
 /// you can include this instead of the "full" headers to help reduce compile
 /// times.
 
+#include <memory>
+
```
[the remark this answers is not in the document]
manuzhang: Fixed in d82d9e2. I moved FileScanTaskIterator out of type_fwd.h into a dedicated public header that includes the complete Iterator definition. Both public API headers now include that header, and Iterator itself remains forward-declared in type_fwd.h. I also added a regression test that destroys FileScanTaskIterator while FileScanTask is still incomplete.

src/iceberg/table_scan.cc
```diff
@@ -64,6 +65,95 @@ const std::vector<std::string> kScanColumnsWithStats = [] {
...
+                                           projected_id_set.end());
+  std::ranges::sort(projected_field_ids);
+
+  std::vector<std::string> projected_field_names;
+  projected_field_names.reserve(projected_field_ids.size());
+  for (int32_t field_id : projected_field_ids) {
+    ICEBERG_ASSIGN_OR_RAISE(auto field_name, schema_ptr->FindColumnNameById(field_id));
+    ICEBERG_CHECK(field_name.has_value(), "Projected field {} not found in schema",
+                  field_id);
+    projected_field_names.emplace_back(*field_name);
```
wgtmac: These vectors are already materialized, but {begin(), end()} copies every ManifestFile before the lazy iterator is returned. Can we move them into ManifestGroup::Make() to avoid doubling peak metadata memory?

src/iceberg/manifest/manifest_group.h
```diff
@@ -136,6 +137,16 @@ class ICEBERG_EXPORT ManifestGroup : public ErrorCollector {
   /// \brief Plan scan tasks for all matching data files.
   Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles();
```
wgtmac: This iterator is lazy for data manifests, but Build() still reads all delete manifests before the first Next(). Could we state that here so “stop planning early” is not read as avoiding delete-manifest I/O?

src/iceberg/manifest/manifest_group.cc
```diff
@@ -131,6 +131,307 @@ ManifestGroup::~ManifestGroup() = default;
...
 ManifestGroup& ManifestGroup::operator=(ManifestGroup&&) noexcept = default;
 
+class ManifestGroup::FilePlanningIterator final
+    : public Iterator<std::shared_ptr<FileScanTask>> {
+ public:
+  static Result<FileScanTaskIterator> Make(std::unique_ptr<ManifestGroup> group) {
+    ICEBERG_RETURN_UNEXPECTED(group->CheckErrors());
+
+    group->delete_index_builder_.WithScanMetrics(group->scan_metrics_);
+    ICEBERG_ASSIGN_OR_RAISE(auto delete_index, group->delete_index_builder_.Build());
```
wgtmac: PrepareStatsProjection() mutates columns_ in place. Reusing the same ManifestGroup for a second plan can change whether stats are retained. Should this temporary projection stay local to each plan?

src/iceberg/manifest/manifest_group.h
```diff
@@ -30,6 +30,7 @@
 #include <vector>
 
 #include "iceberg/delete_file_index.h"
+#include "iceberg/file_scan_task_iterator.h"
 #include "iceberg/iceberg_export.h"
 #include "iceberg/manifest/manifest_entry.h"
```
wgtmac: The returned iterator may outlive this ManifestGroup, but PlanWith() stores only a reference to the executor. Could we document that the executor must outlive the iterator?

src/iceberg/table_scan.cc
```diff
@@ -661,6 +720,69 @@ Result<std::vector<std::shared_ptr<FileScanTask>>> DataTableScan::PlanFiles() co
...
+    scan_metrics->total_data_manifests->Increment(
+        static_cast<int64_t>(data_manifests.size()));
+    scan_metrics->total_delete_manifests->Increment(
+        static_cast<int64_t>(delete_manifests.size()));
+  }
+
+  ICEBERG_ASSIGN_OR_RAISE(
+      auto manifest_group,
+      ManifestGroup::Make(io_, schema_, specs_by_id,
+                          {data_manifests.begin(), data_manifests.end()},
```
wgtmac: These vectors are already materialized, but {begin(), end()} copies every ManifestFile before the lazy iterator is returned. Can we move them into ManifestGroup::Make() to avoid doubling peak metadata memory?

src/iceberg/manifest/manifest_group.h
```diff
@@ -136,6 +137,16 @@ class ICEBERG_EXPORT ManifestGroup : public ErrorCollector {
   /// \brief Plan scan tasks for all matching data files.
   Result<std::vector<std::shared_ptr<FileScanTask>>> PlanFiles();
 
+  /// \brief Lazily plan scan tasks for matching data files.
+  ///
+  /// The returned iterator owns the planning state and may outlive this ManifestGroup.
+  /// It reads one bounded manifest batch at a time instead of materializing all manifest
+  /// entries and scan tasks. When PlanWith() configures an executor, entry iterators for
+  /// manifests in each batch are opened in parallel, while entries are consumed one
+  /// manifest at a time. Creating the iterator consumes this group's configuration, so
```
wgtmac: This iterator is lazy for data manifests, but Build() still reads all delete manifests before the first Next(). Could we state that here so “stop planning early” is not read as avoiding delete-manifest I/O?

src/iceberg/manifest/manifest_group.cc
```diff
@@ -373,6 +679,18 @@ Result<std::unique_ptr<ManifestReader>> ManifestGroup::MakeReader(
...
 }
 
+bool ManifestGroup::PrepareStatsProjection(bool has_equality_deletes) {
+  // The caller's projection records whether stats were requested. Equality-delete
+  // matching may add stats temporarily, but they should still be dropped from the
+  // result when the original projection did not request them. Keeping this decision
+  // here ensures eager and iterator planning use identical semantics.
+  const bool drop_stats = ManifestReader::ShouldDropStats(columns_);
+  if (has_equality_deletes) {
+    columns_ = ManifestReader::WithStatsColumns(columns_);
```
wgtmac: PrepareStatsProjection() mutates columns_ in place. Reusing the same ManifestGroup for a second plan can change whether stats are retained. Should this temporary projection stay local to each plan?

src/iceberg/manifest/manifest_reader.h
```diff
@@ -42,13 +43,23 @@ class ICEBERG_EXPORT ManifestReader {
...
   virtual Result<std::vector<ManifestEntry>> Entries() = 0;
 
   /// \brief Read only live (non-deleted) manifest entries.
   virtual Result<std::vector<ManifestEntry>> LiveEntries() = 0;
 
+  /// \brief Lazily read manifest entries.
+  ///
+  /// Implementations using SupportsManifestEntryIteration stream entries lazily. Other
+  /// implementations are adapted from Entries() for compatibility.
+  Result<std::unique_ptr<Iterator<ManifestEntry>>> EntriesIterator();
```
wgtmac: Could we rename the new pull APIs before making them public?

These APIs return a fallible, single-pass source:
`Next()` returns `Result<std::optional<T>>`.
This is different from a standard C++ iterator or range.

Suggested names:

- `PlanFilesIterator()` → `PlanFilesStream()`
- `EntriesIterator()` → `EntriesStream()`
- `LiveEntriesIterator()` → `LiveEntriesStream()`
- `FileScanTaskIterator` → `FileScanTaskStream`

Example:

```cpp
auto stream = scan->PlanFilesStream();
auto next = stream->Next();
```

This keeps `PlanFiles()` for eager planning and makes the streaming API explicit.

manuzhang: I renamed Iterator and its subclasses to *Stream to be consistent.

src/iceberg/manifest/manifest_reader_internal.h
```diff
@@ -43,7 +43,7 @@ namespace iceberg {
 /// This implementation supports lazy reader creation and filtering based on
 /// partition expressions, row expressions, and partition sets. Following the
 /// Java implementation pattern.
-class ManifestReaderImpl : public ManifestReader {
+class ManifestReaderImpl : public ManifestReader, public SupportsManifestEntryStreaming {
```
wgtmac: Why do we need this extra `SupportsManifestEntryStreaming` mixin at all? We don't have to consider API or ABI backward compatibility before the 1.0 release.

src/iceberg/manifest/manifest_group.cc
```diff
@@ -124,13 +123,320 @@ ManifestGroup::ManifestGroup(
...
+      : group_(std::move(group)),
+        delete_index_(std::move(delete_index)),
+        data_file_evaluator_(std::move(data_file_evaluator)),
+        columns_(std::move(columns)),
+        drop_stats_(drop_stats) {}
+
+  using TaggedEntry = std::pair<int32_t, ManifestEntry>;
+  using TaggedStream = std::pair<int32_t, ManifestEntryStreamPtr>;
+
+  // TODO: Perhaps refactor this concurrent/sequential stream state machine into a generic
```
wgtmac:
```suggestion
  // FIXME: Perhaps refactor this concurrent/sequential stream state machine into a generic
```

wgtmac (approved):

src/iceberg/manifest/manifest_group.cc
```diff
@@ -124,13 +123,320 @@ ManifestGroup::ManifestGroup(
...
+        delete_index_(std::move(delete_index)),
+        data_file_evaluator_(std::move(data_file_evaluator)),
+        columns_(std::move(columns)),
+        drop_stats_(drop_stats) {}
+
+  using TaggedEntry = std::pair<int32_t, ManifestEntry>;
+  using TaggedStream = std::pair<int32_t, ManifestEntryStreamPtr>;
+
+  // FIXME: Perhaps refactor this concurrent/sequential stream state machine into a generic
+  // reusable ParallelStream<T> utility, similar to Iceberg Java's ParallelIterable.
```
wgtmac:
```suggestion
  // FIXME: Perhaps refactor this concurrent/sequential stream state machine into a
  // generic reusable ParallelStream<T> utility, similar to Iceberg Java's
  // ParallelIterable.
```

wgtmac:
Thanks @manuzhang for improving this!

PR #874, opened 2026-08-06
test: cover write, evolve and read back of a defaulted column end to end

huan233usc:
Writes a data file, adds a column with an `initial-default` via `UpdateSchema`, then reads the pre-evolution file back with the evolved schema and checks the old rows materialize the default.

Parametrized over both formats (Parquet, Avro) and over `long` / `string` / `boolean` / `date` / `decimal(9,2)` / `timestamp` / `uuid` — 14 cases. The read-path tests build the projected schema by hand and the evolution tests never touch a data file, so the seam between the two was uncovered; per-type coverage only existed at the `ToArrowScalar` unit layer, not through a real write and read (`uuid` in particular goes through the `arrow.uuid` extension type).

`data_test` passes (169). Verified the assertions bite by feeding wrong expected values.

### DIFF
--- a/src/iceberg/test/CMakeLists.txt
+++ b/src/iceberg/test/CMakeLists.txt
@@ -249,6 +249,7 @@ if(ICEBERG_BUILD_BUNDLE)
                    arrow_c_data_util_test.cc
                    arrow_row_builder_test.cc
                    data_writer_test.cc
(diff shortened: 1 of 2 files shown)


wgtmac (approved):
Thanks @huan233usc!

PR #875, opened 2026-08-06
feat: cache immutable metadata content in FileIO

manuzhang:
## Summary

- add a shared, bounded `FileIO` content cache for metadata JSON, manifest lists, and manifests
- route only immutable metadata reads through the cache, leaving data-file I/O unchanged
- configure caching with the established Iceberg `io.manifest.cache-*` properties across registry-created and catalog-injected `FileIO` instances
- add cache behavior and concurrency regression coverage

## Why

Repeated table loads and scans currently reread immutable Iceberg metadata files. This adds cross-query byte reuse while preserving normal data-file I/O paths.

The cache properties and defaults match Iceberg Java so existing Iceberg catalog configuration can be reused. REST-created `FileIO` instances are configured through the registry, while SQL and in-memory catalogs accept injected `FileIO` instances and configure them during catalog creation. In every path, the cache is owned by `FileIO` and remains disabled unless `io.manifest.cache-enabled` is explicitly set.

## Follow-up

Parsed manifest-list reuse has been split into #948.

## Validation

- `pre-commit run --show-diff-on-failure --all-files`
- focused C++23 syntax checks for the cache, `FileIO`, and cache tests


src/iceberg/metadata_cache.cc
```diff
@@ -0,0 +1,346 @@
...
+
+#include "iceberg/util/macros.h"
+#include "iceberg/util/string_util.h"
+
+namespace iceberg {
+
+namespace {
+
+Result<bool> ParseBoolean(std::string_view key, const std::string& value) {
+  if (value == "true") {
```
wgtmac: [P1] Parse cache booleans case-insensitively. Java's PropertyUtil.propertyAsBoolean ultimately accepts TRUE/FALSE in any case, so a catalog property such as io.manifest.cache-enabled=TRUE is valid and enables caching in Java. This implementation rejects it with InvalidArgument during FileIO/catalog configuration, creating a cross-engine configuration compatibility regression. Please normalize the value (or otherwise match the Java boolean parsing contract) before validating it.

src/iceberg/file_io.cc
```diff
@@ -69,21 +190,110 @@ Result<std::string> FileIO::ReadFile(const std::string& file_location,
...
+      return InvalidArgument("File length {} exceeds int64_t max", *length);
+    }
+    size = static_cast<int64_t>(*length);
+  } else {
+    ICEBERG_ASSIGN_OR_RAISE(size, input_file->Size());
+  }
+  if (size < 0) {
+    return Invalid("Invalid negative file size {} for {}", size, file_location);
+  }
+  if (std::cmp_greater(size, cache->options().max_content_length)) {
```
wgtmac: [P2][defensive] Also bypass the cached wrapper when the known file size exceeds max_total_bytes. With max_total_bytes smaller than max_content_length, this branch still wraps the file; CachedInputFile::Open then reads the whole file before MetadataCache::Get drops it because it cannot fit. A file can therefore cause a transient allocation above the configured total budget and repeat the full read on every open. Returning the underlying InputFile when size exceeds either bound would preserve the intended memory guard.

wgtmac:
Leave a comment to not make it stale.

PR #878, opened 2026-08-07
fix: reject out-of-range integers when parsing JSON literals

### LINKED ISSUE #877: bug: JSON literal parsing silently accepts integers beyond the signed 64-bit range
**Summary**

`LiteralFromJson` accepts JSON integers beyond the signed 64-bit range and turns them into a wrong literal instead of returning a parse error. nlohmann reports unsigned integers as `is_number_integer()`, and `get<int64_t>()` converts a value above INT64_MAX silently rather than throwing, so:

- the `kInt` branch runs its int32 range check on the already-wrapped value, and
- the `kLong` branch has no range check at all.

`18446744073709551615` parses as `Literal::Int(-1)`; `9223372036854775808` parses as `Literal::Long(INT64_MIN)`. The untyped overload, which is the one the REST expression parser actually reaches, has the same hole.

**Root Cause**

```cpp
case TypeId::kInt: {
  if (!json.is_number_integer()) {   // true for unsigned nodes too
    return JsonParseError(...);
  }
  auto val = json.get<int64_t>();    // wraps silently above INT64_MAX
  if (val < INT32_MIN || val > INT32_MAX) {   // checks the wrapped value
    return JsonParseError(...);
  }
  return Literal::Int(static_cast<int32_t>(val));
}

case TypeId::kLong:
  if (!json.is_number_integer()) { ... }
  return Literal::Long(json.get<int64_t>());   // no range check
```

Java guards the same paths explicitly: `SingleValueParser` uses `canConvertToInt()` / `canConvertToLong()`, and `ExpressionParser.asObject` uses `canConvertToLong()`, all of which throw on out-of-range input.

**Impact**

The live path today is table metadata: `initial-default` / `write-default` go through the type-aware parser (`json_serde.cc` `FieldFromJson`), `ValidateDefault` has no integer range check, and the value is later materialized into a returned column via `MakeDefaultArray`. So a metadata file with an out-of-range integer default reads back as `-1` in C++ while Java rejects the file, which is an observable cross-engine difference.

The expression path is latent rather than live: a residual filter parsed from a REST response reaches `ReaderOptions::filter`, but no reader evaluates that field yet (`file_scan_task_reader.h` still has the TODO), and the scan-planning response parsers aren't wired into `RestCatalog`. Worth fixing now so the hole isn't inherited when evaluation does land.

Note this is a correctness/parser-hardening issue, not a security one: `SECURITY-THREAT-MODEL.md` treats catalog-supplied metadata as trusted input.

**Proposed Fix**

Check the range before the conversion, mirroring `canConvertToLong()`:

```cpp
Result<int64_t> GetInt64Checked(const nlohmann::json& json) {
  if (json.is_number_unsigned() &&
      json.get<uint64_t>() > static_cast<uint64_t>(std::numeric_limits<int64_t>::max())) {
    return JsonParseError("Cannot parse {} as an integer value: out of range", ...);
  }
  return json.get<int64_t>();
}
```

and call it from the `kInt` branch, the `kLong` branch, and the untyped overload. The `is_number_unsigned()` half matters: `get<uint64_t>()` on a negative node yields its two's-complement value, which would compare above INT64_MAX and reject every negative literal.

**Out of scope (follow-up)**

`GetTypedJsonValue` in `src/iceberg/util/json_util_internal.h` truncates out-of-range integers the same way (nlohmann's `get<int32_t>` does not throw either), so `FieldFromJson({"id": 2147483648, ...})` yields `field_id = -2147483648` with no error. That helper has on the order of 80 call sites across the repo and deserves its own PR.

I have a fix and regression tests ready and will open a PR.

LuciferYang:
## What

`LiteralFromJson` accepted JSON integers beyond the signed 64-bit range and produced a wrong literal instead of a parse error. nlohmann reports unsigned integers as `is_number_integer()`, and `get<int64_t>()` converts values above INT64_MAX silently rather than throwing. So the `kInt` branch ran its int32 range check on the already-wrapped value, and the `kLong` branch had no range check at all: `18446744073709551615` parsed as `Literal::Int(-1)`, and `9223372036854775808` as `Literal::Long(INT64_MIN)`. The untyped overload had the same hole.

Table metadata is the path where this changes returned data. `initial-default` / `write-default` go through this parser, `ValidateDefault` does not check integer range, and the value later gets materialized into a column. A metadata file that Java rejects reads back as `-1` here. The expression path is latent for now, since nothing evaluates `ReaderOptions::filter` yet.

Fixes #877.

## How

`GetInt64Checked` rejects unsigned nodes above INT64_MAX before the conversion, and the `kInt` branch, the `kLong` branch and the untyped overload all go through it. Java guards the same paths with `canConvertToInt()` / `canConvertToLong()` in `SingleValueParser` and `ExpressionParser`.

The `is_number_unsigned()` half of the guard is load-bearing: `get<uint64_t>()` on a negative node yields its two's-complement value, which compares above INT64_MAX and would reject every negative literal.

Two smaller things on lines this fix already touches: the int32 narrowing on the `kInt` path gained the coverage it never had, and the two out-of-range messages now use the same wording instead of saying "int" in one and "long" in the other.

## Testing

`expression_test` (526 tests) and the full `ctest` suite (18/18) pass. Each new test was checked against a mutation of the code it guards:

- `>` to `>=` in the guard: only `LongMax` fails. The `ULL` suffix there matters, since a signed INT64_MAX node skips the unsigned branch entirely.
- `is_number_unsigned() &&` dropped: `LongMin`, `IntNegative` and `AcceptsNegativeIntegerUntyped` fail.
- int32 range check deleted: `IntAboveInt32Max` and `IntBelowInt32Min` fail.

The three overflow-rejection cases were also confirmed to fail without the fix and pass with it.

## Out of scope

`GetTypedJsonValue` in `src/iceberg/util/json_util_internal.h` truncates out-of-range integers the same way, so `FieldFromJson({"id": 2147483648, ...})` yields `field_id = -2147483648` silently. That helper has roughly 80 call sites, so it belongs in its own PR.

### DIFF
--- a/src/iceberg/expression/json_serde.cc
+++ b/src/iceberg/expression/json_serde.cc
@@ -305,6 +305,21 @@ Result<nlohmann::json> ToJson(const Literal& literal) {
   }
 }
 
+/// \brief Read an integral JSON node as int64_t, rejecting out-of-range values.
+///
+/// nlohmann reports unsigned integers as is_number_integer(), and get<int64_t>()
+/// converts values above INT64_MAX silently rather than throwing, so the range
+/// has to be checked before the conversion. Mirrors Java's canConvertToLong().
+Result<int64_t> GetInt64Checked(const nlohmann::json& json) {
+  if (json.is_number_unsigned() &&
+      json.get<uint64_t>() > static_cast<uint64_t>(std::numeric_limits<int64_t>::max()))
+      [[unlikely]] {
+    return JsonParseError("Cannot parse {} as a long value: out of range",
+                          SafeDumpJson(json));
+  }
+  return json.get<int64_t>();
+}
+
 Result<Literal> LiteralFromJson(const nlohmann::json& json, const Type* type) {
   // If {"type": "literal", "value": <actual>} wrapper is present, unwrap it first.
   if (json.is_object() && json.contains(kType) &&
@@ -326,7 +341,7 @@ Result<Literal> LiteralFromJson(const nlohmann::json& json, const Type* type) {
       if (!json.is_number_integer()) [[unlikely]] {
         return JsonParseError("Cannot parse {} as an int value", SafeDumpJson(json));
       }
-      auto val = json.get<int64_t>();
+      ICEBERG_ASSIGN_OR_RAISE(auto val, GetInt64Checked(json));
       if (val < std::numeric_limits<int32_t>::min() ||
           val > std::numeric_limits<int32_t>::max()) [[unlikely]] {
         return JsonParseError("Cannot parse {} as an int value: out of range",
@@ -335,11 +350,13 @@ Result<Literal> LiteralFromJson(const nlohmann::json& json, const Type* type) {
       return Literal::Int(static_cast<int32_t>(val));
     }
 
-    case TypeId::kLong:
+    case TypeId::kLong: {
       if (!json.is_number_integer()) [[unlikely]] {
         return JsonParseError("Cannot parse {} as a long value", SafeDumpJson(json));
       }
-      return Literal::Long(json.get<int64_t>());
+      ICEBERG_ASSIGN_OR_RAISE(auto val, GetInt64Checked(json));
+      return Literal::Long(val);
+    }
 
     case TypeId::kFloat:
       if (!json.is_number_float()) [[unlikely]] {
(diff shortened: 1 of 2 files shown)


LuciferYang:
friendly ping @wgtmac

wgtmac (approved):
Thanks @LuciferYang!

LuciferYang:
Than you @wgtmac

PR #879, opened 2026-08-07
feat: add Nimble to FileFormatType

zzhao0:
Working on integrating Nimble https://github.com/facebookincubator/nimble to iceberg. This requires extending the FileFormatType enum.

Add `kNimble` to the `FileFormatType` enum along with its `ToString` and `FileFormatTypeFromString` mappings, so tables whose data files are written in the Nimble columnar format can be represented and round-tripped through manifest and REST JSON deserialization.

### DIFF
--- a/src/iceberg/file_format.h
+++ b/src/iceberg/file_format.h
@@ -37,6 +37,7 @@ enum class ICEBERG_EXPORT FileFormatType {
   kAvro,
   kOrc,
   kPuffin,
+  kNimble,
 };
 
 /// \brief Convert a FileFormatType to a string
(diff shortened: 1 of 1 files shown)


wgtmac:
IMHO, we cannot make this change until nimble has been formally accepted by the Iceberg spec. It would be good to start a discussion on dev@iceberg.apache.org to see what's the requirement to add a new file format. Vortex is doing this: https://docs.google.com/document/d/1g0EJDr3DkSmfAce8C94zA2JyWRj2SeNeAeiYOotnSww

zzhao0:
Thanks for pointing me to the direction. Let me read through this.

PR #880, opened 2026-08-08
feat: add v3 geometry and geography support

wgtmac:
- align geospatial type semantics and schema JSON with the v3 spec
- add bound encoding and Java-compatible intersection evaluation
- represent WKB as Arrow binary and support Avro byte round trips
- generate and validate Parquet geospatial logical types
- preserve count metrics while omitting geospatial byte-order bounds


src/iceberg/geospatial.cc
```diff
@@ -0,0 +1,269 @@
...
+  auto x = ReadLittleEndian<double>(bytes.data());
+  auto y = ReadLittleEndian<double>(bytes.data() + kDoubleSize);
+  if (bytes.size() == 2 * kDoubleSize) {
+    return XY(x, y);
+  }
+
+  auto z = ReadLittleEndian<double>(bytes.data() + 2 * kDoubleSize);
+  if (bytes.size() == 3 * kDoubleSize) {
+    return XYZ(x, y, z);
+  }
```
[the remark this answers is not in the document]
wgtmac: This intentionally matches Java's GeospatialBound.fromByteBuffer behavior. Java
accepts a 24-byte XYZ encoding with a NaN Z value, and subsequently treats that
Z dimension as absent. Rejecting it here would make the C++ reader stricter than
Java, so the current permissive normalization is intentional.

wgtmac:
@zhjwpku @manuzhang PTAL :)

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -60,6 +60,193 @@ std::optional<int32_t> GetFieldId(const ::parquet::arrow::SchemaField& parquet_f
...
+          ::parquet::schema::GroupNode::Make("key_value", ::parquet::Repetition::REPEATED,
+                                             {std::move(key), std::move(value)});
+      return ::parquet::schema::GroupNode::Make(
+          std::string(field.name()), repetition, {std::move(repeated)},
+          ::parquet::LogicalType::Map(), field.field_id());
+    }
+    case TypeId::kVariant:
+      return NotSupported("Cannot write Iceberg variant type to Parquet");
+    default:
+      return InvalidSchema("Expected nested Iceberg type, got {}", type);
```
zhjwpku, line 246: nit: `Variant` is not a nested type in spec, maybe we can use unreachable here.

zhjwpku (approved):

src/iceberg/parquet/parquet_schema_util.cc
```diff
@@ -105,16 +292,39 @@ void SelectAnchorColumnIfEmpty(
...
+                         expected_type);
+  }
+  if (descr.logical_type() == nullptr) {
+    return InvalidSchema("Missing Parquet leaf column metadata for Iceberg type {}",
+                         expected_type);
+  }
+  if (descr.physical_type() != ::parquet::Type::BYTE_ARRAY) {
+    return InvalidSchema("Iceberg type {} requires Parquet BYTE_ARRAY", expected_type);
+  }
+  if (expected_type.type_id() == TypeId::kGeometry) {
```
manuzhang, line 309: This (and the following) validates only whether the logical type is GEOMETRY or GEOGRAPHY; it does not compare the logical-type parameters with the expected Iceberg type. As a result, the reader currently accepts cases such as:

```
geometry(EPSG:3857) projected as geometry(OGC:CRS84)
geography(OGC:CRS84, spherical) projected as geography(EPSG:4326, karney)
```

Because WKB does not encode the CRS or edge-interpolation algorithm, this can silently interpret the same bytes using different semantics.

manuzhang:
cc @huan233usc who implemented geometry and geography support in iceberg java

src/iceberg/parquet/parquet_schema_util.cc
wgtmac: This actually aligns with the Java implementation: https://github.com/apache/iceberg/blob/main/parquet/src/main/java/org/apache/iceberg/data/parquet/BaseParquetReaders.java#L229-L240

I think the issue is that we don't have the ability to check if two CRS are logically equivalent.

blackmwk (approved):

src/iceberg/parquet/parquet_schema_util.cc
manuzhang: @wgtmac I also submitted a fix https://github.com/apache/iceberg/pull/17578 at Java side.

wgtmac: We can create a separate PR here to address this.

manuzhang (approved):

wgtmac:
Thanks everyone!

PR #885, opened 2026-08-12
refactor: protect merging snapshot update hooks

manuzhang:
## Summary

- move `MergingSnapshotUpdate`'s `Apply`, `CleanUncommitted`, and `Summary` overrides from public to protected
- keep the destructor public
- explicitly re-expose the protected hooks only in test subclasses that need white-box access

## Why

These methods implement internal extension hooks that are already protected in `SnapshotUpdate`. Declaring the overrides public unnecessarily widens the `MergingSnapshotUpdate` API. Keeping them protected aligns the derived class with the base-class contract while preserving access for subclasses and virtual dispatch from `SnapshotUpdate`.

External code that directly called these internal hooks through a concrete merging update will need to use the public snapshot-update workflow instead. The virtual function layout and runtime behavior are unchanged.

## Validation

- `clang-format --dry-run --Werror src/iceberg/update/merging_snapshot_update.h src/iceberg/test/merging_snapshot_update_test.cc`
- `cmake --build build --target table_update_test -j2`
- `build/src/iceberg/test/table_update_test --gtest_filter='MergingSnapshotUpdateTest.*:MergingSnapshotUpdateV1Test.*'` (79 tests passed)
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/src/iceberg/update/merging_snapshot_update.h
+++ b/src/iceberg/update/merging_snapshot_update.h
@@ -64,6 +64,7 @@ class ICEBERG_EXPORT MergingSnapshotUpdate : public SnapshotUpdate {
  public:
   ~MergingSnapshotUpdate() override = default;
 
+ protected:
   // SnapshotUpdate overrides
   Result<std::vector<ManifestFile>> Apply(
       const TableMetadata& metadata_to_update,
@@ -73,7 +74,6 @@ class ICEBERG_EXPORT MergingSnapshotUpdate : public SnapshotUpdate {
 
   std::unordered_map<std::string, std::string> Summary() override;
 
- protected:
   /// \brief Constructor; reads merge configuration from table properties.
   explicit MergingSnapshotUpdate(std::string table_name,
                                  std::shared_ptr<TransactionContext> ctx);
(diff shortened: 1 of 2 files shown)


src/iceberg/update/merging_snapshot_update.h
```diff
@@ -64,6 +64,7 @@ class ICEBERG_EXPORT MergingSnapshotUpdate : public SnapshotUpdate {
  public:
   ~MergingSnapshotUpdate() override = default;
 
+ protected:
   // SnapshotUpdate overrides
   Result<std::vector<ManifestFile>> Apply(
```
wgtmac: [P2][API compatibility] This changes the exported concrete MergingSnapshotUpdate hooks from public to protected, so downstream code that directly calls Apply/CleanUncommitted/Summary through this concrete type will stop compiling. The PR intentionally narrows the API, but please confirm the project's source-compatibility policy and document the migration to the SnapshotUpdate workflow (or provide a public wrapper if this API is supported).

wgtmac (approved):

PR #889, opened 2026-08-16
refactor(io): make FileIO resolution registry-driven

wgtmac:
## Problem

The existing FileIO design embeds scheme ownership in `ResolvingFileIO` and
registers the resolver itself as a special implementation. This duplicates
backend-specific scheme knowledge, prevents custom FileIOs from declaring the
schemes they support, and couples automatic routing to a special registry
entry.

Cached delegates are also returned as raw pointers while credential refresh
can invalidate the cache. In addition, treating `oss://` as an S3 alias is not
safe without provider-specific endpoint and compatibility validation.

## Changes

- Make `FileIORegistry::Factory` contain a required `create` callback and an
  optional `accepts` callback.
- Resolve normalized schemes from registered factories, with later
  registrations overriding earlier ones.
- Construct the default `ResolvingFileIO` directly while preserving explicit
  `io-impl` precedence.
- Use Java-compatible first-colon scheme parsing and remove hard-coded scheme
  mapping from the resolver.
- Keep cached delegates alive with `shared_ptr` across credential refreshes.
- Forward complete credential lists to delegates that support them and rebuild
  delegates after refresh.
- Limit Arrow S3 routing and credential prefix handling to `s3`, `s3a`, and
  `s3n`; defer OSS/COS support to a separate change.
- Add focused registry, resolver, REST, and S3 coverage.
- Document built-in and custom FileIO usage.


src/iceberg/arrow/s3/s3_properties.h
```diff
@@ -22,6 +22,8 @@
 /// \file iceberg/arrow/s3/s3_properties.h
 /// \brief Define S3 configuration property keys.
 
+#include <algorithm>
+#include <array>
 #include <string_view>
```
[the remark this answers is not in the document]
wgtmac: std::ranges::contains is declared in `<algorithm>`, which this header already includes.

src/iceberg/file_io_registry.cc
```diff
@@ -19,56 +19,98 @@
...
-void FileIORegistry::Register(const std::string& name, Factory factory) {
+void FileIORegistry::Register(std::string name, Factory factory) {
   auto& state = State();
   std::lock_guard lock(state.mutex);
-  state.registry[name] = std::move(factory);
+  std::erase_if(state.registrations, [&name](const RegistryState::Entry& entry) {
+    return entry.name == name;
+  });
+  state.registrations.emplace_back(std::move(name), std::move(factory));
 }
```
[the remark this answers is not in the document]
wgtmac: This is valid in C++23; parenthesized aggregate initialization has been supported since C++20, and all CI compilers build it successfully.

wgtmac:
I've made some refactoring to the FileIO resolution. Let me know what you think. @plusplusjiajia

wgtmac:
cc @MisterRaindrop

src/iceberg/util/location_util.cc
```diff
@@ -17,20 +17,16 @@
...
-#include "iceberg/result.h"
+#include "iceberg/util/location_util.h"
 
 namespace iceberg {
 
-/// \brief The FileIORegistry name of the implementation serving `location`,
-/// chosen by its URI scheme. Exported so tests can link it in shared builds.
-ICEBERG_EXPORT Result<std::string_view> ResolveFileIOName(std::string_view location);
+std::string_view LocationUtil::ParseScheme(std::string_view location) {
+  const auto colon = location.find(':');
```
zhjwpku, line 25: Should we find `://` here? Since Windows local path can contain `:`, not sure if we should care much about Windows platform though :-)

zhjwpku (approved):

src/iceberg/util/location_util.cc
wgtmac: Indeed I have tried this which ended up with messy code here. I deliberately align this logic with Java with the assumption that local file is not supposed to be used in production so we should not bother with them just for testing purposes.

plusplusjiajia:
> I've made some refactoring to the FileIO resolution. Let me know what you think. @plusplusjiajia

@wgtmac Thanks for looping me in — registry-driven resolution is right, and `shared_ptr` delegates close a real use-after-free window.

credential prefixes — agree.  `s3`/`s3a`/`s3n` matches Java's `S3FileIO`.

routing — I'd keep `oss`. This is stricter than Java: `S3URI` applies no scheme whitelist, and its javadoc says it "supports any valid URI schemes ... allows users to use S3FileIO with other S3-compatible object storage services". PyIceberg routes `oss` the same way ([map](https://github.com/apache/iceberg-python/blob/03ad0d6c/pyiceberg/io/__init__.py#L322), [init](https://github.com/apache/iceberg-python/blob/03ad0d6c/pyiceberg/io/pyarrow.py#L450)). `ArrowS3FileIO` is equally capable — `ResolvePath` already handles foreign schemes — so `kS3Schemes` is a new gate, and an unmatched scheme is a hard `NotSupported`, which makes such stores unusable.

src/iceberg/util/location_util.cc
zhjwpku: Fair enough.

src/iceberg/arrow/s3/arrow_s3_file_io.cc
```diff
@@ -91,11 +91,10 @@ std::string SplitEndpointScheme(std::string_view endpoint,
   return std::string(endpoint);
 }
 
-// Location prefixes this FileIO can serve: must cover every scheme
-// ResolveFileIOName routes here, or such a credential would be dropped.
+// Location prefixes this FileIO can serve.
 bool IsS3FileIOCredentialPrefix(std::string_view prefix) {
   return prefix == "s3" || prefix.starts_with("s3://") || prefix.starts_with("s3a://") ||
-         prefix.starts_with("s3n://") || prefix.starts_with("oss://");
+         prefix.starts_with("s3n://");
```
MisterRaindrop: We’d be better off using a macro for this？

src/iceberg/resolving_file_io.cc
```diff
@@ -32,33 +35,22 @@ ResolvingFileIO::ResolvingFileIO(std::unordered_map<std::string, std::string> pr
...
 
-Result<std::string_view> ResolveFileIOName(std::string_view location) {
-  const auto pos = location.find("://");
-  if (pos == std::string_view::npos) {
-    return FileIORegistry::kArrowLocalFileIO;
-  }
+Result<std::shared_ptr<FileIO>> ResolvingFileIO::FileIOForPath(
+    std::string_view location) {
+  const auto scheme = StringUtils::ToLower(LocationUtil::ParseScheme(location));
+  ICEBERG_ASSIGN_OR_RAISE(const auto name, FileIORegistry::Resolve(scheme));
```
MisterRaindrop, line 41: ResolvingFileIO does not have a bound stable registry snapshot therefore, the parsing results returned by FileIO may be affected by the process-wide mutable registry ?
example:  
Step 1: A(s3) Registration
Step 2: Resolver resolves s3 -> A

Time 3: A is replaced by a record with the same name.
Result: Resolver still hit an old cache entry named "A".

I just feel maybe there is this kind of problem.

src/iceberg/arrow/s3/arrow_s3_file_io.cc
wgtmac: I'm not sure if I understand this. Could you give an example?

wgtmac:
@plusplusjiajia I‘d prefer postponing the support of `oss` scheme to a separate PR. Correctly we have no evidence that it works. Perhaps we can provide a cli for FileIO to verify its connection.

src/iceberg/arrow/s3/arrow_s3_file_io.cc
MisterRaindrop: "macro" was not the right term. I feel moving this logic to kS3Schemes in s3_properties.h looks better
like this
```
inline bool IsS3CredentialPrefix(std::string_view prefix) {
    if (prefix == S3Properties::kS3Schema) {
      return true;
    }

    const auto delimiter = prefix.find("://");
    return delimiter != std::string_view::npos &&
           IsS3Scheme(prefix.substr(0, delimiter));
  }
```

src/iceberg/resolving_file_io.cc
wgtmac: I think the contract is that FileIO registration should be done before anything. A created ResolvingFileIO should just use its cached instances to serve followup calls.

src/iceberg/arrow/s3/arrow_s3_file_io.cc
wgtmac: Addressed. Thanks!

blackmwk (approved):

MisterRaindrop (approved):

plusplusjiajia:
> @plusplusjiajia I‘d prefer postponing the support of `oss` scheme to a separate PR. Correctly we have no evidence that it works. Perhaps we can provide a cli for FileIO to verify its connection.

Agreed — no objection to this going in as is. I'll bring `oss` back in #893 with the test and a `fileio.md`

wgtmac:
Thanks everyone!

PR #890, opened 2026-08-17
feat(logging): add transaction commit lifecycle logging

kamcheungting-db:
## Summary

Add operational logging to `Transaction::Commit` so retry behavior and commit
outcomes are easier to diagnose.

## Changes

- Log a `WARN` for each retry, including the table identifier, attempt number,
  and previous error.
- Log an `INFO` only after a real metadata commit succeeds.
- Include the number of attempts and newly committed snapshot details in the
  success log when available.
- Do not report empty or no-op transactions as successful commits.
- Do not classify `CommitStateUnknown` as a confirmed failure.
- Format snapshot details through the lazy logging path so logging failures
  cannot affect the commit result.

The retry policy and retry decisions are unchanged.

## Validation

- Built `table_update_test`
- Ran the transaction retry and no-op transaction tests
- 9 tests passed
- `clang-format`, `pre-commit`, and `git diff --check` passed

### DIFF
--- a/src/iceberg/table_metadata.cc
+++ b/src/iceberg/table_metadata.cc
@@ -38,6 +38,7 @@
 #include "iceberg/exception.h"
 #include "iceberg/file_io.h"
 #include "iceberg/json_serde_internal.h"
+#include "iceberg/logging/log_macros.h"
 #include "iceberg/metrics_config.h"
 #include "iceberg/partition_field.h"
 #include "iceberg/partition_spec.h"
@@ -1099,6 +1100,8 @@ Status TableMetadataBuilder::Impl::AddSnapshot(std::shared_ptr<Snapshot> snapsho
     metadata_.next_row_id += add_rows.value();
   }
 
+  ICEBERG_LOG_DEBUG("Added snapshot {} (sequence number {}) to table metadata",
+                    snapshot->snapshot_id, snapshot->sequence_number);
   return {};
 }
 
--- a/src/iceberg/transaction.cc
+++ b/src/iceberg/transaction.cc
@@ -20,9 +20,11 @@
 
 #include <format>
 #include <memory>
+#include <string>
 
 #include "iceberg/catalog.h"
 #include "iceberg/location_provider.h"
+#include "iceberg/logging/log_macros.h"
 #include "iceberg/schema.h"
 #include "iceberg/snapshot.h"
 #include "iceberg/statistics_file.h"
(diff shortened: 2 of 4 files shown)


src/iceberg/transaction.cc
```diff
@@ -375,15 +377,55 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
             return result;
           });
 
+  if (commit_result.has_value()) {
+    // Name the resulting snapshot only when this commit produced one (current
+    // snapshot advanced); metadata-only commits report a plain success.
+    std::string detail;
+    if (auto snapshot = commit_result.value()->metadata()->Snapshot();
+        snapshot.has_value() &&
+        snapshot.value()->snapshot_id != base_current_snapshot_id) {
```
wgtmac: This can mislabel commits: a metadata-only retry may report another writer’s snapshot, while StageOnly/ToBranch commits report none. Can we detect newly added snapshots instead of comparing current IDs?

src/iceberg/transaction.cc
```diff
@@ -375,15 +377,55 @@ Result<std::shared_ptr<Table>> Transaction::Commit() {
...
+                      op != summary.end() ? op->second : "unknown");
+    }
+    if (attempt > 1) {
+      ICEBERG_LOG_INFO("Transaction commit succeeded after {} attempts{}", attempt,
+                       detail);
+    } else {
+      ICEBERG_LOG_INFO("Transaction commit succeeded{}", detail);
+    }
+  } else {
+    ICEBERG_LOG_ERROR("Transaction commit failed after {} attempt(s): {}", attempt,
```
wgtmac: Java propagates the final commit error without a generic ERROR log. Also, `CommitStateUnknown` is not a confirmed failure. Could we remove this log or distinguish the unknown commit state explicitly?

src/iceberg/table_metadata.cc
```diff
@@ -1106,6 +1107,8 @@ Status TableMetadataBuilder::Impl::AddSnapshot(std::shared_ptr<Snapshot> snapsho
     metadata_.next_row_id += add_rows.value();
   }
 
+  ICEBERG_LOG_DEBUG("Added snapshot {} (sequence number {}) to table metadata",
```
wgtmac: Could we drop this DEBUG log to stay closer to Java? `AddSnapshot` is a low-level builder operation and may run for temporary or retry metadata that is never committed; Java logs only after the snapshot commit succeeds.

src/iceberg/transaction.cc
kamcheungting-db: Done

src/iceberg/transaction.cc
kamcheungting-db: Done

src/iceberg/table_metadata.cc
kamcheungting-db: Done

kamcheungting-db:
@wgtmac could you help me review this pr one more time?

wgtmac (approved):
I refined the commit logging to keep it aligned with the actual commit behavior. Empty or no-op transactions no longer emit a misleading success message, retry logs now use RetryRunner's attempt counter and include the table identifier, and snapshot details are formatted lazily through the logging path so logging failures cannot affect the commit result. This keeps the operational logs trustworthy during retries without changing the commit semantics.

PR #893, opened 2026-08-19
 feat(io): restore the oss scheme with a round trip test and docs

plusplusjiajia:
Per the suggestion on #889: `oss` comes back with a test and documentation.

  `oss` returns to `kS3Schemes`, which also restores the `oss://` credential prefix now that `IsS3CredentialPrefix` derives from that list. A bare `oss` prefix stays rejected, so nothing widens beyond what main had before #889. `CanonicalizeS3Scheme` now derives from the same list too — an alias missing there would not
  fail, it would silently stop matching its credential and fall back to the default one.

  `rest_arrow_file_io_test.cc` and its CMake target are restored; #889 removed them, leaving `REST -> ResolvingFileIO -> registry -> Arrow FileIO` uncovered. The new test writes, reads back and deletes an object addressed as `oss://`, against `ICEBERG_TEST_S3_URI` — the MinIO the AWS jobs already start — and skips when
  unset. It passes in 99 ms locally and fails with `URI scheme 'oss' is not supported for FileIO resolution` once `oss` leaves the scheme table. It proves routing, not credential matching: the same keys also reach the default client through the AWS chain.

  `file-io.md` gains an S3 property table and a section on S3-compatible storage. The OSS example is verified, not transcribed: the signing region is `cn-hangzhou` while the endpoint carries the `oss-` prefix, and path-style is rejected with `SecondLevelDomainForbidden: Please use virtual hosted style to access`.

### DIFF
--- a/mkdocs/docs/file-io.md
+++ b/mkdocs/docs/file-io.md
@@ -29,7 +29,7 @@ implementations:
 | Registry name | Schemes |
 |---|---|
 | `arrow-fs-local` | paths without a scheme, `file` |
-| `arrow-fs-s3` | `s3`, `s3a`, `s3n` |
+| `arrow-fs-s3` | `s3`, `s3a`, `s3n`, `oss` |
 
 The S3 implementation requires Arrow S3 support.
 
@@ -55,6 +55,48 @@ auto file_io = iceberg::FileIORegistry::Load(
 For a REST catalog, set `io-impl` to the registry name. If it is omitted, the
 REST catalog uses `ResolvingFileIO` and selects a registered implementation for
 each file location's scheme.
+
+## Configure S3
+
+| Key | Example | Description |
+|---|---|---|
+| `s3.access-key-id` | `admin` | Static access key ID; must be set together with the secret key |
+| `s3.secret-access-key` | `password` | Static secret access key |
+| `s3.session-token` | `AQoDYXdzEJr...` | Session token, for temporary credentials. Ignored unless both static keys are set |
+| `client.region` | `us-east-1` | Region to sign requests for |
+| `s3.endpoint` | `https://127.0.0.1:9000` | Endpoint to use instead of the AWS one |
+| `s3.path-style-access` | `true` | Address buckets as a path (`endpoint/bucket`) instead of a virtual host (`bucket.endpoint`). Only takes effect together with `s3.endpoint` |
+| `s3.ssl.enabled` | `true` | Scheme to use for the endpoint, overriding the one it carries |
+| `s3.connect-timeout-ms` | `1000` | Connection timeout |
+| `s3.socket-timeout-ms` | `5000` | Request timeout. Ignored outside Windows and macOS |
+
+Without credentials, the AWS default credential chain is used, which covers
+environment variables, the shared configuration file, and the various role and
+identity providers.
+
+### S3-compatible storage
+
+Stores that speak the S3 API are served by the same implementation. The scheme
+selects it; `s3.endpoint` decides where requests actually go. A location keeps
+its own scheme and is canonicalized internally, so a credential vended for the
+`s3` prefix applies to it.
+
+For Alibaba Cloud OSS, point `s3.endpoint` at the S3-compatible endpoint of the
+bucket's region and set `s3.path-style-access` to `false`: with a custom
+endpoint, buckets are addressed as a path unless told otherwise, and the
+service rejects that with
+`SecondLevelDomainForbidden: Please use virtual hosted style to access`:
+
+```cpp
+auto file_io = iceberg::FileIORegistry::Load(
+    iceberg::FileIORegistry::kArrowS3FileIO,
+    {{std::string(iceberg::arrow::S3Properties::kEndpoint),
+      "https://s3.oss-cn-hangzhou.aliyuncs.com"},
+     {std::string(iceberg::arrow::S3Properties::kClientRegion), "cn-hangzhou"},
+     {std::string(iceberg::arrow::S3Properties::kPathStyleAccess), "false"}});
+
+file_io.value()->NewInputFile("oss://bucket/path/to/file.parquet");
+```
 
 ## Register a custom FileIO
 
--- a/mkdocs/docs/file-io.md
+++ b/mkdocs/docs/file-io.md
@@ -64,8 +64,14 @@ each file location's scheme.
 | `s3.secret-access-key` | `password` | Static secret access key |
 | `s3.session-token` | `AQoDYXdzEJr...` | Session token, for temporary credentials. Ignored unless both static keys are set |
 | `client.region` | `us-east-1` | Region to sign requests for |
-| `s3.endpoint` | `https://127.0.0.1:9000` | Endpoint to use instead of the AWS one |
-| `s3.path-style-access` | `true` | Address buckets as a path (`endpoint/bucket`) instead of a virtual host (`bucket.endpoint`). Only takes effect together with `s3.endpoint` |
+| `s3.endpoint` | `https://127.0.0.1:9000` | Endpoint to use instead of the AWS one. When absent, the `AWS_ENDPOINT_URL_S3` / `AWS_ENDPOINT_URL` environment variables are consulted |
+| `s3.path-style-access` | `true` | Address buckets as a path (`endpoint/bucket`) instead of a virtual host (`bucket.endpoint`). Only takes effect together with a custom endpoint |
+
+The following keys are specific to iceberg-cpp; they are not part of the Java
+Iceberg or REST specification property set:
+
+| Key | Example | Description |
+|---|---|---|
 | `s3.ssl.enabled` | `true` | Scheme to use for the endpoint, overriding the one it carries |
 | `s3.connect-timeout-ms` | `1000` | Connection timeout |
 | `s3.socket-timeout-ms` | `5000` | Request timeout. Ignored outside Windows and macOS |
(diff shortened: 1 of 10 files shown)


wgtmac:
How about still removing oss support from https://github.com/apache/iceberg-cpp/pull/889 and then you add it back by adding test and documentation to fileio.md to instruct users how to config for oss?

plusplusjiajia:
> How about still removing oss support from #889 and then you add it back by adding test and documentation to fileio.md to instruct users how to config for oss?

@wgtmac Works for me — I'll keep this PR and grow it into that: once #889 lands I'll rebase, restore the scheme, and add `fileio.md` covering how to configure it.

wgtmac:
#889 has been merged. You may want to proceed with a rebase.

plusplusjiajia:
> #889 has been merged. You may want to proceed with a rebase.

@wgtmac Done — rebased and grown into the restore we discussed: `oss` back in `kS3Schemes`, the round-trip test restored, and `file-io.md` extended with S3 configuration and an OSS section. The test also caught a Windows regression on its first run — `ParseScheme` took drive letters for URI schemes — fixed minimally in`location_util.cc`

src/iceberg/arrow/s3/arrow_s3_file_io.cc
```diff
@@ -177,10 +177,17 @@ Result<std::shared_ptr<::arrow::fs::FileSystem>> BuildArrowS3FileSystem(
...
+// fail, it would silently stop matching its credential.
 std::string CanonicalizeS3Scheme(std::string_view location) {
-  for (std::string_view scheme : {"s3a://", "s3n://"}) {
-    if (location.starts_with(scheme)) {
-      return std::string("s3://").append(location.substr(scheme.size()));
+  for (std::string_view scheme : kS3Schemes) {
+    if (scheme == S3Properties::kS3Schema) {
+      continue;
+    }
+    if (location.starts_with(scheme) &&
```
wgtmac: `FileIORegistry::Resolve` lowercases the URI scheme, so `OSS://...` reaches this S3 FileIO, but this comparison only recognizes lowercase aliases. The credential prefix then misses and `FileIOForPath` silently falls back to default credentials. Please normalize the scheme before matching.

src/iceberg/util/location_util.cc
```diff
@@ -19,14 +19,45 @@
...
+#include <cctype>
+
 namespace iceberg {
 
+namespace {
+
+/// Whether `candidate` is a syntactically valid URI scheme (RFC 3986 section
+/// 3.1): a letter followed by letters, digits, `+`, `-` or `.`.
+bool IsValidScheme(std::string_view candidate) {
+  if (candidate.empty() || !std::isalpha(static_cast<unsigned char>(candidate.front()))) {
```
wgtmac: `std::isalpha` and `std::isalnum` are locale-sensitive, but RFC 3986 schemes are ASCII-only, as in Java's `LocationUtil`. With a non-C locale, a path such as `é:file` can be misclassified as a scheme; please use explicit ASCII checks.

src/iceberg/util/location_util.cc
```diff
@@ -19,14 +19,45 @@
...
-  return location.substr(0, colon);
+  const auto candidate = location.substr(0, colon);
+  // Cannot be a scheme -> a path whose first segment has a colon, such as the
+  // extended-length Windows form `\\?\C:\...`.
+  if (!IsValidScheme(candidate)) {
+    return {};
+  }
+#ifdef _WIN32
+  // A single letter before the colon is a drive, not a scheme.
+  if (candidate.size() == 1) {
```
wgtmac, line 62: Java does not special-case Windows here: `ResolvingFileIO.scheme()` returns the text before the first colon, and `LocationUtil.hasScheme()` only applies RFC 3986 character checks. Please remove the `_WIN32` single-letter drive handling and keep `ParseScheme` aligned with Java.

mkdocs/docs/file-io.md
```diff
@@ -56,6 +56,48 @@ For a REST catalog, set `io-impl` to the registry name. If it is omitted, the
...
+
+| Key | Example | Description |
+|---|---|---|
+| `s3.access-key-id` | `admin` | Static access key ID; must be set together with the secret key |
+| `s3.secret-access-key` | `password` | Static secret access key |
+| `s3.session-token` | `AQoDYXdzEJr...` | Session token, for temporary credentials. Ignored unless both static keys are set |
+| `client.region` | `us-east-1` | Region to sign requests for |
+| `s3.endpoint` | `https://127.0.0.1:9000` | Endpoint to use instead of the AWS one |
+| `s3.path-style-access` | `true` | Address buckets as a path (`endpoint/bucket`) instead of a virtual host (`bucket.endpoint`). Only takes effect together with `s3.endpoint` |
+| `s3.ssl.enabled` | `true` | Scheme to use for the endpoint, overriding the one it carries |
```
wgtmac, line 75: These three keys (`s3.ssl.enabled`, `s3.connect-timeout-ms`, and `s3.socket-timeout-ms`) predate this PR, but they are not Java Iceberg or REST-spec properties. Since this PR exposes them as public S3 configuration, please distinguish them from the standard keys.

src/iceberg/test/rest_arrow_file_io_test.cc
```diff
@@ -0,0 +1,198 @@
...
+
+  auto io =
+      MakeTableFileIO({{"warehouse", "logical_warehouse_name"}}, /*table_config=*/{},
+                      {{.prefix = "oss://bucket/table", .config = {{"k", "v"}}}});
+  ASSERT_THAT(io, IsOk());
+
+  // Opening only builds the delegate, so just the pre-network failure modes
+  // are asserted: kNotSupported for a routing break, the warning for a drop.
+  auto input = io.value()->NewInputFile("oss://bucket/table/data/file.parquet");
+  EXPECT_THAT(input, ::testing::Not(IsError(ErrorKind::kNotSupported)));
```
wgtmac: `NewInputFile` only resolves the URI and constructs a handle; it does not use credentials or hit the network. With `{"k", "v"}`, this only checks that the prefix was not skipped. Please replace it with a real round-trip using an `oss://` credential prefix and scrubbed default credentials, or remove it as redundant.

src/iceberg/test/arrow_s3_file_io_test.cc
```diff
@@ -191,8 +191,8 @@ TEST_F(ArrowS3FileIOTest, SkipsNonS3CredentialPrefix) {
 // credential that is silently skipped leaves S3 access on the default
 // credentials, which only surfaces much later as an auth error.
 TEST_F(ArrowS3FileIOTest, AppliesEveryS3CompatibleCredentialPrefix) {
-  for (std::string_view prefix :
-       {"s3", "s3://bucket/table", "s3a://bucket/table", "s3n://bucket/table"}) {
+  for (std::string_view prefix : {"s3", "s3://bucket/table", "s3a://bucket/table",
+                                  "s3n://bucket/table", "oss://bucket/table"}) {
```
wgtmac: This only checks that `SetStorageCredentials` emits no warning; it does not verify that `FileIOForPath` selects the credentialed delegate. Please cover `oss` in the real round-trip matrix, or rename this test to reflect prefix acceptance rather than application.

wgtmac (commented):
I just took an initial review. Have you checked it against a real OSS setup to make sure it works?

src/iceberg/arrow/s3/arrow_s3_file_io.cc
plusplusjiajia: @wgtmac Fixed — matching is now case-insensitive for URI-form schemes (centralized next to kS3Schemes), with `OSS://` in the test matrix.

src/iceberg/util/location_util.cc
plusplusjiajia: @wgtmac  Done — explicit ASCII checks.

src/iceberg/test/rest_arrow_file_io_test.cc
plusplusjiajia: @wgtmac Removed — the gated round trip covers it for real, and `RegistersBuiltInFileIOs` now resolves `oss` as the ungated routing guard.

src/iceberg/test/arrow_s3_file_io_test.cc
plusplusjiajia: @wgtmac  Both — renamed to `Accepts…` and added a gated `AppliesOssCredentialInRealRoundTrip` that fails unless the oss credential is canonicalized and selected.

mkdocs/docs/file-io.md
plusplusjiajia: @wgtmac   Done — moved to their own table, marked iceberg-cpp-specific.

src/iceberg/util/location_util.cc
plusplusjiajia: @wgtmac Good point on Java parity — I looked into why Java gets away without it: drive-letter handling lives below it in Hadoop's `Path`. We have no such layer, so without this guard every absolute Windows path fails
  resolution (`No FileIO registered for URI scheme 'c'`, https://github.com/apache/iceberg-cpp/actions/runs/32685994120). I've kept it for now, but happy to move it elsewhere if you'd prefer `ParseScheme` to stay literal.

plusplusjiajia:
> I just took an initial review. Have you checked it against a real OSS setup to make sure it works?

@wgtmac Thanks for the review! Yes — verified against a real OSS bucket (that's where the addressing/endpoint guidance in file-io.md comes from).

wgtmac (approved):
LGTM. Thanks @plusplusjiajia for working on this!

Issue #256, opened 2025-10-10
Implement bundle option in Meson configuration


WillAyd:
Per the discussion in https://github.com/apache/iceberg-cpp/pull/233/files#r2387234806 it would be great to add the bundle option to Meson, but we should realistically wait until the following two items are complete:

1. Apache Arrow 22.0 release
2. Avro becomes available in the Meson wrapdb (upstream work in https://github.com/mesonbuild/wrapdb/pull/2270)

PR #896, opened 2026-08-23
build: keep CMake and Meson definitions in sync

### LINKED ISSUE #894: CMake and Meson build definitions are Out Of Sync
## Description

The CMake and Meson build definitions are maintained separately and have drifted. On current `main` (`e0399c6945326dd5281da07de8ed14a61c9f2cac`), equivalent configurations do not install the same public headers or run the same core tests.

Examples:

- CMake installs public headers by globbing each directory, while Meson lists them manually. As a result:
  - CMake installs `iceberg/puffin_dv_io.h` and `iceberg/catalog/rest/auth/token_refresh_scheduler.h`, but Meson does not.
  - Meson installs `iceberg/catalog/catalog_util.h`, `iceberg/catalog/session_catalog.h`, and `iceberg/catalog/session_context.h`, but CMake does not install headers from `src/iceberg/catalog/`.
- CMake includes `snapshot_summary_builder_test.cc` in `table_test`, but Meson does not include that test source, so the Meson CI path does not exercise it.
- Vendored/fallback dependency versions differ between the two build systems:

  | Dependency | CMake | Meson |
  |---|---:|---:|
  | CRoaring | 4.4.3 | 4.3.11 |
  | nlohmann/json | 3.11.3 | 3.12.0 |
  | GoogleTest | 1.15.2 | 1.17.0 |

  Other shared dependencies, such as nanoarrow 0.9.0, utf8proc 2.10.0, spdlog 1.15.3, cpr 1.14.1, and Google Benchmark 1.8.4, are currently aligned but can drift because their pins are also declared independently.
- Dependency resolution policy also differs: CMake requires system Zlib and Curl, while Meson can fall back to pinned wraps (Zlib 1.3.1 and Curl 8.10.1). Any intentional difference here should be documented and tested explicitly.

The Meson bundle gap is already tracked by #256 and is outside the scope of this issue. Other intentional feature differences should likewise remain explicit rather than appearing as accidental drift.

---

**This to track current status and initiated a discussion on how we can maintain and sync the build tools in the long term.**

Sakshamm-Goyal:
## Summary

- install the missing public headers in Meson and the top-level catalog headers in CMake
- include `snapshot_summary_builder_test.cc` in Meson's `table_test`
- align the CRoaring, nlohmann/json, and GoogleTest fallback versions across CMake and Meson
- add a dependency-free parity regression and run it in CI

Closes #894.

## Validation

- `python3 dev/test_build_definition_parity.py`
- `pre-commit run --files .github/workflows/test.yml cmake_modules/IcebergThirdpartyToolchain.cmake src/iceberg/catalog/CMakeLists.txt src/iceberg/catalog/rest/auth/meson.build src/iceberg/meson.build src/iceberg/test/meson.build dev/test_build_definition_parity.py`
- CMake configure with bundle/REST disabled, followed by `cmake --build build-cmake-parity --target table_test --parallel 2`
- `ctest --test-dir build-cmake-parity -R '^table_test$' --output-on-failure`
- Meson configure with REST/tests enabled, followed by `meson compile -C build-meson-parity table_test -j 2`
- `meson test -C build-meson-parity table_test --print-errorlogs`
- inspected both generated install manifests for the corrected public headers

## AI assistance disclosure

AI-assisted tooling was used for initial regression-test scaffolding and mechanical build-definition edits. I reviewed the change end-to-end, verified the dependency choices and generated manifests, and ran the validations listed above. I am not aware of remaining uncertainty in the affected build paths.

### DIFF
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -443,8 +443,16 @@ function(resolve_croaring_dependency)
     endif()
 
     set(CROARING_VENDORED TRUE)
-    set_target_properties(roaring PROPERTIES OUTPUT_NAME "iceberg_vendored_croaring"
-                                             POSITION_INDEPENDENT_CODE ON)
+    # CRoaring 4.3.11 exposes its header-only helper targets in roaring's install
+    # interface. Iceberg installs roaring in its own export without those helper
+    # targets, so downstream find_package(iceberg) would otherwise reject the
+    # package. Preserve the build dependencies without leaking them to consumers.
+    set_target_properties(roaring
+                          PROPERTIES OUTPUT_NAME "iceberg_vendored_croaring"
+                                     POSITION_INDEPENDENT_CODE ON
+                                     INTERFACE_LINK_LIBRARIES
+                                     "$<BUILD_INTERFACE:roaring-headers>;$<BUILD_INTERFACE:roaring-headers-cpp>"
+    )
     install(TARGETS roaring
             EXPORT iceberg_targets
             RUNTIME DESTINATION "${ICEBERG_INSTALL_BINDIR}"
--- a/cmake_modules/IcebergThirdpartyToolchain.cmake
+++ b/cmake_modules/IcebergThirdpartyToolchain.cmake
@@ -422,7 +422,7 @@ function(resolve_croaring_dependency)
     set(CROARING_URL "$ENV{ICEBERG_CROARING_URL}")
   else()
     set(CROARING_URL
-        "https://github.com/RoaringBitmap/CRoaring/archive/refs/tags/v4.4.3.tar.gz")
+        "https://github.com/RoaringBitmap/CRoaring/archive/refs/tags/v4.3.11.tar.gz")
   endif()
 
   fetchcontent_declare(croaring
@@ -532,7 +532,7 @@ function(resolve_nlohmann_json_dependency)
     set(NLOHMANN_JSON_URL "$ENV{ICEBERG_NLOHMANN_JSON_URL}")
   else()
     set(NLOHMANN_JSON_URL
-        "https://github.com/nlohmann/json/releases/download/v3.11.3/json.tar.xz")
+        "https://github.com/nlohmann/json/releases/download/v3.12.0/json.tar.xz")
   endif()
 
   fetchcontent_declare(nlohmann_json
@@ -828,7 +828,7 @@ function(resolve_gtest_dependency)
 
   fetchcontent_declare(googletest
                        GIT_REPOSITORY https://github.com/google/googletest.git
-                       GIT_TAG b514bdc898e2951020cbdca1304b75f5950d1f59 # release-1.15.2
+                       GIT_TAG 52eb8108c5bdec04579160ae17225d66034bd723 # release-1.17.0
                        FIND_PACKAGE_ARGS
                        NAMES
                        GTest)
--- a/dev/test_build_definition_parity.py
+++ b/dev/test_build_definition_parity.py
@@ -84,6 +84,11 @@ class BuildDefinitionParityTest(unittest.TestCase):
         }
         self.assertEqual({}, mismatches)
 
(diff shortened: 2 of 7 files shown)


Sakshamm-Goyal:
The failing Ubuntu Debug and macOS **Build Example** steps were caused by CRoaring 4.3.11 exposing `roaring-headers` and `roaring-headers-cpp` in the vendored target install interface. Iceberg exports `roaring` without exporting those build-only helper targets, so downstream `find_package(iceberg)` marked the package not found.

I pushed `09fabab` to keep both helper targets behind `BUILD_INTERFACE` and added parity coverage for the export invariant. Validation completed locally:

- build-definition parity: 4/4 passed
- repository pre-commit on the changed files: passed
- real Debug CMake build + install: passed
- installed targets contain no `roaring-headers` references
- downstream `example/` configure and build against the installed package: passed

Replacement CI is now running on the updated head.

manuzhang:
@Sakshamm-Goyal Thanks for the PR! I opened #894 to track current status and initiated a discussion on how we can maintain and sync the build tools in the long term. Before taking actions, we'd better gather feedbacks from the community, especially from meson maintainers like @WillAyd. Meanwhile, there are prerequisite issues like #256 we need to resolve first.

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -422,7 +422,7 @@ function(resolve_croaring_dependency)
     set(CROARING_URL "$ENV{ICEBERG_CROARING_URL}")
   else()
     set(CROARING_URL
-        "https://github.com/RoaringBitmap/CRoaring/archive/refs/tags/v4.4.3.tar.gz")
+        "https://github.com/RoaringBitmap/CRoaring/archive/refs/tags/v4.3.11.tar.gz")
```
zhjwpku, line 425: I'm not sure downgrade the third-party version is the right way to go :-(

cmake_modules/IcebergThirdpartyToolchain.cmake
```diff
@@ -532,7 +540,7 @@ function(resolve_nlohmann_json_dependency)
     set(NLOHMANN_JSON_URL "$ENV{ICEBERG_NLOHMANN_JSON_URL}")
   else()
     set(NLOHMANN_JSON_URL
-        "https://github.com/nlohmann/json/releases/download/v3.11.3/json.tar.xz")
+        "https://github.com/nlohmann/json/releases/download/v3.12.0/json.tar.xz")
```
zhjwpku, line 543: I'd suggest upgrade these versions in a separate PR.

dev/test_build_definition_parity.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def extract(pattern: str, relative_path: str) -> str:
+    match = re.search(pattern, read(relative_path))
+    if match is None:
+        raise AssertionError(f"Could not find {pattern!r} in {relative_path}")
+    return match.group(1)
+
+
+class BuildDefinitionParityTest(unittest.TestCase):
```
zhjwpku, line 37: Can this check cover all the relevant files? ISTM we could simply add a section to `AGENTS.md` instructing AI coding tools to keep the CMake and Meson build files in sync. Wouldn’t that be enough?

dev/test_build_definition_parity.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertIn(
+            "token_refresh_scheduler.h",
+            read("src/iceberg/catalog/rest/auth/meson.build"),
+        )
+        self.assertIn(
+            "iceberg_install_all_headers(iceberg/catalog)",
+            read("src/iceberg/catalog/CMakeLists.txt"),
+        )
+
+    def test_core_tests_are_run_by_both_build_systems(self) -> None:
```
zhjwpku, line 49: This doesn't look right to me. We shouldn't run a test for a previously missing file.

zhjwpku:
We should close this PR for now, as we've removed Meson support #935 . We can revisit it once someone with Meson expertise is available to help maintain it.

PR #897, opened 2026-08-24
test(io): stop probing the EC2 metadata service in S3 tests

plusplusjiajia:
Off EC2, every S3 client build in the test suite waits several seconds for the instance metadata probe to time out before falling back. No test in this binary wants those credentials, and the wait dominated the run: `file_io_test` goes from ~54s to 0.3s locally, in CI included.

 Set without overwriting, so a run that does want instance credentials can still export `AWS_EC2_METADATA_DISABLED=false


wgtmac (approved):
Seems reasonable for local run speedup. In the future, perhaps we can introduce a RAII approach to set and restore envs just like Arrow C++ does.

PR #898, opened 2026-08-24
refactor(io): hand out S3 delegates by shared_ptr and build them off the lock

plusplusjiajia:
ArrowS3FileIO's per-credential delegates were raw references into an unsynchronized vector: `SetStorageCredentials` rebuilt it while a concurrent operation could still hold one, and building an S3 client — which can reach the network for bucket-region discovery — happened with no locking at all.

  This gives ArrowS3FileIO the treatment #889 gave ResolvingFileIO: delegates held and handed out by `shared_ptr`, state under a `shared_mutex`, delegate construction moved off the lock, and `DeleteFiles` matching every location against one snapshot so a batch sees a single delegate generation.`SupportsStorageCredentials::credentials()` returns by value, since a reference can dangle once a concurrent install replaces the vector.


src/iceberg/file_io.h
```diff
@@ -193,8 +193,11 @@ class ICEBERG_EXPORT SupportsStorageCredentials {
   virtual Status SetStorageCredentials(
       const std::vector<StorageCredential>& storage_credentials) = 0;
 
-  /// \brief Return currently installed storage credentials.
-  virtual const std::vector<StorageCredential>& credentials() const = 0;
+  /// \brief Return the storage credentials this FileIO holds.
+  ///
+  /// By value because a concurrent install may replace them. An implementation
+  /// that delegates may report what was installed on it.
+  virtual std::vector<StorageCredential> credentials() const = 0;
```
wgtmac: This changes a public virtual return type. External implementations stop compiling, and existing binaries break. Is that intentional?

src/iceberg/arrow/s3/arrow_s3_file_io.cc
```diff
@@ -209,27 +212,67 @@ class ArrowS3FileIO final : public FileIO, public SupportsStorageCredentials {
...
-      file_io_by_prefix_;
+  DelegatesByPrefix file_io_by_prefix_;
 };
 
 Status ArrowS3FileIO::SetStorageCredentials(
     const std::vector<StorageCredential>& storage_credentials) {
-  std::vector<std::pair<std::string, std::unique_ptr<ArrowFileSystemFileIO>>>
-      file_io_by_prefix;
-  file_io_by_prefix.reserve(storage_credentials.size());
+  ICEBERG_ASSIGN_OR_RAISE(auto delegates, BuildDelegates(storage_credentials));
```
wgtmac: REST installs credentials only at FileIO creation. C++ has no refresh path. What production path reinstalls them on a live S3 FileIO?

src/iceberg/arrow/s3/arrow_s3_file_io.cc
```diff
@@ -209,27 +212,67 @@ class ArrowS3FileIO final : public FileIO, public SupportsStorageCredentials {
...
+      std::vector<std::pair<std::string, std::shared_ptr<ArrowFileSystemFileIO>>>;
+
+  /// \brief Longest-prefix match against one consistent view of the delegates.
+  static std::shared_ptr<ArrowFileSystemFileIO> MatchDelegate(
+      const std::shared_ptr<ArrowFileSystemFileIO>& fallback,
+      const DelegatesByPrefix& by_prefix, std::string_view location);
+
+  /// \brief Build a delegate for each credential this FileIO can serve.
+  ///
+  /// Lock-free on purpose: building an S3 client can reach out to discover a
```
wgtmac: ResolvingFileIO still builds the first S3 client under its lock. Region discovery can block other operations. Can we move it out?

src/iceberg/arrow/s3/arrow_s3_file_io.cc
```diff
@@ -244,62 +287,91 @@ Status ArrowS3FileIO::SetStorageCredentials(
...
+  // One snapshot so the whole batch matches the same delegate generation; only
+  // ever a handful of delegates, so a linear scan beats hashing.
+  std::shared_ptr<ArrowFileSystemFileIO> fallback;
+  DelegatesByPrefix by_prefix;
+  {
+    std::shared_lock lock(mutex_);
+    fallback = default_file_io_;
+    by_prefix = file_io_by_prefix_;
+  }
+  std::vector<std::pair<std::shared_ptr<ArrowFileSystemFileIO>, std::vector<std::string>>>
```
wgtmac: This retains the pre-existing fail-fast behavior. Java attempts all batches and counts failures. Is matching that behavior in scope?

src/iceberg/test/arrow_s3_file_io_test.cc
```diff
@@ -237,6 +239,69 @@ TEST_F(ArrowS3FileIOTest, WarnsWhenNoCredentialApplies) {
...
+    return StorageCredential{
+        .prefix = std::string(prefix),
+        .config = {{std::string(S3Properties::kAccessKeyId), std::string(access_key)},
+                   {std::string(S3Properties::kSecretAccessKey), "secret"}}};
+  };
+  ASSERT_THAT(credentialed->SetStorageCredentials({credential("s3://bucket-a", "key-a"),
+                                                   credential("s3://bucket-b", "key-b")}),
+              IsOk());
+
+  auto status = result.value()->DeleteFiles({"s3://bucket-a/%ZZ.parquet",
```
wgtmac: The first URI fails parsing, so bucket-b is never reached. Can this test verify that both prefixes receive deletes?

src/iceberg/test/arrow_s3_file_io_test.cc
```diff
@@ -237,6 +239,69 @@ TEST_F(ArrowS3FileIOTest, WarnsWhenNoCredentialApplies) {
...
+  ASSERT_THAT(credentialed->SetStorageCredentials({credential("first")}), IsOk());
+
+  std::atomic<bool> stop = false;
+  std::atomic<int> failures = 0;
+  std::vector<std::thread> operations;
+  operations.reserve(4);
+  for (int i = 0; i < 4; ++i) {
+    operations.emplace_back([&] {
+      while (!stop.load()) {
+        if (!result.value()->NewInputFile("s3://bucket/key").has_value()) {
```
wgtmac: NewInputFile only creates a wrapper. Could this test hold a file across an install and perform an actual read?

src/iceberg/arrow/s3/arrow_s3_file_io.cc
```diff
@@ -192,7 +195,7 @@ class ArrowS3FileIO final : public FileIO, public SupportsStorageCredentials {
  public:
   ArrowS3FileIO(std::shared_ptr<::arrow::fs::FileSystem> arrow_fs,
                 std::unordered_map<std::string, std::string> default_properties)
-      : default_file_io_(std::move(arrow_fs)),
+      : default_file_io_(std::make_shared<ArrowFileSystemFileIO>(std::move(arrow_fs))),
```
wgtmac: Why do we need to change this?

PR #904, opened 2026-09-02
feat(hive): add Iceberg-to-HMS schema and record conversion helpers

MisterRaindrop:
Next step of the HiveCatalog split (follows #796): two conversion layers between Iceberg's catalog model and the HMS record shape. No Thrift, no I/O, fully unit tested.

- **`hive_schema`** — renders an Iceberg `Schema` as the Hive DDL column list HMS stores, nested types included.
- **`hive_utils`** — `Namespace` / `TableIdentifier` ↔ HMS `Database` / `Table`, the Iceberg-on-HMS parameter keys (`metadata_location`, `table_type`, `EXTERNAL`), and the LazySimpleSerDe descriptor Spark and Trino expect. Plain structs rather than Thrift objects, so `HmsClient` stays the only translation unit that includes the generated bindings.

Where the two reference implementations disagree:

- `timestamptz` → `timestamp with local time zone`, following iceberg-java; iceberg-rust rejects zoned timestamps outright, which would make such tables uncreatable. Java's sub-Hive-3 fallback is not reproduced — no Hive-version probe here, and the vendored IDL is Hive 4.0.1.
- A database owner is carried by `hive.metastore.database.owner` / `hive.metastore.database.owner-type`, the keys both references use; a bare `owner` is the *table* owner key.

Iceberg v3 types that neither reference maps to a real Hive type return `kNotImplemented` rather than guessing a spelling. Nothing calls any of this yet — `HiveCatalog` remains a stub; `HmsClient`'s Thrift wrappers and the catalog CRUD paths come next.

33 new unit tests; `hive_catalog_test` goes from 17 to 50 cases.

### DIFF
--- a/src/iceberg/catalog/hive/hive_schema.cc
+++ b/src/iceberg/catalog/hive/hive_schema.cc
@@ -1,0 +1,134 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "iceberg/catalog/hive/hive_schema.h"
+
+#include <format>
+#include <string>
+#include <vector>
+
+#include "iceberg/schema.h"
+#include "iceberg/schema_field.h"
+#include "iceberg/type.h"
+#include "iceberg/util/checked_cast.h"
+#include "iceberg/util/macros.h"
+
+namespace iceberg::hive {
+
+namespace {
+
+// Hive doesn't have a precise counterpart to every Iceberg primitive
+// type; the closest equivalent is chosen for each case below. Mapping
+// matches iceberg-rust's `HiveSchemaBuilder::primitive`.
+Result<std::string> PrimitiveToHive(const Type& type) {
+  switch (type.type_id()) {
+    case TypeId::kBoolean:
+      return std::string("boolean");
+    case TypeId::kInt:
+      return std::string("int");
+    case TypeId::kLong:
+      return std::string("bigint");
+    case TypeId::kFloat:
+      return std::string("float");
+    case TypeId::kDouble:
+      return std::string("double");
+    case TypeId::kDate:
+      return std::string("date");
+    case TypeId::kTimestamp:
+      return std::string("timestamp");
+    case TypeId::kTimestampTz:
+      // Hive's TIMESTAMPLOCALTZ, the spelling iceberg-java emits on
+      // Hive 3+. Engines reading through HMS must see the tz-adjusted
+      // type, otherwise they read an Iceberg timestamptz column as a
+      // naive timestamp. See the header for why the Hive 2 fallback is
+      // not reproduced.
+      return std::string("timestamp with local time zone");
+    case TypeId::kTime:
+    case TypeId::kString:
+    case TypeId::kUuid:
+      return std::string("string");
+    case TypeId::kBinary:
+    case TypeId::kFixed:
+      return std::string("binary");
+    case TypeId::kDecimal: {
+      const auto& decimal = internal::checked_cast<const DecimalType&>(type);
+      return std::format("decimal({},{})", decimal.precision(), decimal.scale());
+    }
+    default:
+      return NotImplemented("Unhandled primitive Iceberg type id: {}",
+                            static_cast<int>(type.type_id()));
+  }
+}
+
+Result<std::string> StructToHive(const StructType& type) {
+  std::string out = "struct<";
+  bool first = true;
+  for (const auto& field : type.fields()) {
+    if (!first) {
+      out += ",";
+    }
+    ICEBERG_ASSIGN_OR_RAISE(auto field_type, TypeToHiveString(*field.type()));
+    out += std::string(field.name());
+    out += ":";
+    out += field_type;
+    first = false;
+  }
+  out += ">";
+  return out;
+}
+
+Result<std::string> ListToHive(const ListType& type) {
+  ICEBERG_ASSIGN_OR_RAISE(auto inner, TypeToHiveString(*type.element().type()));
+  return std::format("array<{}>", inner);
+}
+
+Result<std::string> MapToHive(const MapType& type) {
+  ICEBERG_ASSIGN_OR_RAISE(auto key_type, TypeToHiveString(*type.key().type()));
+  ICEBERG_ASSIGN_OR_RAISE(auto value_type, TypeToHiveString(*type.value().type()));
+  return std::format("map<{},{}>", key_type, value_type);
+}
+
+}  // namespace
+
+Result<std::string> TypeToHiveString(const Type& type) {
+  switch (type.type_id()) {
+    case TypeId::kStruct:
+      return StructToHive(internal::checked_cast<const StructType&>(type));
+    case TypeId::kList:
+      return ListToHive(internal::checked_cast<const ListType&>(type));
+    case TypeId::kMap:
+      return MapToHive(internal::checked_cast<const MapType&>(type));
+    default:
+      return PrimitiveToHive(type);
+  }
+}
+
+Result<std::vector<HiveColumn>> SchemaToHiveColumns(const Schema& schema) {
+  std::vector<HiveColumn> columns;
+  columns.reserve(schema.fields().size());
+  for (const auto& field : schema.fields()) {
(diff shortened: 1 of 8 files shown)


src/iceberg/catalog/hive/hive_utils.h
```diff
@@ -0,0 +1,177 @@
...
+inline constexpr std::string_view kTableTypeIceberg = "ICEBERG";
+inline constexpr std::string_view kExternalKey = "EXTERNAL";
+inline constexpr std::string_view kExternalTrue = "TRUE";
+
+/// \brief Iceberg namespace / table property keys used during conversion.
+inline constexpr std::string_view kCommentProperty = "comment";
+inline constexpr std::string_view kLocationProperty = "location";
+
+/// \brief HMS table owner property key.
+inline constexpr std::string_view kOwnerProperty = "owner";
```
wgtmac: Java uses `hive.metastore.table.owner` as the table owner key. This code uses `owner`. With the Java key, `HiveTable.owner` stays empty. With `owner`, behavior differs from Java. Use the Java key here and update the test. If no owner is set, Java uses the current user.

src/iceberg/catalog/hive/hive_schema.cc
```diff
@@ -0,0 +1,134 @@
...
+      return std::string("string");
+    case TypeId::kBinary:
+    case TypeId::kFixed:
+      return std::string("binary");
+    case TypeId::kDecimal: {
+      const auto& decimal = internal::checked_cast<const DecimalType&>(type);
+      return std::format("decimal({},{})", decimal.precision(), decimal.scale());
+    }
+    default:
+      return NotImplemented("Unhandled primitive Iceberg type id: {}",
```
wgtmac: Java maps `VARIANT` to Hive `unknown`. This default path returns `kNotImplemented`. That blocks a valid v3 table that Java can create. Add `case TypeId::kVariant: return std::string("unknown");` and a test.

src/iceberg/catalog/hive/hive_utils.cc
```diff
@@ -0,0 +1,217 @@
...
+namespace {
+
+// Keys lifted into dedicated `Database` fields, so they must not be
+// duplicated into `parameters`.
+constexpr std::array<std::string_view, 4> kDatabaseReservedKeys = {
+    kCommentProperty, kLocationProperty, kHmsDbOwnerProperty, kHmsDbOwnerTypeProperty};
+
+// The table equivalent: `location` and `owner` have dedicated `Table`
+// fields, and `comment` is not part of the Iceberg-on-HMS contract.
+constexpr std::array<std::string_view, 3> kTableReservedKeys = {
```
wgtmac: `comment` is a standard Iceberg table property. Java forwards it to HMS parameters. Filtering it here loses the property. Keep it in `parameters` and update the test.

src/iceberg/catalog/hive/hive_utils.cc
```diff
@@ -0,0 +1,217 @@
...
+  table.output_format = std::string(kFileOutputFormat);
+
+  // Mandatory Iceberg-on-HMS marker parameters.
+  table.parameters.emplace(std::string(kMetadataLocationKey), metadata_location);
+  table.parameters.emplace(std::string(kTableTypeKey), std::string(kTableTypeIceberg));
+  table.parameters.emplace(std::string(kExternalKey), std::string(kExternalTrue));
+
+  // Forward any user-supplied table properties that aren't reserved.
+  for (const auto& [key, value] : table_properties) {
+    if (!IsTableReservedKey(key) && !table.parameters.contains(key)) {
```
wgtmac: Java translates `gc.enabled` to `external.table.purge`. It also handles `engine.hive.enabled`. This code adds the raw key and always uses LazySimpleSerDe. Add these translations and the storage-handler branch, or reject the flag clearly.

src/iceberg/catalog/hive/hive_utils.cc
```diff
@@ -0,0 +1,217 @@
...
+}
+
+Result<HiveDatabase> ConvertToHiveDatabase(
+    const Namespace& ns, const std::unordered_map<std::string, std::string>& properties) {
+  ICEBERG_RETURN_UNEXPECTED(ValidateNamespace(ns));
+  ICEBERG_RETURN_UNEXPECTED(ValidateOwnerSettings(properties));
+
+  HiveDatabase database;
+  database.name = ns.levels[0];
+  database.description = GetOrEmpty(properties, kCommentProperty);
```
wgtmac: Java sets a default database location and owner. This code leaves both empty. Pass the warehouse and current user into this layer, or fill them in the caller. Document the contract.

src/iceberg/catalog/hive/hive_utils.cc
```diff
@@ -0,0 +1,217 @@
...
+  database.location_uri = GetOrEmpty(properties, kLocationProperty);
+  database.owner_name = GetOrEmpty(properties, kHmsDbOwnerProperty);
+  database.owner_type = GetOrEmpty(properties, kHmsDbOwnerTypeProperty);
```
wgtmac: The validator accepts `USER`, `GROUP`, and `ROLE` case-insensitively. This line keeps the original case. The Thrift owner type is a `PrincipalType` enum. Normalize the value to `USER`, `GROUP`, or `ROLE`.

src/iceberg/catalog/hive/hive_utils.cc
```diff
@@ -0,0 +1,217 @@
...
+  table.owner = GetOrEmpty(table_properties, kOwnerProperty);
+  table.table_type = "EXTERNAL_TABLE";
+  table.location = std::string(location);
+  table.columns = columns;
+  table.serde = std::string(kLazySimpleSerDe);
+  table.input_format = std::string(kFileInputFormat);
+  table.output_format = std::string(kFileOutputFormat);
+
+  // Mandatory Iceberg-on-HMS marker parameters.
+  table.parameters.emplace(std::string(kMetadataLocationKey), metadata_location);
```
wgtmac: An empty `metadata_location` returns success here. `GetMetadataLocation` treats the same value as missing. Return `InvalidArgument` for empty input. Add a test.

src/iceberg/catalog/hive/hive_schema.cc
```diff
@@ -0,0 +1,134 @@
...
+      return std::string("date");
+    case TypeId::kTimestamp:
+      return std::string("timestamp");
+    case TypeId::kTimestampTz:
+      // Hive's TIMESTAMPLOCALTZ, the spelling iceberg-java emits on
+      // Hive 3+. Engines reading through HMS must see the tz-adjusted
+      // type, otherwise they read an Iceberg timestamptz column as a
+      // naive timestamp. See the header for why the Hive 2 fallback is
+      // not reproduced.
+      return std::string("timestamp with local time zone");
```
wgtmac: Java uses `timestamp with local time zone` only on Hive 3+. It uses `timestamp` on Hive 2. This code always emits the Hive 3+ type. Add version detection and fallback, or reject Hive 2 clearly.

src/iceberg/test/hive_utils_test.cc
```diff
@@ -0,0 +1,194 @@
...
+
+  const auto round_tripped = ConvertFromHiveDatabase(*original);
+  EXPECT_EQ(round_tripped.ns.levels, std::vector<std::string>{"warehouse"});
+  EXPECT_EQ(round_tripped.properties.at("comment"), "primary");
+  EXPECT_EQ(round_tripped.properties.at("hive.metastore.database.owner"), "alice");
+  EXPECT_EQ(round_tripped.properties.at("team"), "data");
+  EXPECT_FALSE(round_tripped.properties.contains("hive.metastore.database.owner-type"));
+}
+
+TEST(ConvertToHiveTableTest, SetsIcebergMarkerParametersAndStorageDescriptor) {
```
wgtmac: These tests do not try to override `metadata_location`, `table_type`, or `EXTERNAL`. The production code blocks overrides. Add one negative test for each key.

src/iceberg/test/hive_schema_test.cc
```diff
@@ -0,0 +1,189 @@
...
+std::string ToHive(const std::shared_ptr<Type>& type) {
+  auto result = TypeToHiveString(*type);
+  EXPECT_TRUE(result.has_value())
+      << "TypeToHiveString failed: " << (result ? "" : result.error().message);
+  return result.value_or("");
+}
+
+}  // namespace
+
+TEST(TypeToHiveStringTest, PrimitivesMapToHiveScalarTypes) {
```
wgtmac: The schema tests cover happy paths. They do not cover rejected v3 types. Add negative tests for `timestamp_ns`, `timestamptz_ns`, `geometry`, `geography`, and `unknown`.

src/iceberg/catalog/hive/hive_schema.h
```diff
@@ -0,0 +1,93 @@
...
+/// `timestamp with local time zone` (TIMESTAMPLOCALTZ). Engines that
+/// read the table through HMS -- Hive, Spark, Trino -- take their
+/// column types from this DDL, so downgrading to a naive `timestamp`
+/// would silently misrepresent a tz-adjusted column to every one of
+/// them. iceberg-rust instead rejects zoned timestamps outright, which
+/// would make such tables uncreatable; that behaviour is not copied.
+///
+/// Java picks the spelling at runtime and falls back to plain
+/// `timestamp` below Hive 3, where TIMESTAMPLOCALTZ does not exist.
+/// That fallback is not reproduced: iceberg-cpp has no Hive-version
```
wgtmac: Do we want to use a config to control this just in case?

PR #907, opened 2026-09-03
chore: bump apache avro version to 1.12.2

zhjwpku:


zhjwpku:
This seems wrong, we should wait for 1.13 :-(

PR #910, opened 2026-09-05
docs: clarify contribution conventions for agents

zhjwpku:
use Conventional Commits for commit messages and run pre-commit for every PR, fixing any reported issues before review.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,3 +28,17 @@ When assessing potential vulnerabilities or calibrating automated security
 findings, use [`SECURITY-THREAT-MODEL.md`](SECURITY-THREAT-MODEL.md) as the
 authoritative detailed description of this repository's security boundaries,
 trust assumptions, and non-boundaries.
+
+## Build System Consistency
+
+Keep CMake and Meson build definitions in sync. When adding, removing, or
+renaming source files, headers, tests, or build targets, or changing dependencies
+or build options, update the corresponding CMake (`CMakeLists.txt` and CMake
+modules) and Meson (`meson.build` and `meson.options`) definitions in the same
+change.
+
+## PR & Commit Conventions
+
+- Use Conventional Commits for commit messages.
+- Commit messages describe the what and why, not implementation details.
+- Run `pre-commit` for every PR and fix any reported issues.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,5 +40,4 @@ change.
 ## PR & Commit Conventions
 
 - Use Conventional Commits for commit messages.
-- Commit messages describe the what and why, not implementation details.
 - Run `pre-commit` for every PR and fix any reported issues.


AGENTS.md
```diff
@@ -28,3 +28,17 @@ When assessing potential vulnerabilities or calibrating automated security
...
+Keep CMake and Meson build definitions in sync. When adding, removing, or
+renaming source files, headers, tests, or build targets, or changing dependencies
+or build options, update the corresponding CMake (`CMakeLists.txt` and CMake
+modules) and Meson (`meson.build` and `meson.options`) definitions in the same
+change.
+
+## PR & Commit Conventions
+
+- Use Conventional Commits for commit messages.
+- Commit messages describe the what and why, not implementation details.
```
wgtmac: I would recommend keeping PR title and description in sync with latest commit. And then we can directly use those to be the final git commit message.

AGENTS.md
```diff
@@ -28,3 +28,17 @@ When assessing potential vulnerabilities or calibrating automated security
...
+- Run `pre-commit` for every PR and fix any reported issues.
```
wgtmac, line 43: I think this is enforced by ci so we don't need to mention it?

AGENTS.md
zhjwpku: This line comes from https://github.com/apache/iceberg/blob/main/AGENTS.md?plain=1#L169, I'm not sure how can the AI coding tools know which is the latest commit, there can be other comments to be addressed after the latest commit.

AGENTS.md
zhjwpku: Yeah, this is enforced by ci, but I think it would be great if the agent can run pre-commit locally before creating the PR.

zhjwpku: Maybe the agent is capable enough to detect `.pre-commit-config.yaml` and run the pre-commit automatically. If you still think this is unnecessary, I'll remove it.

manuzhang: > Maybe the agent is capable enough to detect .pre-commit-config.yaml and run the pre-commit automatically.

no,  it doesn't. My PR often fails due to pre-commit CI failure.

AGENTS.md
manuzhang: I think @wgtmac's concern is that PR title and description could be out of sync when the PR contents are updated based on reviews. On the other hand, I'm not worried about Commit message itself, where AI has always done a good job in describing what and why.

zhjwpku: Yeah, I agree that the PR title and description can get out of sync, but I'm not sure how we could reliably instruct the agent to update them once the PR is ready to merge. I think that's a point where human review is still necessary.

I will delete this one since llm should be trained to do a good job.

zhjwpku:
```suggestion
```

zhjwpku:
I've removed the Build System Consistency section since Meson support has been removed. This PR now only adds guidance on using Conventional Commits and running pre-commit before review. Please take a look again. @wgtmac @manuzhang

manuzhang (approved):

wgtmac (approved):
I think now this is clear. Let's move forward. We can always improve once we've got better idea.

PR #912, opened 2026-09-05
feat(meson): align library modules with CMake

### LINKED ISSUE #256: Implement bundle option in Meson configuration

zhjwpku:
Provide the same library modules through Meson and CMake, including the bundle, Hive, and SQL catalogs, so applications can use either build system.

Keep installed static and shared libraries usable by downstream consumers and cover their public interfaces with installation checks.

Closes #256

AI assistance was used in this change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -208,7 +208,7 @@ jobs:
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
     name: Meson - ${{ matrix.title }}
     runs-on: ${{ matrix.runs-on }}
-    timeout-minutes: 30
+    timeout-minutes: 60
     strategy:
       max-parallel: 15
       fail-fast: false
@@ -221,9 +221,22 @@ jobs:
             meson-setup-args: -Drest_integration_test=enabled
           - title: AMD64 Windows 2025
             runs-on: windows-2025
-            meson-setup-args: --vsenv
+            meson-setup-args: --vsenv --buildtype=release --cmake-prefix-path=C:/vcpkg/installed/x64-windows-static-md
           - title: AArch64 macOS 26
             runs-on: macos-26
+          - title: AMD64 Ubuntu 26.04, all catalogs and library variants
+            runs-on: ubuntu-26.04
+            CC: gcc-14
+            CXX: g++-14
+            sql-clients: true
+            benchmarks: true
+            meson-setup-args: --default-library=both --buildtype=release -Db_ndebug=true --force-fallback-for=arrow,avro -Dhive=enabled -Dsql_catalog=enabled -Dsql_sqlite=enabled -Dsql_postgresql=enabled -Dsql_mysql=enabled -Dbenchmarks=enabled
+          - title: AMD64 Ubuntu 26.04, bundle disabled
+            runs-on: ubuntu-26.04
+            CC: gcc-14
+            CXX: g++-14
+            bundle: disabled
+            meson-setup-args: -Dbundle=disabled
     steps:
       - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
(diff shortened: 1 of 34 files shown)


zhjwpku:
I'm not a meson expert, I use codex to help to align meson build with CMake, @WillAyd I'd appreciate it if you can help to review this change.

meson.options
```diff
@@ -30,13 +30,68 @@
 # and ICEBERG_INSTALL_DOCDIR correspond to Meson's --libdir / --bindir /
 # --includedir / --datadir arguments, respectively
 
+option(
+    'bundle',
```
WillAyd, line 34: Meson provides functionality that CMake doesn't have for wrangling dependencies and choosing how they should (or shouldn't) be packaged in a source distribution / install. As such, I don't think this option makes a lot of sense to implement in Meson

meson.options
```diff
@@ -30,13 +30,68 @@
...
     'rest',
     type: 'feature',
     description: 'Build rest catalog client',
     value: 'enabled',
 )
 
+option(
+    'hive',
+    type: 'feature',
+    value: 'disabled',
```
WillAyd, line 50:
```suggestion
```

I don't think its worth repeating these everywhere; best to keep as auto

meson.build
```diff
@@ -17,11 +17,13 @@
 
 project(
     'iceberg',
+    'c',
     'cpp',
     version: '0.4.0',
     license: 'Apache-2.0',
-    meson_version: '>=1.3.0',
+    meson_version: '>=1.8.3',
```
WillAyd, line 24: Do you know why this needs to be bumped? Generally better to keep support for older versions unless there's a critical feature

meson.build
```diff
@@ -17,11 +17,13 @@
...
     default_options: [
+        'default_library=static',
```
WillAyd, line 26: Do you know why this needs to change?

subprojects/packagefiles/arrow/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+cmake_minimum_required(VERSION 3.25)
+project(IcebergArrow LANGUAGES C CXX)
+
+# Meson's CMake module imports library targets but not header installation.
```
WillAyd, line 21: Why does we need a CMakeLists.txt file for arrow in the meson subprojects space? There is already a Meson wrapdb entry for arrow

src/iceberg/CMakeLists.txt
```diff
@@ -184,6 +184,7 @@ list(APPEND
      roaring::roaring)
 list(APPEND
      ICEBERG_STATIC_INSTALL_INTERFACE_LIBS
+     ZLIB::ZLIB
```
WillAyd, line 187: Is this related to meson?

src/iceberg/test/meson.build
```diff
@@ -23,10 +23,54 @@ configure_file(
...
+        deps = [get_variable(parent + '_test_dep')] + deps
+    endif
+    test_args = spec.get('compile_args', [])
+    if host_machine.system() == 'windows'
+        test_args += '-D' + name.to_upper() + '_STATIC'
+    endif
+    if get_option('default_library') != 'shared'
+        lib = get_variable(name + '_static_lib')
+    elif host_machine.system() == 'windows'
+        lib = static_library(
```
WillAyd, line 45: You can avoid this branching if you use the generic `build_target` function and specify the `target_type`

.github/workflows/test.yml
```diff
@@ -221,9 +221,22 @@ jobs:
...
+            meson-setup-args: --vsenv --buildtype=release --cmake-prefix-path=C:/vcpkg/installed/x64-windows-static-md
           - title: AArch64 macOS 26
             runs-on: macos-26
+          - title: AMD64 Ubuntu 26.04, all catalogs, S3 and library variants
+            runs-on: ubuntu-26.04
+            CC: gcc-14
+            CXX: g++-14
+            sql-clients: true
+            benchmarks: true
+            meson-setup-args: --default-library=both --buildtype=release -Db_ndebug=true --force-fallback-for=arrow,avro -Dhive=enabled -Dsql_catalog=enabled -Dsql_sqlite=enabled -Dsql_postgresql=enabled -Dsql_mysql=enabled -Dbenchmarks=enabled -Ds3=enabled -Dsigv4=enabled
```
WillAyd, line 233: This would be better expressed via `-Dauto_features=enabled` instead of specifying each option; that way the CI scales across all features as they get added

WillAyd (changes requested):
Thanks @zhjwpku . This is a rather large change - my suggestion would be to break it up into smaller pieces.

zhjwpku:
Closed this in favor of #935.

PR #925, opened 2026-09-07
feat: expose ReplacePartitions on Table and Transaction

shangxinli:
PR2 of 2 for #775, closes the ReplacePartitions item in #637.

Adds `NewReplacePartitions()` to `Table` and `Transaction`, with the `StaticTable`
override returning `NotSupported`. Placement follows `SetSnapshot`: exposed on
`Transaction`, and on `Table` for the standalone-commit path.

Tests now go through `Table::NewReplacePartitions()` instead of constructing the
operation directly, plus a transaction-scoped test and a multi-file-per-partition test.


wgtmac (approved):

wgtmac:
Thanks for completing this!

PR #926, opened 2026-09-08
feat: add CherryPickOperation and wire SnapshotManager::Cherrypick

shangxinli:
Adds `CherryPickOperation` and implements `SnapshotManager::Cherrypick()`, which was a TODO stub.

Follows Java's `CherryPickOperation`: appends re-apply their added files, dynamic overwrites (`replace-partitions`) re-apply adds and deletes and are rejected if a replaced partition changed, and anything else can only be fast-forward. Records `source-snapshot-id` and `published-wap-id`, and rejects a WAP id that was already published.

Like Java, the operation is reachable only through `Transaction`, not `Table`.

One divergence worth a look: `Apply()` here always builds a new snapshot, so it cannot express Java's fast-forward, which returns an existing one. `SnapshotManager` instead routes a fast-forward to `SetSnapshot`, with the predicate in `SnapshotUtil::CanFastForward()` so there is a single definition. `SetSnapshot::RequireFastForward()` re-checks it at apply time so a commit retry cannot silently turn the fast-forward into a branch move.

14 tests ported from Java `TestSnapshotManager` and `TestWapWorkflow`.

Part of #637.


zhjwpku:
The Meson build has been removed. I'll review this PR once the conflicts are resolved. Thanks!

shangxinli:
@zhjwpku thanks. it is resolved now.

src/iceberg/update/cherry_pick_operation.cc
```diff
@@ -0,0 +1,333 @@
...
+      ICEBERG_BUILDER_RETURN_IF_ERROR(DeleteDataFile(removed));
+    }
+  }
+
+  return *this;
+}
+
+Status CherryPickOperation::ValidateNonAncestor(const TableMetadata& metadata,
+                                                int64_t snapshot_id) const {
+  ICEBERG_ASSIGN_OR_RAISE(bool is_ancestor,
```
zhjwpku, line 252: If we create a new branch on an empty table, append some data to it, and then cherry-pick that append to main, this check fails with `No current snapshot`[1]. ISTM that main has no history yet, should we allow this case?

[1] https://github.com/apache/iceberg-cpp/blob/main/src/iceberg/table_metadata.cc#L284-L286

PR #932, opened 2026-09-11
refactor: expose snapshot utilities in a public header

zhjwpku:
Making SnapshotUtil public lets downstream integrations reuse snapshot traversal, time lookup, and schema resolution logic. Java impl already uses it outside the core library: Spark uses `schemaFor`, and Flink uses `ancestorsBetween`.
Exposing the C++ equivalent would support similar use cases without duplicating that logic.

### DIFF
--- a/src/iceberg/table_scan.cc
+++ b/src/iceberg/table_scan.cc
@@ -39,7 +39,7 @@
 #include "iceberg/table_metadata.h"
 #include "iceberg/util/content_file_util.h"
 #include "iceberg/util/macros.h"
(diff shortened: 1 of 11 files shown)


smaheshwar-pltr (approved):
+1 on this change! (I also wanted to do this actually - https://github.com/smaheshwar-pltr/iceberg-cpp/pull/4)

wgtmac (approved):
Thanks @zhjwpku and @smaheshwar-pltr!

PR #933, opened 2026-09-11
ci: use Quay and GitHub releases for MinIO

smaheshwar-pltr:
Related:

- https://github.com/apache/iggy/pull/4148
- https://github.com/apache/iceberg/pull/18071
- https://github.com/apache/iceberg-rust/pull/3202
- https://github.com/apache/doris/pull/67897

Docker Hub is no longer serving `minio/minio` (https://hub.docker.com/r/minio/minio), preventing the Docker-based S3 test environment from starting on clean machines, including Linux CI and open PRs. The native binary downloads from `dl.min.io` also return HTTP 410, stopping bucket setup on Linux and macOS.

This PR switches to the corresponding Quay image, as the related PRs above do - those other projects are encountering the same issue as us. It also downloads the MinIO client and Windows server from pinned GitHub releases and fails immediately on HTTP download errors.

We should track moving away from MinIO as a follow-up, like [PyIceberg did](https://github.com/apache/iceberg-python/pull/3928) - but that's a larger discussion; this fix is temporary to unblock CI, PRs, etc.

AI Disclosure: Assisted by Codex (GPT-6).


kevinjqliu (approved):
LGTM ty!

PR #935, opened 2026-09-15
build: remove Meson build support

### LINKED ISSUE #930: [Discussion] Remove meson build support
I would like to propose removing meson build support from iceberg-cpp.

We currently have meson build support for:

- iceberg: the core library
- iceberg_data: the data reading and writing layer
- iceberg_rest: the REST catalog client

But sadly, meson support is still not aligned with CMake, missing modules include iceberg_bundle (Arrow, Avro, and Parquet integrations), the Hive catalog, and the SQL catalog.

A major challenge is that we do not seem to have enough meson expertise available for ongoing maintenance and review. AI can help us write build definitions quickly, but we still need knowledgeable reviewers to verify that the changes work correctly across supported configurations and for downstream consumers.

For example, I used AI to help implement PR #912, which aims to align meson's library modules with CMake. The implementation was relatively quick, but the PR has not received any reviews so far. Without sufficient meson expertise, it is difficult to gain confidence in the implementation and maintain it over time.

Meanwhile, we run meson CI jobs on Linux, Windows, and macOS. These consume CI resources and require maintenance even though our meson support remains incomplete. Apache Iceberg already has limited CI capacity, so I think we should consider whether maintaining a second build system is a worthwhile use of those resources.

As far as I am aware, we do not currently have users relying on Meson to build iceberg-cpp. Given the maintenance burden, limited review capacity, and CI cost, I propose removing the meson build and associated CI jobs.

If you are using meson to build iceberg-cpp, or are willing to help maintain and review its build support, please share your use case and thoughts. All feedback and alternative suggestions are welcome.

cc @wgtmac @manuzhang @WillAyd @abnobdoss @HuaHuaY

zhjwpku:
Closes #930.


manuzhang (approved):

manuzhang:
@zhjwpku please rebase and resolve conflicts.

wgtmac (approved):

PR #936, opened 2026-09-17
feat: make the public headers consumable as C++20

### LINKED ISSUE #928: Public headers require C++23, forcing the same on every consumer
`CMakeLists.txt:35` sets `CMAKE_CXX_STANDARD 23`, which is a fine choice for the
library's own sources. The requirement is not confined to them, though.
`src/iceberg/result.h` includes `<expected>` and `<format>`, and defines:

```cpp
template <typename T, typename E = typename DefaultError<T>::type>
using Result = std::expected<T, E>;

using Status = Result<void>;
```

`Result<T>` is the return type of most public entry points —
`Catalog::ListNamespaces`, `ListTables`, `LoadTable`, `CreateTable`,
`StageCreateTable` and so on. Every consumer translation unit that calls them
must therefore compile as C++23 as well. `std::expected` needs libstdc++ 12 or
libc++ 16 and `<format>` needs libstdc++ 13, so a consumer on an older but still
widely deployed toolchain cannot include the headers at all.

This matters for the engines the library is meant to be embedded in: Velox
builds as C++20, Arrow as C++17, DuckDB as C++11. Integrating leaves two
options — move the whole engine to C++23, or add an isolation layer whose only
purpose is keeping iceberg-cpp headers out of the rest of the build. We are
looking at iceberg-cpp for an Iceberg connector in
[Axiom](https://github.com/facebookincubator/axiom), which builds on Velox at
C++20, and would rather do neither.

Would you consider making the public error type portable while keeping the API
shape unchanged?

```cpp
#if defined(__cpp_lib_expected)
template <typename T, typename E = typename DefaultError<T>::type>
using Result = std::expected<T, E>;
#else
// vendored fallback with the same interface
#endif
```

`arrow::Result`, `absl::StatusOr` and `tl::expected` all exist for this reason.
The library's own sources could keep building as C++23; only the public headers
would need to hold a lower baseline.

Happy to send a patch if the direction is agreeable.

cc @PingLiuPing, who is proposing the Iceberg connector for Axiom.

PingLiuPing:
Enable C++20 consumers to include and use iceberg-cpp’s installed public headers while keeping iceberg-cpp implementation code and internal workflows on C++23.

Restore the vendored iceberg::expected backport — brings back src/iceberg/expected.h from #40 (removed by [#139](https://github.com/PingLiuPing/axiom/issues/139)), with its test and LICENSE entry. unexpected's comparison operators now go through error().

Built and ran public_headers_cxx20 as an external C++20 package consumer; all 199 installed public headers compiled and linked.

Fix #928


zhjwpku:
It seems there are some conflicts, please rebase the main branch.

example/CMakeLists.txt
```diff
@@ -20,13 +20,65 @@ cmake_minimum_required(VERSION 3.25)
 
 project(example)
 
-set(CMAKE_CXX_STANDARD 23)
```
zhjwpku, line 23: Can we make C++23  still as default and let it accept user supplied option so that C++20 can be test manually locally.

PingLiuPing: @zhjwpku Thank you for the comments.

Making it configurable is a good idea. Have different thoughts on making C++23 as default though.

With this patch, it changes the minimum supported C++ standard from C++23 to C++20 for downstream consumers. And C++20 is the interface contract between the consumer and iceberg-cpp library, and the contract should be tested continuously. 

Setting the default to C++20 ensures the minimum supported standard (contract) is continuously exercised. Defaulting it to C++23 would let C++20 only breakages slip through.

One refinement is that C++23 compatibility should still be tested separately. C++23 should accepts C++20 code, but we can enhance this by provide an optional example configuration for C++23, for example, expose an `ICEBERG_EXAMPLE_CXX_STANDARD` cache setting that defaults to 20 and accepts 23; then update CI to build both.
And also refine the document to state clearly that the minimum C++ standard is C++20 for public headers. What do you think?

Happy to make changes either way.

zhjwpku: Make sense to me, I think we should build both for compatibility purpose.

src/iceberg/expected.h
```diff
@@ -0,0 +1,2401 @@
+/*
```
manuzhang, line 1: Do we have to put this file here? It will be better to be placed in a `third_lib` directory and have a README to explain its purpose.

PingLiuPing: @manuzhang Thanks.

Wondering which one you mean: `<repo_root>/src/iceberg/third_lib` or `<repo_root>/third_lib`?

manuzhang: I see there's already a `thirdparty` directory, but I'm not sure that's the best place. @wgtmac should have more background.

zhjwpku: I think it's fine to put `expected.h` here. We already have `murmurhash3_internal.h` in `src/iceberg/util`. Similarly, the thrift idls in `thirdparty` generate code in `src/iceberg/catalog/hive/gen-cpp`, and those generated files are also tracked in git.

manuzhang: I'm worried about long term maintenance. We may forget what it is, where it comes from and when to drop over time.

PingLiuPing: Would it help to add a few comments to the file header (after the license) explaining the purpose of this file and providing a brief history?
Or I can update the commit message or separate it to another PR and explain in the PR description.

PingLiuPing: @manuzhang @zhjwpku Added comments in expected.h. Let's me know if that works for you.
Happy to make further changes.

zhjwpku: Work for me, I think the `zeus-cpp` section in LICENSE should tell people the history of `expected.h`, but no objection for the comments in `expected.h`.

example/CMakeLists.txt
```diff
@@ -20,13 +20,79 @@ cmake_minimum_required(VERSION 3.25)
...
+
+if(NOT ICEBERG_PUBLIC_INCLUDE_DIR)
+  message(FATAL_ERROR "Could not locate iceberg's installed public headers")
+endif()
+
+file(GLOB_RECURSE
+     ICEBERG_PUBLIC_HEADERS
+     CONFIGURE_DEPENDS
+     "${ICEBERG_PUBLIC_INCLUDE_DIR}/iceberg/*.h"
+     "${ICEBERG_PUBLIC_INCLUDE_DIR}/iceberg/*.hpp")
```
wgtmac, line 79: nit: we don't have `.hpp`, do we?

example/CMakeLists.txt
PingLiuPing: @zhjwpku Exposed `ICEBERG_EXAMPLE_CXX_STANDARD` and updated document accordingly in de7b278ec124f61defb6fca39fc8e2bfdf6be339.

src/iceberg/expected.h
```diff
@@ -0,0 +1,2443 @@
...
+/// must stay consumable from C++20 translation units. `std::expected` is a
+/// C++23 library feature, so `iceberg::Result<T>` (see iceberg/result.h) is
+/// defined in terms of `iceberg::expected` instead. The API mirrors
+/// `std::expected` (including the monadic `and_then`, `or_else`, `transform`
+/// and `transform_error`), so code can be written as if against the standard
+/// type.
+///
+/// History:
+/// - apache/iceberg-cpp#40 vendored this header, adapted from
+///   https://github.com/zeus-cpp/expected (MIT), while the project targeted
```
wgtmac, line 39: Could we pin the vendored source to a specific upstream tag/commit and record the local delta here? The current history only links the repository, so future audits and upstream syncs will not know which version this 2.4k-line file came from.

src/iceberg/util/error_collector.h
```diff
@@ -101,6 +104,13 @@ class ICEBERG_EXPORT ErrorCollector {
   ErrorCollector(const ErrorCollector&) = default;
   ErrorCollector& operator=(const ErrorCollector&) = default;
 
+// C++23 uses deducing `this` so that `return AddError(...)` keeps returning the
```
wgtmac, line 107: Both this class and `SnapshotUpdate` now duplicate the full API and comment blocks across the C++20/C++23 branches. Could we centralize the deducing-`this` feature test and avoid maintaining two implementations, or at least keep one API shape unless derived-type fluent chaining is a required compatibility guarantee?

example/CMakeLists.txt
```diff
@@ -20,13 +20,79 @@ cmake_minimum_required(VERSION 3.25)
...
-set(CMAKE_CXX_STANDARD 23)
+# C++20 is the minimum standard iceberg-cpp's public headers support, so the
+# example builds as C++20 by default to keep that contract exercised. Set this to
+# 23 to also check the headers from a C++23 consumer.
+set(ICEBERG_EXAMPLE_CXX_STANDARD
+    20
+    CACHE STRING "C++ standard used to build the example (20 or 23)")
+set_property(CACHE ICEBERG_EXAMPLE_CXX_STANDARD PROPERTY STRINGS 20 23)
+if(NOT ICEBERG_EXAMPLE_CXX_STANDARD MATCHES "^(20|23)$")
+  message(FATAL_ERROR "ICEBERG_EXAMPLE_CXX_STANDARD must be 20 or 23, got "
```
wgtmac, line 31: Do we really need to specify 20 or 23 here? We need to update this file as well when we support C++26.

mkdocs/docs/getting-started.md
```diff
@@ -23,10 +23,14 @@
 
 **Required:**
 
-- C++23 compliant compiler (GCC 14+, Clang 18+, MSVC 2022+)
+- C++23 compliant compiler (GCC 14+, Clang 18+, MSVC 2022+) to build iceberg-cpp itself
```
wgtmac, line 26: It seems better to keep here unchanged to indicate that we officially support C++23. Then we can add a dedicated section below for the contract of C++20 compatibility.

example/CMakeLists.txt
```diff
@@ -20,13 +20,79 @@ cmake_minimum_required(VERSION 3.25)
...
+  string(APPEND ICEBERG_PUBLIC_HEADER_CHECK_SOURCE
+         "#include <${ICEBERG_PUBLIC_HEADER_RELATIVE_PATH}>\n")
+endforeach()
+string(APPEND ICEBERG_PUBLIC_HEADER_CHECK_SOURCE "\nint main() { return 0; }\n")
+
+file(GENERATE
+     OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/public_headers_check.cc"
+     CONTENT "${ICEBERG_PUBLIC_HEADER_CHECK_SOURCE}")
+
+add_executable(public_headers_check "${CMAKE_CURRENT_BINARY_DIR}/public_headers_check.cc")
```
wgtmac, line 96: IMO, a better alternative is to add a dedicated test executable built with C++20. It takes extra steps to install iceberg libraries and then build the example. We can add a non-installed header file (e.g. src/iceberg/cpp20_compatibility_internal.h) to include all public headers and then use it in the test case. The challenge is to make this header file in sync when we add new header files. We can update AGENTS.md to add this as an advice.

PR #945, opened 2026-09-19
refactor: rename SnapshotCache to SnapshotReader

zhjwpku:
Name the class after its role of reading a snapshot's manifests, and document that it loads and caches the manifest list on first access. Update call sites and local variable names to match, preserving behavior.


manuzhang (approved):
This looks like a clean refactor. The new name makes more sense to me. Thanks @zhjwpku

wgtmac (approved):

wgtmac:
Nice clean up! Thanks @zhjwpku and @manuzhang!

PR #947, opened 2026-09-20
fix(manifest): preserve snapshot ID inheritance semantics

wgtmac:
## Purpose

C++ converted an unassigned manifest snapshot ID from `null` to `-1`.
This made inheritance-ready manifests indistinguishable from manifests
containing a literal `-1`.

As a result, `FastAppend` and `MergeAppend` could directly reuse a manifest
with explicit `-1` entry snapshot IDs instead of rewriting those entries
with the new commit snapshot ID.

## Changes

- Change `ManifestFile::added_snapshot_id` to `std::optional<int64_t>`.
- Preserve `null` from `ManifestWriter::ToManifestFile`.
- Directly reuse manifests only when `added_snapshot_id` is null.
- Copy manifests with an explicit `-1` and assign the commit snapshot ID.
- Continue requiring `added_snapshot_id` when writing manifest lists.
- Update committed-manifest consumers to handle the optional in-memory field.
- Add FastAppend and MergeAppend coverage for null inheritance and explicit `-1`.


zhjwpku (approved):
I checked the Java FastAppend impl [0]. It reuses the original manifest only when `canInheritSnapshotId() && snapshotId() == null`. For an explicit `-1`, it rewrites the manifest. These changes align the C++ behavior with Java, thanks.

+1

[0] https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/FastAppend.java#L120-L127

manuzhang (approved):

blackmwk (approved):

